PR #3, opened 2016-07-26
A framework for N-dimension points

andyferris:
**Don't merge yet**

Playing around with some ideas on how to deal with tangent spaces of points in N-dimensional spaces that might not be linear (i.e. parameterized by Cartesian coordinates). There is a `Point` type for Affine points but in general there exists an `AbstractPoint` class with the ability to add/subtract vectors to points and to subtract points from each other. There are clever tricks to e.g. allow the UTMZ type from Geodesy to keep it's zone intact.

Not yet sure if this approach is the right one. It seems useful for UTM types which are quasi-vector-like, but the real problem we are trying to solve is getting access to the tangent space of arbitrary coordinate parameterizations and the kinds of subtractions defined here make less sense for e.g. spherical coordinates (since displacement vectors don't have to be small). Perhaps this could be a lighter-weight interface involving `size()` and `add_derivs()` / `get_derivs()` or similar... or there could be some kind of wrapper type for a point and its tangent space.

@c42f

### DIFF
--- a/src/Point.jl
+++ b/src/Point.jl
@@ -1,0 +1,301 @@
+# This file introduces the concept of an "AbstractPoint", which is a
+# generalizatoin of an Affine space where you can add vectors to points and
+# subtract points to get displacement vectors, but you may not directly add
+# points, etc.
+#
+# Points should have a subset of vector behaviour, including: scalar indexing,
+# subtraction two points (giving a vector), addition or subtraction by a vector.
+
+"""
+    abstract AbstractPoint{N, T}
+
+An `AbstractPoint` represents the location of a point in an `N`-dimensional
+space. In general this space is not assumed to be Cartesian, and the "origin" of
+the space (where the coordinates equal zero) is not assumed to have any special
+signficance.
+
+However, `AbstractPoint`s do have a well-defined tangent space about their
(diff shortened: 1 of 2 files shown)


src/Point.jl
```diff
@@ -0,0 +1,301 @@
...
+        if i == get(eltype_param_number)
+            T_out = T_out{T}
+        else
+            T_out = T_out{SA.parameters[i]}
+        end
+    end
+    return T_out
+end
+
+@pure function super_eltype_param_point(T)
```
c42f: heh, `super_duper_mega_eltype_param_point` ?

c42f:
As I said in person, I'm not quite sure *what* to think about this yet :-) Need to go dredging in ~~memory~~wikipedia for some differential geometry that I've forgotten.

c42f:
Oh, and btw, I do think this prototyping is great, but I don't think it really belongs in StaticArrays - it's more about differential geometry than it is about small arrays.

andyferris:
Yes, agreed. It was more there for the similarities of implementation and
convenience of not having to make *yet another* new package.

andyferris:
Yeah, let's do it.

PR #15, opened 2016-08-11
Non-allocating fill

timholy:
Also turns on precompilation (hope that's OK!).


andyferris:
Yes, obviously better. Thanks @timholy!

andyferris:
@timholy, if you want I can make you a collaborator? I trust your judgement :)

PR #16, opened 2016-08-12
Improve inference of eltype in matrix multiplication

timholy:
I was using this in image processing and multiplying a matrix of `Float32` times a vector of `RGB{Float32}`.

The only risk here is that `zero(T)` might fail. I'm not sure what else to do, though.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -115,18 +115,6 @@ end
 # @generated function Base.zeros!{SA <: StaticArray}(a::SA)
 # @generated function Base.ones!{SA <: StaticArray}(a::SA)
 
-@generated function Base.fill!{SA <: StaticArray}(a::SA, val)
-    l = length(SA)
-    T = eltype(SA)
-    exprs = [:(@inbounds a[$i] = valT) for i = 1:l]
-    return quote
-        $(Expr(:meta, :inline))
-        valT = convert($T, val)
-        $(Expr(:block, exprs...))
-        return a
-    end
-end
-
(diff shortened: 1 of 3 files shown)


timholy:
Any issues here?

andyferris:
Sorry, no problem at all - just a busy weekend :)

andyferris:
But to be *perfectly* safe should it also have an addition?

andyferris:
Out of curiosity - why didn't `promote_type(Float64, RGB{Float32})` work?

schmrlng:
I think we're down to zero `fill!` methods now - maybe revert b5459da?

andyferris:
Haha thanks :)

timholy:
> Sorry, no problem at all - just a busy weekend :)

Didn't mean to pressure you, from your comments on my issue I just assumed it might have gotten missed. (Hope the weekend was fun.)

> Out of curiosity - why didn't promote_type(Float64, RGB{Float32}) work?

It did: it gave `Any` :smile:.

But `promote_type` can't really do the job, because the right answer depends on the operator as well as the types. For example `vcat(::Float64, ::RGB{Float32})` should return a `Vector{Any}`, `::Float64 + ::RGB{Float32}`is not defined, and `::Float64 * ::RGB{Float32}` should return a `RGB{Float64}`. `promote_type` is really only suitable for things like concatenation, where you're looking for a type that can hold elements of both types.

timholy:
> But to be perfectly safe should it also have an addition?

True. I don't know of any cases where that's necessary now, but it might be even better.

PR #20, opened 2016-08-24
Make Array-of-SArray `op` Number infer the right output Array type

c42f:
For some reason, there's special rules using Base.promote_op to infer
the correct type for operations involving Number and arrays of other
types.  This means we need to overload promote_op or we'll get an Array
of Any in these cases.

See https://github.com/SimonDanisch/FixedSizeArrays.jl/issues/156 for a similar issue.


andyferris:
Great, thanks!

Issue #26, opened 2016-09-05
Allow @SVector comprehension to eval range with locals?


dbeach24:
Because the `@SVector` macro does not evaluate the range using the local scope, I find myself transforming code like this:

``` jl
function mytransform{N}(x::SVector{N})
    @SVector [x[i] * i for i=1:N]
end
```

into this:

``` jl
function mytransform{N,T}(x::SVector{N,T})
    SVector{N,T}(ntuple(i -> x[i] * i, N) ...)
end
```

I'm not sure if the second form is as efficient, but in either case, the first version is far easier to read.

Is there any chance that `@SVector` comprehensions can be made to support local variables in the range?  Or, is there some other way I should be writing this?

c42f:
Normally `@SVector` unrolls the generator expression, but that's not possible in this case without extra compiler support since `N` isn't a literal known to the macro.

However, the second version can be written directly with a generator which isn't too bad:

```julia
function mytransform{N,T}(x::SVector{N,T})
    SVector{N,T}(x[i]*i for i=1:N)
end
```

From a quick look at code_native, it seems that a call is still being made so this probably won't be super fast.  It may be possible to improve the codegen with an extra generated function, to unroll once `N` is known?

andyferris:
@dbeach24 This is something I'd *like* to do but as @c42f mentions its a little difficult with current compiler support. I have an idea for one hack which would end up using a call and a loop yet AFAIK it still should be faster to write
```
function mytransform{N,T}(x::SVector{N,T})
    @inbounds return SVector{N,T}(x[i]*i for i=1:N)
end
```
as @c42f we need an `@inbounds` somewhere for the `x[i]` (otherwise I don't know why it would call, but if that doesn't work then I should investigate).

Another approach is to have a *statically valued* thing like `SUnitRange{1, N}() <: StaticVector{Int}` and call `map` on it. This will probably be implemented sometime soonish.

To my understanding, this one:
```
function mytransform{N,T}(x::SVector{N,T})
    SVector{N,T}(ntuple(i -> x[i] * i, N) ...)
end
```
should be performant up until N = 16 and slow afterwards, The compiler people are interested in making this fast for all N (a longer term goal, however).

andyferris:
I've started to look at this in #30

andyferris:
OK I've had a thought. First we implement function initialization for all `StaticArrays` such that
```julia
SVector{N}(f::Function)
```
[...]

PR #30, opened 2016-09-15
WIP: static ranges for smarter macro comprehensions

andyferris:
To address #26, we need some way of getting constant values in a comprehension like

``` julia
function f{N}(::SomeType{N})
    ...
    vec = @SVector [2x for x = 1:N]
    ...
end
```

Here I play with new types `StaticIndex{Idx,T}` and the much simpler `StaticOneTo{N}`. The latter is hooked into the `SVector` macro to achieve the above. These ideas need generalizing and making robust, and I'd appreciate any opinions on how we might want to deal with _statically valued_ arrays?

### DIFF
--- a/src/SVector.jl
+++ b/src/SVector.jl
@@ -53,6 +53,18 @@ macro SVector(ex)
             error("Use a one-dimensional comprehension for @SVector")
         end
 
+        # If its a 1:N we allow N to be a constant (e.g. a type parameter)
+        rng_expr = ex.args[2].args[2]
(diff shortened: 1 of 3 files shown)


andyferris:
>  any opinions on how we might want to deal with statically valued arrays?

E.g. Is statically-valued worth its own type tree like `StaticArray`, complete with many generic methods? What is the interface? Is it a trait? 

@timholy @c42f

c42f:
I think this is superseded by #150 ?  Please reopen if I'm wrong!

PR #33, opened 2016-09-25
Fix copy-paste error in some getindexes

nbaum:
Looks like the bounds-checking code was copied from functions where `m` is a matrix. Here that parameter is `a`. :)


andyferris:
Thanks @nbaum for catching this.

Looks good to me.

PR #38, opened 2016-09-29
add example of converting `Array` to `SArray`

jingpengwu:


andyferris:
Good idea. The `Int` should be unnecessary, though.

jingpengwu:
theoretically, we don't need the `Int`. but it does not work!

```
julia> sa = SArray{(2,2)}(m)
ERROR: MethodError: no method matching promote_tuple_eltype(::Type{Array{Int64,2}})
Closest candidates are:
  promote_tuple_eltype{T<:Tuple}(::Union{T<:Tuple,Type{T<:Tuple}}) at /Users/jpwu/.julia/v0.5/StaticArrays/src/util.jl:14
 in Type(...) at /Users/jpwu/.julia/v0.5/StaticArrays/src/SArray.jl:42

```

andyferris:
Oh, well I'd better fixed that first. I'm on vacation right now, but I'll have a look as soon as I can.

PR #41, opened 2016-10-04
Speedier AbstractArray conversion

andyferris:
This needs to test (tests don't finish on my Windows machine...)


c42f:
Looks good to me (Other than the fact that there's yet one more generated function :-P )

andyferris:
> Other than the fact that there's yet one more generated function 

Right! Oh well.

andyferris:
Tests on my linux machine don't terminate either. Needs investigation

andyferris:
So there might be an issue with compilation, see JuliaLang/julia#18794.

For now I've worked around this by renaming some variable names to avoid rebinding. Should be good to go, if tests pass.

PR #43, opened 2016-10-04
Add zero function

### LINKED ISSUE #42: Incorrect output type for zero function
Similar to #6, the `zero` function returns an MVector when the input is an SVector.  For example:

```julia
julia> zero(SVector(1, 2))
2-element StaticArrays.MVector{2,Int64}:
 0
 0
```

danielmatz:
Fixes #42.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -36,7 +36,7 @@ Base.promote_op{Op,T<:Number,A<:StaticArray}(op::Op, ::Type{T}, ::Type{A}) = sim
         $(Expr(:call, SA, Expr(:tuple, v...)))
     end
 end
-Base.zero{SA <: StaticArray}(a::Union{SA,Type{SA}}) = zeros(a)
+@inline Base.zero{SA <: StaticArray}(a::Union{SA,Type{SA}}) = zeros(a)
 
 @generated function Base.ones{SA <: StaticArray}(::Union{SA,Type{SA}})
     s = size(SA)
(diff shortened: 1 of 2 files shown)


andyferris:
Thanks @danielmatz for looking at this. Thought, actually, this only makes sense for `zero()` (the additive identity) and not for `one()` (defined as the multiplicative identity).

Similarly for `Base.Array` we have that `zero()` works and `one()` does not. 

```
julia> v = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> zero(v)
3-element Array{Int64,1}:
 0
 0
 0

julia> one(v)
ERROR: MethodError: no method matching one(::Array{Int64,1})
Closest candidates are:
  one(::BitArray{2}) at bitarray.jl:413
  one(::Number) at number.jl:63
  one{T<:Number}(::Type{T<:Number}) at number.jl:64
  ...
```

danielmatz:
Doh!  Yes, I'll change that.

danielmatz:
OK, this should be good to go now.

src/arraymath.jl
```diff
@@ -36,6 +36,7 @@ Base.promote_op{Op,T<:Number,A<:StaticArray}(op::Op, ::Type{T}, ::Type{A}) = sim
         $(Expr(:call, SA, Expr(:tuple, v...)))
     end
 end
+Base.zero{SA <: StaticArray}(a::Union{SA,Type{SA}}) = zeros(a)
```
andyferris: I've been adding `@inline` explicitly everywhere since I'm not completely sure the heuristic inlining will work or not after several layers of inlining.

andyferris:
Thank you very much @danielmatz - appreciate it :)

Issue #45, opened 2016-10-07
FixedSizeArrays --> StaticArrays


SimonDanisch:
The transition so far has been more painful then expected :(
I thought I should start a list with problems:
- [x] normalize is slow
- [ ] constructors not inherited [decision needed]
- [x] missing `op(::Array{SVector}, ::SVector)`, e.g. `a.+b` or `(+).(a,b)`

I repeat my concern from #11 :
Types inheriting from `StaticVector` don't inherit all constructors. Having a consistent set of constructors for different fixed arrays was an important goal for usability and because its not that trivial to implement them all correctly. Are there any plans to offer this? I must admit I got myself into quite a mess with that (the constructor code in FixedSizeArrays is probably the ugliest code I've written so far)... But maybe we could offer a macro that one could execute for a custom type?

I feel a bit like writing a new package targeting 0.6 (after it got a few more features we need) with all the goodies from `FixedSizeArrays` and `StaticArrays` :D

KristofferC:
> I feel a bit like writing a new package targeting 0.6 (after it got a few more features we need) with all the goodies from `FixedSizeArrays` and `StaticArrays`

and then there was 3... :P

c42f:
Looks like normalize is slow because it's going via Base and calling `similar()` resulting in an `MVector`.  I guess #46 addresses that for your use case?  (I didn't bother with the general p-norm btw.)

c42f:
The issue of `op(::Array{SVector}, ::SVector)` is a particularly thorny one, because it clashes with broadcasting semantics in a fairly fundamental way.  For example, broadcasting already gives a meaning to the interaction of a vector-of-vectors with a vector, and it's not the one you probably want:

```julia
julia> A = [[0,1], [1,2]]
2-element Array{Array{Int64,1},1}:
 [0,1]
 [1,2]

julia> A .+ [1,-1]
2-element Array{Array{Int64,1},1}:
 [1,2]
 [0,1]
```

In this case, `1` is added to `[0,1]`, and `-1` is added to `[1,2]`.  This is the definitive and well defined broadcasting rules for abstract arrays... but it's not what *I* generally want when using `Vector` as a simple *container* of short vectors rather than a numeric array.  Rather I'd prefer map-like behaviour, where the second short vector is added to each element of `A`.

```julia
julia> B = [[0,1], [1,2], [1,0]]
3-element Array{Array{Int64,1},1}:
 [0,1]
 [1,2]
 [1,0]

[...]

PR #46, opened 2016-10-07
Fast path for commonly used version of normalize()

c42f:
Also a couple of minimal tests for normalize and vecnorm

Addresses part of #45

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -301,6 +301,8 @@ end
     end
 end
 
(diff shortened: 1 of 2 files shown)


SimonDanisch:
thanks! :)

c42f:
Right, I thought that perhaps speed was more important than checking for special cases here.

andyferris:
I totally agree

PR #47, opened 2016-10-09
Support M*CartesianIndex, min, max, promotion

timholy:
Multiplication by a CartesianIndex is nice in multidimensional loops.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -163,3 +163,11 @@ end
         return a
     end
 end
+
+Base.min{T1<:Real,T2<:Real}(a::StaticArray{T1}, b::StaticArray{T2}) = _min(promote(a, b)...)
+Base.min{SA<:StaticArray}(a::SA, b::SA) = _min(promote(a, b)...)
+_min{SA<:StaticArray}(a::SA, b::SA) = SA(map(min, convert(Tuple, a), convert(Tuple, b)))
+
+Base.max{T1<:Real,T2<:Real}(a::StaticArray{T1}, b::StaticArray{T2}) = _max(promote(a, b)...)
+Base.max{SA<:StaticArray}(a::SA, b::SA) = _max(promote(a, b)...)
+_max{SA<:StaticArray}(a::SA, b::SA) = SA(map(max, convert(Tuple, a), convert(Tuple, b)))
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -1,4 +1,4 @@
-import Base: .+, .-, .*, ./
+import Base: .+, .-, .*, ./, +, -
 
 # Support for elementwise ops on AbstractArray{S<:StaticArray} with Number
 Base.promote_op{Op,A<:StaticArray,T<:Number}(op::Op, ::Type{A}, ::Type{T}) = similar_type(A, promote_op(op, eltype(A), T))
@@ -23,6 +23,13 @@ Base.promote_op{Op,T<:Number,A<:StaticArray}(op::Op, ::Type{T}, ::Type{A}) = sim
 @inline .-(a1::Number, a2::StaticArray) = broadcast(-, a1, a2)
 @inline .*(a1::Number, a2::StaticArray) = broadcast(*, a1, a2)
 @inline ./(a1::Number, a2::StaticArray) = broadcast(/, a1, a2)
+
+for op in (:+, :-, :.+, :.-, :.*, :./)
+    @eval begin
+        @inline $op(a1::StaticVector, a2::CartesianIndex) = $op(a1, convert(SVector, a2))
+        @inline $op(a1::CartesianIndex, a2::StaticVector) = $op(convert(SVector, a1), a2)
+    end
+end
 
 @generated function Base.zeros{SA <: StaticArray}(::Union{SA,Type{SA}})
     s = size(SA)
--- a/src/core.jl
+++ b/src/core.jl
@@ -165,3 +165,10 @@ function size{SA <: StaticArray}(::Type{SA})
             SMatrix{3,3}(m) # correct - size is inferrable
         """)
 end
+
+function Base.promote_rule{SA1<:StaticArray, SA2<:StaticArray}(::Type{SA1}, ::Type{SA2})
+    T = promote_type(eltype(SA1), eltype(SA2))
+    _promote_rule(similar_type(SA1, T), similar_type(SA2, T))
+end
+_promote_rule{SA<:StaticArray}(::Type{SA}, ::Type{SA}) = SA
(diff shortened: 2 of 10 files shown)


src/SMatrix.jl
```diff
@@ -123,6 +123,7 @@ end
     Base.unsafe_convert(Ptr{T}, Base.data_pointer_from_objref(m))
 end
 
+Base.promote_rule{S1,S2,T,U,L}(::Type{SMatrix{S1,S2,T,L}}, ::Type{SMatrix{S1,S2,U,L}}) = SMatrix{S1,S2,promote_type(T,U),L}
```
KristofferC: Is there anything gained by marking these pure?

src/SArray.jl
```diff
@@ -91,6 +91,8 @@ end
     Base.unsafe_convert(Ptr{T}, Base.data_pointer_from_objref(a))
 end
 
+Base.promote_rule{S,N,T,U,L}(::Type{SArray{S,T,N,L}}, ::Type{SArray{S,U,N,L}}) = SArray{S,promote_type(T,U),N,L}
```
andyferris: Can we use `similar_type` here and catch all `StaticArrays`?

src/SMatrix.jl
andyferris: Does it do anything? The compiler treats type transformations as pure since the types come out of type inference anyway. (i.e. we get `Type{SMatrix{S1,S2...}}` out of this which is *not* widened in the compiler to `::DataType`)

If `promote_type(T,U)` was sufficiently complex (uninferable) perhaps `@pure` would be useful, but I think its up to the devs of `T` and `U` to decide if they make their `promote_rule` pure.

timholy: `promote_rule` gets called by `promote_type`, and `promote_type` is annotated with `@pure`. So it wouldn't change anything.

src/SArray.jl
timholy: Good idea, see revised PR.

src/arraymath.jl
```diff
@@ -163,3 +163,11 @@ end
         return a
     end
 end
+
+Base.min{T1<:Real,T2<:Real}(a::StaticArray{T1}, b::StaticArray{T2}) = _min(promote(a, b)...)
```
andyferris: Hmm... the definition of `min` here is just plain *odd*. I see now that `Base` defines this, but wouldn't `min.(a, b)` work better and semantically make *much* more sense? And codegen for this is already great with *StaticArrays*.

Personally, I see this as one of the autovectorization operations that accidentally has not been pruned from `Base` yet. Can whatever method that calls this be updated to use the dot-call syntax?

src/core.jl
```diff
@@ -165,3 +165,10 @@ function size{SA <: StaticArray}(::Type{SA})
             SMatrix{3,3}(m) # correct - size is inferrable
         """)
 end
+
+function Base.promote_rule{SA1<:StaticArray, SA2<:StaticArray}(::Type{SA1}, ::Type{SA2})
+    T = promote_type(eltype(SA1), eltype(SA2))
```
andyferris: This doesn't provide a way of defining what to do when the sizes aren't the same. Currently we get this error:

```
julia> [SVector(1,2,3), SVector(1,2)]
ERROR: The size of type `StaticArrays.SVector{S,Int64}` is not known.

If you were trying to construct (or `convert` to) a `StaticArray` you
may need to add the size explicitly as a type parameter so its size is
inferrable to the Julia compiler (or performance would be terrible). For
example, you might try

    m = zeros(3,3)
    SMatrix(m)      # this error
    SMatrix{3,3}(m) # correct - size is inferrable

 in size(::Type{StaticArrays.SVector{S,Int64}}) at /home/ferris/.julia/v0.5/StaticArrays/src/core.jl:155
 in setindex!(::Array{StaticArrays.SVector{S,Int64},1}, ::StaticArrays.SVector{3,Int64}, ::Int64) at ./array.jl:415
 in copy!(::Array{StaticArrays.SVector{S,Int64},1}, ::Tuple{StaticArrays.SVector{3,Int64},StaticArrays.SVector{2,Int64}}) at ./abstractarray.jl:479
 in vect(::StaticArrays.SVector{3,Int64}, ::Vararg{Any,N}) at ./abstractarray.jl:27
```

It turns out that this is rather awkward!! I think we might need definitions of conversion to abstract `StaticArray`s which are no-ops in the appropriate cases. Bugger.

andyferris:
I note that `Base` doesn't support `Matrix * CartesianIndex`. Is there a reason for that?

src/core.jl
timholy: This is fixed on this branch; that's the whole point of the method just below this:
```julia
julia> [SVector(1,2,3), SVector(1,2)]
ERROR: no promotion exists for StaticArrays.SVector{3,Int64} and StaticArrays.SVector{2,Int64}
 in _promote_rule(::Type{StaticArrays.SVector{3,Int64}}, ::Type{StaticArrays.SVector{2,Int64}}) at /home/tim/.julia/v0.5/StaticArrays/src/core.jl:174
 in promote_type at ./promotion.jl:140 [inlined]
 in promote_typeof(::StaticArrays.SVector{3,Int64}, ::StaticArrays.SVector{2,Int64}, ::Vararg{StaticArrays.SVector{2,Int64},N}) at ./promotion.jl:156
 in vect(::StaticArrays.SVector{3,Int64}, ::Vararg{Any,N}) at ./abstractarray.jl:24
```
It's slightly counterintuitive, but it turns out to be much easier to check that the rest of the parameters are the same after you've harmonized the element type.

This won't catch things like `promote_type(SArray{(2,),Float32,1,2}, SVector{2,Int})`, but that's a separate issue.

src/arraymath.jl
timholy: That's reasonable. Do you want deprecation/error stubs that direct users to use `min.(a, b)` etc?

timholy:
CartesianIndex is an indexing type, not a math type---it supports a few mathematical operations that are frequently used in indexing, but in core julia `M*I` would be a very strange operation---it really only makes sense in the context of affine transformations applied to arrays (i.e., `ImageTransformation.jl`). But this package basically exists to blur the line between tuples and vectors, so there's a certain sense to it. 

However, in the most recent commit I also just added support for `convert(SVector, I::CartesianIndex)`. We could take a much more minimalistic approach and not support any math operations with CartesianIndex directly, and just ask folks to `convert`. I think that would be a fine option, personally.

timholy:
See an alternative in #49. Pick one, not both!

timholy:
Superseded by #49.

PR #49, opened 2016-10-10
Define conversion and promotion for CartesianIndex

timholy:
This is an alternative---more minimalist---version of #47. It's not quite as convenient for the user, but it has one big advantage: this approach prevents "mission creep" in needing to define more and more methods for `CartesianIndex`. On balance I slightly favor this approach, but it's a close thing.

### DIFF
--- a/src/SVector.jl
+++ b/src/SVector.jl
@@ -53,6 +53,12 @@ end
     Base.unsafe_convert(Ptr{T}, Base.data_pointer_from_objref(v))
 end
 
+# Converting a CartesianIndex to an SVector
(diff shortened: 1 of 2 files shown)


c42f:
The fact that this diff is about 5x shorter than #47 is pretty convincing.  Is the only downside that people will have to explicitly call the `SVector` constructor in some cases?

andyferris:
I'd lean slightly towards this one, to be honest? I'm ok if we merge this.

timholy:
Yes, the explicit call is the only downside. Which isn't a big deal.

PR #50, opened 2016-10-12
Added a `Scalar` type for 0-dimensional array

andyferris:
Better than `Array{T,0}` and `Ref` as scalar containers since this is immutable. It's useful for controlling broadcasting. Here is an example:

```
julia> using StaticArrays

julia> s = Scalar(SVector(1,2,3))
0-dimensional StaticArrays.Scalar{StaticArrays.SVector{3,Int64}}:
[1,2,3]

julia> v = [SVector(1,2,3), SVector(0,0,0), SVector(10,10,10)]
3-element Array{StaticArrays.SVector{3,Int64},1}:
 [1,2,3]   
 [0,0,0]   
 [10,10,10]

julia> v .+ s
3-element Array{StaticArrays.SVector{3,Int64},1}:
 [2,4,6]   
 [1,2,3]   
 [11,12,13]
```

See some relevant discussions at:

https://github.com/JuliaArrays/StaticArrays.jl/issues/45#issuecomment-252975952
https://github.com/JuliaLang/julia/issues/18379
https://github.com/JuliaLang/julia/pull/18766#issuecomment-253031994

### DIFF
--- a/src/Scalar.jl
+++ b/src/Scalar.jl
@@ -1,0 +1,20 @@
+"""
+    Scalar{T}(x::T)
+
+Construct a statically-sized 0-dimensional array that contains a single element,
+`x`. This type is particularly useful for influencing broadcasting operations.
+"""
+immutable Scalar{T} <: StaticArray{T,0}
+    data::T
+end
+
+@inline (::Type{Scalar}){T}(x::Tuple{T}) = Scalar{T}(x[1])
+
+@pure size(::Type{Scalar}) = ()
+@pure size{T}(::Type{Scalar{T}}) = ()
+
+@inline function getindex(v::Scalar)
+    v.data
+end
+
+@inline Tuple(v::Scalar) = (v.data,)
--- a/src/Scalar.jl
+++ b/src/Scalar.jl
@@ -8,13 +8,20 @@ immutable Scalar{T} <: StaticArray{T,0}
     data::T
 end
 
-@inline (::Type{Scalar}){T}(x::Tuple{T}) = Scalar{T}(x[1])
+@inline (::Type{Scalar{T}}){T}(x::Tuple{T}) = Scalar{T}(x[1])
 
 @pure size(::Type{Scalar}) = ()
 @pure size{T}(::Type{Scalar{T}}) = ()
 
-@inline function getindex(v::Scalar)
+getindex(v::Scalar) = v.data
+@inline function getindex(v::Scalar, i::Int)
+    @boundscheck if i != 1
+        error("Attempt to index Scalar at index $i")
+    end
     v.data
 end
 
 @inline Tuple(v::Scalar) = (v.data,)
+
+# A lot more compact than the default array show
+Base.show{T}(io::IO, ::MIME"text/plain", x::Scalar{T}) = print(io, "Scalar{$T}(", x.data, ")")
--- a/README.md
+++ b/README.md
@@ -229,6 +229,18 @@ easier to define the types without the extra tuple characters (compare
 because it is so easy to define new `StaticArray` subtypes, and they naturally
 work together.
 
+### `Scalar`
+
+Sometimes you want to broadcast an operation, but not over one of your inputs.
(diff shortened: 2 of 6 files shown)


andyferris:
I still need to do tests and documentation, but feedback welcome

TotalVerb:
@andyferris My implementation had these methods also. I don't remember why all of them were necessary, or if they even are.

```julia
Base.show{T}(io::IO, x::Scalar{T}) = print(io, "Scalar{$T}(", x.val, ")")
Base.show{T}(io::IO, ::MIME"text/plain", x::Scalar{T}) = show(io, x)
Base.indices(x::Scalar) = ()
Base.size(x::Scalar) = ()
Base.getindex(x::Scalar) = x.val
Base.getindex(x::Scalar, i::Integer) = i == 1 ? x.val : throw(BoundsError(x, i))
```

Perhaps add these to the test cases?

```julia
@test Scalar(2) .* [1, 2, 3] == [2, 4, 6]
@test Scalar([1 2; 3 4]) .+ [[1 1; 1 1], [2 2; 2 2]] == [[2 3; 4 5], [3 4; 5 6]]
```

src/Scalar.jl
```diff
@@ -0,0 +1,20 @@
...
+    Scalar{T}(x::T)
+
+Construct a statically-sized 0-dimensional array that contains a single element,
+`x`. This type is particularly useful for influencing broadcasting operations.
+"""
+immutable Scalar{T} <: StaticArray{T,0}
+    data::T
+end
+
+@inline (::Type{Scalar}){T}(x::Tuple{T}) = Scalar{T}(x[1])
```
TotalVerb: Doesn't this prevent scalarizing tuples? Can this maybe be a `convert` method instead?

src/Scalar.jl
```diff
@@ -0,0 +1,20 @@
...
+
+@pure size(::Type{Scalar}) = ()
+@pure size{T}(::Type{Scalar{T}}) = ()
+
+@inline function getindex(v::Scalar)
```
TotalVerb: Are these `@inline` annotations actually needed? Also, this function can probably be defined as `getindex(v::Scalar) = v.data`.

andyferris:
> My implementation had these methods also

The *StaticArrays* interface provides many methods for free

>  add these to the test cases

Thanks, will do!

TotalVerb:
This looks great!

andyferris:
done

andyferris:
Thanks @TotalVerb

andyferris:
I'm still amused that a good way of making a scalar it to make it a subtype of `AbstractArray`.

src/Scalar.jl
andyferris: The StaticArrays interface requires a constructor that accepts a Tuple. I may have messed this up a little though. 

(PS - side effect is small - it only does this for a one-tuple which is scalar-like anyway).

TotalVerb: This code looks like it's preventing scalarizing any tuple, though.

src/Scalar.jl
andyferris: As a rule I haven't been trusting the heuristics of the compiler anywhere (at times a very large amount of code is inlined). But in this case you are right, I think it will always work.

src/Scalar.jl
c42f: I got caught up on this too, but I think I'm ok after reading Andy's comment - `Tuple{T}` being the tuple containing exactly one `T` (as quite distinct from `Array{T}` containing many `T`).

TotalVerb: Ah right. The new version does not have the issue with even tuples containing one `T` though, which is even better :smile:.

c42f:
I think this is good and it fits the needs discussed in #45. As discussed there the name `Scalar` has the potential to be confusing, but probably not more than the other possibilities. It also seems to be the terminology used in the discussions in `Base`.  So :+1: to merge this.

andyferris:
Thanks for the reviews

PR #51, opened 2016-10-13
Solvers for small problems

andreasnoack:
I needed these in an application so I thought I better upstream them. It's my first PR here so please let me know how things should look like.

Update: I should probably add that I'm aware that the versions here a slightly less precise than possible but they are quite a bit simpler and since the problems are only 2x2 and 3x3 I don't think the extra precision is critical.


andyferris:
Thanks Andreas. I'll have a close look.

> I'm aware that the versions here a slightly less precise than possible

I'm happy for methods to be pretty fast and without branching to make things fast. Currently the 3x3 eig won't work properly for certain situations like 2 zero degenerate eigenvalues. Here your code won't work properly for singular (or near singular) matrices even if a solution exists. I *think* that's OK - but what is other people's opinion?

c42f:
I think `StaticArrays` should target algorithms in the "very fast and moderately accurate" category.  It's easy to convert to `Base.Array` and use the more stable alternatives (at some cost) if necessary.

As far as I've observed (eg, in FixedSizeArrays, and other small sized array libraries in the past, mainly in graphics) this is a bit of a sweet spot for small sized array libraries.

c42f:
Perhaps at some stage in the future we could even have a macro in the spirit of `@fastmath` for selecting small matrix algorithms?  There's usually several possibilities along the speed<->accuracy spectrum.

src/solve.jl
```diff
@@ -0,0 +1,16 @@
+@inline function (\){T}(A::SMatrix{2,2,T,4}, b::SVector{2,T})
+    d = det(A)
+    @inbounds x = SVector( A[2,2]*b[1] - A[1,2]*b[2],
+                           A[1,1]*b[2] - A[2,1]*b[1] ) / d
+    return x
+end
+
+@inline function (\){T}(A::SMatrix{3,3,T,9}, b::SVector{3,T})
```
c42f: As the package exists now, there's no nice way to overload this for all 3x3 `StaticMatrix` subtypes.  If you look at the code of `inv`, you'll see the unfortunate hardcoded trait-like workaround that we have so far.  What we really need is a proper size trait for fixed size dispatch.  In the meantime, I'd suggest copying the `inv` implementation, otherwise this won't work for `SArray`, `MMatrix`, `MArray`.  (We'd like to remove some of this zoo of types, but it's hard to achieve nice statically sized type names with current compiler capabilities.)

src/solve.jl
```diff
@@ -0,0 +1,16 @@
...
+    d = det(A)
+    @inbounds begin
+        x = SVector( A[1,2]*A[2,3]*b[3] - A[1,2]*A[3,3]*b[2] - A[1,3]*A[2,2]*b[3] + A[1,3]*A[3,2]*b[2] + A[2,2]*A[3,3]*b[1] - A[2,3]*A[3,2]*b[1],
+                    -A[1,1]*A[2,3]*b[3] + A[1,1]*A[3,3]*b[2] + A[1,3]*A[2,1]*b[3] - A[1,3]*A[3,1]*b[2] - A[2,1]*A[3,3]*b[1] + A[2,3]*A[3,1]*b[1],
+                     A[1,1]*A[2,2]*b[3] - A[1,1]*A[3,2]*b[2] - A[1,2]*A[2,1]*b[3] + A[1,2]*A[3,1]*b[2] + A[2,1]*A[3,2]*b[1] - A[2,2]*A[3,1]*b[1] ) / d
```
c42f: So this is basically inverse via cofactors, with the matvec multiply written out explicitly?  Seems reasonable to me, though did you compare speed-wise with `inv(A)*b`?

andreasnoack: I've changed the version a bit to reduce the number of flops by factoring out some `b`s. Now, this version has the same number of additions and multiplications as a version based on `inv` but this version only has three division whereas the `inv` based version has six and because they are much more expensive I think I see a 2x difference in the 3x3 case.

andreasnoack:
@c42f Ok, now I've committed the `@generated` sin as well. PR is updated and should work for all the `StaticArray`s.

src/solve.jl
c42f: Good point about the divisions.  A 2x performance improvement is very convincing.

src/solve.jl
c42f: Looks good, thanks.

src/solve.jl
```diff
@@ -0,0 +1,40 @@
...
+                        (A[1,3]*A[2,1] - A[1,1]*A[2,3])*b[3],
+                (A[2,1]*A[3,2] - A[2,2]*A[3,1])*b[1] +
+                    (A[1,2]*A[3,1] - A[1,1]*A[3,2])*b[2] +
+                        (A[1,1]*A[2,2] - A[1,2]*A[2,1])*b[3] ) / d
+        end
+    else
+        # FixMe! Unsatisfactory ineffective but requires some infrastructure
+        # to make efficient so we fall back on inv for now
+        quote
+            inv(A)*b
```
andyferris: OK I just confirmed that the code in Base is a little useless at the moment (uses LU or QR factorization methods, neither of which work natively with static arrays. So this needs to go on our TODO but not for this PR.

andyferris:
I will tag

andyferris:
https://github.com/JuliaLang/METADATA.jl/pull/6752

andreasnoack:
Thanks. The base code was written mainly to interface with LAPACK and has been made more generic incrementally. We should start looking into avoiding the hard coding of `Vector` in the LinAlg types but I fear the extra type parameters a bit. It can be a little complicated to handle many of them but you are probably too aware of that.

andyferris:
Yes, right. But I don't think the extra type parameter in `Eigen`, for example, is too scary at all.

`Diagonal` is the next `Base` type in my crosshairs. 

I also think traits would help here a bit... things like determining if a `Symmetric` is statically-sized or not is just a little bit more difficult to dispatch on. We would need this for diagonal, upper and lower diagonal types, etc. Grr. The current system seems to force me to reimplement all of `Base.LinAlg`!

andyferris:
@andreasnoack There seems to be a win64 error at: https://ci.appveyor.com/project/JuliaArrays/staticarrays-jl/build/job/msnol5phf8li35bu

Is that right?

c42f:
Seems to have been a spurious error, I restarted the build and it seems fine now.

PR #57, opened 2016-10-31
Inherit from AbstractArray not DenseArray

andyferris:
More properly deal with the many types of statically-sized arrays by not inheriting from `DenseArray`. We do loose some free functionality from `Base`, but most methods were not type-stable and need to be reimplemented anyway.

(also some other minor fiddling)

CC @c42f


c42f:
Looks good to me.  This is surprisingly small :+1:

andyferris:
Yes and passed all the tests. Hope it doesn't break user code.

PR #58, opened 2016-10-31
`Size` traits

andyferris:
Adds a trait-based scheme for dispatch on the size of a statically-sized arrays. The `Size{(m,n,...)}` type represents the trait (and currently isn't constructed). It's abstract supertype `SizeOf` helps implement the "traitor" style of automated trait dispatch whereby a supertype is used to construct a trait.

One standout feature is that it allows me to relax the type signature of a size-dispatched method to `AbstractArray`. As an example, similar methods to these are now defined:

```julia
det(x::StaticMatrix) = det(SizeOf(x), x)
det(::Type{Size{(1,1)}}, x::AbstractMatrix) = x[1,1]
det(::Type{Size{(2,2)}}, x::AbstractMatrix) = x[1,1]*x[2,2] - x[1,2]*x[2,1]
# and other definitions as necessary
```

Using BenchmarkTools we see a 6.5x speedup on the 2x2 case:
```

julia> mat = [1.0 2.0; 3.0 4.0]
2×2 Array{Float64,2}:
 1.0  2.0
 3.0  4.0

julia> @benchmark det(mat)
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     455
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  320.00 bytes
  allocs estimate:  7
  minimum time:     235.00 ns (0.00% GC)
  median time:      243.00 ns (0.00% GC)
  mean time:        280.85 ns (10.09% GC)
  maximum time:     6.76 μs (93.27% GC)

julia> @benchmark det(Size(2,2),mat)
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     997
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  16.00 bytes
  allocs estimate:  1
  minimum time:     36.00 ns (0.00% GC)
  median time:      37.00 ns (0.00% GC)
  mean time:        38.60 ns (1.35% GC)
  maximum time:     1.34 μs (96.21% GC)
```
There was also a 7x speedup for the 3x3 case. Obviously static arrays were already getting this benefit but this might be easier to use than `MMatrix` for example.

CC @c42f @timholy @SimonDanisch

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -13,6 +13,8 @@ export Scalar, SArray, SVector, SMatrix
 export MArray, MVector, MMatrix
 export FieldVector, MutableFieldVector
 
+export Size, SizeOf
+
 export @SVector, @SMatrix, @SArray
 export @MVector, @MMatrix, @MArray
 
@@ -21,6 +23,7 @@ export similar_type
 include("util.jl")
 
 include("core.jl")
+include("traits.jl")
 include("Scalar.jl")
 include("SVector.jl")
 include("FieldVector.jl")
@@ -29,7 +32,6 @@ include("SArray.jl")
 include("MVector.jl")
 include("MMatrix.jl")
 include("MArray.jl")
-
 
 include("indexing.jl")
 include("abstractarray.jl")
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -13,7 +13,7 @@ export Scalar, SArray, SVector, SMatrix
 export MArray, MVector, MMatrix
 export FieldVector, MutableFieldVector
 
-export Size, SizeOf
+export Size
 
 export @SVector, @SMatrix, @SArray
 export @MVector, @MMatrix, @MArray
--- a/src/det.jl
+++ b/src/det.jl
@@ -1,4 +1,4 @@
-@inline det(A::StaticMatrix) = det(SizeOf(A), A)
+@inline det(A::StaticMatrix) = _det(Size(A), A)
 
 """
     det(Size(m,m), mat)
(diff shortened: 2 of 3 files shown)


andyferris:
An alternative (perhaps more Julian?) approach to specialized dispatch would be something like

```julia
wrapped_mat = Size(2,2)(mat)  # Some kind of StaticArray subtype which wraps mat with a size
det(wrapped_mat)
```

c42f:
As the `det()` implementation shows, this is going to be a big improvement. I think just `Size` by itself would be fine; I don't think having `SizeOf` separate really buys us much here.

As to whether we should use a wrapper instead, I'm not sure.  It solves the issue of argument lists getting out of control with extra traits everywhere... but it also feels somehow unnatural to me.  Perhaps because the wrapper will not be true to the type hierarchy of the wrapped type.  As an implementation tool for things like `det` that doesn't seem to be a problem, but for more general usage it might be bad?

andyferris:
Thanks for the input

> I think just Size by itself would be fine; I don't think having SizeOf separate really buys us much here.

True, makes sense

> but it also feels somehow unnatural to me. Perhaps because the wrapper will not be true to the type hierarchy of the wrapped type.

I'm thinking of having
```julia
immutable SizedArray{S,T,N} <: DenseArray{T,N}
    data::Array{T,N}
end
```
since other structures (from sparse arrays and views to arrayfire/distributed arrays) really aren't going to benefit (speedwise) as much from these algorithms. Maybe that is too restrictive but we can revisit this in the future, if there is a good reason.

c42f:
> `SizedArray{S,T,N} <: StaticArray{T,N}`

Sure, that'd work in this case.  It was more of a general meta-objection that wrapping types might not be a good general tool for enhancing types with additional traits. Dunno, would be interesting to see what it looks like.

andyferris:
OK things should be simpler now.

PR #60, opened 2016-11-01
Add `SizedArray` wrapper for `Array`

andyferris:
Now allows `Array` to use the (faster) algorithms from StaticArrays.jl, if the user knows the dimensions at compile-time. A convenience constructor from calling a `Size` instance is provided, e.g.
```julia
julia> Size(2,2)(rand(2,2))
2×2 StaticArrays.SizedArray{(2,2),Float64,2}:
 0.367002  0.592863
 0.725256  0.151296

julia> det(ans)  # Uses a StaticArrays `det()` method
-0.3744516359087295
```

### DIFF
--- a/src/SizedArray.jl
+++ b/src/SizedArray.jl
@@ -1,0 +1,46 @@
+
+"""
+    SizedArray{(dims...)}(array)
+
+Wraps an `Array` with a static size, so to take advantage of the (faster)
+methods defined by the static array package. The size is checked once upon
+construction.
+
+(Also, `Size(dims...)(array)` acheives the same thing)
+"""
+immutable SizedArray{S,T,N} <: StaticArray{T,N}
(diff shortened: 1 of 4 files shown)


andyferris:
@c42f

c42f:
Interesting.  How does `SizedArray` compare with `MArray` in performance?  Do we need both?  They seem quite similar.

andyferris:
My feeling/experience is that performance can be better for either in different situations, but systematic benchmarking would be helpful. At least we don't have `setindex!` problems with GC etc with `SizedArray`.

I can definitely see `SizedArray` as useful in some situations so the question is whether it leads to `MArray` being removed, so on that logic I'm going to merge this.

c42f:
Sure, fine by me.

PR #66, opened 2016-11-03
Use `Size` more consistently

andyferris:
Begin implementing use of `Size` where it makes sense ,and related changes.

Also added some *BenchmarkTools.jl* performance tests (get unusual results).


andyferris:
Merging

PR #67, opened 2016-11-08
FixedSizeArray tests

SimonDanisch:
This doesn't really need to be a PR, but It's an easy way to keep track of what is already working and discuss things.
Looks very good so far, thanks for putting so much work into this!

I put the things not working into the comments. Here's them extracted:

### Need to find an equivalent in StaticArrays:
```Julia
testset "scalar nan" begin
    for (p, r) in (
            (Point{2, Float32}(NaN, 1), true),
            (Point{2, Float64}(1, NaN), true),
            (Vec{11, Float64}(NaN), true),
            (Point{2, Float32}(1, 1), false),
            (RGB{Float32}(NaN), true),
        )
        @fact isnan(p) == r
    end
end
```

# methods I needed to define:

This is basically type pirating. I think it's hard to get the 1 valued constructor
otherwise (at least last time I tried). I love it though, so maybe still do it?
It's pretty much my most favorite constructor.
```Julia
Base.convert{N, T}(::Type{NTuple{N}}, x::T) = ntuple(i-> T(x), Val{N})
Base.convert{N, T}(::Type{NTuple{N, T}}, x::Real) = ntuple(i-> T(x), Val{N})
```

Probably problematic: treating SVector as scalar... On the other hand
this should be fairly ambiguity free:

```Julia
function Base.extrema{T <: StaticVector}(a::AbstractVector{T})
    reduce((x, v)-> (min.(x[1], v), max.(x[2], v)), a)
end
function Base.minimum{T <: StaticVector}(a::AbstractVector{T})
    reduce((x, v)-> min.(x[1], v), a)
end
function Base.maximum{T <: StaticVector}(a::AbstractVector{T})
    reduce((x, v)-> max.(x[1], v), a)
end
```

# not passing, will port later

### Unit not implemented
```Julia
unit(Vec4d, 1)
```

### rand not all versions implemented
```Julia
rand(Vec{7, Int}, 1:7)
randn(Mat{4,2, Complex{Float64}})
@test typeof(rand(Mat4d, -20f0:0.192f0:230f0)) == Mat4d
@test typeof(rand(Mat{4,21,Float32}, -20f0:0.192f0:230f0)) == Mat{4,21,Float32}
```

### one not implemented
```Julia
x = one(Mat{4,2, Int})
```

### complex yields different results:
```Julia
v2 = Vec(6.0,5.0,4.0)
v1 = Vec(1.0,2.0,3.0)
vi = Vec(1,2,3)
v2 = Vec(6.0,5.0,4.0)
v1c = Vec(6.0+3.0im,5.0-2im,4.0+0.0im)
v2c = v1 + v2*im
v2c = Vec(1.0 + 6.0im, 2.0 + 5.0im, 3.0 + 4.0im)

@testset "Complex Ops" begin
    @testset "dot product" begin

        @test dot(v1c,v2c) == dot([6.0+3.0im,5.0-2im,4.0+0.0im], [1.0,2.0,3.0] + [6.0,5.0,4.0]*im)
        @test Vector(transpose(v1c)*v2c) == [6.0+3.0im 5.0-2im 4.0+0.0im]*([1.0,2.0,3.0] + [6.0,5.0,4.0]*im)
        @test Matrix(v2c*transpose(v1c)) == ([1.0,2.0,3.0] + [6.0,5.0,4.0]*im)*[6.0+3.0im 5.0-2im 4.0+0.0im]
    end
end
```

### setindex is slow

```Julia
m = Mat{4,4}(
    1,2,3,4,
    5,6,7,8,
    9,10,11,12,
    13,14,15,16
)
setindex(m, 42.0, 2,2)
```

### No 2D cross
```Julia
a,b = Vec2d(0,1), Vec2d(1,0)
@test cross(a,b) == -1.0
@test isa(cross(a,b), Float64) == true
```

### slow
```Julia
all(x-> x==1, x) == true
```

I call it a day for now ;)

Next important item on the list:
Implementing the Face type from GeometryTypes!


andyferris:
OK, nice work Simon. This is awesome! I can't process it all right now but I'll comment separately.

andyferris:
I see there are some things which extend stuff from base, for instance `one` on non-square matrices (use `eye` for that), and `cross` on 2D vectors. I've never seen `unit` before - could you explain? 

Everything else seems like simple fixes :)

andyferris:
I'm also finishing up some stuff which uses my new `Size` trait for dispatch. I'll try and merge that soon, and then turn my attention to this branch. I think we can merge this... the more tests the better, right?

andyferris:
> This is basically type pirating. I think it's hard to get the 1 valued constructor
otherwise (at least last time I tried). I love it though, so maybe still do it?
It's pretty much my most favorite constructor.
```
Base.convert{N, T}(::Type{NTuple{N}}, x::T) = ntuple(i-> T(x), Val{N})
Base.convert{N, T}(::Type{NTuple{N, T}}, x::Real) = ntuple(i-> T(x), Val{N})
```

Right... why on earth there aren't any `Tuple` constructors defined in base is a mystery to me!

I had lots and lots of trouble with the 1 element constructor also. At the moment it seems to be working, but it might be fragile
```julia
julia> SVector(1)
1-element StaticArrays.SVector{1,Int64}:
 1
```
Is there something I've missed?

andyferris:
> ```
function Base.extrema{T <: StaticVector}(a::AbstractVector{T})
    reduce((x, v)-> (min.(x[1], v), max.(x[2], v)), a)
end
function Base.minimum{T <: StaticVector}(a::AbstractVector{T})
    reduce((x, v)-> min.(x[1], v), a)
end
function Base.maximum{T <: StaticVector}(a::AbstractVector{T})
    reduce((x, v)-> max.(x[1], v), a)
end
```

OK this is interesting. Arguably these two behaviours of base are inconsistent:
```julia
julia> min([1,2], [2,-1])
2-element Array{Int64,1}:
  1
 -1

julia> minimum([[1,2], [2,-1]])
ERROR: ArgumentError: ordering is not well-defined for arrays
 in _mapreduce(::Base.#identity, ::Base.#scalarmin, ::Base.LinearFast, ::Array{Array{Int64,1},1}) at ./reduce.jl:154
 in minimum(::Array{Array{Int64,1},1}) at ./reduce.jl:317
```

I'm wondering if the first will change to `min.()` in the future. In which case, we'll need to find a different name for these functions. Otherwise, we should fix `Base.minimum(x)` to become `reduce(min, x)`. I'll file an issue.

andyferris:
See https://github.com/JuliaLang/julia/issues/19275

andyferris:
For scalar `isnan`, why not use
```julia
any(isnan.(array))
```
?

andyferris:
`hypot` seems to be broadcasting. Just use `hypot(a1, a2)`.

andyferris:
```julia
# Why is this not in Base???
@inline function randn{T}(mt::MersenneTwister, ::Type{Complex{T}})
    k = sqrt(T(2))
    Complex{T}(randn(mt, T)/k, randn(mt,T)/k)
end
```

andyferris:
Your `rand` might be backwards:

```
julia> v = rand(1:7, SVector{3}) # new method
3-element StaticArrays.SVector{3,Int64}:
 1
 6
 7

julia> rand(1:7, 3)
3-element Array{Int64,1}:
 7
 6
 2
```

andyferris:
OK, setindex` is now defined (for a single element change). Not sure how to make it fast though... I guess I'll look at FixedSizeArrays.

I've made progress here, might be time to rebase on master and check your list again.

SimonDanisch:
@andyferris wow you're fast! ;)
```

**Comment by SimonDanisch** [comment]
```Julia
any(isnan.(array))
```
That's how I figured out, that `any` seems to be slow ;)
Mapreduce is the way forward, as long as it gets inlined I suppose. Mapreduce doesn't work right now: ` MethodError: mapreduce(::Base.#isnan, ::Base.#|, ::StaticArrays.SVector{4,Float64}) is ambiguous. Candidates`

SimonDanisch:
> I had lots and lots of trouble with the 1 element constructor also. At the moment it seems to be working, but it might be fragile
> Is there something I've missed?

Sorry, wasn't very clear from my side:
```Julia
SVector{3}(1) -> SVector(1, 1, 1)
```

andyferris:
`any` should be faster on master now

SimonDanisch:
Hm weird, the generated code looks much better, but it's still not as fast as it could be... Well, not that important right now ;)

SimonDanisch:
I added a 2D cross... I know, this has been discussed, but I've seen this in multiple math boards arguing for that and I've run into enough uses of 2D cross to make it seem justifiable ;)
On a side node, why is `cross` a generated function? Is this again forcing the compiler to specialize? Wouldn't using `::T` be enough for that, though?

SimonDanisch:
`unit` creates a unit vector for axis `i`: 
```Julia
function @inline unit{SV <: StaticVector}(::Type{SV}, i::Integer)
    T = eltype(SV);
    map(j-> ifelse(i == j, T(1), T(0)), SV)
end
```
What's the preferred way to implement this? I see that there is no `map(idx -> ..., Type{SVector})` ... `ntuple` would be another way, but will the length propagate statically? ;)

SimonDanisch:
I'm a bit surprised about the heavy use of generated functions... I tried to restrict them to just a few places in FixedSizeArrays. Is this really necessary that almost every function is a generated one?
I'm not sure if it matters much, but restricting the usage of them was (at least last time I heard about it) the consent in Base.

andyferris:
So the approach here is kind-of insane, aiming for maximal speed for smallish arrays, where "smallish" is up to the user. We inline everything and unroll all loops, generating code up to linear in the number of elements in the array(s) (the matrix multiplication code being a complicated exception). So a great many (but not all) functions have a need to be generated, and many others ended up that way via copy-paste :)

Generated functions have the following negative properties

1. Make it difficult to statically compile a program (though not impossible).
2. Are slower to compile than regular functions.
3. Harder to read, etc.

Since you can't usefully use the package without calling any generated functions, item 1 didn't seem important. Item 2 is justified by the fact that we are targetting fast run-time and more-or-less ignoring compile-time cost. I justified item 3 (and 2) to myself by the fact that it *might possibly* avoid specialization issues where the compiler simplfies "complicated" types like `NTuple{1000, Float64}`. I found via trial-and-error that functions which where both inlined and generated caused the fewest codegen issues. 

On the other hand. just yesterday I found problems with missing inbounds and inlining in various places, so it is possible the "bad' codegen I was seeing was sometimes caused by such omissions.  I also developed my intuition on pre-RC releases of 0.5 (since which some codegen issues have been fixed).  Unfortunately, I simply do not understand the compiler heuristics well enough so I decided to *never* trust the heuristics unless it was known to be safe.

Not sure if that helps or explains things. I actually want to remove **all** of the generated functions when the codegen is the same (even for large arrays, for instance `ntuple` dies for length 20, which is too small). There's even an issue open listing the required compiler support, and it comes down to just 2 or 3 compiler features. I've been trying to switch to BenchmarkTools.jl to detect regressions reliably so we can remove individual generated functions with some "proof" that it is OK.

andyferris:
So I made `setindex` fast :) Have a look at this:

```julia
julia> v = rand(1:10, SVector{3})
3-element StaticArrays.SVector{3,Int64}:
 10
  6
  7

ulia> f(x) = @inbounds return setindex(x, 100, 2)

julia> f(v)
3-element StaticArrays.SVector{3,Int64}:
  10
 100
   7

julia> @code_native f(v)
	.text
Filename: REPL[11]
	pushq	%rbp
	movq	%rsp, %rbp
Source line: 73
	movq	(%rsi), %rax
	movq	16(%rsi), %rcx
Source line: 1
	movq	%rax, (%rdi)
	movq	$100, 8(%rdi)
	movq	%rcx, 16(%rdi)
	movq	%rdi, %rax
	popq	%rbp
	retq
	nop
```
It turns out that `ifelse` is quite powerful :)

SimonDanisch:
Great! Yeah that's how I'd implement this nowadays :)

After your thorough explanation, I think we're on the same page with generated functions. Just wanted to be sure of that ;)

SimonDanisch:
There is still: https://github.com/SimonDanisch/FixedArrayBench.jl
Where I started trying to do exactly this thorough comparison with BenchmarkTools ;)
There is also https://github.com/JuliaCI/PkgBenchmark.jl which should help us with that

SimonDanisch:
Finally got around to the facetype.
I noticed this behavior:
```Julia
A = [1,2,3,4]
A[(1,2,3)] -> SVector(1,2,3)
A[SVector(1,2,3)] -> [1,2,3]
```
I guess I see where that's coming from, but shouldn't we change that to e.g.:
```Julia
A = [1,2,3,4]
A[(1,2,3)] -> (1,2,3)
A[SVector(1,2,3)] -> SVector(1,2,3)
```
:P 
And the tuple version should probably be moved to base?

andyferris:
+100!

Yes, I agree the tuple indexing was a mistake but I didn't see it until afterwards that a static array makes much more sense. 

I also think indexing with a tuple -> tuple should go into base.

SimonDanisch:
Great! Let me see if I can submit a fix!

SimonDanisch:
Lol, what did I do to this PR while I wasn't paying attention? It suddenly needed a lot of fixing... Well, it's fixed now! Hope we can move forward with this.
Sorry for the mess in the constructor macro (and tests)! Hope it's fine like that. Maybe you can peek over it, to see if there are some low hanging fruits for more elegance.

andyferris:
So this passes `FixedSizeArrays`'s tests?

Are there some unit tests which we should copy over? (though from memory you were using a different test framework)

SimonDanisch:
Not all tests, but the important ones ;) And GLVisualize works with this!

SimonDanisch:
Should we merge this? Or are there still blocking items?

andyferris:
Thanks Simon for this.

src/linalg.jl
```diff
@@ -298,6 +298,16 @@ end
...
+    elseif length(a) === 2 && length(b) === 2
+        return quote
+            $(Expr(:meta, :inline))
+            promote_type(eltype(a), eltype(b))(a[1]*b[2]-a[2]*b[1])
+        end
+    elseif length(a) === 2 && length(b) === 2
+        return quote
+            $(Expr(:meta, :inline))
+            promote_type(eltype(a), eltype(b))(a[1]*b[2]-a[2]*b[1])
+        end
```
c42f: Oops, just noticed this block of code somehow got added twice!

andyferris:
I just learnt that the "2D cross product" has another name: "perp dot product". 

http://mathworld.wolfram.com/PerpDotProduct.html

(@SimonDanisch I've noticed people getting confused why a cross product isn't returning a vector, so was kind-of wondering if there was a better name for this operation or anyway to improve our API to make it clearer)

Issue #76, opened 2016-11-10
Error during SArray initialization with array comprehension containing two SVector iterates


wsshin:
Consider two SVector objects with 2 and 3 entries:
```
julia> vec2 = @SVector [1,2]
2-element StaticArrays.SVector{2,Int64}:
 1
 2

julia> vec3 = @SVector [1,2,3]
3-element StaticArrays.SVector{3,Int64}:
 1
 2
 3
```
I can use one of these two SVector objects as iterates in array comprehension to initialize SArray.  For example, 
```
julia> @SArray [1 for x = vec3, y = vec3]
3×3 StaticArrays.SArray{(3,3),Int64,2,9}:
 1  1  1
 1  1  1
 1  1  1
```
However, as soon as I mix two SVector objects, an error is generated:
```
julia> @SArray [1 for x = vec2, y = vec3]
ERROR: The size of type `StaticArrays.SVector{S,Int64}` is not known.

If you were trying to construct (or `convert` to) a `StaticArray` you
may need to add the size explicitly as a type parameter so its size is
inferrable to the Julia compiler (or performance would be terrible). For
example, you might try

    m = zeros(3,3)
    SMatrix(m)      # this error
    SMatrix{3,3}(m) # correct - size is inferrable
```
Is this an expected behavior?

Here is the `versioninfo()` output:
```
julia> versioninfo()
Julia Version 0.5.1-pre+4
Commit 887ad05* (2016-10-28 19:39 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin16.0.0)
  CPU: Intel(R) Core(TM)2 Duo CPU     T9900  @ 3.06GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Penryn)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, penryn)
```

andyferris:
Thanks @wsshin for the report. Comprehensions (and generators) certainly are only partially supported at the moment, and we want to support them better in the future, but it is rather odd that one works and the other doesn't.

c42f:
This seems to be due to a bizarre interaction between type inference with `size`:

```julia
julia> using StaticArrays

julia> vec2 = @SVector([1,2]); vec3 = @SVector([1,2,3])
3-element StaticArrays.SVector{3,Int64}:
 1
 2
[...]

PR #77, opened 2016-11-12
Workaround for comprehensions built from svectors

c42f:
This doesn't address the root cause, but works around the issue reported in #76.


andyferris:
Thanks, Chris, for getting to the bottom of this one :)

PR #82, opened 2016-11-26
Fix fill

### LINKED ISSUE #81: fill() allocates when the number of elements is greater than 14
```julia
@allocated fill(0., SMatrix{1, 14, Float64})
```
results in 0 (after JIT compilation). However,
```julia
@allocated fill(0., SMatrix{1, 15, Float64})
```
results in 128. Sizes greater than 15 also allocate.

I'll open a PR to fix this by going through the `tuple` constructor in a minute.

tkoolen:
Fixes #81. 

I'm actually not 100% sure that using `@allocated` in a test is a good idea (considering various Julia startup switches), but I didn't know how else to test this. Let me know if you have  a better idea. I made sure the test set passes locally with `--code-coverage` at least.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -104,7 +104,7 @@ end
     return quote
         $(Expr(:meta, :inline))
         valT = convert($T, val)
-        SA($(expr...))
+        $(Expr(:call, SA, Expr(:tuple, expr...)))
     end
 end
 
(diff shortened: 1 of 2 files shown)


andyferris:
Thanks for fixing this!

Since the problem is inference bailing out when types are "too complex", you can diagnose this with `@code_warntype` (which gave `Any` as the output for sizes greater than 14).

andyferris:
Not sure what's going on with the 32-bit windows build... seems unrelated though.

PR #83, opened 2016-11-28
change tuple indexing to return tuple, SVector indexing -> SVector

SimonDanisch:
Tests seem to pass.. I guess it'll need a bit more polishing + more tests ;)


test/indexing.jl
```diff
@@ -74,8 +74,8 @@
         m = [1.0 2.0; 3.0 4.0]
 
         @test m[(1,2), (1,2)] === @SMatrix [1.0 2.0; 3.0 4.0]
-        @test m[1, (1,2)] === @SVector [1.0, 2.0]
-        @test m[(1,2), 1] === @SVector [1.0, 3.0]
+        @test m[1, (1,2)] ===  (1.0, 2.0)
+        @test m[(1,2), 1] ===  (1.0, 3.0)
```
c42f: This certainly seems useful and sensible as it stands.  My only question is how the type change here generalizes to more dimensions.  That is, what type should the following return?

```
m[(1,2), (1,2)]
```

Maybe there's no good answer to this / maybe we don't care!  I don't claim to have thought deeply about it.  What do you guys think?

SimonDanisch: Yeah I noted this as a todo ;) I don't really have a good answer, though :( Tuple of tuple would make sense, but who would like to work with that?!
I think in the end the only reasonable thing to do is to not allow it, and force the user to use some static array for indexing.

andyferris:
Thanks a lot for this Simon!

I am really thinking we should drop tuple indexing altogether - for the reason that it doesn't really belong here.

Indexing by a static array to get an `SVector` or whatever makes a lot of sense to me, so this part is great :)

SimonDanisch:
Okay lets do it that way :) btw I'm almost done with the migration... Only a few bugs are left!

andyferris:
That's great. I hope this wasn't too painful?

andyferris:
BTW, while I have you here, I must say I am impressed with the progress in GLVisualize. It makes me want to stop what I'm doing and start to write some visualization stuff from Julia :) 

I also agree with your arguments on Discourse to just implement more UI widgets in GLVisualize. A great majority of what libraries like Qt offer (cross-platform compatibility, a convenient set of libraries and classes, etc) we already get for *free* in Julia, so if you don't need a "native look" then something like GLVisualize should be perfect. (Unfortunately, I'm unlikely to contribute at this stage).

SimonDanisch:
Thanks! :) Guess this means that I have to start sabotaging whatever you're working on right now :P

andyferris:
Bump.

We had discussed removing tuple indexing and then merging this, is that still on the agenda?

I'm actually looking forward to using `SVector`s for indexing... I just came across something today where that made perfect sense.

SimonDanisch:
Yeah lets do this!  I was just coming to the conclusion, that I should start merging everything into StaticArrays, tag it, then merge my GLVisualize branches into master... and then look at Plots.jl, which started using FixedSizeArrays as well -.-

SimonDanisch:
So actually, I have a branch which combines all these changes... Would it be easier to just merge that branch?

SimonDanisch:
I think I merged these changes into sd/fsa. Should we just merge this first and then I rebase sd/fsa?

andyferris:
So, wait, you want to temporarily merge the tuple indexing -> tuple functions?

Maybe we've gotten each other confused: I was asking that this branch delete *all* the tuple indexing functions, and that the indexing with `Tuple` tests change to indexing with `SVector` tests. Is that something you'd like to see happen? IMO the tuple->tuple indexing deserves its own package + a PR to Julia `Base`, rather than living here.

Specifically, I was thinking to delete these lines of code (or, if you feel like it, add deprecation warnings).

https://github.com/JuliaArrays/StaticArrays.jl/blob/sd/index/src/indexing.jl#L5L39

as well as the `Tuple` part of the `Union` here:
 
https://github.com/JuliaArrays/StaticArrays.jl/blob/sd/index/src/indexing.jl#L118

and here: 

https://github.com/JuliaArrays/StaticArrays.jl/blob/sd/index/src/indexing.jl#L145

and change these `setindex!` methods to accept `StaticVector`s not `Tuple`s:

https://github.com/JuliaArrays/StaticArrays.jl/blob/sd/index/src/indexing.jl#L209L293

plus just switch over any tuple indexing tests to using static vectors as indices. 

Sorry, I realize this is a pain!! Especially if you think deprecations are worthwhile. Either way, I'd tag this as 0.2.0 since it is a breaking change, and I don't know who relies on these functions (when I started writing *StaticArrays* I thought I'd be using them internally, but I was wrong...).

SimonDanisch:
I kinda feel like I did that in https://github.com/JuliaArrays/StaticArrays.jl/pull/67. Jeez, I shouldn't have put such a big refactor on hold. I run out of time for these changes, will look at what I did later! ;)

SimonDanisch:
Nope I didn't! :D

andyferris:
I merged #67... what is the status with this one? (They appear to conflict)

andyferris:
Simon, just checking - should this PR still be open?

PR #85, opened 2016-12-02
Fix definition of prod()

umgrmxsvaf:
Seems to just be a typo. :)

This is my first pull request, so hopefully I did it correctly.


KristofferC:
Generally when you fix a bug like this it is always an extra bonus if there is an accompanying test such that the bug doesn't creep back up and to make sure that the fix actually works :).

umgrmxsvaf:
Hmmm, I'll try to figure out such tests go.

umgrmxsvaf:
Sorry about that. ^_^; Thanks for the advice.

c42f:
Great, thanks for the patch, and the tests :-)

PR #86, opened 2016-12-12
Implement versions of maximum(), minimum(), diff() with Val{dim} as second argument

wsshin:
- Implement `reduce{D}(op, a, ::Type{Val{D}})` that takes `Val{D}` indicating the direction of reduction.
- Implement `maximum{D}(a, ::Type{Val{D}})` and `minimum{D}(a, ::Type{Val{D}})` using `reduce{D}(op, a, ::Type{Val{D}})`.
- Implement `diff{D}(a, ::Type{Val{D}})` separately, because it cannot be implemented using `reduce{D}(op, a, ::Type{Val{D}})`.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -90,6 +90,36 @@ end
     end
 end
 
+@generated function reduce{D}(op, a::StaticArray, ::Type{Val{D}})
+    S = size(a)
+    if S[D] == 1
+        return :(return a)
+    else
+        N = ndims(a)
+        Snew = ([n==D ? 1:S[n] for n = 1:N]...)
+        newtype = similar_type(a, Snew)
+
+        exprs = Array{Expr}(Snew)
+        itr = [1:n for n = Snew]
+        for i = Base.product(itr...)
+            ik = copy([i...])
+            ik[D] = 2
+            expr = :(op(a[$(i...)], a[$(ik...)]))
+            for k = 3:S[D]
+                ik[D] = k
+                expr = :(op($expr, a[$(ik...)]))
+            end
+
+            exprs[i...] = expr
+        end
+
+        return quote
+            $(Expr(:meta,:inline))
+            @inbounds return $(Expr(:call, newtype, Expr(:tuple, exprs...)))
+        end
+    end
+end
+
 # These are all similar in Base but not @inline'd
 @inline sum{T}(a::StaticArray{T}) = reduce(+, zero(T), a)
 @inline prod{T}(a::StaticArray{T}) = reduce(*, one(T), a)
@@ -101,8 +131,29 @@ end
 @inline sumabs2{T}(a::StaticArray{T}) = mapreduce(abs2, +, zero(T), a)
 @inline minimum(a::StaticArray) = reduce(min, a) # base has mapreduce(idenity, scalarmin, a)
 @inline maximum(a::StaticArray) = reduce(max, a) # base has mapreduce(idenity, scalarmax, a)
+@inline minimum{D}(a::StaticArray, dim::Type{Val{D}}) = reduce(min, a, dim)
+@inline maximum{D}(a::StaticArray, dim::Type{Val{D}}) = reduce(max, a, dim)
 
+@generated function diff{D}(a::StaticArray, ::Type{Val{D}}=Val{1})
+    S = size(a)
+    N = ndims(a)
+    Snew = ([n==D ? S[n]-1:S[n] for n = 1:N]...)
+    newtype = similar_type(a, Snew)
 
+    exprs = Array{Expr}(Snew)
+    itr = [1:n for n = Snew]
+
+    for i1 = Base.product(itr...)
+        i2 = copy([i1...])
+        i2[D] = i1[D] + 1
+        exprs[i1...] = :(a[$(i2...)] - a[$(i1...)])
+    end
+
+    return quote
+        $(Expr(:meta,:inline))
+        @inbounds return $(Expr(:call, newtype, Expr(:tuple, exprs...)))
+    end
+end
 
 ###############
 ## mapreduce ##
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -90,13 +90,13 @@ end
     end
 end
 
-@generated function reduce{D}(op, a::StaticArray, ::Type{Val{D}})
+@generated function reducedim{D}(op, a::StaticArray, ::Type{Val{D}})
     S = size(a)
     if S[D] == 1
         return :(return a)
     else
         N = ndims(a)
-        Snew = ([n==D ? 1:S[n] for n = 1:N]...)
+        Snew = ([n==D ? 1 : S[n] for n = 1:N]...)
         newtype = similar_type(a, Snew)
 
         exprs = Array{Expr}(Snew)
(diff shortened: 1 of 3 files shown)


andyferris:
Looks good, thanks @wsshin 

Can I bother you to add some unit tests, please? (It's just that I have a habit of breaking things so tests are good ;) )

wsshin:
Just added tests for the newly implemented functions.  Hope this helps!

andyferris:
Thanks for the tests :) They help a lot.

One difficulty in all of this is that `newtype` is chosen with the input's element type, even if `op` returns a different type.

This poses difficulties, for instance if the reduced dimension is length-1 then `op` won't even be called. For StaticArrays, at least this case could be checked statically, but it does seem like a pain...

I'm tempted to ignore this in the interest of sanity and expediency, and just merge this PR. Any contrary opinions? @c42f?

src/mapreduce.jl
```diff
@@ -90,6 +90,36 @@ end
     end
 end
 
+@generated function reduce{D}(op, a::StaticArray, ::Type{Val{D}})
+    S = size(a)
+    if S[D] == 1
+        return :(return a)
+    else
+        N = ndims(a)
+        Snew = ([n==D ? 1:S[n] for n = 1:N]...)
```
c42f: The spacing around the ternary here is pretty confusing, as `1:S[n]` ends up looking like a range.  I'd space that out a bit.

wsshin: Good point.  I will make the change!

src/mapreduce.jl
```diff
@@ -90,6 +90,36 @@ end
     end
 end
 
+@generated function reduce{D}(op, a::StaticArray, ::Type{Val{D}})
```
c42f: Shouldn't this extend `Base.reducedim`?

andyferris: Oh right... I didn't know about that function. Certainly makes sense to extend them instead.

c42f:
I had a quick look.  `Base.reduce` has much the same type inference problem that you have here, so you could potentially reuse some of the functionality from there, eg `r_promote` and `mr_empty`.  As you can see in Base, getting it right is not really an easy task.

However, I don't think there's much harm in merging this without any fancy type inference, as long as the API is consistent with Base.  I assume the fancy stuff could be added later without too much disruption.

andyferris:
Agreed. Just move `reduce` -> `reducedims` and I'll merge this one.

src/mapreduce.jl
wsshin: I didn't know about `Base.reducedim` either.  Thanks for pointing this out!

To mimic `Base.reducedim` more closely, we should be able to pass a vector of dimensions (rather than a single dimension) in which reduction occurs.  This means that we would need to pass a vector of `Val{D}`'s, which requires the function to be parametrized as `reducedim{D1,D2,...,Dn}` for an arbitrary `n`.

Is specifying an arbitrary number of parameters possible in Julia?  Or we just opt with a single dimension?

andyferris: If a single dimensionis is sufficient, that will do for now.

One way of doing this is `reducedim(op, a, Val{(D1,D2,...)})` so later we can check if `D` is an integer or a tuple and expand the tuple. Alternatively they could be applied in sequence, like `reducedim(op, a, Val{D1}, Val{D2}, ...)`. 

I don't think this is necessary for this PR however.

wsshin:
Done!  Thanks for all the comments.  This was my first PR ever, and also the first time using `@generated`.  Learned a lot through this experience.  Hope to have more chances to contribute to this project in the future!

andyferris:
Thanks @wsshin we appreciate the effort and glad you found it worthwhile. Contributions are always welcome :)

wsshin:
Just found another instance of the confusing spacing around the ternary operator that @c42f pointed out, and fixed it, but this PR is already merged...  Is it OK to generate a new PR at this point?

c42f:
Yes, no problem at all - if you've got a small tweak to make, feel free to open a new PR

c42f:
This is interesting, check out the code difference between the fastmath version and the normal `max`:

```julia
A = rand(SMatrix{3,3})

@code_native reducedim(max, A,Val{2})
# vs
@code_native reducedim((x,y)->(@fastmath max(x,y)), A,Val{2})
```

The assembly of the non-fastmath version is a mess, presumably due to a lot of careful NaN handling :-/

wsshin:
Interesting...  I find that we have a similar result from `reduce(max, A)` vs. `reduce(@fastmath(max), A)` as well.

KristofferC:
Just want to chime in that looking at assembly code can be useful but, as always, benchmarking is king.

wsshin:
@c42f See https://github.com/JuliaArrays/StaticArrays.jl/pull/87 for the change from `a ? b:c` to `a ? b : c`.

c42f:
@KristofferC very true. Peeking at the assembly can be interesting, but I don't want to suggest that it can replace actual benchmarking with a realistic workload on the target architecture.

c42f:
In the spirit of avoiding misinformation, I went ahead and benchmarked the Base implementation of `reducedim`, which should have similar issues regarding `NaN` correctness.

```
julia> A = rand(1000,1000);

julia> using BenchmarkTools

julia> t1 = @benchmark minimum(A, 1)
BenchmarkTools.Trial:
  memory estimate:  8.17 kb
  allocs estimate:  9
  --------------
  minimum time:     2.978 ms (0.00% GC)
  median time:      3.020 ms (0.00% GC)
  mean time:        3.032 ms (0.00% GC)
  maximum time:     3.406 ms (0.00% GC)
  --------------
  samples:          1648
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> t2 = @benchmark reducedim(min, A, 1, typemax(eltype(A)))
BenchmarkTools.Trial:
  memory estimate:  8.17 kb
  allocs estimate:  9
  --------------
  minimum time:     2.980 ms (0.00% GC)
  median time:      3.004 ms (0.00% GC)
  mean time:        3.022 ms (0.00% GC)
  maximum time:     3.426 ms (0.00% GC)
  --------------
  samples:          1654
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> t3 = @benchmark reducedim(@fastmath(min), A, 1, typemax(eltype(A)))
BenchmarkTools.Trial:
  memory estimate:  8.17 kb
  allocs estimate:  9
  --------------
  minimum time:     829.793 μs (0.00% GC)
  median time:      842.073 μs (0.00% GC)
  mean time:        850.967 μs (0.00% GC)
  maximum time:     1.100 ms (0.00% GC)
  --------------
  samples:          5861
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> judge(median(t2), median(t1))
BenchmarkTools.TrialJudgement:
  time:   -0.08% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

julia> judge(median(t3), median(t1))
BenchmarkTools.TrialJudgement:
  time:   -71.99% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
```

So in this particular case, the feeling of "that looks weird" from the assembly was correct, and the version without NaN handling is a lot faster.

PR #90, opened 2016-12-23
Only define broadcasting operators for Julia versions predating

andreasnoack:
automatic rewriting to `broadcast`


SimonDanisch:
looks good to me!

c42f:
Yes, this looks great, thanks.

c42f:
Eek, it looks like the generated code for `StaticArrays` dotted operators is super inefficient on current julia-0.6.  That's not much to do with the change here, mind you.

andreasnoack:
@c42f Not necessarily. It was pointed out to me [here](https://github.com/JuliaLang/julia/issues/19706) that this should probably be analyzed differently. Did wrap the expression in a function before calling `@code_llvm`? 

```jl
@code_llvm A .* 2.0
```
looks terrible, but
```jl
f(x) = x .* 2.0
@code_llvm f(A)
```
looks similar on 0.5 and master.

c42f:
Yes, I did wrap it in a function.  However, getting the latest julia-0.6 master seems to have fixed the problem.  I guess something about my local build had gotten into an "interesting" state, or there was a transient codegen issue on master.  Never mind, nothing to see here!

andreasnoack:
Good news. Would it be okay for you to tag a patch release?

c42f:
Sure.  In the process of fiddling with this I also noticed #91, I'll get that one into the release as well.

c42f:
Ok, I'm trying out @attobot, let's see what happens.

https://github.com/JuliaArrays/StaticArrays.jl/releases/tag/v0.1.3

PR #92, opened 2016-12-30
Bugfix: Avoid emitting Expr(:meta, :inline) in a macro (!)

c42f:
Emitting the inline meta inside a macro causes unsuspecting user
functions to become inline, leading to surprises in type stability.

@andyferris it's all those generated functions, they've made you :meta :inline trigger happy :-P

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -209,7 +209,6 @@ macro MArray(ex)
         end
 
         return quote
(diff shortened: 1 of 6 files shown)


andyferris:
Lol. Oops. I blame late nights and excessive copy-paste!

c42f:
Heh :-)

PR #96, opened 2017-01-24
Update to work on Julia 0.6

andreasnoack:
Define `similar_type` per type to avoid hooking into type internals.

Restructure code a bit to avoid old world problems

Disable FixedSizeArrays since current approach is not feasible on Julia 0.6

This is a minimal fix for the package on Julia 0.6. The fix of 265 requires that all methods used in generated functions are defined before generated functions that use the methods. Consequently, I've restructured the code a bit such that some of the generated functions are defined later. I guess the way generated functions are used could makes it harder to support user defined functions since they are defined later than the generated functions but the tests for `FieldVector`s pass.

### DIFF
--- a/src/FieldVector.jl
+++ b/src/FieldVector.jl
@@ -21,10 +21,10 @@ abstract FieldVector{T} <: StaticVector{T}
 # Is this a good idea?? Should people just define constructors that accept tuples?
 @inline (::Type{FV}){FV<:FieldVector}(x::Tuple) = FV(x...)
 
-@pure size{FV<:FieldVector}(::FV) = (length(FV.types),)
-@pure size{FV<:FieldVector}(::Type{FV}) = (length(FV.types),)
-@pure length{FV<:FieldVector}(::FV) = size(FV)[1]
-@pure length{FV<:FieldVector}(::Type{FV}) = size(FV)[1]
+@pure length{FV<:FieldVector}(::FV) = nfields(FV)
+@pure length{FV<:FieldVector}(::Type{FV}) = nfields(FV)
+@pure size{FV<:FieldVector}(::FV) = (length(FV),)
+@pure size{FV<:FieldVector}(::Type{FV}) = (length(FV),)
 
 @inline getindex(v::FieldVector, i::Integer) = getfield(v, i)
 @inline setindex!(v::FieldVector, x, i::Integer) = setfield!(v, i, x)
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -37,8 +37,8 @@ include("MMatrix.jl")
 include("MArray.jl")
 include("SizedArray.jl")
 
-include("indexing.jl")
 include("abstractarray.jl")
+include("indexing.jl")
 include("mapreduce.jl")
 include("arraymath.jl")
 include("linalg.jl")
@@ -50,7 +50,9 @@ include("inv.jl")
 include("eigen.jl")
 include("cholesky.jl")
 
-include("FixedSizeArrays.jl")
+if VERSION < v"0.6.0-dev.1671"
+    include("FixedSizeArrays.jl")
+end
 include("ImmutableArrays.jl")
 
 # TODO list
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ notifications:
   email: false
 matrix:
   allow_failures:
-    - julia: nightly
+    # - julia: nightly
     - os: osx
 # uncomment the following lines to override the default test script
 #script:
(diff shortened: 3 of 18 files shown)


src/FieldVector.jl
```diff
@@ -21,10 +21,10 @@ abstract FieldVector{T} <: StaticVector{T}
 # Is this a good idea?? Should people just define constructors that accept tuples?
 @inline (::Type{FV}){FV<:FieldVector}(x::Tuple) = FV(x...)
 
-@pure size{FV<:FieldVector}(::FV) = (length(FV.types),)
-@pure size{FV<:FieldVector}(::Type{FV}) = (length(FV.types),)
-@pure length{FV<:FieldVector}(::FV) = size(FV)[1]
-@pure length{FV<:FieldVector}(::Type{FV}) = size(FV)[1]
+@pure length{FV<:FieldVector}(::FV) = nfields(FV)
```
andyferris: Is `nfields` effectively `@pure` anyway?

andyferris:
You beat me to it :) Thanks.

I might need a bit of time to review at the moment, sorry.

src/FieldVector.jl
andreasnoack: I would think so but I'm never sure about these things.

src/indexing.jl
```diff
@@ -95,7 +95,7 @@ end
     end
 end
 
-# MAYBE: fixed-size indexing with bools? They would have to be Val's...
+# MAYBE: fixed-wx indexing with bools? They would have to be Val's...
```
andyferris: what is `wx`?

andreasnoack: A typo. I thought I had force pushed to change that back. Let me check my branch.

andreasnoack:
No problem. We use the package for a project where we have to use master so I had to do these but for now, we can just use my fork.

I think the main/only controversial change is the `similar_type` which would now have to be defined for each type. It is not that many definitions, though. At least not make the tests pass. However, I think it would be a good idea to avoid hooking into the type internals.

src/FieldVector.jl
andyferris: Oh good, it's not just me! :)

src/indexing.jl
andreasnoack: I hadn't committed the last changes. I've just pushed the version without `wx`. It is almost identical to the last version.

SimonDanisch:
I think that's the way we wanted to go anyways ;)

SimonDanisch:
@andyferris is it okay if I go ahead and review & merge this?

andyferris:
Yes please, Simon, I did have a bit of a look and it all seemed good to me. Sorry Andreas - its been a bit hectic for me lately.

Keno:
```
ERROR: LoadError: MethodError: convert(::Type{StaticArrays.SVector{2,Float64}}, ::StaticArrays.SVector{2,Float64}) is ambiguous. Candidates:
  convert{SA<:StaticArrays.StaticArray}(::Type{SA}, sa::StaticArrays.StaticArray) in StaticArrays at /global/cscratch1/sd/kfischer/.julia/v0.6/StaticArrays/src/core.jl:69
  convert{SA<:StaticArrays.StaticArray}(::Type{SA}, a::AbstractArray) in StaticArrays at /global/cscratch1/sd/kfischer/.julia/v0.6/StaticArrays/src/core.jl:72
  convert{SA<:StaticArrays.StaticArray}(::Type{SA}, x1) in StaticArrays at /global/cscratch1/sd/kfischer/.julia/v0.6/StaticArrays/src/core.jl:66
  convert{T}(::Type{T}, x::T) in Base at essentials.jl:29
```

src/StaticArrays.jl
```diff
@@ -50,7 +50,7 @@ include("inv.jl")
 include("eigen.jl")
 include("cholesky.jl")
 
-include("FixedSizeArrays.jl")
+# include("FixedSizeArrays.jl")
```
andyferris: @SimonDanisch, what is your opinion about this?

There seems to be some modifications to `FixedSizeArrays.jl`. Would some more modification "rescue" this feature? 

Do many people use the FixedSizeArrays feature?

SimonDanisch: Sorry, @andreasnoack and i got a bit sidetracked... My only complaint was the `where`, since its not 0.5 compatible... I was fine with just removing the whole thing for now. Porting geometrytypes ( which uses this) to 0.6 will be a bit of work anyways... Lets just seperate it!

andreasnoack:
The problem that caused the inference explosion has just been fixed on master so when nightlies are up, we can restart the tests here.

c42f:
I restarted the travis build - seems to have worked this time.

SimonDanisch:
Sweet. I will go ahead to merge this, since the breaking stuff is mostly my problem ;)

PR #99, opened 2017-02-09
RFC: add colwise function

tkoolen:
I needed something like this in my own package, and thought it might be useful for others as well.

Note that this is not the same as `broadcast`, since `broadcast` expects a function that takes a number of scalars as its arguments, whereas `colwise` expects a function that takes two vectors as its arguments.

I can add a `rowwise` too if you like.

This is inspired by Eigen's [`.colwise()`](http://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#a7e5a0184a798ffb18c586c3ca43fde68) functionality.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -22,6 +22,7 @@ export @SVector, @SMatrix, @SArray
 export @MVector, @MMatrix, @MArray
 
 export similar_type, setindex
+export colwise
 
 include("util.jl")
 
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -428,3 +428,24 @@ end
         end
     end
 end
+
+#############
+## colwise ##
+#############
+@generated function colwise(f, vec::StaticVector, mat::StaticArray)
+    length(vec) == size(mat, 1) || throw(DimensionMismatch())
+    exprs = [:(f(vec, mat[:, $j])) for j = 1:size(mat, 2)]
+    return quote
+        $(Expr(:meta, :inline))
+        @inbounds return $(Expr(:call, hcat, exprs...))
+    end
+end
+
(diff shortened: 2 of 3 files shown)


c42f:
I think this is more or less a specific version of the functionality exposed by `Base.mapslices`, so we could extend `mapslices` instead.  It looks like the equivalence would be

```julia
colwise(f, v, m)
# vs
mapslices(c->f(c, v), m, Val{2})
```

The second thing here is definitely a bit ugly as written in this particular case, but it's quite general and is consistent with `Base`.  Thoughts?

andyferris:
We generally haven't implemented many new functions that don't exist in `Base`. Have you considered a PR there? This operation would make sense for non-static arrays also.

How does everyone else feel about this? The question is, does this function belong somewhere else, or is it OK to start accepting new functionality here?

PS this is a yucky form of broadcast if you consider `mat` as a svector of svectors, something along the lines of:
```
tmp1 = reinterpret(SVector{3,SVector{3, Float64}}, mat)
tmp2 = f.(Scalar(vec),mat)
reinterpret(SMatrix{3,3,Float64,9}, tmp2)
```
This is a bit disgusting, though :) But in general, using `Scalar` is a useful trick so that `broadcast` acts on elements which are `SVector`s.

andyferris:
I would be happy with implementing specializations of `mapslices` and so-on.

tkoolen:
I wasn't even aware of the existence of `mapslices`. Specializing that in `StaticArrays` sounds like the way to go, yes. I'll close this PR. Thanks for your help!

c42f:
TBH I had to dig `mapslices` out of the documentation ;-) I had a vague feeling something was there but I'd forgotten the details. Somewhat related are `reducedim` and `mapreducedim`.

PR #102, opened 2017-02-13
Remove indexing of AbstractArray with Tuple

### LINKED ISSUE #101: "Tuple indexing into AbstractArray" methods are not okay
You need to remove these immediately, or make them specific to your own types. Packages should not change the behavior of all other array code when they're imported. If I had seen these sooner I wouldn't have allowed this package to be registered.

c42f:
Two pirates walk into a bar dragging a heavy chest.
As they open it the barman reels back in horror, glimpsing the unchanging rectangular pattern inside.
"Get out you filthy scum, we don't serve your type of pirates here"

Resolves #101

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -13,31 +13,6 @@
     end
 end
 
-# Tuple indexing into AbstractArray. TODO, move into base
-@generated function getindex{S}(a::AbstractArray, inds::NTuple{S,Integer})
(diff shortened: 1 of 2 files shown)


SimonDanisch:
excellent :) I'll go ahead and close: https://github.com/JuliaArrays/StaticArrays.jl/pull/83#issuecomment-279268551

PR #104, opened 2017-02-13
fix #103

### LINKED ISSUE #103: Size bug in + (introduced in 0.1.5)
``` julia
julia> a = rand(SMatrix{1, 2})
1×2 StaticArrays.SMatrix{1,2,Float64,2}:
 0.964606  0.400012

julia> b = rand(SMatrix{1, 2})
1×2 StaticArrays.SMatrix{1,2,Float64,2}:
 0.134356  0.179786

julia> c = a + b
2×1 StaticArrays.SMatrix{2,1,Float64,2}:
 1.09896 
 0.579798
```

SimonDanisch:


test/mapreduce.jl
```diff
@@ -72,5 +72,7 @@
 
         broadcast!(+, mm, v1, M)
         @test mm == @MMatrix [3 4; 7 8; 11 12; 15 16]
+        # issue #103
+        @test map(+, M, M) == @SMatrix [2 4; 6 8; 10 12; 14 16]
```
tkoolen: I would change the right hand side to just `[2 4; 6 8; 10 12; 14 16]` (without the `@SMatrix`) to catch mistakes where `StaticArrays` is internally consistent but still wrong.

tkoolen:
Thanks for the quick response!

PR #105, opened 2017-02-15
Rewrite all `@pure` functions

andyferris:
* Now we use `Size` more consistently and rely less on `@pure` implementations of `similar_type`.

 * `similar` and `similar_type` both use `Size` not integers to discuss size

 * `similar` and `similar_type` are now streamlined. Users only need to
   override one definition, just like in Base.

 * Associated documentation and test changes.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -94,12 +94,10 @@ end
 ## MArray methods ##
 ####################
 
-similar_type{M,T,N,L,S}(::Type{MArray{M,T,N,L}}, ::Type{S}) = MArray{M,S,N,L}
-
-@pure size{Size}(::Type{MArray{Size}}) = Size
-@pure size{Size,T}(::Type{MArray{Size,T}}) = Size
-@pure size{Size,T,N}(::Type{MArray{Size,T,N}}) = Size
-@pure size{Size,T,N,L}(::Type{MArray{Size,T,N,L}}) = Size
+@pure Size{S}(::Type{MArray{S}}) = Size(S)
+@pure Size{S,T}(::Type{MArray{S,T}}) = Size(S)
+@pure Size{S,T,N}(::Type{MArray{S,T,N}}) = Size(S)
+@pure Size{S,T,N,L}(::Type{MArray{S,T,N,L}}) = Size(S)
 
 function getindex(v::MArray, i::Integer)
     Base.@_inline_meta
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -1,15 +1,13 @@
 typealias StaticScalar{T} StaticArray{T,0}
 
-@pure length{T<:StaticArray}(a::T) = prod(size(a))
-@pure length{T<:StaticArray}(a::Type{T}) = prod(size(a))
-@pure length{T<:StaticScalar}(a::T) = 1
-@pure length{T<:StaticScalar}(a::Type{T}) = 1
+length{T<:StaticArray}(a::T) = prod(Size(T))
+length{T<:StaticArray}(a::Type{T}) = prod(Size(T))
 
-@pure size{T<:StaticArray}(::T, d::Integer) = size(T, d)
-@pure function size{T<:StaticArray}(a::Type{T}, d::Integer)
-    s = size(a)
-    return d <= length(s) ? s[d] : 1
-end
+size{T<:StaticArray}(::T) = get(Size(T))
+size{T<:StaticArray}(::Type{T}) = get(Size(T))
+
+size{T<:StaticArray}(::T, d::Integer) = Size(T)[d]
+size{T<:StaticArray}(::Type{T}, d::Integer) = Size(T)[d]
 
 # This has to be defined after length and size because it is generated
 @generated function convert{SA<:StaticArray}(::Type{SA}, a::AbstractArray)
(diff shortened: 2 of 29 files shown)


andyferris:
This was a little tricky to get inference working well on v0.5. I had to use `@generated` in two places where on v0.6- `@pure` was sufficient - so it's good that the compiler is improving :)

But in general I think this reorganization is a very good thing.

andyferris:
Test failure is caused by https://github.com/JuliaLang/julia/issues/20620 on Julia master (v0.5 works).

(Ideally we would deal with this more fully but seems orthogonal to this PR).

src/MArray.jl
```diff
@@ -94,12 +94,10 @@ end
...
-similar_type{M,T,N,L,S}(::Type{MArray{M,T,N,L}}, ::Type{S}) = MArray{M,S,N,L}
-
-@pure size{Size}(::Type{MArray{Size}}) = Size
-@pure size{Size,T}(::Type{MArray{Size,T}}) = Size
-@pure size{Size,T,N}(::Type{MArray{Size,T,N}}) = Size
-@pure size{Size,T,N,L}(::Type{MArray{Size,T,N,L}}) = Size
+@pure Size{S}(::Type{MArray{S}}) = Size(S)
+@pure Size{S,T}(::Type{MArray{S,T}}) = Size(S)
+@pure Size{S,T,N}(::Type{MArray{S,T,N}}) = Size(S)
+@pure Size{S,T,N,L}(::Type{MArray{S,T,N,L}}) = Size(S)
```
c42f: Do you know whether this kind of repetition go away with the new type system work in 0.6?

src/abstractarray.jl
```diff
@@ -39,169 +37,88 @@ Base.linearindexing{T<:StaticArray}(::Type{T}) = Base.LinearFast()
...
+size (or else a dynamically sized array will be returned).
+"""
+similar{SA<:StaticArray}(::SA) = similar(SA,eltype(SA))
+similar{SA<:StaticArray}(::Type{SA}) = similar(SA,eltype(SA))
 
-# And also similar() returning mutable StaticArrays
-@inline similar{SV <: StaticVector}(::SV) = MVector{length(SV),eltype(SV)}()
-@inline similar{SV <: StaticVector, T}(::SV, ::Type{T}) = MVector{length(SV),T}()
+similar{SA<:StaticArray,T}(::SA,::Type{T}) = similar(SA,T,Size(SA))
+similar{SA<:StaticArray,T}(::Type{SA},::Type{T}) = similar(SA,T,Size(SA))
```
c42f: So now we've got `similar` based on input `StaticArray` type as well as value.  I guess that makes sense as they all still return a value which can be written into. Did did we have it before anywhere?

src/core.jl
```diff
@@ -9,30 +9,23 @@
...
+  type (or type constructor) that accepts a flat tuple of data.
 
-For mutable containers you may need to define the following:
+For mutable containers you may also need to define the following:
 
- - `similar()` returns a *mutable* container of similar type, defaults to using
-   a `MVector`, `MMatrix` or `MArray`. You will benefit from overloading this if
-   your container is already mutable!
- - `setindex!` on mutable arrays (e.g. those returned by `similar`)
+ - `setindex!` for a single elmenent (linear indexing).
```
c42f: elmenent

src/deque.jl
```diff
@@ -1,7 +1,5 @@
-export push, pop, shift, unshift, insert, deleteat, setindex
-
 @generated function push(vec::StaticVector, x)
-    newtype = similar_type(vec, (length(vec) + 1 ,))
+    newtype = similar_type(vec, Size(length(vec) + 1 ,))
```
c42f: Minor tidy up: trailing comma no longer required here and in the rest of `deque.jl`.

src/linalg.jl
```diff
@@ -54,7 +54,7 @@ end
         error("Dimension mismatch")
     end
 
-    exprs = [i == j ? :(a.λ + b[$(sub2ind(size(a), i, j))]) : :(b[$(sub2ind(size(a), i, j))]) for i = 1:n, j = 1:n]
+    exprs = [i == j ? :(a.λ + b[$(sub2ind(size(b), i, j))]) : :(b[$(sub2ind(size(b), i, j))]) for i = 1:n, j = 1:n]
```
c42f: I guess this never worked before?

test/FieldVector.jl
```diff
@@ -18,19 +18,21 @@
         @test length(p) === 3
         @test eltype(p) === Float64
 
+        @test (p + p)::Point3D ≈ Point3D(2.0, 4.0, 6.0)
```
c42f: Can use `===` here, I think, to check that both the elements and type are the same.  (Elements should sum exactly, as the result is exactly representable in `Float64`.)

test/linalg.jl
```diff
@@ -5,69 +5,101 @@
...
+        @test @inferred(transpose(@SVector([1, 2, 3]))) === @SMatrix([1 2 3])
+        @test @inferred(transpose(@SMatrix([1 2; 0 3]))) === @SMatrix([1 0; 2 3])
+        @test @inferred(transpose(@SMatrix([1 2 3; 4 5 6]))) === @SMatrix([1 4; 2 5; 3 6])
 
-        @test @SVector([1, 2, 3])' === @SMatrix([1 2 3])
-        @test @SMatrix([1 2; 0 3])' === @SMatrix([1 0; 2 3])
-        @test @SMatrix([1 2 3; 4 5 6])' === @SMatrix([1 4; 2 5; 3 6])
+        @test @inferred(ctranspose(@SVector([1, 2, 3]))) === @SMatrix([1 2 3])
+        @test @inferred(ctranspose(@SMatrix([1 2; 0 3]))) === @SMatrix([1 0; 2 3])
+        @test @inferred(ctranspose(@SMatrix([1 2 3; 4 5 6]))) === @SMatrix([1 4; 2 5; 3 6])
```
c42f: That's a lot of `@inferred`.  You need a custom `@testinferred` macro or some kind of helper to make these easier to look at ;-)

test/linalg.jl
```diff
@@ -82,4 +114,8 @@
 
         @test normalize(SVector(1,2,3)) ≈ normalize([1,2,3])
     end
+
+    @testset "trace" begin
+        @test trace(@SMatrix [1.0 2.0; 3.0 4.0]) ≈ 5.0
```
c42f: Can be exact equality, not \approx

c42f:
I assume the CI failures in latest 0.6 are expected for now?

src/MArray.jl
andyferris: I think `Size(::Type{T} where T <: MArray{S}) where S = Size(S)` might be sufficient for all of the above, but I'd like to check. We could probably successfully move the `@pure` to the `Size` constructor and drop it from here, also.

src/abstractarray.jl
andyferris: I think so (or we should have) but it was all a bit dodgier...

src/linalg.jl
andyferris: No... we need to work to improve test coverage.

test/FieldVector.jl
andyferris: I never know when to trust floating point arithmetic

test/linalg.jl
andyferris: :)

test/FieldVector.jl
c42f: The rules say that basic floating point arithmetic operations should be correctly rounded.  In particular, this means that if *all intermediate results* of a computation are exactly representable, then the final result is exact.

This certainly holds true when you're adding, subtracting and multiplying exact integers using floating point arithmetic, up until `maxintfloat(Float64) == 9.007199254740992e15`.

The reason I bring this up is that `isapprox` uses quite a generous default for `rtol`, a lot looser than you'd really want in many tests.

Issue #106, opened 2017-02-16
Dispatch issues with size() on v0.6-dev


rdeits:
I think this is probably a Julia bug, but since I encountered it here first, I wanted to check with you all before I made noise over at Julia itself. 

The following works in Julia v0.5:

```juila
julia> using StaticArrays

julia> immutable Foo{N, T} <: StaticVector{T}
              data::NTuple{N, T}
          end

julia> Base.size{N, T}(::Type{Foo{N, T}}) = (N,)

julia> Base.getindex(f::Foo, i) = f.data[i]

julia> f = Foo(1,2,3)
3-element Foo{3,Int64}:
 1
 2
 3
```

Running the same code on v0.6-dev gives:

```julia
julia> versioninfo()
Julia Version 0.6.0-dev.2831
Commit cb1aae9 (2017-02-15 11:49 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin15.6.0)
  CPU: Intel(R) Core(TM) i7-2860QM CPU @ 2.50GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, sandybridge)

julia> using StaticArrays

julia> immutable Foo{N, T} <: StaticVector{T}
              data::NTuple{N, T}
          end

julia> Base.size{N, T}(::Type{Foo{N, T}}) = (N,)

julia> Base.getindex(f::Foo, i) = f.data[i]

julia> f = Foo(1,2,3)
3-element Foo{3,Int64}:
Error showing value of type Foo{3,Int64}:
ERROR: The size of type `Foo{3,Int64}` is not known.

If you were trying to construct (or `convert` to) a `StaticArray` you
may need to add the size explicitly as a type parameter so its size is
inferrable to the Julia compiler (or performance would be terrible). For
example, you might try

    m = zeros(3,3)
    SMatrix(m)      # this error
    SMatrix{3,3}(m) # correct - size is inferrable

Stacktrace:
 [1] size(::Type{Foo{3,Int64}}) at /Users/rdeits/.julia/v0.6/StaticArrays/src/core.jl:126
[...]

PR #107, opened 2017-02-17
Use new inner constructor and type declaration syntax

andreasnoack:
The new syntax for inner constructors doesn't parse on 0.5 so it will be ugly and more demanding to support on both versions. Furthermore, 0.6 will be out soon and nobody will care about 0.5. If they care anyway and we add features then we can backport those features to a 0.5 branch.

### DIFF
--- a/src/SMatrix.jl
+++ b/src/SMatrix.jl
@@ -66,7 +66,7 @@ end
         SMatrix{S1, S2, $T, L}(x)
     end
 end
-SMatrixNoType{S1, S2, L, T} = SMatrix{S1, S2, T, L}
+@compat SMatrixNoType{S1, S2, L, T} = SMatrix{S1, S2, T, L}
 @generated function (::Type{SMatrixNoType{S1, S2, L}}){S1,S2,L}(x::NTuple{L,Any})
     T = promote_tuple_eltype(x)
     return quote
--- a/src/SMatrix.jl
+++ b/src/SMatrix.jl
@@ -17,14 +17,14 @@ unknown to the compiler (the element type may optionally also be specified).
 immutable SMatrix{S1, S2, T, L} <: StaticMatrix{T}
     data::NTuple{L, T}
 
-    function SMatrix{S1,S2,T,L}(d::NTuple{L,T}) where {S1,S2,T,L}
+    function (::Type{SMatrix{S1,S2,T,L}}){S1,S2,T,L}(d::NTuple{L,T})
         check_smatrix_params(Val{S1}, Val{S2}, T, Val{L})
-        new(d)
+        new{S1,S2,T,L}(d)
     end
 
-    function SMatrix{S1,S2,T,L}(d::NTuple{L,Any}) where {S1,S2,T,L}
+    function (::Type{SMatrix{S1,S2,T,L}}){S1,S2,T,L}(d::NTuple{L,Any})
         check_smatrix_params(Val{S1}, Val{S2}, T, Val{L})
-        new(convert_ntuple(T, d))
+        new{S1,S2,T,L}(convert_ntuple(T, d))
     end
 end
 
--- a/src/SMatrix.jl
+++ b/src/SMatrix.jl
@@ -17,12 +17,12 @@ unknown to the compiler (the element type may optionally also be specified).
 immutable SMatrix{S1, S2, T, L} <: StaticMatrix{T}
     data::NTuple{L, T}
 
-    function SMatrix(d::NTuple{L,T})
+    function SMatrix{S1,S2,T,L}(d::NTuple{L,T}) where {S1,S2,T,L}
         check_smatrix_params(Val{S1}, Val{S2}, T, Val{L})
         new(d)
     end
 
-    function SMatrix(d::NTuple{L,Any})
+    function SMatrix{S1,S2,T,L}(d::NTuple{L,Any}) where {S1,S2,T,L}
         check_smatrix_params(Val{S1}, Val{S2}, T, Val{L})
         new(convert_ntuple(T, d))
     end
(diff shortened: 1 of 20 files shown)


andyferris:
LGTM.  Thanks Andreas. 

Just so you know, I was planning on merging and releasing #105 first, before merging this one. This plus major changes due to #106 are required before this package is usable on v0.6. Those other changes are backward compatible so maybe they should be released for v0.5 too (otherwise it would block backporting). 

Would a `@static if VERSION < ...` version of this PR be a sensible approach? Or is it better all-round to focus development on a single version of Julia?

andyferris:
CI is complaint about `TupleN`

andreasnoack:
The problem is that the syntax doesn't parse at all so we'll end up with a lot of definitions like https://github.com/Evizero/UnicodePlots.jl/blob/master/src/graphics/bargraphics.jl#L1-L44 which I think is ugly. Personally, I think it is better to focus development hours on 0.6 and thereafter instead of compatibility.

There is a with the build bots so CI is running with old binaries that don't have the latest syntax changes. Tests pass locally on latest master.

c42f:
I think merging this might be a bit premature - are the change here required to make the package work on 0.6, or are they syntactic niceties?  If possible it would be nice for master to continue working with 0.5 for a while.

andreasnoack:
The inner constructor syntax is deprecated so it gives warnings on master. The only way to make it work across 0.5 and 0.6 is to write that code as strings and use `include_string` conditionally on the version such as in the link I posted above.

timholy:
There is always the `(::Type{MyNewType{T}}){T}(arg) = new{T}(arg)` syntax, which works on both.

andreasnoack:
Thanks for pointing that out. I wasn't aware of that. I'll update the PR.

andyferris:
Thanks Andreas (and good idea, Tim).

I still think it would be better to delay this for just the moment and first address #106, which completely and utterly destroys *any* custom subtypes of `StaticArray`, is a much larger amount of work, and the fix is compatible with Julia v0.5.

At that point we can base development on v0.6 (and backport crucial fixes to v0.5 - but not fixing #106 first will make any backports less trivial). This will let us (in no particular order):

 * Not require `Compat`.
 * Make use of `where` syntax to simplify dispatch (e.g. [this](https://github.com/JuliaArrays/StaticArrays.jl/pull/105#discussion_r101935226)).
 * Let us take advantage of compiler (specifically, inference and constant propagation) improvements (this will mean, less `@generated` functions).
 * Keep code cleaner.

Basically, this package is unusually low-level and reliant on certain compiler behavior - I'd rather not have this package be artificially limited by continuing to support two versions of Julia.

I do apologize about the deprecation warning noise between now and then - I would suggest to keep rebasing this branch on the other changes for the next little while, and use this branch for downstream development work targeting v0.6 in the meantime.

I am also open to dropping v0.5 support *now* if people think this is a better route (how many backports will there really be?). But please, do keep in mind how completely broken custom static arrays are - deprecation warnings might be a *good* warning-sign for users :)

andreasnoack:
I don't really understand why merging these small deprecation warning fixes could be a problem for fixing custom static arrays. Notice, that the PR doesn't drop 0.5 anymore.

andyferris:
I suppose it should be OK now that I'm tagging the `similar_type` changes. I'll be ripping out `Compat` sometime soon when we go to v0.6-only.

@andreasnoack this PR now has conflicts (my fault)

c42f:
Nice, thanks for putting up with 0.5 for a little longer :-)

test/fixed_size_arrays.jl
```diff
@@ -8,15 +8,15 @@ import StaticArrays.FixedSizeArrays: @fixed_vector
...
-typealias Vec1d Vec{1, Float64}
-typealias Vec2d Vec{2, Float64}
-typealias Vec3d Vec{3, Float64}
-typealias Vec4d Vec{4, Float64}
-typealias Vec3f Vec{3, Float32}
-
-typealias Mat2d Mat{2,2, Float64, 4}
-typealias Mat3d Mat{3,3, Float64, 9}
-typealias Mat4d Mat{4,4, Float64, 16}
+@compat const Vec1d = Vec{1, Float64}
```
andyferris: The sometimes `const` and sometimes not syntax isn't really great (not your fault, mind you).

c42f: Ah yes, I noticed this syntax quirk too.  Hard to know what to do about it.  In packages, it would be nice if all bindings at global scope were implicitly const by default.  But for simple scripting that would be fairly horrible.

andyferris: Could be a special property of `Main`? (Or some property that could be set on a module-by-module basis).

andyferris:
Travis errors are *odd*...

src/SMatrix.jl
```diff
@@ -66,7 +66,7 @@ end
         SMatrix{S1, S2, $T, L}(x)
     end
 end
-typealias SMatrixNoType{S1, S2, L, T} SMatrix{S1, S2, T, L}
+@compat SMatrixNoType{S1, S2, L, T} = SMatrix{S1, S2, T, L}
```
andyferris: I don't know why, but I don't remember this existing, which is why I couldn't fathom the test failure... In theory, `MMatrix` should be symmetric. But don't worry about this for this PR.

andreasnoack:
On 0.6, there seems to be an issue with `@inferred` and dot broadcast syntax. Looking into it now.

andyferris:
I made an issue for that

andyferris:
https://github.com/JuliaLang/julia/issues/20620

andreasnoack:
Oh, I missed that one. I think there is an easy fix.

andyferris:
I just merged this, Andreas. I still plan to remove *Compat* and v0.5 support sometime soonish.

c42f:
Yes, let's make a release before we break everything with #113 (I guess this will end up dropping 0.5 support... honestly I was hoping to last a little longer!)

andyferris:
I was wondering who the target audience was? There is no functional changes for v0.5 users and v0.6 users will be using quite a different codebase soon... @andreasnoack is using `master` for v0.6 development reasonable for the time-being?

andreasnoack:
Using a release is slightly simpler in a larger group of collaborators but the main reason is Travis testing where it is pretty inconvenient to use master.

andyferris:
Yes, Travis. That makes sense.

PR #115, opened 2017-03-15
fall back to constructor in convert

SimonDanisch:
... This removes the annoyance, that e.g. `convert(SVector{3, Float64}, (1.0, 1.0, 1.0))` fails with an error message which doesn't make sense!

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -36,7 +36,7 @@ For mutable containers you may also need to define the following:
 # Generic case, with least 2 inputs
 @inline (::Type{SA}){SA<:StaticArray}(x1,x2,xs...) = SA((x1,x2,xs...))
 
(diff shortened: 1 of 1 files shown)


andyferris:
Can this lead to recursions?

SimonDanisch:
Likely :-P

andyferris:
lol... I see how it is... :)

SimonDanisch:
I guess we should be fine if no convert is called in any constructor ;) I skimmed over the code quickly and it doesn't seemed to be used + tests are passing, so we should be fine ....
We could also assert the length, but I guess that assertion would need to happen in a generated function to not loose performance :(

andyferris:
What are we going to do about this?

SimonDanisch:
This change seems to be on master already?

Issue #109, opened 2017-02-21
World-age failure in M * v with Colors


timholy:
I noticed this while trying to get Images to pass tests on 0.6. 

Suppressing depwarns from StaticArrays:
```julia
julia> using Colors, StaticArrays, ColorVectorSpace

julia> @noinline foo(M, v) = M * v
foo (generic function with 1 method)

julia> M = @SMatrix rand(3,3)
3×3 StaticArrays.SMatrix{3,3,Float64,9}:
 0.35674   0.0902514   0.804478
 0.645144  0.00201846  0.354793
 0.940136  0.176271    0.110509

julia> v = SVector{3}(rand(RGB{Float64}, 3))
3-element StaticArrays.SVector{3,ColorTypes.RGB{Float64}}:
 RGB{Float64}(0.0250308,0.813153,0.433093) 
 RGB{Float64}(0.612239,0.88945,0.304911)   
 RGB{Float64}(0.0285772,0.674416,0.0770198)

julia> foo(M, v)
ERROR: MethodError: no method matching zero(::Type{ColorTypes.RGB{Float64}})
The applicable method may be too new: running in world age 21469, while current world is 21474.
Closest candidates are:
  zero(::Type{C<:ColorTypes.AbstractRGB}) where C<:ColorTypes.AbstractRGB at /home/tim/.julia/v0.6/ColorVectorSpace/src/ColorVectorSpace.jl:172 (method too new to be called from this world context.)
  zero(::Type{Base.LibGit2.GitHash}) at libgit2/oid.jl:106
  zero(::Type{Base.Pkg.Resolve.VersionWeights.VWPreBuildItem}) at pkg/resolve/versionweight.jl:80
  ...
Stacktrace:
 [1] promote_matprod(::Type{Float64}, ::Type{ColorTypes.RGB{Float64}}) at /home/tim/.julia/v0.6/StaticArrays/src/matrix_multiply.jl:7
 [2] *(...) at /home/tim/.julia/v0.6/StaticArrays/src/matrix_multiply.jl:135
 [3] foo(::StaticArrays.SMatrix{3,3,Float64,9}, ::StaticArrays.SVector{3,ColorTypes.RGB{Float64}}) at ./REPL[2]:1

julia> zero(RGB{Float64})
RGB{Float64}(0.0,0.0,0.0)
```
[...]

Issue #95, opened 2017-01-23
incorrect `pure` annotations


vtjnash:
The `Base.@pure` annotation means "the result of this function is not dependent on any part of the system other than the arguments". Therefore, any function which can be overloaded should not be marked `@pure`, as inference will not be able to get the correct answer. In general, you should trust that inference can figure out that returning a constant has no side-effects, and permit it to still construct its method dependency edges (e.g. avoid annotating anything as `@pure`)

andyferris:
Hi Jameson,

Thanks for the information. I'm still trying to read between the lines here - is this related to worlds/#JuliaLang/julia#265? As in, the system cannot update functions when a new `@pure` method is defined which is a specialization of an existing one, because a "method dependency edge" isn't generated for the constant inserted by a `@pure` function? Or is it really that inference cannot handle multiple `@pure` methods per function, even if every such method is defined prior to compiling any dependent code?

I'm only asking because I was a little surprised to see "any function which can be overloaded should not be marked `@pure`" so I'm wondering if consecutively defining two `@pure` methods (which are not designed to be overloaded by users, and with no dependent methods in between) is a Bad Thing.

If this is related to enforcing consistent worlds, would it be fundamentally a bad thing if `@pure` functions (and `const` variables for that matter, since you can redefine a `const` at the REPL) could be part of the "many-worlds" system, with dependency edges and annotated ages? I can see a semantic argument to enforcing the definition of `@pure` that you gave above (and `const` could become truly constant), but then I wonder if I should just receive an error when I try to change them. (i.e. just turn undefined behavior into an error).

Two more questions: Can I redefine an `@inline` method and have the new system recompile its dependencies? (i.e. is a method dependency edge generated even when the code is inlined?)

Finally, with "you should trust that inference can figure out that returning a constant has no side-effects" the reason I use `@inline`, `@pure` and `@generated` so much is because I don't understand the rules for how the compiler determines what is side-effect free and thus where it inserts a constant as the output of a standard method. For instance, if I have a (normal) function that returns a type, is it side-effect free if that type hasn't been seen before and therefore is added to the type cache? Is it side-effect free if there is something allocated as an intermediate step to determine this type, but that allocated object doesn't "escape" the method? (Is this is explained somewhere in the docs?)

vtjnash:
It's not specifically related to worlds, although it wasn't as interesting (observable) when everything was treated incorrectly.

[...]

PR #121, opened 2017-03-21
Julia 0.6 rewrite

### LINKED ISSUE #113: Code rewrite for Julia v0.6
Much of the code needs revamping for Jula v0.6 (see e.g. #109). I've made a start on the `julia-0.6` branch, but more work remains. This is a tracking issue.

The requirement is to replace all calls to `size` (or `Size`) from within a `@generated` function *generator* to outside of the generator, such as in the generated body, or by using an outer function which calls `Size` and another `@generated` one which unrolls expressions, etc (given the relevant sizes as an input).

Here's a list of all the files/remaining tasks. Completed files should use v0.6 syntax/dispatch rules/etc (no *Compat*):

- [x] util.jl
- [x] traits.jl
- [x] convert.jl (previously was core.jl)
- [x] SUnitRange.jl (this is new)
- [x] Scalar.jl
- [x] SVector.jl
- [x] FieldVector.jl
- [x] SMatrix.jl
- [x] SArray.jl
- [x] MVector.jl
- [x] MMatrix.jl
- [x] MArray.jl
- [x] SizedArray.jl
- [x] abstractarray.jl
- [x] indexing.jl (now more generic indexing that implements all the APL rules)
- [x] broadcast.jl (new file, now generalized to arbitrarily many inputs)
- [x] mapreduce.jl
- [x] arraymath.jl
- [x] linalg.jl
- [x] matrix_multiply.jl 
- [x] solve.jl
- [x] deque.jl
- [x] det.jl
- [x] inv.jl
- [x] eigen.jl
- [x] cholesky.jl

andyferris:
This was quite a large amount of work. Most of the files have been rewritten, and hopefully code quality has improved a bit. Note these improvements:

 * Full multidimensional APL indexing should now be implemented.
 * Similarly, `broadcast`, `broadcast!`, `map` and `map!` now support arbitrary many static array inputs.
 * Universal use of the `Size` trait for dispatch
 * A `Length` trait and related pure functions
 * Reorganized files a little, tried to upgrade most code to the new v0.6 syntax and be more consistent with formatting and expression munging.
 * Started some work on static ranges - the eventual goal here is to upgrade from talking about sizes to talking about indices, so we can have static arrays with index ranges other than `1:n`. (However, this wasn't exported.)

We have this breaking change:

 * `similar_type` defaults to `SVector`, etc - it does *not* preserve immutability. This helps e.g. with things like `m[i, SVector(1,2,3)]` where `m::Matrix`, and should generally be beneficial for speed.

We have this regression

 * Mixed static array - standard array operations like `SMatrix * Vector` have been removed. We can add these back in, if they are popular, but until we have inlined non-isbits immutables, we shouldn't do the simple thing here and wrap them in a `SizedArray` (well, we could, but it might cost more to allocate a pointer to the heap + gc, than to perform the actual operation).

I don't expect a full code review here. Nonetheless I'll leave this PR open for a day or two. (Please feel free to comment here before or *after* it is merged - we can always improve the code).

Closes #113 #109 #106 #95


rdeits:
This is amazing. Thank you!

src/broadcast.jl
```diff
@@ -0,0 +1,147 @@
...
+                else
+                    current_ind[i] = 1
+                    current_ind[i+1] += 1
+                end
+            else
+                break
+            end
+        end
+        j += 1
+    end
```
c42f: This loop looks exactly the same as in `broadcast`, with the exception of the `dest[$j] =` bit.  Could be factored together?

src/convert.jl
```diff
@@ -0,0 +1,92 @@
...
+end
+
+function convert{T}(::Type{Vector}, sa::StaticVector{T})
+    out = Vector{T}(size(sa))
+    @inbounds for i = 1:length(sa)
+        out[i] = sa[i]
+    end
+    return out
+end
+=#
```
c42f: What's all this commented out stuff? Speculative functionality? Removed stuff?

src/det.jl
```diff
@@ -1,39 +1,39 @@
...
-@inline _det(::Size{(1,1)}, A::AbstractMatrix) = @inbounds return A[1]
+@inline _det(::Size{(1,1)}, A::StaticMatrix) = @inbounds return A[1]
 
-@inline function _det(::Size{(2,2)}, A::AbstractMatrix)
+@inline function _det(::Size{(2,2)}, A::StaticMatrix)
     @inbounds return A[1]*A[4] - A[3]*A[2]
 end
 
-@inline function _det{T<:Unsigned}(::Size{(2,2)}, A::AbstractMatrix{T})
+@inline function _det(::Size{(2,2)}, A::StaticMatrix{T}) where {T <: Unsigned}
```
c42f: Can this `where` be avoided with `StaticMatrix{<:Unsigned}`?

src/det.jl
```diff
@@ -1,39 +1,39 @@
...
+@generated function _det{S,T}(::Size{S}, A::StaticMatrix{T})
     if S[1] != S[2]
         throw(DimensionMismatch("matrix is not square"))
     end
-    T2 = typeof((one(T)*zero(T) + zero(T))/one(T))
     return quote # Implementation from Base
+        T2 = typeof((one(T)*zero(T) + zero(T))/one(T))
         if istriu(A) || istril(A)
-            return convert($T2, det(UpperTriangular(A)))::$T2 # Is this a Julia bug that a convert is not type stable??
+            return convert(T2, det(UpperTriangular(A))) # Is this a Julia bug that a convert is not type stable??
```
c42f: The type assert was removed here.  Obsolete comment?

src/eigen.jl
```diff
@@ -190,7 +193,7 @@ end
 # A small part of the code in the following method was inspired by works of David
 # Eberly, Geometric Tools LLC, in code released under the Boost Software
 # License (included at the end of this file).
-@inline function _eig{T<:Real}(::Size{(3,3)}, A::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale)
+@inline function _eig(::Size{(3,3)}, A::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
```
c42f: Can the `where` pieces be removed in this file and replaced with the shorthand?

src/Scalar.jl
```diff
@@ -6,9 +6,15 @@ Construct a statically-sized 0-dimensional array that contains a single element,
...
 immutable Scalar{T} <: StaticArray{T,0}
     data::T
+
+    Scalar{T}(x::AbstractArray) where {T} = new{T}(convert(T,x))
+    Scalar{T}(x::Tuple{T2}) where {T, T2} = new{T}(convert(T,x[1]))
+    Scalar{T}(x) where {T} = new{T}(convert(T, x))
 end
 
-@inline (::Type{Scalar{T}}){T}(x::Tuple{T}) = Scalar{T}(x[1])
+@inline Scalar(x::Tuple{T}) where {T} = Scalar{T}(x[1])
```
andyferris: The abstract constructors will go `Scalar(x) -> Scalar((x,))`. This then goes to `Scalar{T}(x)`, which is an inner constructor (and is more specific than other constructors).

src/SUnitRange.jl
```diff
@@ -0,0 +1,12 @@
+# `Start` and `End` should be `Int`
+struct SUnitRange{Start,End} <: StaticVector{Int}
+end
+
+@pure Size(::SUnitRange{Start, End}) where {Start,End} = Size(End-Start+1)
+
+@pure @propagate_inbounds function getindex(x::SUnitRange{Start,End}, i::Int) where {Start, End}
+    @boundscheck if i < Start || i > End
```
andyferris: Maybe worth looking into. Though I thought this pattern was used throughout `Base`.

src/MMatrix.jl
```diff
@@ -94,9 +94,9 @@ end
 # Some more advanced constructor-like functions
 @inline one{N}(::Type{MMatrix{N}}) = one(MMatrix{N,N})
 @inline eye{N}(::Type{MMatrix{N}}) = eye(MMatrix{N,N})
-@inline eye{N,M}(::Type{MMatrix{N,M}}) = eye(MMatrix{N,M,Float64})
-@inline zeros{N,M}(::Type{MMatrix{N,M}}) = zeros(MMatrix{N,M,Float64})
-@inline ones{N,M}(::Type{MMatrix{N,M}}) = ones(MMatrix{N,M,Float64})
+#@inline eye{N,M}(::Type{MMatrix{N,M}}) = eye(MMatrix{N,M,Float64})
+#@inline zeros{N,M}(::Type{MMatrix{N,M}}) = zeros(MMatrix{N,M,Float64})
+#@inline ones{N,M}(::Type{MMatrix{N,M}}) = ones(MMatrix{N,M,Float64})
```
andyferris: Yeah, probably should delete these.

What happened here is that I made the generic versions of these functions a bit more powerful.

src/abstractarray.jl
```diff
@@ -71,25 +55,25 @@ similar_type{A<:AbstractArray,T,S}(::A,::Type{T},s::Size{S}) = similar_type(A,T,
...
-@generated default_similar_type{T,S}(::Type{T}, s::Size{S}, ::Type{Val{2}}) = SMatrix{S[1],S[2],T,prod(S)}
-@generated default_similar_type{T,S,D}(::Type{T}, s::Size{S}, ::Type{Val{D}}) = SArray{S,T,D,prod(S)}
+default_similar_type{T,S}(::Type{T}, s::Size{S}, ::Type{Val{1}}) = SVector{S[1],T}
+default_similar_type{T,S}(::Type{T}, s::Size{S}, ::Type{Val{2}}) = SMatrix{S[1],S[2],T,prod(s)}
+default_similar_type{T,S,D}(::Type{T}, s::Size{S}, ::Type{Val{D}}) = SArray{S,T,D,prod(s)}
 
-# mutable things stay mutable
-similar_type{SA<:Union{MVector,MMatrix,MArray},T,S}(::Type{SA},::Type{T},s::Size{S}) = mutable_similar_type(T,s,length_val(s))
+# should mutable things stay mutable?
+#similar_type{SA<:Union{MVector,MMatrix,MArray},T,S}(::Type{SA},::Type{T},s::Size{S}) = mutable_similar_type(T,s,length_val(s))
```
andyferris: Yes. I forgot to mention above. @timholy seemed supportive, so I left it this way.

src/abstractarray.jl
```diff
@@ -120,103 +104,67 @@ similar{SA<:SizedArray,T,S}(::Type{SA},::Type{T},s::Size{S}) = sizedarray_simila
...
-        $(Expr(:meta, :inline))
-        m = sym.data
-        if sym.uplo == 'U'
-            @inbounds return SM($(Expr(:call, SM, Expr(:tuple, exprs_up...))))
-        else
-            @inbounds return SM($(Expr(:call, SM, Expr(:tuple, exprs_lo...))))
-        end
+@inline reshape(a::StaticArray, s::Size) = similar_type(a, s)(Tuple(a))
+@generated function reshape(a::AbstractArray, s::Size{S}) where {S}
+    if IndexStyle(a) == IndexLinear() # TODO this isn't "hyperpure"
```
andyferris: Actually, this one is critical. If `a` is defined after `using StaticArrays`, this will fail.

src/arraymath.jl
```diff
@@ -1,232 +1,149 @@
...
-    @inline .\(a1::Number, a2::StaticArray) = broadcast(\, a1, a2)
-    @inline .%(a1::Number, a2::StaticArray) = broadcast(%, a1, a2)
-    @inline .^(a1::Number, a2::StaticArray) = broadcast(^, a1, a2)
-    @inline .<<(a1::Number, a2::StaticArray) = broadcast(<<, a1, a2)
-    @inline .>>(a1::Number, a2::StaticArray) = broadcast(>>, a1, a2)
-    @inline .==(a1, a2::StaticArray) = broadcast(==, a1, a2)
-    @inline .<(a1, a2::StaticArray) = broadcast(<, a1, a2)
-    @inline .>(a1, a2::StaticArray) = broadcast(>, a1, a2)
-    @inline .<=(a1, a2::StaticArray) = broadcast(<=, a1, a2)
-    @inline .>=(a1, a2::StaticArray) = broadcast(>=, a1, a2)
```
andyferris: :)

src/arraymath.jl
```diff
@@ -1,232 +1,149 @@
...
+        end
     end
 end
 
-@inline Base.fill{SA <: StaticArray}(val, ::SA) = fill(val, SA)
-@generated function Base.fill{SA <: StaticArray}(val, ::Type{SA})
-    l = length(SA)
-    T = eltype(SA)
-    expr = [:valT for i = 1:l]
+@inline fill(val, ::SA) where {SA <: StaticArray} = ones(val, SA)
```
andyferris: oops

src/broadcast.jl
```diff
@@ -0,0 +1,147 @@
...
+        end
+    end
+    newsize = tuple(newsize...)
+
+    exprs = Array{Expr}(newsize)
+    more = true
+    current_ind = ones(Int, length(newsize))
+
+    while more
+        exprs_vals = [(a[i] <: Number ? :(a[$i]) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))])) for i = 1:length(sizes)]
```
andyferris: It seemed OK - might be worth investigating further. You need to make sure you benchmark `map` and `broadcast` with a wrapper function that gets rid of all the `Varargs` - in practice you won't be calling these from toplevel so that's the natural way to benchmark it.

src/broadcast.jl
```diff
@@ -0,0 +1,147 @@
...
+            end
+        end
+    end
+
+    eltype_exprs = [:(eltype($t)) for t ∈ a]
+    newtype_expr = :(Core.Inference.return_type(f, Tuple{$(eltype_exprs...)}))
+
+    return quote
+        @_inline_meta
+        @inbounds return similar_type($first_staticarray, $newtype_expr, Size($newsize))(tuple($(exprs...)))
```
andyferris: I'd like to attack this at some point. Probably worth looking throughout the package in a separate PR

src/det.jl
andyferris: Yes, it can

src/det.jl
andyferris: ahh... I probably should test this...

src/eigen.jl
andyferris: This one uses `T`.

tkoolen:
@rdeits, I think you broke Github.

andyferris:
So it wasn't just me that saw that? I reloaded and it seems fine now.

src/SUnitRange.jl
c42f: It is, but I think it's usually written as 

```julia
@boundscheck checkbounds(A, i)
```

or something.  IIRC I had problems when the expression given to `@boundscheck` wasn't a function call.

timholy: So it looks like you've defined a static version of what I'm calling an `IdentityRange` or `IdempotentRange`, see https://github.com/JuliaArrays/IdentityRanges.jl/issues/1 and https://github.com/JuliaLang/julia/pull/21052. Just to check, you're aware that these are non-1 indices arrays you've defined here? If this is intentional, you should definitely define the `indices` function for them.

If that's not what you're intending, you should be checking against `1 <= i <= End - Start + 1` and then return `i + Start - 1`.

src/abstractarray.jl
timholy: Can you make this safer by dispatching on the `IndexStyle` in "regular" functions?

```julia
reshape(a::AbstractArray, s::Size) = _reshape(IndexStyle(a), a, s)
@generated function _reshape(::IndexLinear, a, s::Size{S}) where {S}
...
end
@generated function _reshape(::IndexStyle, a, s::Size{S}) where {S}
...
end
```

src/abstractarray.jl
```diff
@@ -120,103 +104,67 @@ similar{SA<:SizedArray,T,S}(::Type{SA},::Type{T},s::Size{S}) = sizedarray_simila
...
+        exprs = [:(a[$i]) for i = 1:prod(S)]
+    else
+        exprs = [:(a[$(inds...)]) for inds ∈ CartesianRange(S)]
```
timholy: `inds.I...`? Also, does this imply that `a[4,2]` is defined for a `SMatrix{3, 4}`? Or do you mean to be iterating over `Size(a)`?

src/indexing.jl
```diff
@@ -1,822 +1,312 @@
...
-end
+# Note: all indexing behavior defaults to dense, linear indexing
 
-@generated function getindex{I1 <: Integer,I2 <: Integer, T}(
-        m::AbstractArray{T}, inds1::StaticVector{I1}, inds2::StaticVector{I2}
-    )
-    S1, S2 = length(inds1), lengths(inds2)
-    exprs = [:(m[inds1[$j1], inds2[$j2]]) for j1 = 1:S1, j2 = 1:S2]
-    return Expr(:call, SMatrix{S1, S2, T}, Expr(:tuple, exprs...)) # I guess SMatrix should be fine?
+@propagate_inbounds function getindex(a::StaticArray, inds::Int...)
```
timholy: Do you really want to support `inds::Int...`, or would it be better to support only `i::Vararg{Int, N}` where `N` matches the dimensionality of the array? If you do that, then you get all the goodness of dropped 1s, etc, for free.

Also, you may not need this at all for `IndexStyle(a)::IndexLinear`.

src/indexing.jl
```diff
@@ -1,822 +1,312 @@
...
-            $(Expr(:meta, :inline))
-            a
-        end
-    else
-        l = length(SA)
-        inds = 1:l
-        return quote
-            $(Expr(:meta, :inline, :propagate_inbounds))
-            $(Expr(:call, :getindex, :a, Expr(:tuple, inds...)))
+@generated function _getindex_scalar(::Size{S}, a::StaticArray, inds::Int...) where S
```
timholy: If you need this, split this first segment into a `sub2ind_expr` function, or something?

src/indexing.jl
```diff
@@ -1,822 +1,312 @@
...
-end
-
-function Base.unsafe_setindex!{Sizes,T}(a::MArray{Sizes,T}, v, i...)
-    # First check v and i have correct sizes (ignoring singleton dimensions, mimicking Base.Array)
-    i_sizes = ntuple(j -> i[j] == Colon() ? Sizes[j] : length(i[j]), length(i) )
+    # Iterate over input indices
+    ind_types = inds.parameters
+    current_ind = ones(Int,length(ind_types))
+    more = true
+    while more
```
timholy: This overall logic seems repeated in many places...is it consolidatable?

src/broadcast.jl
c42f: Sure, works for me.

src/SUnitRange.jl
andyferris: Haha thanks @timholy - these definitely are WIP. I think I'll split this off and put it on a branch for now.

These are just meant to be singleton array types, so that `indices(SVector(1,2,3)) === SUnitRange{1,3}()`. AFAICT `IdentityRanges` are meant to be a bit like a safe `:`? (How is it distinct from `start:stop`?)

I suppose that `OneTo` has been able to double-up for both purposes, for one-based arrays?

src/abstractarray.jl
andyferris: Will do :)

src/indexing.jl
andyferris: Hmm... yes this seems like a valid point.

I should check if the implementation here is faster than a basic `sub2ind` one.

src/indexing.jl
andyferris: That might have been better (but it's already written, so...)

src/SUnitRange.jl
timholy: IdentityRanges are indeed a generalization of `:`, one that allows you to slice some or all of an axis while preserving the indices. As to how it differs from `start:stop`...if you're constructing a view `v = view(a, s)` with an AbstractVector `s`, the key identity is that `v[i] == a[s[i]]`. Since `start:stop` is a conventional `UnitRange`, it is indexed starting at 1, so `v = view(a, start:stop)` would be indexed starting at 1 no matter how `a` is indexed. In contrast, `v = view(a, IdentityRange(start:stop))` would be indexed starting with `start`, no matter how `a` is indexed.

andyferris:
@SimonDanisch - what is the plan for `StaticArrays.FixedSizeArrays` on Julia 0.6?

SimonDanisch:
See if we can make it work?! ;) I will have to look into it very soon!

src/matrix_multiply.jl
```diff
@@ -1,676 +1,303 @@
...
-@generated function *(A::StaticMatrix, B::StaticMatrix)
-    TA = eltype(A)
-    TB = eltype(B)
-
-    T = promote_matprod(TA, TB)
-
-    can_mutate = !isbits(A) || !isbits(B) # !isbits implies can get a persistent pointer (to pass to BLAS). Probably will change to !isimmutable in a future version of Julia.
-    can_blas = T == TA && T == TB && T <: Union{Float64, Float32, Complex{Float64}, Complex{Float32}}
+@generated function _A_mul_B(Sa::Size{sa}, Sb::Size{sb}, a::StaticMatrix{Ta}, b::StaticMatrix{Tb}) where {sa, sb, Ta, Tb}
+    can_mutate = a.mutable && b.mutable # TODO this probably isn't safe. Maybe a trait??
```
c42f: This should be fine, shouldn't it? `a` and `b` are `DataType` in the generator context, and even if the field access here was turned into a call to `getfield`, it's an intrinsic (I think?) and can't be extended.

src/matrix_multiply.jl
```diff
@@ -1,676 +1,303 @@
...
-        if can_blas && size(A,1)*size(A,2)*size(B,2) >= 14*14*14
+        if can_blas && sa[1]*sa[2]*sb[2] >= 14*14*14
             return quote
-                $(Expr(:meta, :inline))
-                C = similar(A, $T, $(Size(s)))
-                A_mul_B_blas!(C, A, B)
+                @_inline_meta
+                T = promote_matprod(Ta, Tb)
+                C = similar(a, T, $S)
+                A_mul_B_blas!($S, C, Sa, Sb, a, b)
```
c42f: Nice.

src/matrix_multiply.jl
```diff
@@ -1,676 +1,303 @@
...
-@generated function A_mul_B!(C::StaticMatrix, A::StaticMatrix, B::StaticMatrix)
-    if isbits(C)
-        error("Cannot mutate $C")
-    end
-
-    TA = eltype(A)
-    TB = eltype(B)
-    T = promote_matprod(TA, TB)
-
-    can_blas = T == TA && T == TB && T <: Union{Float64, Float32, Complex{Float64}, Complex{Float32}}
```
c42f: :+1: for less code!

src/solve.jl
```diff
@@ -1,40 +1,28 @@
...
+        ((a[2,2]*a[3,3] - a[2,3]*a[3,2])*b[1] +
+            (a[1,3]*a[3,2] - a[1,2]*a[3,3])*b[2] +
+            (a[1,2]*a[2,3] - a[1,3]*a[2,2])*b[3]) / d,
+        ((a[2,3]*a[3,1] - a[2,1]*a[3,3])*b[1] +
+            (a[1,1]*a[3,3] - a[1,3]*a[3,1])*b[2] +
+            (a[1,3]*a[2,1] - a[1,1]*a[2,3])*b[3]) / d,
+        ((a[2,1]*a[3,2] - a[2,2]*a[3,1])*b[1] +
+            (a[1,2]*a[3,1] - a[1,1]*a[3,2])*b[2] +
+            (a[1,1]*a[2,2] - a[1,2]*a[2,1])*b[3]) / d )
+end
```
c42f: It's a huge improvement to move these to traits.  Nice!

src/traits.jl
```diff
@@ -53,13 +54,32 @@ function Size{SA <: StaticArray}(::Type{SA})
         """)
 end
 
-# Some @pure convenience functions.
 
-@pure get{S}(::Size{S}) = S
-@pure getindex{S}(::Size{S}, i::Int) = i <= length(S) ? S[i] : 1
+struct Length{L}
+    function Length{L}() where L
+        check_length(L)
```
c42f: Oh, can you use that trick you were showing to avoid the need for `check_length`?

c42f: Not needed of course, just a nicety.

src/util.jl
```diff
@@ -28,6 +28,9 @@ end
     end
 end
 
+
+# TODO: the below seems to be type piracy...
+#=
```
c42f: Then let's just delete these, and document it?

c42f: Hard to know whether to depwarn them, really they should never have been in here in the first place.

test/custom_types.jl
```diff
@@ -0,0 +1,7 @@
+@testset "Custom types" begin
+    # Issue 123
+    @eval (struct MyType{N, T} <: StaticVector{T}
```
c42f: You don't need the brackets around `struct`, I think?

test/linalg.jl
```diff
@@ -17,10 +17,10 @@
...
         v4 = [4,3,2,1]
 
-        @test @inferred(v1 - v4) === @SVector [-2, 1, 4, 7]
-        @test @inferred(v3 - v2) === @SVector [-2, 1, 4, 7]
-        @test @inferred(v1 - v4) === @SVector [-2, 1, 4, 7]
-        @test @inferred(v3 - v2) === @SVector [-2, 1, 4, 7]
+        @test_broken @inferred(v1 + v4) === @SVector [6, 7, 8, 9]
+        @test_broken @inferred(v3 + v2) === @SVector [6, 7, 8, 9]
+        @test_broken @inferred(v1 - v4) === @SVector [-2, 1, 4, 7]
+        @test_broken @inferred(v3 - v2) === @SVector [-2, 1, 4, 7]
```
c42f: Could use a comment discussing what happened here.  Same for other newly broken tests, I guess.

c42f: ie, obviously there's a decision to be made about whether we reimplement this functionality (in which case we delete the broken tests).  Just a one liner about the decision to leave it broken for now, perhaps?

src/util.jl
c42f: Personally, I'd lean toward just deleting any type piracy, and documenting it in the release notes.

src/linalg.jl
```diff
@@ -379,36 +315,37 @@ _norm_p0(x) = x == 0 ? zero(x) : one(x)
...
     end
 
-    if s[1] == 0
+    if S[1] == 0
         return zero(eltype(a))
     end
 
-    exprs = [:(a[$(sub2ind(s, i, i))]) for i = 1:s[1]]
+    exprs = [:(a[$(sub2ind(S, i, i))]) for i = 1:S[1]]
     total = reduce((ex1, ex2) -> :($ex1 + $ex2), exprs)
```
KristofferC: That will wreck performance if `exprs` is too long because julia can't handle too many function argument well.

KristofferC: Although I have noticed this reduction pattern to be quite tough on inference and this long compilation times.

src/linalg.jl
```diff
@@ -220,108 +166,96 @@ end
...
+@inline function _cross(::Size{(3,)}, a::StaticVector, b::StaticVector)
+    T = typeof(a[2]*b[3]-a[3]*b[2])
+    @inbounds return similar_type(a, typeof(a[2]*b[3]-a[3]*b[2]))((a[2]*b[3]-a[3]*b[2], a[3]*b[1]-a[1]*b[3], a[1]*b[2]-a[2]*b[1]))
+end
+@inline function _cross(::Size{(2,)}, a::StaticVector{<:Unsigned}, b::StaticVector{<:Unsigned})
+    @inbounds return Signed(a[1]*b[2]) - Signed(a[2]*b[1])
+end
+@inline function _cross(::Size{(3,)}, a::StaticVector{<:Unsigned}, b::StaticVector{<:Unsigned})
+    T = typeof(a[2]*b[3]-a[3]*b[2])
+    @inbounds return similar_type(a, typeof(Signed(a[2]*b[3])-Signed(a[3]*b[2])))(((Signed(a[2]*b[3])-Signed(a[3]*b[2]), Signed(a[3]*b[1])-Signed(a[1]*b[3]), Signed(a[1]*b[2])-Signed(a[2]*b[1]))))
```
andyferris: Yeah, it just happens to be needed for the determinant algorithm, and someone raised an issue about that! Seemed a bit silly but oh well.

src/linalg.jl
```diff
@@ -220,108 +166,96 @@ end
...
-        return quote
-            $(Expr(:meta, :inline))
-            @inbounds return $expr
-        end
-    else
-        error("vecdot() expects arrays of same length. Got sizes $(size(a)) and $(size(b)).")
+    expr = :(conj(a[1]) * b[1])
+    for j = 2:prod(S)
+        expr = :($expr + conj(a[$j]) * b[$j])
+    end
```
andyferris: lol. 

Are we sure that multi-argument `+` is performant?

src/linalg.jl
andyferris: Are we sure that multi-argument + is performant? (There will be a splat somewhere, which potentially ruins performance for vectors and matrices with 16+ elements)

src/mapreduce.jl
```diff
@@ -1,430 +1,214 @@
...
 
-@generated function map{T1,T2}(f, a1::AbstractArray{T1}, a2::StaticArray{T2})
-    exprs = [:(f(a1[$j], a2[$j])) for j = 1:length(a2)]
+@generated function _map(f, ::Size{S}, a::StaticArray...) where {S}
+    exprs = Vector{Expr}(prod(S))
+    for i ∈ 1:prod(S)
+        tmp = [:(a[$j][$i]) for j ∈ 1:length(a)]
+        exprs[i] = :(f($(tmp...)))
+    end
+    eltypes = [eltype(a[j]) for j ∈ 1:length(a)] # presumably, `eltype` is "hyperpure"?
```
andyferris: lol.

Yes - should I make a PR to Julia NEWS.md highlighting our new *hyperpure*™ technology stack?

src/matrix_multiply.jl
andyferris: It's a bit funny... you can have a `StaticArray` struct which is immutable but for which `setfield!` works (such as `SizedArray`), and you can (in theory) have a `mutable struct` and not define `setfield!`. Annoying...

I probably should just specialize on `Union{SizedMatrix{<:BlasFloat}, MMatrix{<:BlasFloat}}`.

test/linalg.jl
andyferris: Well... probably a good idea...

andyferris:
OK, I'll make a release to METADATA sometime over the next day or two - anyone who finds a bug before then, I'd appreciate a report.

src/linalg.jl
c42f: I *thought* it was solved way back in 0.4 via a hack I spotted in inference.

But I've just done a test and you're right to be worried about this - generated code quality still falls off a cliff with more than 7 arguments to `+`.  This is despite the fact that `+` is force inlined via a hack in inference.

Looking at `@code_typed sumit(@SVector(ones(8)))` of

```julia
@generated function sumit(v)
    s = Expr(:call, :+, [:(v[$i]) for i=1:length(v)]...)
    quote
        @inbounds s = $s
        s
    end
end
```

shows some utterly strange things going on as a function of the `SVector` length.

src/matrix_multiply.jl
c42f: Looks good.

dpsanders:
Just wanted to mention that ValidatedNumerics's tests pass with this branch on 0.6 -- many thanks for the hard work!

timholy:
I'm delighted to report that ImageFiltering passes as well! Thanks so very much!

I assume the new release will be called 0.4?

andyferris:
Yep, to make it official, I just tagged v0.4.0 in github - its up to attobot now :)

andyferris:
https://github.com/JuliaLang/METADATA.jl/pull/8494

src/mapreduce.jl
```diff
@@ -1,430 +1,214 @@
...
-        expr = :(op(f(a1[1]), f(a1[2])))
-        for j = 3:length(a1)
-            expr = :(op($expr, f(a1[$j])))
-        end
-        return quote
-            $(Expr(:meta, :inline))
-            @inbounds return $expr
-        end
-    end
+@inline function mapreduce(f, op, a::StaticArray...)
```
tkelman: Little odd, but this allows for `mapreduce(f, op)` which would normally error. `same_size()` isn't defined right now, but it might at some point, by accident or piracy.

REQUIRE
```diff
@@ -1,2 +1 @@
 julia 0.5
-Compat 0.19.0
```
tkelman: `@compat` is still used in `test/fixed_size_arrays.jl`, though it looks like that's now dead code

andyferris: Yes, its not deleted because I think Simon Danisch will revive it sometime

src/mapreduce.jl
andyferris: This will still error :) 

I do see your point (though it seems unlikely that this signature would clash with ones defined by others). Are you requesting that I modify the signature, or suggesting?

> `same_size()` isn't defined right now, but it might at some point, by accident or piracy.

I'm not *quite* sure how to respond, since I'm not even exporting this. I *do* intend to keep watch on new features and breaking changes for Julia 0.7/1.0 (and a new export from `Base` would count as more than bugfix w.r.t. semver). But yes, it would be useful to generally have a function like this exported from `Base` one day.

tkelman: Just trying to think of ways in which the 0-occurrences version of this might be a problem (or not). I guess there could also be other packages that might do the same basic signature but for vararg of their own type, in which case `mapreduce(f, op)` might be ambiguous between here and that other package? Maybe a little safer to have the signature 1-or-more instead of 0-or-more, but maybe I'm just being paranoid.

PR #122, opened 2017-03-22
allow Union entries to SMatrix

marius311:
Currently it looks like you can't do SMatrices with any sort of Union element type. SVectors and SArrays are fine though,

```julia
julia> SVector{1,Complex}(1)
1-element StaticArrays.SVector{1,Complex}:
 1+0im

julia> SArray{(1,),Complex}(1)
1-element StaticArrays.SArray{(1,),Complex,1,1}:
 1+0im

julia> SMatrix{1,1,Complex,1}(1)
ERROR: SMatrix: Parameter T must be a DataType. Got Complex
Stacktrace:
 [1] Type at /home/marius/.julia/v0.6/StaticArrays/src/SMatrix.jl:26 [inlined]
 [2] StaticArrays.SMatrix{1,1,Complex,1}(::Int64) at /home/marius/.julia/v0.6/StaticArrays/src/convert.jl:3
```

From a performance standpoint you would obviously never do this, but it seems it should at least work. The fix is just delete three lines which the comment there indicates might have just been forgotten about, but I've only started using this package in the last couple of days so an expert should check if this is a good fix!

(PS, thanks for the great package, and for the ahead-of-the-curve 0.6 upgrades)

### DIFF
--- a/src/SMatrix.jl
+++ b/src/SMatrix.jl
@@ -29,10 +29,6 @@ immutable SMatrix{S1, S2, T, L} <: StaticMatrix{T}
 end
 
 @generated function check_smatrix_params{S1,S2,L}(::Type{Val{S1}}, ::Type{Val{S2}}, T, ::Type{Val{L}})
-    if !(T <: DataType) # I think the way types are handled in generated fnctions might have changed in 0.5?
-        return :(error("SMatrix: Parameter T must be a DataType. Got $T"))
-    end
-
     if !isa(S1, Int) || !isa(S2, Int) || !isa(L, Int) || S1 < 0 || S2 < 0 || L < 0
         return :(error("SMatrix: Sizes must be positive integers. Got $S1 × $S2 ($L elements)"))
     end
--- a/src/MMatrix.jl
+++ b/src/MMatrix.jl
@@ -35,10 +35,6 @@ type MMatrix{S1, S2, T, L} <: StaticMatrix{T}
 end
 
 @generated function check_MMatrix_params{S1,S2,L}(::Type{Val{S1}}, ::Type{Val{S2}}, T, ::Type{Val{L}})
-    if !(T <: DataType) # I think the way types are handled in generated fnctions might have changed in 0.5?
-        return :(error("MMatrix: Parameter T must be a DataType. Got $T"))
-    end
-
     if !isa(S1, Int) || !isa(S2, Int) || !isa(L, Int) || S1 < 0 || S2 < 0 || L < 0
         return :(error("MMatrix: Sizes must be positive integers. Got $S1 × $S2 ($L elements)"))
     end
--- a/src/util.jl
+++ b/src/util.jl
@@ -11,6 +11,7 @@
         $(Expr(:tuple, exprs...))
     end
 end
+convert_ntuple(T,_) = error("StaticArray type parameter must be a Type, not '$T'")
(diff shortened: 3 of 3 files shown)


src/SMatrix.jl
```diff
@@ -29,10 +29,6 @@ immutable SMatrix{S1, S2, T, L} <: StaticMatrix{T}
 end
 
 @generated function check_smatrix_params{S1,S2,L}(::Type{Val{S1}}, ::Type{Val{S2}}, T, ::Type{Val{L}})
-    if !(T <: DataType) # I think the way types are handled in generated fnctions might have changed in 0.5?
```
andyferris: Probably changing `T <: DataType` to `T isa Type` here work (on v0.6, otherwise try `isa(T, Type)`). Could you check if this works for you?

marius311: Actually looks like the correct thing to put there is `if !(T isa Type{<:Type})` since we're in a generated function. There's no way to do this in 0.5 AFAIK, although you can always just check outside the generated function. This solution of just deleting these lines at least gives the same error messages between SMatrix/Array/Vector when you try to construct them with a non-type T, albeit crytic. I could look more tomorrow into catching this error more directly if you'd like.

andyferris: Good point (I was treating it as a type parameter, but clearly it's not).

If you do want to have a look, feel free to target v0.6 only. We should also try change this to a `@pure` function, if possible (I had trouble on v0.5). Also if we can remove the `Val`s and split it into two functions - one to check `T` and another to check `S1,S2,L`, that would be fantastic.

marius311:
I myself unfortunately don't think I have time to contribute any code reworking much more significant than this, but how about this solution? I believe it works for all the different types and on 0.5. The error looks like, e.g.

```julia
julia> SVector{1,2}(1)
ERROR: StaticArray type parameter must be a Type, not '2'
 in convert_ntuple(::Int64, ::Tuple{Int64}) at /home/marius/.julia/v0.5/StaticArrays/src/util.jl:11
 in StaticArrays.SVector{1,2}(::Int64) at ./sysimg.jl:53
```

andyferris:
Marius - instead of deleting this code, it would be preferable to preserve the functionality.

This can be achieved by using these two methods:  

```julia
function check_smatrix_params{S1,S2,L}(::Type{Val{S1}}, ::Type{Val{S2}}, T, ::Type{Val{L}})
    # error: T is not a type (or else it would dispatch to the below)
end

@generated function check_smatrix_params{S1,S2,T,L}(::Type{Val{S1}}, ::Type{Val{S2}}, ::Type{T}, ::Type{Val{L}})
    # check sizes here
end
```

and similarly for `MMatrix`. I have already implemented something like this in the `julia-0.6` branch.

marius311:
Ah sorry I should have said more explicitly, I pushed two more commits that put back checking the "T" type, also on Vectors and Arrays, I think you might have not seen the commits? That's what I was asking about "hows this solution?" 

Anyway, I see now that you've fixed the Union in SMatrix issue on the julia-0.6 branch like you mention, thanks. Minor point, but it looks like error checking is not happening for Vectors and Arrays. I think a simple solution is what I did, but its entirely up to you, feel 100% free to close this PR!

marius311:
Closing since this all looks fixed on master now, thanks.

PR #124, opened 2017-03-29
Fix indexing with zero-size arrays

Keno:


andyferris:
Looks good - thanks @Keno.

Out of curiosity, under what circumstances did you note this? It probably doesn't happen often that you *need* a vector of known length equal to zero...

Keno:
Celeste has index sets for different objects (stars/galaxies) which through complicated abstractions end up as svectors of indicies when indexing into a hessian. The shape parameters for stars are empty (because they're point sources), but I don't check for that case in the generic code that does both.

andyferris:
Aha, thanks for the explanation.

andyferris:
Was the trailing singletons issue (#125) also related to Celeste/other working code, or just a curiosity?

Keno:
Yes, it is. There's one more issue which is that the new version of this code blows up type inference. I have a fix coming for that as well, but that'll have to wait until tomorrow.

Keno:
This code meaning StaticArrays, not this PR

andyferris:
I look forward to it

PR #126, opened 2017-03-29
modify varargs signatures to ensure at least one input

tkelman:
to avoid defining methods that don't make much sense like `mapreduce(f, op)`


andyferris:
Thanks Tony - I was planning to get around to this at some point, so you've saved me the effort! :)

PR #127, opened 2017-03-30
RFC: Unify SArray,SMatrix,SVector

Keno:
After the recent refactor something that used to be an SMatrix became a
two dimensional SArray (or vice versa - who can keep track). That situation
seems unfortunate and unnecessary, so this is an attempt to unify them.
This is decently breaking, because it requires the SArray dimension tuple
to be e.g. `Tuple{1,2}` rather than `(1,2)`, but should allow us to avoid
quite a bit of duplication. These are just the minimal changes to get this
running. I would imagine, we could remove some further duplication in follow
up patches.


src/SArray.jl
```diff
@@ -19,56 +19,82 @@ immutable SArray{Size, T, N, L} <: StaticArray{T, N}
...
     return quote
         $(Expr(:meta, :inline))
-        SArray{Size,T,$(length(Size)),$(prod(Size))}(x)
+        SArray{Size,T,$(tuple_length(Size)),$(tuple_prod(Size))}(x)
     end
 end
 
-@generated function (::Type{SArray{Size}}){Size, T <: Tuple}(x::T)
+function SArray{Size,T,N}(x::Tuple) where {Size, T, N}
+    Base.depwarn("SArray{(dim1,dim2,...),...} is deprecated, use SArray{Tuple{dim1, dim2, ...}, ...}", :SArray)
```
andyferris: Would it be safer to check that `Size` is indeed an instance of `Tuple{Vararg{Int}}` before this depwarn, and error instead if not?

andyferris:
> After the recent refactor something that used to be an SMatrix became a
two dimensional SArray (or vice versa - who can keep track).

Yeah, I "simplified" `similar_type` so that canonically we used `SVector` for 1D containers and `SMatrix` for 2D containers. I'm happy to review this decision if that helps.

andyferris:
@keno, yesterday you mentioned a type stability issue in some work you were doing. Was this caused by `SArray` unexpectedly turning into `SMatrix` or something?

Keno:
No, that's separate, will be a later PR. It's just a bit tricky, so I wanted to get some more experience with that change, so it's not a PR yet.

src/SArray.jl
```diff
@@ -19,56 +19,82 @@ immutable SArray{Size, T, N, L} <: StaticArray{T, N}
...
+        check_sarray_parameters(Size, T, Val{N}, Val{L})
         new{Size,T,N,L}(convert_ntuple(T, x))
     end
 end
 
-@generated function check_sarray_parameters{Size,T,N,L}(::Type{Val{Size}}, ::Type{T}, ::Type{Val{N}}, ::Type{Val{L}})
-    if !(isa(Size, Tuple{Vararg{Int}}))
-        error("SArray parameter Size must be a tuple of Ints (e.g. `SArray{(3,3)}`)")
+@pure tuple_length(T) = length(T.parameters)
+@pure tuple_prod(T) = *(T.parameters...)
```
Keno: Sounds good.

src/SArray.jl
```diff
@@ -19,56 +19,82 @@ immutable SArray{Size, T, N, L} <: StaticArray{T, N}
...
+function check_sarray_parameters(Size, T, N, L)
+    !isa(Size, DataType) || (Size.name !== Tuple.name) && throw(ArgumentError("SArray parameter Size must be a Tuple type, got $Size"))
+    !isa(T, Type) && throw(ArgumentError("SArray parameter T must be a type, got $T"))
+    !isa(N.parameters[1], Int) && throw(ArgumenError("SArray parameter N must be an integer, got $(N.parameters[1])"))
+    !isa(L.parameters[1], Int) && throw(ArgumentError("SArray parameter L must be an integer, got $(L.parameters[1])"))
+    # shouldn't reach here. Anything else should have made it to the function below
+    error("Internal error. Please file a bug")
+end
+
+@generated function check_sarray_parameters{Size,T,N,L}(::Type{Size}, ::Type{T}, ::Type{Val{N}}, ::Type{Val{L}})
```
Keno: Yes, seems reasonable.

src/SArray.jl
Keno: Maybe. Arguably, this should just be an error immediately. The only concern here would be something that the type system doesn't let you put in Tuples. Happy to go with whatever you prefer.

andyferris:
Well, let me know if you want some help

andyferris:
It's worth pointing out that we had a discussion with @JeffBezanson about `Tuple{3,4}`, etc - the most pertinent quote was "so far, I can't think of anything we could achieve by removing that feature".

I feel we can reasonably trust that this will continue working (or else be superseded by something better, like varargs type parameters in user defined types).

c42f:
+1 for this change if Jeff thinks the ability to (ab)use `Tuple` like this is not going away.  In `FixedSizeArrays` this choice worked reasonably well, I think.  We can also put the fixed size back into the abstract `StaticArray` as a type parameter and reclaim the ability to do dispatch on the size without quite so many traits all over the place.

In all, it's amusing to see this design quirk of FixedSizeArrays come back again :-)

andyferris:
Cool. Sounds good then.

Just to confirm - @SimonDanisch you also have experience with this approach. A good idea, in your opinion?

SimonDanisch:
I never regretted it, besides the occasional fights with @jeffbezanson :)

Keno:
Updated to include MArray/Scalar as well

Keno:
Sorry I pushed some other things that are WIP, will back them out and re-push.

andyferris:
OK, thanks @Keno - awesome work.

One question - would it be beneficial to overload `show` for `show(io::IO, ::Type{SVector{N,T}}) where T` so that we see `SVector`. `SMatrix` and `Scalar` as previously? It seems reasonable (to me) to do this for commonly used type aliases.

src/MArray.jl
```diff
@@ -20,75 +20,63 @@ type MArray{Size, T, N, L} <: StaticArray{T, N}
...
 @generated function (::Type{MArray{Size,T}}){Size,T}()
     return quote
         $(Expr(:meta, :inline))
-        MArray{Size, T, $(length(Size)), $(prod(Size))}()
+        MArray{Size, T, $(tuple_length(Size)), $(tuple_prod(Size))}()
     end
 end
 
-@inline MArray(a::StaticArray) = MArray{size(typeof(a))}(Tuple(a))
+@inline MArray(a::StaticArray) = MArray{size_tuple(typeof(a))}(Tuple(a))
```
andyferris: I think we lost `size_tuple` somehow...

Issue #129, opened 2017-03-30
Readme still contains an example of indexing using a tuple


tkoolen:
From the quickstart guide:
```julia
v1 = SVector(1, 2, 3)
v1[(3,2,1)] === @SVector [3, 2, 1]
```
now results in
```
ERROR: ArgumentError: invalid index: (3, 2, 1)
Stacktrace:
 [1] getindex(::StaticArrays.SVector{3,Int64}, ::Tuple{Int64,Int64,Int64}) at ./abstractarray.jl:875
```

andyferris:
Oh bugger. The README is a bit disorganized, sometimes I forget to update things.

We've moved to indexing using static arrays `v1[SVector(3,2,1)]` rather than tuples. The method we used before could be considered (a) type piracy and (b) more suitable for constructing a tuple, not a static vector.

PR #133, opened 2017-04-01
Update readme examples

tkoolen:
Fixes https://github.com/JuliaArrays/StaticArrays.jl/issues/129.

It would be good to have doctests for this, but that's a bigger project.


andyferris:
lgtm. Thanks @tkoolen!

PR #134, opened 2017-04-02
WIP Add size to abstract StaticArray type

andyferris:
This is a continuation of #127 whereby we add a `Tuple{...}` size descriptor to the abstract `StaticArray` type.

I'm previewing this WIP to get feedback on the interface, such as whether we want `StaticArray{Size, T, N}`, or have the dimensions after `T`, or after `N`. Any comments appreciated, otherwise I'll carry on.

### DIFF
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -1,71 +1,69 @@
 """
-    SArray{Size, T, L}(x::NTuple{L, T})
-    SArray{Size, T, L}(x1, x2, x3, ...)
+    SArray{S, T, L}(x::NTuple{L, T})
+    SArray{S, T, L}(x1, x2, x3, ...)
 
-Construct a statically-sized array `SArray`. Since this type is immutable,
-the data must be provided upon construction and cannot be mutated later. The
-`Size` parameter is a Tuple specifying the dimensions of the array. The
-`L` parameter is the `length` of the array and is always equal to `prod(S)`.
-Constructors may drop the `L` and `T` parameters if they are inferrable
-from the input (e.g. `L` is always inferrable from `Size`).
+Construct a statically-sized array `SArray`. Since this type is immutable, the data must be
+provided upon construction and cannot be mutated later. The `S` parameter is a Tuple-type
+specifying the dimensions, or size, of the array - such as `Tuple{3,4,5}` for a 3×4×5-sized
+array. The `L` parameter is the `length` of the array and is always equal to `prod(S)`.
+Constructors may drop the `L` and `T` parameters if they are inferrable from the input
+(e.g. `L` is always inferrable from `S`).
 
-    SArray{Size}(a::Array)
+    SArray{S}(a::Array)
 
-Construct a statically-sized array of dimensions `Size` using the data from
-`a`. The `Size` parameter is mandatory since the size of `a` is unknown to the
+Construct a statically-sized array of dimensions `S` (expressed as a `Tuple{...}`) using
+the data from `a`. The `S` parameter is mandatory since the size of `a` is unknown to the
 compiler (the element type may optionally also be specified).
 """
-immutable SArray{Size, T, N, L} <: StaticArray{T, N}
+immutable SArray{S, T, N, L} <: StaticArray{S, T, N}
     data::NTuple{L,T}
 
-    function (::Type{SArray{Size,T,N,L}}){Size,T,N,L}(x::NTuple{L,T})
-        check_array_parameters(Size, T, Val{N}, Val{L})
-        new{Size,T,N,L}(x)
+    function (::Type{SArray{S, T, N, L}}){S, T, N, L}(x::NTuple{L,T})
+        check_array_parameters(S, T, Val{N}, Val{L})
+        new{S, T, N, L}(x)
     end
 
-    function (::Type{SArray{Size,T,N,L}}){Size,T,N,L}(x::NTuple{L,Any})
-        check_array_parameters(Size, T, Val{N}, Val{L})
-        new{Size,T,N,L}(convert_ntuple(T, x))
+    function (::Type{SArray{S, T, N, L}}){S, T, N, L}(x::NTuple{L,Any})
+        check_array_parameters(S, T, Val{N}, Val{L})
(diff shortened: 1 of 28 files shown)


andyferris:
OK, bulk of work is done, but I haven't updated the tests so many are failing.

More alarmingly, I'm running into an inference crash in the unit tests; similar things have been reported [elsewhere](https://github.com/JuliaLang/julia/issues/21244).

andyferris:
This is ready for a look at. There is a problem with inference (world age failure) in a testset for solve, but it makes no sense to me.

I would still like some feedback on design here, please. :)

c42f:
Afraid I don't have the energy to look at this tonight... maybe tomorrow.

test/solve.jl
```diff
@@ -1,11 +1,27 @@
 @testset "Solving linear system" begin
 
-    @testset "Problem size: $n x $n. Matrix type: $m. Element type: $elty" for n in (1,2,3,4),
+    # I see an inference error in the combined testset below. I can't reproduce
+    # at the REPL or in a function for individual n, etc... speculatively, it
+    # might be reusing A, b with different types in the same (non-toplevel)
+    # scope, for which I've come accross inference bugs in the past.
```
c42f: Somewhat disturbing.

Seems mitigated by the fact that this test case is rather odd with its loop over types though.  May not be a problem in practice?

src/SArray.jl
```diff
@@ -1,72 +1,70 @@
 """
-    SArray{Size, T, L}(x::NTuple{L, T})
-    SArray{Size, T, L}(x1, x2, x3, ...)
+    SArray{S, T, L}(x::NTuple{L, T})
+    SArray{S, T, L}(x1, x2, x3, ...)
```
andyferris: This I always find a bit weird. In some sense, we are documenting the type *and* the constructors. The one without `T` and `L` is documented below.

The `L` is necessary to know about whenever you want to specify it as an element type of an array or a member of a struct, so this was my effort to get the user to find out about it's existence.

Though now I notice that I forgot the `N`!

src/SMatrix.jl
```diff
@@ -53,9 +53,12 @@ end
     end
 end
 
-@inline convert{S1,S2,T}(::Type{SMatrix{S1,S2}}, a::StaticArray{T}) = SMatrix{S1,S2,T}(Tuple(a))
+@inline convert{S1,S2,T}(::Type{SMatrix{S1,S2}}, a::StaticArray{<:Any, T}) = SMatrix{S1,S2,T}(Tuple(a))
```
andyferris: It does allow reshaping... but primarily it provides the functionality to change `T`.

Should we tighten this up at some point?

src/SUnitRange.jl
```diff
@@ -1,8 +1,23 @@
 # `Start` and `End` should be `Int`
-struct SUnitRange{Start,End} <: StaticVector{Int}
+struct SUnitRange{Start, End, L} <: StaticVector{L, Int}
```
andyferris: I guess we could - user's can use the `@pure` constructor on line 20.

src/SizedArray.jl
```diff
@@ -93,4 +84,4 @@ Creates a `SizedArray` wrapping `array` with the specified statically-known
 `dims`, so to take advantage of the (faster) methods defined by the static array
 package.
 """
-(::Size{S}){S}(a::Array) = SizedArray{S}(a)
+(::Size{S}){S}(a::Array) = SizedArray{Tuple{S...}}(a)
```
andyferris: True. Make a PR :)

src/abstractarray.jl
```diff
@@ -39,32 +38,26 @@ if they wish to overload the default behavior.
...
 
 # Default types
-# Generally, use SVector, etc
+# Generally, use SArray
 similar_type{A<:AbstractArray,T,S}(::Type{A},::Type{T},s::Size{S}) = default_similar_type(T,s,length_val(s))
-
-default_similar_type{T,S}(::Type{T}, s::Size{S}, ::Type{Val{0}}) = Scalar{T}
-default_similar_type{T,S}(::Type{T}, s::Size{S}, ::Type{Val{1}}) = SVector{S[1],T}
-default_similar_type{T,S}(::Type{T}, s::Size{S}, ::Type{Val{2}}) = SMatrix{S[1],S[2],T,prod(s)}
 default_similar_type{T,S,D}(::Type{T}, s::Size{S}, ::Type{Val{D}}) = SArray{Tuple{S...},T,D,prod(s)}
```
andyferris: The length of a `Scalar` is `1`.

Try `SArray{Tuple{},Int,0,1}((3,))`

andyferris: Also `prod([]) === 1`

test/solve.jl
andyferris: I think it's not a problem in practice, or at least I couldn't find any evidence of a problem.

src/SMatrix.jl
andyferris: (perhaps a separate PR)

src/SArray.jl
c42f: Ah, righto.  Yes, I find the constructor/type dichotomy odd too.  I often wish the doc system would provide more help here, in terms of autodocumenting the type signature docs, in a way which is systematic but not intrusive.  Oh ho, look at this trick:

```julia
module A

"""
    $(B)
"""
type B{N,T,X,Y,Z}
end

end

help?> A.B
  A.B{N,T,X,Y,Z}

  B is a thing with some type parameters
```

Kinda close?

c42f: Still, even if this was neater, it relies on package authors being systematic and good at writing docs...

src/SMatrix.jl
c42f: Sure.  Chuck in a TODO?

src/SizedArray.jl
c42f: heh :)

src/abstractarray.jl
c42f: Oh.  Duh!

src/SMatrix.jl
andyferris: issue?

Issue #132, opened 2017-04-01
broadcast is ambiguous with base


Keno:
This method: https://github.com/Keno/StaticArrays.jl/blob/master/src/broadcast.jl#L7
is ambiguous with base:

```
MethodError: broadcast(::Optim.UnivariateProblems.##1#3, ::Float64) is ambiguous. Candidates:
  broadcast(f, a::Union{Number, StaticArrays.StaticArray}, b::Union{Number, StaticArrays.StaticArray}...) in StaticArrays at /global/cscratch1/sd/kfischer/celeste-pkgs/v0.6/StaticArrays/src/broadcast.jl:8
  broadcast(f, x::Number...) in Base.Broadcast at broadcast.jl:16
```
that's bad form

timholy:
The problem here is that Base broadcast requires zero or more arguments whereas here it requires one or more. Presumably just need to get rid of the `a` argument and start straight with `b`, since `Union{A,B}` should always be less specific than `A`.

Overall, `detect_ambiguities` picks up 72 ambiguities when `StaticArrays` is loaded. The good news is that `Base.Dates` seems to no longer be a source of trouble.

andyferris:
This is due to a recent PR by @tkelman: https://github.com/JuliaArrays/StaticArrays.jl/pull/126/files

I suspect we should role back that particular change for `broadcast`?

marius311:
Just ran into this also. Can't say I'm following closely enough to understand why the two separate `a,b` args were needed in the PR mentioned above, but if you want to keep that, you could use the following trick, 

```julia
function broadcast(f, a::Union{_, Number, StaticArray}, b::Union{_, Number, StaticArray}...) where {_<:StaticArray}
```

this will *only* match when there's at least one StaticArray present among the args, so there's not even a question of if its more specific.

marius311:
(See also https://discourse.julialang.org/t/incorrect-ambiguity/2690 since this trick/hack doesn't always work, granted I just checked it does in this case)

andyferris:
Thanks @marius311 - that's a very interesting observation.

dlfivefifty:
Even `exp.(1.0+2.0im)` triggers this....

dlfivefifty:
How would you feel about this (less than elegant) work around for low number of arguments:
``julia
[...]

PR #135, opened 2017-04-02
Make `broadcast` signature compatible with `Base`.

andyferris:
A recent PR made some other signatures not define the zero-input version of certain functions where they are not defined in Base. In this case, this seems somewhat unnecessary since there is already a `broadcast(f, ::Number...)` method in `Base`.

The alternative is to define `broadcast(f, ::Number)` method here to remove the ambiguity. This seems like type piracy...

See #132.


c42f:
Looks good to me.

PR #136, opened 2017-04-03
Make map() take mixtures of StaticArray and AbstractArray

c42f:
This allows StaticArray/AbstractArray linear algebra + and - to return a static array.

Here the versions of `map()` only support StaticArray up to the second argument for mixtures of StaticArray and AbstractArray.  Dealing with the ambiguities for longer argument lists becomes increasingly annoying!

I also upgraded `same_size` to work with mixtures of static and abstract arrays, and put it into traits.jl as it seemed to fit there.

Still todo is to fix matrix*vec to be size inferrable with a smatrix input, but that could be a separate PR.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -1,25 +1,24 @@
-# Returns the common Size of the inputs (or else throws a DimensionMismatch)
-@inline same_size(a1::StaticArray, as::StaticArray...) = _same_size(Size(a1), as...)
-@inline _same_size(s::Size) = s
-@inline function _same_size(s::Size, a1::StaticArray, as::StaticArray...)
-    if s === Size(a1)
-        return _same_size(s, as...)
-    else
-        throw(DimensionMismatch("Dimensions must match. Got inputs with $s and $(Size(a1))."))
-    end
-end
-
 @inline _first(a1, as...) = a1
 
 ################
 ## map / map! ##
 ################
 
-@inline function map(f, a::StaticArray, b::StaticArray...)
-    _map(f, same_size(a, b...), a, b...)
-end
+@inline map(f, as::StaticArray...) =
+    _map(f, same_size(as...), as...)
+# Mixed StaticArray + AbstractArray; various versions to avoid ambiguities.
+# With the versions below, if a StaticArray isn't present in the first two
+# arguments, we'll end up in Base.map() instead.
+@inline map(f, a1::StaticArray, as::AbstractArray...) =
+    _map(f, same_size(a1, as...), a1, as...)
+@inline map(f, a1::AbstractArray, a2::StaticArray, as::AbstractArray...) =
+    _map(f, same_size(a1, a2, as...), a1, a2, as...)
+@inline map(f, a1::StaticArray, a2::AbstractArray, as::AbstractArray...) =
+    _map(f, same_size(a1, a2, as...), a1, a2, as...)
+@inline map(f, a1::StaticArray, a2::StaticArray, as::AbstractArray...) =
+    _map(f, same_size(a1, a2, as...), a1, a2, as...)
 
-@generated function _map(f, ::Size{S}, a::StaticArray...) where {S}
+@generated function _map(f, ::Size{S}, a::AbstractArray...) where {S}
     exprs = Vector{Expr}(prod(S))
     for i ∈ 1:prod(S)
         tmp = [:(a[$j][$i]) for j ∈ 1:length(a)]
@@ -29,7 +28,7 @@ end
     newT = :(Core.Inference.return_type(f, Tuple{$(eltypes...)}))
     return quote
         @_inline_meta
-        @inbounds return similar_type(typeof(_first(a...)), $newT)(tuple($(exprs...)))
+        @inbounds return similar_type(typeof(_first(a...)), $newT, Size(S))(tuple($(exprs...)))
     end
 end
 
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -1,7 +1,11 @@
+@inline _first(a1, as...) = a1
+
 ################
 ## map / map! ##
 ################
 
+# The following type signature for map() matches any list of AbstractArrays,
+# provided at least one is a static array.
 @inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
     _map(f, same_size(as...), as...)
 end
@@ -16,7 +20,7 @@ end
     newT = :(Core.Inference.return_type(f, Tuple{$(eltypes...)}))
     return quote
         @_inline_meta
-        @inbounds return similar_type(typeof(_first_static(a...)), $newT, Size(S))(tuple($(exprs...)))
+        @inbounds return similar_type(typeof(_first(a...)), $newT, Size(S))(tuple($(exprs...)))
     end
 end
 
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -1,22 +1,10 @@
-@inline _first(a1, as...) = a1
-
 ################
 ## map / map! ##
 ################
 
-@inline map(f, as::StaticArray...) =
+@inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
     _map(f, same_size(as...), as...)
-# Mixed StaticArray + AbstractArray; various versions to avoid ambiguities.
-# With the versions below, if a StaticArray isn't present in the first two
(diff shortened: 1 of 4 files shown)


src/traits.jl
```diff
@@ -82,11 +82,12 @@ Length(::Type{SA}) where {SA <: StaticArray} = Length(Size(SA))
 @pure length(::Size{S}) where {S} = length(S)
 @pure length_val{S}(::Size{S}) = Val{length(S)}
 
-@pure Base.:(==){S}(::Size{S}, s::Tuple{Vararg{Int}}) = S == s
-@pure Base.:(==){S}(s::Tuple{Vararg{Int}}, ::Size{S}) = s == S
+# Note - using === here, as Base doesn't inline == for tuples as of julia-0.6
```
andyferris: nice catch :)

src/mapreduce.jl
```diff
@@ -1,25 +1,24 @@
...
-    _map(f, same_size(a, b...), a, b...)
-end
-
-@generated function _map(f, ::Size{S}, a::StaticArray...) where {S}
+@inline map(f, as::StaticArray...) =
+    _map(f, same_size(as...), as...)
+# Mixed StaticArray + AbstractArray; various versions to avoid ambiguities.
+# With the versions below, if a StaticArray isn't present in the first two
+# arguments, we'll end up in Base.map() instead.
+@inline map(f, a1::StaticArray, as::AbstractArray...) =
```
andyferris: Yesterday, I saw an unofficial "trick" to get "at least one of these" that goes something like:
```julia
map(Union{SA, AbstractArray}...) where {SA <: StaticArray} = ...
```
It works because the method is ignored unless `SA` can be found and determined (all method parameters must be specified in order to be counted as a match). Not sure if it will work for us (what if there are multiple `SA` types) and not sure if it's a good idea (it could be fragile, also Keno or someone was saying the requirement for `SA` to be determinable might be dropped - but I didn't see how the method body could make sense in this case).

src/traits.jl
```diff
@@ -104,3 +105,27 @@ Length(::Type{SA}) where {SA <: StaticArray} = Length(Size(SA))
...
+"""
+@inline _size(a) = size(a)
+@inline _size(a::StaticArray) = Size(a)
+
+# Return first static Size from a set of arrays
+@inline _first_static_size(a1::StaticArray, as...) = Size(a1)
+@inline _first_static_size(a1, as...) = _first_static_size(as...)
+@inline _first_static_size() = throw(ArgumentError("No StaticArray found in argument list"))
+
+# Returns the common Size of the inputs (or else throws a DimensionMismatch)
```
andyferris: Thanks for the re-org, this is better.

src/mapreduce.jl
c42f: Very neat, I'll try this :-)

c42f: This seems to work!  I'll go with it.

c42f:
Thanks for the comments Andy.  I've reorganized to improve codegen for `same_size()`.  I think the problem we were seeing this afternoon was mostly due to me not understanding the slurping penalty when looking at the generated code.  However, I've rearranged it anyway to put the error formatting out of the way on a cold code path.  Generated assembly looks much better now, eg:

```asm
julia> @code_native SVector(1,2,3) + [1,2,3]
        .text
Filename: linalg.jl
        pushq   %rbp
        movq    %rsp, %rbp
        pushq   %r15
        pushq   %r14
        pushq   %r12
        pushq   %rbx
        subq    $48, %rsp
        movq    %rdx, %r15
        movq    %rsi, %rbx
        movq    %rdi, %rax
;; check sizes (roughly starting here...)
        movabsq $140194643302176, %r14  # imm = 0x7F819BE9D720
        movq    %fs:0, %rdi
        addq    $-10904, %rdi           # imm = 0xD568
        vxorps  %xmm0, %xmm0, %xmm0
        vmovups %xmm0, -56(%rbp)
        movq    $0, -40(%rbp)
        movq    $6, -72(%rbp)
        movq    (%rdi), %rcx
        movq    %rcx, -64(%rbp)
        leaq    -72(%rbp), %rcx
        movq    %rcx, (%rdi)
Source line: 129
        movq    51744128(%r14), %rcx
        cmpq    24(%r15), %rcx
        jne     L158
;; Add vectors
Source line: 19
        movq    (%r15), %rcx
        movq    (%rbx), %rdx
        movq    8(%rbx), %rsi
        addq    (%rcx), %rdx
        addq    8(%rcx), %rsi
        movq    16(%rbx), %rbx
        addq    16(%rcx), %rbx
Source line: 12
        movq    %rdx, (%rax)
        movq    %rsi, 8(%rax)
        movq    %rbx, 16(%rax)
        movq    -64(%rbp), %rcx
        movq    %rcx, (%rdi)
        addq    $48, %rsp
        popq    %rbx
        popq    %r12
        popq    %r14
        popq    %r15
        popq    %rbp
        retq
;; Call error handling, only if necessary.
L158:
        leaq    -56(%rbp), %r12
Source line: 126
        movq    %r14, -56(%rbp)
        movabsq $jl_gc_pool_alloc, %rax
        movl    $1472, %esi             # imm = 0x5C0
        movl    $32, %edx
        callq   *%rax
        leaq    124130416(%r14), %rcx
        movq    %rcx, -8(%rax)
        movq    16(%rbx), %rcx
        movq    %rcx, 16(%rax)
        vmovups (%rbx), %xmm0
        vmovups %xmm0, (%rax)
        movq    %rax, -48(%rbp)
        movq    %r15, -40(%rbp)
        addq    $157996272, %r14        # imm = 0x96AD4F0
        movabsq $jl_invoke, %rax
        movl    $3, %edx
        movq    %r14, %rdi
        movq    %r12, %rsi
        callq   *%rax
        ud2
        nop
```

c42f:
The appveyor build error is very odd, I don't think it's related to this PR.
https://ci.appveyor.com/project/andyferris/staticarrays-jl-b6vmi/build/1.0.154/job/un7afg2qcn4w56rs

Possibly related is a build failure on master on Travis
https://travis-ci.org/JuliaArrays/StaticArrays.jl/jobs/218327551

Both of these are in the 2x2 linear system solvers, but with different datatypes, and possibly completely unrelated according to the error messages. I can't reproduce any of this on my machine yet :-/

andyferris:
The solver tests are very naive and can result in singular matrices. This should be fixed separately, not blocking this PR.

src/mapreduce.jl
```diff
@@ -1,25 +1,12 @@
...
-
-@inline _first(a1, as...) = a1
-
 ################
 ## map / map! ##
 ################
 
-@inline function map(f, a::StaticArray, b::StaticArray...)
-    _map(f, same_size(a, b...), a, b...)
+@inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
```
andyferris: I feel this signature is tricky enough that a code comment might be useful.

src/mapreduce.jl
```diff
@@ -29,7 +16,7 @@ end
     newT = :(Core.Inference.return_type(f, Tuple{$(eltypes...)}))
     return quote
         @_inline_meta
-        @inbounds return similar_type(typeof(_first(a...)), $newT)(tuple($(exprs...)))
+        @inbounds return similar_type(typeof(_first_static(a...)), $newT, Size(S))(tuple($(exprs...)))
```
andyferris: Is `_first_static` the ideal choice here? (Just asking - I'm not sure what the answer is).

Until/unless we have a proper promotion mechanism, having some memorable rule like `similar` and `similar_type` being called on the first argument has certain advantages. 

It's hard to predict what the user wants here, and we don't typically have the option of `map!` for when the user wants to make the choice explicitly, so perhaps a certain API for `map` (or `map_to_type`, etc) could be useful.

andyferris:
As a general comment in the `map` space, currently `Base.map` allows sizes that aren't `==` to match by ignoring trailing singleton dimensions. 

```julia
julia> map(+, rand(3), rand(3,1))
3×1 Array{Float64,2}:
 0.790038
 1.68035 
 1.32206
```

Personally, I don't love this, but it also makes sense to mirror `Base` behavior.

src/mapreduce.jl
c42f: You're right, I wasn't sure about this change.  I'm happy to change it back to the first argument.

src/mapreduce.jl
c42f: Sure, done.

c42f:
I think I'm happy with this now.

PR #137, opened 2017-04-04
Define `size` more explicitly, helping the compiler

Keno:


Keno:
Sorry, this isn't quite good enough. Will have an update in a sec.

andyferris:
Can we do this generically with #134?

Keno:
Yes, we just have to be a bit careful. The compiler has some aversions here. These functions are very carefully written such that constant propagates down to llvm level. Check with e.g.:
```
f(A) = size(A,2)
@code_llvm f(A)
```
for any static array. Should be a constant. If you see anything else you're in trouble.

PR #138, opened 2017-04-06
bring back fixedsizearrays

SimonDanisch:


SimonDanisch:
Uh, this seems to still pass, but I pulled in newest master and it doesn't pass anymore...
Also master doesn't pass on newest Julia master. Is there some branch fixing this that I should include here?
Error:
```Julia
Base.ArgumentError(msg="Size mismatch in Static Array parameters. Got size Tuple{1,2}, dimension 2 and length 1.")
```
@andyferris @c42f @Keno

SimonDanisch:
Lol, just spent way too much time to figure out, that `test_broken` exists, doesn't leave a stacktrace, but turns up as a broken test in the summary -.-
I think I fixed the errors I mention above by putting the error message of the generated function in quotes: https://github.com/JuliaArrays/StaticArrays.jl/blob/b1714d9d19b98238901320532185795ce6fe1d05/src/util.jl#L42, so that they don't make the body of the staged function impure!

src/FixedSizeArrays.jl
```diff
@@ -134,37 +148,46 @@ macro fixed_vector(name, parent)
...
         eltype_or(::Type{$(name)}, or) = or
-        eltype_or{T}(::Type{$(name){TypeVar(:S), T}}, or) = T
-        eltype_or{S}(::Type{$(name){S, TypeVar(:T)}}, or) = or
-        eltype_or{S, T}(::Type{$(name){S, T}}, or) = T
 
-        size_or(::Type{$(name)}, or) = or
-        size_or{T}(::Type{$(name){TypeVar(:S), T}}, or) = or
-        size_or{S}(::Type{$(name){S, TypeVar(:T)}}, or) = (S,)
-        size_or{S, T}(::Type{$(name){S, T}}, or) = (S,)
+        if VERSION < v"0.6.0-dev"
```
c42f: StaticArrays as a whole is now 0.6 only.  Is there any use in having this version check?

src/util.jl
```diff
@@ -39,22 +39,22 @@ end
...
 
 @generated function check_array_parameters{Size,T,N,L}(::Type{Size}, ::Type{T}, ::Type{Val{N}}, ::Type{Val{L}})
     if !all(x->isa(x, Int), Size.parameters)
-        throw(ArgumentError("Static Array parameter Size must be a tuple of Ints (e.g. `SArray{Tuple{3,3}}` or `SMatrix{3,3}`)."))
+        return :(throw(ArgumentError("Static Array parameter Size must be a tuple of Ints (e.g. `SArray{Tuple{3,3}}` or `SMatrix{3,3}`).")))
     end
 
     if L != tuple_prod(Size) || L < 0 || tuple_minimum(Size) < 0 || tuple_length(Size) != N
-        throw(ArgumentError("Size mismatch in Static Array parameters. Got size $Size, dimension $N and length $L."))
+        return :(throw(ArgumentError("Size mismatch in Static Array parameters. Got size $Size, dimension $N and length $L.")))
```
c42f: I think I'm still a bit confused about why this is absolutely necessary.  But I'm not surprised that this is another concession required for the new `@generated` rules :-)

src/util.jl
```diff
@@ -39,22 +39,22 @@ end
 @pure tuple_minimum(T::Tuple) = minimum(T)
 
 # Something doesn't match up type wise
-function check_array_parameters(Size, T, N, L)
-    (!isa(Size, DataType) || (Size.name !== Tuple.name)) && throw(ArgumentError("Static Array parameter Size must be a Tuple type, got $Size"))
-    !isa(T, Type) && throw(ArgumentError("Static Array parameter T must be a type, got $T"))
-    !isa(N.parameters[1], Int) && throw(ArgumenError("Static Array parameter N must be an integer, got $(N.parameters[1])"))
-    !isa(L.parameters[1], Int) && throw(ArgumentError("Static Array parameter L must be an integer, got $(L.parameters[1])"))
+@generated function check_array_parameters(Size, T, N, L)
```
c42f: I don't think this needs to be `@generated`, because this method only exists as a fallback signature for when the valid `check_array_parameters` signature further down fails to match.  So codegen does not need to be good for this one.

It took me five minutes to figure out why we even have two versions of `check_array_parameters` though.  Would be good to add a comment along the lines of

```
# This version of check_array_parameters() is a fallback signature to catch cases when invalid type
# parameters are passed.  It exists only for improved error reporting and should not be called by
# correct code.
```

src/StaticArrays.jl
```diff
@@ -93,7 +93,7 @@ include("eigen.jl")
 include("cholesky.jl")
 include("deque.jl")
 
-#include("FixedSizeArrays.jl")  # Currently defunct
+include("FixedSizeArrays.jl")  # Currently defunct
```
c42f: Obsolete comment.

src/util.jl
SimonDanisch: Oh.... That probably wasn't intended... I got a bit confused because I started testing with an old version of master, where this was still a generated function...

SimonDanisch:
>Which parts of this do you still need?

All!

I wouldn't mind to have this come as standard functionality of StaticArrays!

c42f:
> I wouldn't mind to have this come as standard functionality of StaticArrays

Certainly it would be great to have any distinct functionality which still lives in here as part of StaticArrays proper, provided it doesn't conflict with `Base.AbstractArray` conventions.  (For example, `Base.isnan` is an elementwise operation in julia 0.5, but `FixedSizeArrays.isnan` is a reduction. You could/should use `any(isnan.(v))` instead.)

src/util.jl
```diff
@@ -40,22 +40,20 @@ end
...
 # Something doesn't match up type wise
 function check_array_parameters(Size, T, N, L)
-    (!isa(Size, DataType) || (Size.name !== Tuple.name)) && throw(ArgumentError("Static Array parameter Size must be a Tuple type, got $Size"))
-    !isa(T, Type) && throw(ArgumentError("Static Array parameter T must be a type, got $T"))
-    !isa(N.parameters[1], Int) && throw(ArgumenError("Static Array parameter N must be an integer, got $(N.parameters[1])"))
-    !isa(L.parameters[1], Int) && throw(ArgumentError("Static Array parameter L must be an integer, got $(L.parameters[1])"))
+    (!isa(Size, DataType) || (Size.name !== Tuple.name)) && return throw(ArgumentError("Static Array parameter Size must be a Tuple type, got $Size"))
+    !isa(T, Type) && return throw(ArgumentError("Static Array parameter T must be a type, got $T"))
+    !isa(N.parameters[1], Int) && return throw(ArgumenError("Static Array parameter N must be an integer, got $(N.parameters[1])"))
+    !isa(L.parameters[1], Int) && return throw(ArgumentError("Static Array parameter L must be an integer, got $(L.parameters[1])"))
```
c42f: These `return`s don't do anything - leftovers from previous changes, I think?

src/util.jl
```diff
@@ -40,22 +40,20 @@ end
...
     # shouldn't reach here. Anything else should have made it to the function below
-    error("Internal error. Please file a bug")
+    return error("Internal error. Please file a bug")
 end
 
-@generated function check_array_parameters{Size,T,N,L}(::Type{Size}, ::Type{T}, ::Type{Val{N}}, ::Type{Val{L}})
+function check_array_parameters{Size,T,N,L}(::Type{Size}, ::Type{T}, ::Type{Val{N}}, ::Type{Val{L}})
```
c42f: This is the version of `check_array_parameters()` which will be called, and I think there was a reason to have it as a `@generated` function - it's fairly important that this does compile down to nothing in the case that the checks pass.

Did you check that this actually compiles down to nothing, even when it's not a generated function? I'd suggest you just leave this as it was, but include the fix you had before, to throw in the generated code, rather than in the generator itself.

SimonDanisch: Oh I see where my confusion comes from... Only one of them where `@generated` .
It's a pity, that Julia's dead code elimination just removes this in theory...

SimonDanisch:
>FixedSizeArrays.isnan is a reduction. You could/should use any(isnan.(v)) instead.)

Not 100% sure about that, now that people should use isnan.(Vector) for element wise application, which should disambiguate this case... Also, I feel like there are tons of examples, where StaticVectors are rightfully treated as scalars.
But I guess I can remove it, if that would make it hard to merge for you!

c42f:
> Not 100% sure about that, now that people should use isnan.(Vector) for element wise application, which should disambiguate this case

True, once 0.6 deprecations are of the way `isnan(::Vector)` could take on a new meaning in Base, in principle.  Though I'm not sure it should - as a Base function, would it mean `all(isnan.(v))` or `any(isnan.(v))` for instance?  Seems potentially ambiguous.

> But I guess I can remove it, if that would make it hard to merge for you!

Not at all, I'm fine merging it inside the `StaticArrays.FixedSizeArrays` submodule. I just meant this as an example of where the utilities in `StaticArrays.FixedSizeArrays` diverge from `AbstractArray`, which would make me uncomfortable about including them in the main `StaticArrays` module.

src/util.jl
c42f: Right, this is why I suggested adding that explanatory comment in my previous review of this PR ;-)

SimonDanisch:
Should we merge it as is and get rid of the fixedsizearray module in a later release?

andyferris:
FYI the reductions generally take a function, like `any(isnan, vector)` (much faster than `any(isnan.(vector))`)

src/util.jl
andyferris: Yeah, ditch the `return`s.

src/util.jl
andyferris: Yeah this is a bit annoying. Generally the compiler tries to compute things which are either pure or "effect free" and perform dead branch elimination, but I think this function is just too complex. However inference is getting better all the time at this so its worth keeping an eye on in the future.

andyferris:
Simon - I'm happy that you have worked on this. Let's fix (revert?) `util.jl` and merge.

andyferris:
Actually, can you explain the `broadcast` changes?

SimonDanisch:
> Actually, can you explain the broadcast changes?
(Would it have been better to have generic bugfixes in a separate PR?)

Yes I guess... Problem is, it's much nicer (for me) to have all fixes for GeometryTypes/GLVisualize in one branch....

The change makes `round.(Int, vec)` work and opens up support for similar behaviour to what base is doing - treating everything not an array as a scalar!

andyferris:
I thought so - got it!

Base in v0.6 actually has a rather fancy implementation of `broadcast`. Ideally we'd hook into this (assuming it is inlined).

SimonDanisch:
ooooh, that'd be sweet :) 

I checked out utils.jl and only put the error in quotes for the generated function... otherwise the error looks just plain horrible!

c42f:
> I checked out utils.jl and only put the error in quotes for the generated function

Thanks Simon, this looks a lot better now.

src/util.jl
andyferris: I also don't get why this helps... I had actually removed the quotes from the errors in most parts of the package because they seemed to not be required...

c42f: Yes arguably it's better to throw an error early rather than late (ie, in the generator rather than the generated code - so to leave this without the quotes).  I just figured it was some funky new limitation of `@generated`... @SimonDanisch what exactly happens without this?

SimonDanisch:
Okay, that was way too easy :D

andyferris:
This needs a rebase

SimonDanisch:
Done!

PR #139, opened 2017-04-07
Workaround for type inference problem in `index_sizes`

martinholters:
Wokraround for JuliaLang/julia#21244.

I briefly tried adding a test case similar to the one from JuliaLang/julia#21244, but it seems to need the `struct Foo` which cannot be defined inside a `@testset`. Is it worth to define such a type globally in `runtests.jl`?

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -52,18 +52,17 @@ end
 ## Indexing utilities  ##
 #########################
 
-@pure increment(::Type{Val{N}}) where {N} = Val{N+1}
+@pure tail(::Type{Size{S}}) where {S} = Size{Base.tail(S)}
+@inline tail(::S) where {S<:Size} = tail(S)()
 
-@inline index_sizes(s::Size, inds...) = _index_sizes(s, Val{1}, (), inds...)
-@inline _index_sizes(s::Size, ::Type{Val{N}}, x::Tuple) where {N} = x
-@inline _index_sizes(s::Size, v::Type{Val{N}}, x::Tuple, ::Int, inds...) where {N} = _index_sizes(s, increment(v), (x..., Size()), inds...)
-@inline _index_sizes(s::Size, v::Type{Val{N}}, x::Tuple, a::StaticArray, inds...) where {N} = _index_sizes(s, increment(v), (x..., Size(a)), inds...)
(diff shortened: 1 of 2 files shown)


martinholters:
Ah, just after (of course!) submitting the PR, I realized two things:
1. The new `index_sizes(s::Size, inds...)` will error if `s` contains fewer elements than `inds`. Is that a problem? The fix should be trivial.
2. `code_warntype(DevNull, StaticArrays.index_sizes, Tuple{Vararg{Any}})` is able to trigger the error so would make a good test. Update incoming...

martinholters:
Fixed the the case for `s` containing fewer elements than `inds` (in case that matters).

andyferris:
Thanks for working on this!

I'm on vacation for another week, so I can't have an in depth look at the moment. I can say I'm very happy this is being looked at.

I assume someone has (or will) check the generated code? Other than that, I think I'm ok with this PR.

martinholters:
I don't think I'm sufficiently familiar with this package to check the generated code. I've done some very basic benchmarking of indexing:
```julia
sm = @SMatrix [1 2 3; 4 5 6]
sv = @SVector [1, 2]
@benchmark $sm[$sv, 1]
@benchmark $sm[1, $sv]
@benchmark $sm[:, $sv]
@benchmark $sm[$sv, :]
@benchmark $sm[1, :]
@benchmark $sm[1, 1]
@benchmark $sm[:, 1]
@benchmark $sm[:, :]
```
They are all within measurement between master and this PR. I've also checked `@code_warntype sm[:, 1]` as one example and it is identical between master and this PR. But these probably aren't the critical cases?

c42f:
Thanks for tackling this one. It's very pleasing that your new implementation avoids any mention of `Val` :-)

Regarding your `struct Foo` problem, the way I've been dealing with this is as follows:

```julia
@testset "blah" begin

# ... lots of other tests ...

# Now the test which needs a custom type
@eval struct Foo
    i::Int
end

# Some test using Foo
@test Foo(1).i == 1

end
```

Of course the type `Foo` is still defined globally within the current module, but at least the code for it is next to the test which needs it.

martinholters:
I don't think it would be worth the effort, so: no. If anyone disagrees, I'd be happy to add it, but not in the next couple of days.

c42f:
Sure no problems. I had a quick look but it seems to be obscure behavior which would be better as a compiler test.

Issue #131, opened 2017-03-31
SVector + Vector: Julia 0.5 vs 0.6


krcools:
Not sure if this is by design or not but I noticed that on julia-0.5 a `SVector` added to an `Array` results in an `SVector`, wheras on julia-0.6 it results in an `Array`.

c42f:
Yes, these were dropped in #121 (see the description), but perhaps they should be reimplemented.

It should be very easy to implement the map-like operations again, I'll have a go at that first.

andyferris:
I support reimplementing these.

c42f:
Started in #136

c42f:
Fixed in #136 (see also #140)

PR #140, opened 2017-04-11
Reinstate StaticMatrix * AbstractVector -> StaticVector

c42f:
This was present prior to the big 0.6 rewrite, and should be reinstated for convenience as the output can be size inferred.  This PR is a bit copy-and-paste-ish, though not a lot worse than what's there already, I guess.

This is an associated fix for #131


c42f:
I think I'll merge this - it's a minor addition which should be pretty harmless.

andyferris:
Thanks Chris - LGTM

PR #145, opened 2017-04-24
Define `size` more explicitly, helping the compiler

c42f:
@Keno this is an updated version of your PR.

Resolves #137

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -1,11 +1,13 @@
 length(a::SA) where {SA <: StaticArray} = prod(Size(SA))
 length(a::Type{SA}) where {SA <: StaticArray} = prod(Size(SA))
 
-size(::StaticArray{S}) where {S} = get(Size(S))
-@pure size(::Type{<:StaticArray{S}}) where {S} = get(Size(S))
-
-size(::SA, d::Int) where {SA <: StaticArray} = Size(SA)[d]
-@pure size(::Type{SA}, d::Int) where {SA <: StaticArray} = Size(SA)[d]
+@pure size(::Type{<:StaticArray{S}}) where S = tuple(S.parameters...)
+@inline function size(t::Type{<:StaticArray}, d::Int)
+    S = size(t)
+    d > length(S) ? 1 : S[d]
+end
+@inline size(a::StaticArray) = size(typeof(a))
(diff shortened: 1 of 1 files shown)


src/abstractarray.jl
```diff
@@ -1,11 +1,13 @@
...
 length(a::Type{SA}) where {SA <: StaticArray} = prod(Size(SA))
 
-size(::StaticArray{S}) where {S} = get(Size(S))
-@pure size(::Type{<:StaticArray{S}}) where {S} = get(Size(S))
-
-size(::SA, d::Int) where {SA <: StaticArray} = Size(SA)[d]
-@pure size(::Type{SA}, d::Int) where {SA <: StaticArray} = Size(SA)[d]
+@pure size(::Type{<:StaticArray{S}}) where S = tuple(S.parameters...)
+@inline function size(t::Type{<:StaticArray}, d::Int)
+    S = size(t)
```
andyferris: Is generated code better with `Size(t)`?

c42f:
Yep, it's better.  I didn't spend a lot of time on this, but I did check the the following using @Keno's original example:

Without patch:
```julia
julia> foo(A) = size(A,2)
julia> @code_llvm foo(ones(SMatrix{2,3}))

define i64 @julia_foo_60424(%SArray* nocapture readonly dereferenceable(48)) #0 !dbg !5 {
top:
  %1 = call i64 @julia_size_60425(%SArray* nocapture nonnull readonly %0, i64 2)
  ret i64 %1
}
```

With patch:
```julia
julia> foo(A) = size(A,2)
julia> @code_llvm foo(ones(SMatrix{2,3}))

define i64 @julia_foo_60360(%SArray* nocapture readonly dereferenceable(48)) #0 !dbg !5 {
top:
  ret i64 3
}
```

It'll be a bit of a grind, but we need to put some solid work into benchmarking StaticArrays. This kind of spot checking of the assembly is all very well, but I'm pretty sure we're going to fall into a massive performance hole sometime soon without realizing it :-/

c42f:
I checked a few other things like `bar(A) = size(A)[1] + size(A)[2]` - all looks good afaict.

Keno:
This looks fine to me.

andyferris:
Thanks guys!!

PR #147, opened 2017-04-25
Clean up and fix pointer code

### LINKED ISSUE #146: Concatenating arrays of FieldVectors fails due to ambiguity
On Julia version 0.6.0-pre.beta.293 (2017-04-23 19:58 UTC), and master of StaticArrays:

```
julia> struct Point{T} <: FieldVector{T}
         x::T
         y::T
       end

julia> [[Point(3,4)]; [Point(1,2), Point(3,4)]]
ERROR: MethodError: Base.unsafe_convert(::Type{Ptr{Void}}, ::Ptr{Point{Int64}}) is ambiguous. Candidates:
  unsafe_convert(::Type{P}, x::Ptr) where P<:Ptr in Base at essentials.jl:154
  unsafe_convert(::Type{Ptr{T}}, m::Ref{FV}) where {T, FV<:StaticArrays.FieldVector} in StaticArrays at /Users/ajkeller/.julia/v0.6/StaticArrays/src/FieldVector.jl:30
Possible fix, define
  unsafe_convert(::Type{Ptr{T}}, ::Ptr{FV<:StaticArrays.FieldVector})
Stacktrace:
 [1] vcat(::Array{Point{Int64},1}, ::Array{Point{Int64},1}) at ./array.jl:1120
```

This works fine on Julia 0.5.1 and StaticArrays 0.3.1.

andyferris:
Fix #146

### DIFF
--- a/src/FieldVector.jl
+++ b/src/FieldVector.jl
@@ -23,8 +23,8 @@ abstract type FieldVector{N, T} <: StaticVector{N, T} end
 @propagate_inbounds setindex!(v::FieldVector, x, i::Int) = setfield!(v, i, x)
 
 # See #53
-Base.cconvert{T}(::Type{Ptr{T}}, v::FieldVector) = Ref(v)
-Base.unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, m::Ref{FV}) =
+Base.cconvert{T}(::Type{Ptr{T}}, v::FieldVector) = Base.RefValue(v)
+Base.unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, m::Base.RefValue{FV}) =
     _unsafe_convert(Ptr{T}, eltype(FV), m)
-_unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, ::Type{T}, m::Ref{FV}) =
+_unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, ::Type{T}, m::Base.RefValue{FV}) =
          Ptr{T}(Base.unsafe_convert(Ptr{FV}, m))
--- a/src/MMatrix.jl
+++ b/src/MMatrix.jl
@@ -101,10 +101,6 @@ end
 
 @inline Tuple(v::MMatrix) = v.data
 
-@inline function Base.unsafe_convert{N,M,T}(::Type{Ptr{T}}, m::MMatrix{N,M,T})
-    Base.unsafe_convert(Ptr{T}, Base.data_pointer_from_objref(m))
-end
-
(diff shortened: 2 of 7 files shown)


c42f:
Looks good, especially the deduplication.  Though I don't really understand the ambiguity with out digging further.

One quibble - this might have API implications if someone was using `ArrayRef` via `Ref(::AbstractArray, ::Int)`, and expecting to pass that to a ccall?

src/FieldVector.jl
```diff
@@ -23,8 +23,8 @@ abstract type FieldVector{N, T} <: StaticVector{N, T} end
...
 
 # See #53
-Base.cconvert{T}(::Type{Ptr{T}}, v::FieldVector) = Ref(v)
-Base.unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, m::Ref{FV}) =
+Base.cconvert{T}(::Type{Ptr{T}}, v::FieldVector) = Base.RefValue(v)
+Base.unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, m::Base.RefValue{FV}) =
     _unsafe_convert(Ptr{T}, eltype(FV), m)
-_unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, ::Type{T}, m::Ref{FV}) =
+_unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, ::Type{T}, m::Base.RefValue{FV}) =
          Ptr{T}(Base.unsafe_convert(Ptr{FV}, m))
```
c42f: Isn't this broken when `T` is not a primitive type?

No more broken than before, but still...

andyferris: Not sure why you are saying that?

andyferris:
Isn't `ArrayRef` just for `Array`?

c42f:
> Isn't ArrayRef just for Array?

Oddly, it's defined for all `AbstractArray`

src/FieldVector.jl
c42f: Sorry, not primitive type.  Broken when `T` is some kind of `mutable struct`?

A `FieldVector` of these `T` is laid out as an array of pointers

```C
struct FV
{
    T* data[N];
};
```

Then creating a `RefValue{FV}` basically gives a `FV*`, which is a `T**`, which can't be meaningfully converted to a `T*` using `Ptr{T}`.  (Sorry for the mix of C and julia notation)

c42f:
After looking at this for a little, I must admit I'm a little confused ;-)

On the whole, I'm somewhat inclined to just merge it - I don't *think* there's anything broken here which wasn't broken before, and it fixes a known problem in practice.

PR #150, opened 2017-04-27
WIP Return `SUnitRange` from `indices(::StaticArray)`

andyferris:
Needs testing.

Eventually we might be able to move to generic indexing like `OffsetArrays`. But even without that, it's nice to be able to discuss and introspect the indices (at compile time).

@timholy are there some caveats to doing this? (I already ran into a few surprising things - perhaps you know of any other gotchas?)

### DIFF
--- a/src/SUnitRange.jl
+++ b/src/SUnitRange.jl
@@ -35,3 +35,26 @@ end
 Base.unsafe_length(r::SUnitRange) = length(r)
 @inline first(r::SUnitRange{Start}) where {Start} = Start # matches Base.UnitRange when L == 0...
 @inline endof(r::SUnitRange{Start, L}) where {Start, L} = L
+
+(==)(::SUnitRange{Start, L}, ::SUnitRange{Start, L}) where {Start, L} = true
+(==)(::SUnitRange{Start1, L1}, ::SUnitRange{Start2, L2}) where {Start1, Start2, L1, L2} = false
+
+(==)(::SUnitRange{1, L}, r::Base.OneTo) where {L} = L == r.stop
+(==)(::SUnitRange, ::Base.OneTo) = false
+(==)(r::Base.OneTo, ::SUnitRange{1, L}) where {L} = L == r.stop
+(==)(::Base.OneTo, ::SUnitRange) = false
+
+start(r::SUnitRange) = 1
+next(r::SUnitRange, i) = (r[i], i+1)
+done(r::SUnitRange{Start, L}, i) where {Start, L} = i > L
+
+@pure Base.UnitRange{Int}(::SUnitRange{Start, L}) where {Start, L} = Start : (Start + L - 1)
+@pure Base.Slice(::SUnitRange{Start, L}) where {Start, L} = Base.Slice(Start : (Start + L - 1)) # TODO not optimal
+
+function Base.checkindex(::Type{Bool}, ::SUnitRange{Start, L}, r::UnitRange{Int}) where {Start, L}
+    return first(r) < Start | last(r) >= Start + L
+end
+
+function Base.checkindex(::Type{Bool}, ::SUnitRange{Start, L}, r::Base.Slice{UnitRange{Int}}) where {Start, L}
+    return first(r) < Start | last(r) >= Start + L
+end
--- a/src/SUnitRange.jl
+++ b/src/SUnitRange.jl
@@ -30,3 +30,8 @@ show(io::IO, ::Type{SUnitRange}) = print(io, "SUnitRange")
 function show(io::IO, ::MIME"text/plain", ::SUnitRange{Start, L}) where {Start, L}
     print(io, "SUnitRange($Start,$(Start + L - 1))")
 end
+
+# For this type to be usable as `indices`, they need to support some more stuff
+Base.unsafe_length(r::SUnitRange) = length(r)
+@inline first(r::SUnitRange{Start}) where {Start} = Start # matches Base.UnitRange when L == 0...
+@inline endof(r::SUnitRange{Start, L}) where {Start, L} = L
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -11,7 +11,9 @@ import Base: getindex, setindex!, size, similar, vec, show,
              fill!, det, inv, eig, eigvals, trace, vecnorm, norm, dot, diagm,
              sum, diff, prod, count, any, all, sumabs, sumabs2, minimum,
              maximum, extrema, mean, copy, rand, randn, randexp, rand!, randn!,
-             randexp!, normalize, normalize!, indices, first, endof
+             randexp!, normalize, normalize!, indices, first, endof, start, next, done
+
+import Base: ==
 
 export StaticScalar, StaticArray, StaticVector, StaticMatrix
 export Scalar, SArray, SVector, SMatrix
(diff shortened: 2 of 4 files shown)


c42f:
Looks sensible so far.

andyferris:
I somehow broke a *macro* with this code...

andyferris:
whow - so much stack overflow
```
[119997] != at ./operators.jl:129 [inlined]
[119998] ==(::StaticArrays.SUnitRange{1,2}, ::StaticArrays.SUnitRange{1,2}) at ./abstractarray.jl:1513
[119999] ==(::Tuple{StaticArrays.SUnitRange{1,2}}, ::Tuple{StaticArrays.SUnitRange{1,2}}) at ./tuple.jl:267
```

andyferris:
Getting deeper... now it's `rediced_indices0` and conversions to `Base.Slice`. Oh dear.

andyferris:
it appears to me that we should make `AbstractUnitRange` into a trait...

andyferris:
The alternative here is to define a new `AbstractUnitRange` which is a static version of `OneTo`, such that these are the indices of a static array.

(My eventual "take over the world" goal here is to use a trait to describe if an array is statically sized by asking the question if all it's indices are static, and organize dispatch around this).

src/SUnitRange.jl
```diff
@@ -30,3 +30,31 @@ show(io::IO, ::Type{SUnitRange}) = print(io, "SUnitRange")
...
+@inline first(r::SUnitRange{Start}) where {Start} = Start # matches Base.UnitRange when L == 0...
+@inline endof(r::SUnitRange{Start, L}) where {Start, L} = L
+
+(==)(::SUnitRange{Start, L}, ::SUnitRange{Start, L}) where {Start, L} = true
+(==)(::SUnitRange{Start1, L1}, ::SUnitRange{Start2, L2}) where {Start1, Start2, L1, L2} = false
+
+(==)(::SUnitRange{1, L}, r::Base.OneTo) where {L} = L == r.stop
+(==)(::SUnitRange, ::Base.OneTo) = false
+(==)(r::Base.OneTo, ::SUnitRange{1, L}) where {L} = L == r.stop
+(==)(::Base.OneTo, ::SUnitRange) = false
```
timholy: You may also want a variant for generic `AbstractUnitRange`.

src/SUnitRange.jl
```diff
@@ -30,3 +30,31 @@ show(io::IO, ::Type{SUnitRange}) = print(io, "SUnitRange")
...
+
+start(r::SUnitRange) = 1
+next(r::SUnitRange, i) = (r[i], i+1)
+done(r::SUnitRange{Start, L}, i) where {Start, L} = i > L
+
+@pure Base.UnitRange{Int}(::SUnitRange{Start, L}) where {Start, L} = Start : (Start + L - 1)
```
timholy: Any particular reason this needs to be `@pure`?

src/SUnitRange.jl
```diff
@@ -30,3 +30,31 @@ show(io::IO, ::Type{SUnitRange}) = print(io, "SUnitRange")
...
+@pure Base.Slice(::SUnitRange{Start, L}) where {Start, L} = Base.Slice(Start : (Start + L - 1)) # TODO not optimal
```
timholy: This is not wrong but is perhaps pushing the limits. `Base.Slice` is a (deliberately) limited version of [IdentityRange](https://github.com/JuliaArrays/IdentityRanges.jl), and we only say that two arrays are equal if they have *both* the same values and the same indices. Since this is a constructor rather than a conversion, you might be able to get away with this, but I'd also add
```julia
convert(::Type{Base.Slice}, r::SUnitRange{1,L}) = Base.Slice(r)
convert(::Type{Base.Slice}, r::SUnitRange) = error("$r cannot be converted to Base.Slice, consider `Base.Slice(r)` instead")
```
(`convert` has an implicit promise of preserving at least approximate equality relationships, see the many discussions about whether it's OK to strip physical units e.g., `convert(Int, 3meter)`.)

src/SUnitRange.jl
```diff
@@ -30,3 +30,31 @@ show(io::IO, ::Type{SUnitRange}) = print(io, "SUnitRange")
...
+
+function Base.checkindex(::Type{Bool}, ::SUnitRange{Start, L}, r::UnitRange{Int}) where {Start, L}
```
timholy: Any particular reason not to define one method for `AbstractUnitRange`?

timholy:
> it appears to me that we should make AbstractUnitRange into a trait...

Oh, wow, I see what you mean. Yikes, this is indeed an issue. I think your plan to define a whole new static type that subtypes from `AbstractUnitRange` is probably the best solution for now.

andyferris:
Cool - thanks Tim for the feedback! You've confirmed everything I thought you would say ;)

timholy:
I take predictability as a complement. (In great contrast to my current president...)

vlc1:
I'm really interested in using `SUnitRange`s, and was wondering whether this PR had led to anything?

Also, is there a reason why `SOneTo` is a subtype of `AbstractUnitRange` while `SUnitRange` derives from `StaticVector`?

andyferris:
@c42f, thoughts on whether we are close to ready for doing this?

> Also, is there a reason why SOneTo is a subtype of AbstractUnitRange while SUnitRange derives from StaticVector?

Yes - you can only have one supertype. Really we need multiple inheretence or traits or interfaces or something to do this better. Given we have to choose one... trust me - it's far easier to replicate `AbstractUnitRange` functionality, rather than `StaticVector`.

vlc1:
Thanks for your feedback. It sounds that doing things right requires refactoring quite a few things, I'm not sure I have the skills to do this.

However a quick way to get what I need would be to start from the `static-indices` branch while making `SUnitRange` a subtype of `StaticVector` (for the reasons you mentioned).

Would that make sense?

c42f:
> @c42f, thoughts on whether we are close to ready for doing this?

Last I looked, I just had the general feeling that these changes are both highly desirable and have the potential to be somewhat breaking. Other than that, I feel like I don't have any special insight on how hard it would be to do this, and I'd only gain some by trying to do the actual work ;-)

hyrodium:
I'll close this PR because:
* This seems stale.
* Currently, `axes(::StaticArray)` returns tuple of `SOneTo`. (`indices` was deprecated in Julia v0.7.0, and removed in Julia v1.0.0.)
  * Note that on Julia v0.6.0, `indices(::SMatrix)` returns tuple of `Base.OneTo`.

```julia
julia> versioninfo()
Julia Version 0.6.0
Commit 9036443 (2017-06-19 13:05 UTC)
Platform Info:
  CPU: AMD Ryzen 7 2700X Eight-Core Processor
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Barcelona)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, generic)

julia> m = @SMatrix rand(3,3)
3×3 StaticArrays.SArray{Tuple{3,3},Float64,2,9}:
 0.277928  0.778035  0.645978
 0.315307  0.682505  0.176565
 0.984043  0.505622  0.843649

julia> indices(m)
(Base.OneTo(3), Base.OneTo(3))
```

```julia
julia> versioninfo()
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  CPU: AMD Ryzen 7 2700X Eight-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, znver1)

julia> m = @SMatrix rand(3,3)
 0.994084   0.435498   0.579111
 0.120314   0.935061   0.435616
 0.0265949  0.0811534  0.772815

julia> axes(m)
(SOneTo(3), SOneTo(3))
```

PR #157, opened 2017-05-05
Capture complex outer product

### LINKED ISSUE #156: Outer product return type change with non-Real scalar types
On StaticArrays 0.3.0 (Julia 0.5.1),
```julia
using StaticArrays
x = zero(SVector{3, Number})
x * x'
```
results in
```
3×3 StaticArrays.SMatrix{3,3,Int64,9}:
 0  0  0
 0  0  0
 0  0  0
```
while on StaticArrays 0.5.0 (latest Julia nightly) it results in
```
3×3 Array{Int64,2}:
 0  0  0
 0  0  0
 0  0  0
```
because the generic
```
*(vec::AbstractArray{T,1} where T, rowvec::RowVector) in Base.LinAlg at linalg/rowvector.jl:183
```
is called.
The scalar type (`Number` above) doesn't have to be abstract; I ran into this with the concrete `SymPy.Sym` type.

andyferris:
Fixes #156


andyferris:
This is missing a line of code that got put into #158. If we merge both, we'll be OK.

c42f:
Ok, I've reviewed this over at #158 instead.

PR #158, opened 2017-05-05
Promote hcat vcat

### LINKED ISSUE #155: hcat type promotion
```julia
using StaticArrays
m1 = rand(SMatrix{1, 1, Float32})
m2 = rand(SMatrix{1, 1, Float64})
[m1 m2]
```
results in
```
1×2 StaticArrays.SArray{Tuple{1,2},Float32,2,2}:
 0.594061  0.283404
```
Should be a `Float64` matrix.

andyferris:
Fixes #155. (Based on #157).

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -103,7 +103,7 @@ end
 
     return quote
         @_inline_meta
-        @inbounds return similar_type(a, Size($Snew))(tuple($(exprs...)))
+        @inbounds return similar_type(a, promote_type(eltype(a), eltype(b)), Size($Snew))(tuple($(exprs...)))
     end
 end
 # TODO make these more efficient
@@ -129,7 +129,7 @@ end
 
     return quote
         @_inline_meta
-        @inbounds return similar_type(a, Size($Snew))(tuple($(exprs...)))
+        @inbounds return similar_type(a, promote_type(eltype(a), eltype(b)), Size($Snew))(tuple($(exprs...)))
     end
 end
 # TODO make these more efficient
@@ -297,8 +297,8 @@ end
     end
 end
 
-# TODO same for `RowVector`?
 @inline Size(::Union{RowVector{T, SA}, Type{RowVector{T, SA}}}) where {T, SA <: StaticArray} = Size(1, Size(SA)[1])
+@inline Size(::Union{RowVector{T, CA}, Type{RowVector{T, CA}}} where CA <: ConjVector{<:Any, SA}) where {T, SA <: StaticArray} = Size(1, Size(SA)[1])
 @inline Size(::Union{Symmetric{T,SA}, Type{Symmetric{T,SA}}}) where {T,SA<:StaticArray} = Size(SA)
 @inline Size(::Union{Hermitian{T,SA}, Type{Hermitian{T,SA}}}) where {T,SA<:StaticArray} = Size(SA)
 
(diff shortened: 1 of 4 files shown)


andyferris:
Test failure was on Julia 0.7, I'm tempted to ignore it.

src/linalg.jl
```diff
@@ -297,8 +297,8 @@ end
     end
 end
 
-# TODO same for `RowVector`?
 @inline Size(::Union{RowVector{T, SA}, Type{RowVector{T, SA}}}) where {T, SA <: StaticArray} = Size(1, Size(SA)[1])
+@inline Size(::Union{RowVector{T, CA}, Type{RowVector{T, CA}}} where CA <: ConjVector{<:Any, SA}) where {T, SA <: StaticArray} = Size(1, Size(SA)[1])
```
andyferris: Oops. This line of code was meant for #157. I suggest we review all changes here and just merge both (or this).

fredrikekre:
failure looks to be this: https://github.com/JuliaLang/julia/pull/21684#issuecomment-299362899

src/matrix_multiply.jl
```diff
@@ -36,11 +36,13 @@ promote_matprod{T1,T2}(::Type{T1}, ::Type{T2}) = typeof(zero(T1)*zero(T2) + zero
 @inline *(A::StaticMatrix, B::StaticMatrix) = _A_mul_B(Size(A), Size(B), A, B)
 @inline *(A::StaticVector, B::StaticMatrix) = *(reshape(A, Size(Size(A)[1], 1)), B)
 @inline *(A::StaticVector, B::RowVector{<:Any, <:StaticVector}) = _A_mul_B(Size(A), Size(B), A, B)
+@inline *(A::StaticVector, B::RowVector{<:Any, CV} where CV <: ConjVector{<:Any, <:StaticVector}) = _A_mul_B(Size(A), Size(B), A, B)
```
c42f: Fancy.  This can also be written without the `where`, with `B::RowVector{<:Any,<:ConjVector{<:Any,<:StaticVector}}`

test/matrix_multiply.jl
```diff
@@ -44,6 +44,13 @@
         m = @SMatrix [1 2 3 4]
         v = @SVector [1, 2]
         @test @inferred(v*m) === @SMatrix [1 2 3 4; 2 4 6 8]
+
+        # Outer product
+        v2 = zeros(SVector{3, Int})
+        @test v2 * v2' === zeros(SMatrix{3, 3, Int})
+
+        v3 = zeros(SVector{3, Complex{Int}})
+        @test v3 * v3' === zeros(SMatrix{3, 3, Complex{Int}})
```
c42f: Worth testing test a little more specifically, I think.  With `zeros`, the code could do all sorts of things and these tests will still pass.

src/linalg.jl
c42f: I'll review it all here.

test/matrix_multiply.jl
andyferris: I can - this was meant to be a dispatch test (keeping in mind that the matrix-matrix code that this relies upon is checked thoroughly elsewhere)

c42f: I was wondering whether the  `_A_mul_B(::Size{sa}, ::Size{sb}, a::StaticVector{<: Any, Ta}, b::RowVector{Tb, CV} where CV <: ConjVector{<:Any, <:StaticVector})` was specifically tested?

andyferris: Well, it was going to a base method before and now it isn't...

Should tests test for end behaviours or specific things like which method gets invoked?

c42f: In this case, I'd test the end behaviour - the tests will be less brittle that way, and the call chain from public interface to private implementation is not complicated to follow.

In general it's a tradeoff:
* Testing the implementation details can let you write more specific tests for smaller blocks of code.  When things break, isolating the problem is easier.
* Testing the interface lets you refactor the implementation without breaking the tests.  Refactoring very detailed implementation tests can be a lot of work, and isn't much fun.

c42f: So, after all that - is this version of `_A_mul_B` tested elsewhere?  If not, we should test this (but via the public interface) to make sure we have coverage for that code path.

PR #160, opened 2017-05-10
Add matrix exponential for 1x1 and 2x2, and fallback to Base (fixes #159)

### LINKED ISSUE #159: Matrix exponential
Is there any plan on adding support for matrix exponential in StaticArrays?

Godisemo:
I have only tried this in julia v0.5, which is why the PR is to the julia-0.5 branch. You could probably cherry pick it to master.

Note: On my computer this gives a speedup of over 300x

```julia
a = randn(SMatrix{2,2})
b = Array(a)

@benchmark expm(a)
# BenchmarkTools.Trial:
#  memory estimate:  48 bytes
#  allocs estimate:  1
#  --------------
#  minimum time:     57.861 ns (0.00% GC)
#  median time:      58.580 ns (0.00% GC)
#  mean time:        65.692 ns (8.97% GC)
#  maximum time:     3.145 μs (96.47% GC)
#  --------------
#  samples:          10000
#  evals/sample:     983
@benchmark expm(b)
# BenchmarkTools.Trial:
#  memory estimate:  5.67 KiB
#  allocs estimate:  76
#  --------------
#  minimum time:     15.294 μs (0.00% GC)
#  median time:      18.114 μs (0.00% GC)
#  mean time:        19.030 μs (4.38% GC)
#  maximum time:     4.279 ms (98.87% GC)
#  --------------
#  samples:          10000
#  evals/sample:     1
```

### DIFF
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -1,64 +1,51 @@
 @inline expm(A::StaticMatrix) = _expm(Size(A), A)
 
-@generated function _expm(::Size{(1,1)}, A::StaticMatrix)
-    @assert size(A) == (1,1)
-    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
+@inline function _expm(::Size{(1,1)}, A::StaticMatrix)
+    T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
     newtype = similar_type(A,T)
 
-    quote
-        $(Expr(:meta, :inline))
-        ($newtype)((exp(A[1]), ))
-    end
+    (newtype)((exp(A[1]), ))
 end
 
-@generated function _expm{S<:Real}(::Size{(2,2)}, A::StaticMatrix{S})
-    @assert size(A) == (2,2)
-    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
+@inline function _expm{S<:Real}(::Size{(2,2)}, A::StaticMatrix{S})
+    T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
     newtype = similar_type(A,T)
 
-    quote
-        $(Expr(:meta, :inline))
+    @inbounds a = A[1]
+    @inbounds c = A[2]
+    @inbounds b = A[3]
+    @inbounds d = A[4]
 
-        @inbounds a = A[1]
-        @inbounds c = A[2]
-        @inbounds b = A[3]
-        @inbounds d = A[4]
+    v = (a-d)^2 + 4*b*c
 
-        v = (a-d)^2 + 4*b*c
-
-        # if v == 0
-        z1::$T = 1.0
-        z2::$T = 0.5
-
-        if v > 0
-          z = sqrt(v)
-          z1 = cosh(z / 2)
-          z2 = sinh(z / 2) / z
-        elseif v < 0
-          z = sqrt(-v)
-          z1 = cos(z / 2)
-          z2 = sin(z / 2) / z
-        end
-
-        r = exp((a + d) / 2)
-        m11 = r * (z1 + (a - d) * z2)
-        m12 = r * 2b * z2
-        m21 = r * 2c * z2
-        m22 = r * (z1 - (a - d) * z2)
-
-        ($newtype)((m11, m21, m12, m22))
+    if v > 0
+        z = sqrt(v)
+        z1 = cosh(z / 2)
+        z2 = sinh(z / 2) / z
+    elseif v < 0
+        z = sqrt(-v)
+        z1 = cos(z / 2)
+        z2 = sin(z / 2) / z
+    else # if v == 0
+        z1 = T(1.0)
+        z2 = T(0.5)
     end
+
+    r = exp((a + d) / 2)
+    m11 = r * (z1 + (a - d) * z2)
+    m12 = r * 2b * z2
+    m21 = r * 2c * z2
+    m22 = r * (z1 - (a - d) * z2)
+
+    (newtype)((m11, m21, m12, m22))
 end
 
 # TODO add complex valued expm
 # TODO add special case for 3x3 matrices
 
-@generated function _expm(::Size, A::StaticArray)
-  T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
-  newtype = similar_type(A,T)
+@inline function _expm(s::Size, A::StaticArray)
+    T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
+    newtype = similar_type(A,T)
 
-  quote
-      $(Expr(:meta, :inline))
-      ($newtype)(Base.expm(Array(A)))
-  end
+    s(Base.expm(Array(A)))
 end
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -43,9 +43,4 @@ end
 # TODO add complex valued expm
 # TODO add special case for 3x3 matrices
 
-@inline function _expm(s::Size, A::StaticArray)
-    T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
-    newtype = similar_type(A,T)
-
-    s(Base.expm(Array(A)))
-end
+@inline _expm(s::Size, A::StaticArray) = s(Base.expm(Array(A)))
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -11,7 +11,7 @@
     end
 end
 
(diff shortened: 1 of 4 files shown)


src/expm.jl
```diff
@@ -0,0 +1,64 @@
+@inline expm(A::StaticMatrix) = _expm(Size(A), A)
+
+@generated function _expm(::Size{(1,1)}, A::StaticMatrix)
+    @assert size(A) == (1,1)
```
andyferris: Probably overkill, given the above.

src/expm.jl
```diff
@@ -0,0 +1,64 @@
...
+    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
```
andyferris: In Julia v0.6, you can only perform "hyper"-pure operations inside the function generator. Since the eltype might not be defined before static arrays is loaded, this just isn't "pure enough" for the assumptions made by the Julia 0.6 compiler.

However, it's safe to move this into the body of the function (inside the `quote`) - where possible inference will perform type calculations at compile time.

Also - why `sqrt(one)` not `exp(zero)`? (zero is slightly more flexible from a group vs. algebra perspective, i.e. in theory `eltype(A)` might be a generator that supports addition and `zero` only.)

src/expm.jl
```diff
@@ -0,0 +1,64 @@
...
+    newtype = similar_type(A,T)
```
andyferris: Also move this to body - it's "free"

src/expm.jl
```diff
@@ -0,0 +1,64 @@
...
+    quote
+        $(Expr(:meta, :inline))
+        ($newtype)((exp(A[1]), ))
+    end
+end
+
+@generated function _expm{S<:Real}(::Size{(2,2)}, A::StaticMatrix{S})
+    @assert size(A) == (2,2)
+    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
+    newtype = similar_type(A,T)
```
andyferris: Move these two lines to the body.

src/expm.jl
```diff
@@ -0,0 +1,64 @@
...
+
+        @inbounds a = A[1]
+        @inbounds c = A[2]
+        @inbounds b = A[3]
+        @inbounds d = A[4]
+
+        v = (a-d)^2 + 4*b*c
+
+        # if v == 0
+        z1::$T = 1.0
```
andyferris: These are unnecessary since `if` statements don't introduce a new scoping block.

src/expm.jl
```diff
@@ -0,0 +1,64 @@
...
+        ($newtype)((m11, m21, m12, m22))
+    end
+end
+
+# TODO add complex valued expm
+# TODO add special case for 3x3 matrices
+
+@generated function _expm(::Size, A::StaticArray)
+  T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
+  newtype = similar_type(A,T)
```
andyferris: Move these two lines to the generated body

src/expm.jl
```diff
@@ -0,0 +1,64 @@
...
+
+  quote
+      $(Expr(:meta, :inline))
+      ($newtype)(Base.expm(Array(A)))
```
andyferris: Generally, I've been using `SizedArray` for this purpose with `eig`, `inv`, etc. The static size of the array is not forgotten, the data doesn't need to be copied to the stack unnecessarily, and the user gets feedback that a fully "native" static arrays method was not used.

However, it does violate `similar_type`, which doesn't seem that desirable...

src/expm.jl
Godisemo: you mean the assertion?

andyferris: Yeah

src/expm.jl
Godisemo: Moving
```julia  
        T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
        newtype = similar_type(A,T)
```
into the body removes the need for using a generated functions in the first place, right?

andyferris: actually, yes! I think the `eig` and `inv` code works like that.

src/expm.jl
Godisemo: The reason for this was to ensure type stability. Moved it to the else-block instead.

src/expm.jl
```diff
@@ -0,0 +1,51 @@
+@inline expm(A::StaticMatrix) = _expm(Size(A), A)
+
+@inline function _expm(::Size{(1,1)}, A::StaticMatrix)
+    T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
```
andyferris: Why are we promoting with `Float32` here?

src/expm.jl
```diff
@@ -0,0 +1,51 @@
...
+
+    (newtype)((m11, m21, m12, m22))
+end
+
+# TODO add complex valued expm
+# TODO add special case for 3x3 matrices
+
+@inline function _expm(s::Size, A::StaticArray)
+    T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
+    newtype = similar_type(A,T)
```
andyferris: `newtype` unused

src/expm.jl
```diff
@@ -0,0 +1,51 @@
...
+    if v > 0
+        z = sqrt(v)
+        z1 = cosh(z / 2)
+        z2 = sinh(z / 2) / z
+    elseif v < 0
+        z = sqrt(-v)
+        z1 = cos(z / 2)
+        z2 = sin(z / 2) / z
+    else # if v == 0
+        z1 = T(1.0)
```
andyferris: `zero(T)`

src/expm.jl
Godisemo: I probably copied from the cholesky source

src/expm.jl
Godisemo: `one(T)`?

andyferris: Or maybe not. I dunno.

Godisemo: since there is no `half(T)`, I would prefer not to use `one(T)`

src/expm.jl
```diff
@@ -0,0 +1,51 @@
+@inline expm(A::StaticMatrix) = _expm(Size(A), A)
+
+@inline function _expm(::Size{(1,1)}, A::StaticMatrix)
+    T = promote_type(typeof(exp(zero(eltype(A)))), Float32)
+    newtype = similar_type(A,T)
+
+    (newtype)((exp(A[1]), ))
+end
+
+@inline function _expm{S<:Real}(::Size{(2,2)}, A::StaticMatrix{S})
```
andyferris: `StaticMatrix` has new type parameters, `StaticMatrix{M,N,T}`.

Try `_expm(::Size{(2,2)}, A::StaticMatrix{<:Any, <:Any, <:Real})`

Godisemo: Then it won't work on 0.5 though. How to handle backwards compatibility?

andyferris: > How to handle backwards compatibility?

We don't :) They have diverged.

It's not an ideal situation, but that's where we are.

andyferris: I think there is a branch somewhere where we can merge backports. It's just more work.

Godisemo: Ok, up until the last commit, it worked on julia 0.5. I think I might have messed up when changing target branch for this PR to master.

andyferris: I see... there seems to be some conflicts.

One way of doing this one PR to `julia-0.5` and one PR to `master`. The difference will be trivial.

Godisemo: yeah, I changed back to the julia-0.5 branch. Will copy the files over to master and create a new PR. You will have to revert the last commit though and pull 1f2ff2b

Godisemo:
Is this PR and #161 not ready for merging?

andyferris:
Travis is showing errors...

Godisemo:
You can clearly see that the errors are caused by previous commits... The tests that fail are related to `Binary IO` and `write`.

andyferris:
I'm totally confused.

I re-ran travis... what exactly are these errors about?

c42f:
Yes, these errors are really weird.  Looks like I forgot a `Compat` dependency for `take!` when doing the IO tests... but as to why the tests pass locally I'm really confused.

c42f:
After merging #174 the test failures here are now only spurious failures on nightly - tests for julia-0.5 pass so I'm going to merge this.

PR #163, opened 2017-05-15
More tests

kshyatt:


test/eigen.jl
```diff
@@ -43,6 +44,10 @@
         @test vals::SVector ≈ vals_a
         @test eigvals(m) ≈ vals
         @test (vecs*diagm(vals)*vecs')::SMatrix ≈ m
+        
+        m_d = randn(3)
+        m = @SMatrix diagm(m_d)
```
andyferris: Sorry, this isn't supported by the macro. I think `diagm(m_d)` works fine, however.

andyferris:
Cool, thanks very much Kate! (Katharine?) I've always felt bad about the coverage here...

BTW,  I briefly read your "build a MERA" paper. Nice work - I've had discussions before with people whether this would be a good thing to do - seems fun. Is Bela your supervisor?

test/eigen.jl
andyferris: I mean `diagm(randn(SVector{3}))`

andyferris: (the API I want but haven't done yet is `m_d = rand(Size(3))`)

test/chol.jl
```diff
@@ -0,0 +1,25 @@
...
+        m_a = randn(2,2)
+        #non hermitian
+        @test_throws ArgumentError chol(SMatrix{2,2}(m_a))
+        m_a = m_a*m_a'
+        m = @SMatrix m_a
+        @test chol(m) ≈ chol(m_a)
+    end
+
+    @testset "3×3" for i = 1:100
+        m_a = randn(3,3)
```
c42f: How about calling `srand()` somewhere so that these are fully reproducible?

Ideally we'd do large scale randomized testing (fuzzing, I guess) independently from the unit tests.

c42f:
This is great, we do need more tests!

kshyatt:
@andyferris either Katie or Katharine is fine :)

Thanks re: the paper. Yes, Bela's my supervisor. Let me know if you're ever in the mood to chat about it!

I'll work on fixing up that failing test.

test/eigen.jl
```diff
@@ -43,6 +44,9 @@
         @test vals::SVector ≈ vals_a
         @test eigvals(m) ≈ vals
         @test (vecs*diagm(vals)*vecs')::SMatrix ≈ m
+       
+        m_d = randn(SVector{3}); m = diagm(m_d)
+        @test eigvals(m) ≈ m_d
```
andyferris: @test eigvals(m) ≈ sort(m_d)

kshyatt: Just caught that myself - long day!

andyferris:
> Let me know if you're ever in the mood to chat about it!

OK, I'll find you at JuliaCon, then?

kshyatt:
@andyferris Sure! I'm happy to chat before then as well if you'd like (Skype, Hangouts, email).

Can this be merged?

PR #168, opened 2017-05-18
fix a method ambiguity error in unsafe_convert for FieldVectors

mweastwood:
I'm not 100% sure this change is correct, but without it I see a method ambiguity error when trying to pass a `FieldVector` through `ccall`:

```
  MethodError: Base.unsafe_convert(::Type{Ptr{LibHealpix.UnitVector}}, ::Base.RefValue{LibHealpix.UnitVector}) is ambiguous. Candidates:
    unsafe_convert(::Type{Ptr{T}}, m::Base.RefValue{FV}) where {T, FV<:StaticArrays.FieldVector} in StaticArrays at /home/michael/.julia/v0.6/StaticArrays/src/FieldVector.jl:27
    unsafe_convert(P::Type{Ptr{T}}, b::Base.RefValue{T}) where T in Base at refpointer.jl:56
```

where `LibHealpix.UnitVector <: FieldVector{3, Float64}`.

### DIFF
--- a/src/FieldVector.jl
+++ b/src/FieldVector.jl
@@ -24,7 +24,5 @@ abstract type FieldVector{N, T} <: StaticVector{N, T} end
 
 # See #53
 Base.cconvert{T}(::Type{Ptr{T}}, v::FieldVector) = Base.RefValue(v)
-Base.unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, m::Base.RefValue{FV}) =
-    _unsafe_convert(Ptr{T}, eltype(FV), m)
-_unsafe_convert{T, FV <: FieldVector}(::Type{Ptr{T}}, ::Type{T}, m::Base.RefValue{FV}) =
-         Ptr{T}(Base.unsafe_convert(Ptr{FV}, m))
+Base.unsafe_convert(::Type{Ptr{T}}, m::Base.RefValue{FV}) where {N,T,FV<:FieldVector{N,T}} =
+    Ptr{T}(Base.unsafe_convert(Ptr{FV}, m))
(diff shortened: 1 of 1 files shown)


c42f:
Thanks for the patch!

I suppose the ambiguity is due to the fact that we've set up `unsafe_convert` here to be able to convert a RefValue{FV} into a pointer to the elements, whereas Base wants to convert a RefValue{T} into a pointer to T itself.  The fact that the Base version dispatches on both arguments together is making it no less specific than our version, despite the fact that we've pinned down the second argument more precisely.

I'd say the fix here isn't quite right yet as we do want to support types other than `Number`. A non-elegant fix would be to add an overload which is more specific than both.

However, we might be able to do better - what about something like something like

```julia
Base.unsafe_convert(::Type{Ptr{T}}, m::Base.RefValue{<:FieldVector{N,T}}) where {N,T} =
    Ptr{T}(Base.unsafe_convert(Ptr{FV}, m))
```

mweastwood:
That solution works as well! I've updated the PR and included a cosmetic change so that the whole file is using the new `where` syntax (the inconsistency was bothering me...)

c42f:
In 0.6, you don't even need the `where` syntax when a type is a pure placeholder which occurs in a single location.  For example:

```julia
Base.cconvert(::Type{Ptr{<:Any}}, v::FieldVector) = Base.RefValue(v)
```

Should work just fine.

c42f:
Or even more succinct, I think you could get away with

```julia
Base.cconvert(::Type{<:Ptr}, v::FieldVector) = Base.RefValue(v)
```

I'm still getting used to the new possibilities :-)

mweastwood:
That is going to take some getting used to. I'm still not used to the fact that `Ptr{Int} <: Ptr{T} where T` now. I've updated the last commit to use the even more compact syntax.

c42f:
We've had `Ptr{Int} <: Ptr` for a while, and

```
julia> Ptr == (Ptr{T} where T)
true
```

Technically in the `unsafe_convert` case you only really need to mention a single type parameter `T` in order to express the constraint:

```julia
Base.unsafe_convert(::Type{Ptr{T}}, m::Base.RefValue{<:FieldVector{<:Any,T}}) where {T}
```

Anyway, this looks good, I'm happy to merge it.

PR #179, opened 2017-05-24
Fix ctranspose bug & add test

kshyatt:
Didn't actually work for complex arrays...


andyferris:
Lol - thanks!

andyferris:
Let's never speak of this again...

PR #180, opened 2017-05-24
Let eigen work for "big" Herm/symm matrices

kshyatt:
Fixed some bugs along the way...

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -12,6 +12,7 @@ end
 end
 
 @inline _eigvals(::Size{(1,1)}, a, permute, scale) = @inbounds return SVector(real(a.data[1]))
+@inline _eigvals(::Size{(1, 1)}, a::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real} = @inbounds return SVector(real(parent(a).data[1]))
 
 @inline function _eigvals(::Size{(2,2)}, A::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
     a = A.data
@@ -102,6 +103,10 @@ end
     return SVector(eig1, eig2, eig3)
 end
 
+@inline function _eigvals(s::Size, A::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
+    vals = eigvals(Hermitian(Array(parent(A))))
+    return SVector{s[1], T}(vals)
+end
 
 
 
@@ -124,8 +129,8 @@ end
 end
 
 @inline function _eig(s::Size, A::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
-    eigen = eigfact(Hermitian(Array(parent(A))); permute=permute, scale=scale)
-    return (s(eigen.values), s(eigen.vectors)) # Return a SizedArray
+    eigen = eigfact(Hermitian(Array(parent(A))))
+    return (SVector{s[1], T}(eigen.values), SMatrix{s[1], s[2], T}(eigen.vectors)) # Return a SizedArray
 end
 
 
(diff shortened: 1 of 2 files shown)


src/eigen.jl
```diff
@@ -124,8 +129,8 @@ end
 end
 
 @inline function _eig(s::Size, A::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
-    eigen = eigfact(Hermitian(Array(parent(A))); permute=permute, scale=scale)
-    return (s(eigen.values), s(eigen.vectors)) # Return a SizedArray
+    eigen = eigfact(Hermitian(Array(parent(A))))
```
andyferris: Is there a reason for removing `permute` and `scale`?

src/eigen.jl
```diff
@@ -124,8 +129,8 @@ end
...
+    return (SVector{s[1], T}(eigen.values), SMatrix{s[1], s[2], T}(eigen.vectors)) # Return a SizedArray
```
andyferris: It was intentional that this returned an `SizedArray` rather than a `SArray`.

I'm not saying it was defensible - but it served as an easy way to see that an `Array` was constructed during the process. If we don't want to keep this, then a third possibility is to use `similar_type` to get the output type.

src/eigen.jl
kshyatt: The corresponding method in Base doesn't take these two args for Hermitian/symmetric matrices.

src/eigen.jl
kshyatt: Whoops! It was unintentional that I changed this - whatever you'd prefer is fine with me :)

src/eigen.jl
andyferris: OK, great! (I didn't have REPL access when I asked)

src/eigen.jl
andyferris: This has come up a few times. @c42f, any opinion?

src/eigen.jl
```diff
@@ -12,6 +12,7 @@ end
 end
 
 @inline _eigvals(::Size{(1,1)}, a, permute, scale) = @inbounds return SVector(real(a.data[1]))
+@inline _eigvals(::StaticArrays.Size{(1, 1)}, a::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real} = @inbounds return SVector(real(parent(a).data[1]))
```
c42f: The explicit `StaticArrays` can be removed from the first parameter here?

src/eigen.jl
c42f: No strong opinion from me. It's hard to imagine a really *good* reason for the user of the API to care about this one way or another. Using it as an indication that some allocation has happened seems a bit of a blunt instrument to me.  It's not hard to imaging a different algorithm where sufficient postprocessing occurs that it's better to return an SVector.

Perhaps, for consistency, it would be best if we always just return something immutable. It seems bad to me that larger vs smaller sizes might return containers with different mutability.

kshyatt: It looks like the 2x2 and 3x3 methods use a `SVector` for `vals` as well. That's the variable we're looking at, right?

test/eigen.jl
```diff
@@ -141,4 +141,33 @@
         @test vecs*diagm(vals)*vecs' ≈ m
         @test eigvals(m) ≈ vals
     end
+    
+    @testset "4×4" for i = 1:100
+        m_a = randn(4,4)
+        m_a = m_a*m_a'
+        m = SMatrix{4,4}(m_a)
+
+        (vals_a, vecs_a) = eig(m)
```
c42f: Was this meant to be `eig(m_a)`

src/eigen.jl
c42f: Right.  So I think it's reasonable to leave it like you have it here.

kshyatt:
Looks like the CI fail wasn't my fault. Is this ready to go?

PR #183, opened 2017-05-25
use 0.6.0-pre for minimum julia version in REQUIRE

tkelman:
`abstract type` won't work on early 0.6.0-dev versions,
so better to stick with the julia-0.5-compatible versions of the package there


andyferris:
Thanks

PR #185, opened 2017-05-26
More macro error tests

kshyatt:


andyferris:
Cool.

I've been too lazy, but I was hoping to get around to making a better macro system, especially now that `SVector` is a type of `SArray`. I had been thinking something like `@SA`.

PR #189, opened 2017-05-27
Add tests for randn! and randexp!

kshyatt:
Didn't do `rand!` because it has some weird ambiguity I don't understand.


andyferris:
Thanks. 

I'll have to look into the ambiguity...

PR #192, opened 2017-05-28
Add simple forward and backward substitution to solve triangular systems

bshall:
Firstly, thanks for great package.

I needed to solve some triangular systems with static arrays and thought that you might find a pull request useful. Below are a couple of quick benchmarks. I'm not sure how you decide on your huristics so I haven't added any fall backs to the methods in Base yet.

```julia
A = randn(n, n) |> t -> chol(t't);
b = A*ones(n, 2);
SA = UpperTriangular(SMatrix{n, n}(A.data));
Sb = SMatrix{n, 2}(b);
```
For `n=3` the results for `@benchmark A \ b` are:
BenchmarkTools.Trial: 
memory estimate:  144 bytes
allocs estimate:  2
minimum time:     261.064 ns (0.00% GC)
median time:      270.437 ns (0.00% GC)
mean time:        284.320 ns (3.11% GC)
maximum time:     4.887 μs (89.11% GC)
samples:          10000
evals/sample:     343

And the results for `@benchmark SA \ Sb` are:
BenchmarkTools.Trial: 
memory estimate:  64 bytes
allocs estimate:  1
minimum time:     60.341 ns (0.00% GC)
median time:      62.549 ns (0.00% GC)
mean time:        67.918 ns (5.45% GC)
maximum time:     1.611 μs (93.13% GC)
samples:          10000
evals/sample:     982

### DIFF
--- a/test/solve.jl
+++ b/test/solve.jl
@@ -25,3 +25,20 @@
         @test m(A)\v(b) ≈ A\b
     end =#
 end
+
+@testset "Solving triangular system" begin
+    for n in (1,2,3,4),
+          (t1, uplo1) in ((UpperTriangular, :U),
+                          (LowerTriangular, :L)),
+            (m, v, u) in ((SMatrix{n, n}, SVector{n}, SMatrix{n, 2}), (MMatrix{n,n}, MVector{n}, SMatrix{n, 2})),
+                elty in (Float32, Float64, Int)
+
+        eval(quote
+            A = $(t1)($elty == Int ? rand(1:7, $n, $n) : convert(Matrix{$elty}, randn($n, $n)) |> t -> chol(t't) |> t -> $(uplo1 == :U) ? t : ctranspose(t))
+            b = convert(Matrix{$elty}, A*ones($n, 2))
+            SA = $t1($m(A.data))
+            @test SA \ $v(b[:, 1]) ≈ A \ b[:, 1]
+            @test SA \ $u(b) ≈ A \ b
+        end)
+    end
+end
--- a/test/solve.jl
+++ b/test/solve.jl
@@ -27,18 +27,28 @@
 end
 
 @testset "Solving triangular system" begin
-    for n in (1,2,3,4),
-          (t1, uplo1) in ((UpperTriangular, :U),
-                          (LowerTriangular, :L)),
-            (m, v, u) in ((SMatrix{n, n}, SVector{n}, SMatrix{n, 2}), (MMatrix{n,n}, MVector{n}, SMatrix{n, 2})),
-                elty in (Float32, Float64, Int)
+    for n in (1, 2, 3, 4),
+          (t, uplo) in ((UpperTriangular, :U),
+                        (LowerTriangular, :L)),
+              (m, v, u) in ((SMatrix{n,n}, SVector{n}, SMatrix{n, 2}),
+                            (MMatrix{n,n}, MVector{n}, SMatrix{n, 2})),
+                  eltya in (Float32, Float64, BigFloat, Complex64, Complex128, Complex{BigFloat}, Int),
+                      eltyb in (Float32, Float64, BigFloat, Complex64, Complex128, Complex{BigFloat})
 
         eval(quote
-            A = $(t1)($elty == Int ? rand(1:7, $n, $n) : convert(Matrix{$elty}, randn($n, $n)) |> t -> chol(t't) |> t -> $(uplo1 == :U) ? t : ctranspose(t))
-            b = convert(Matrix{$elty}, A*ones($n, 2))
-            SA = $t1($m(A.data))
-            @test SA \ $v(b[:, 1]) ≈ A \ b[:, 1]
-            @test SA \ $u(b) ≈ A \ b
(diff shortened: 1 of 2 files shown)


andyferris:
Awesome, thanks @bshall!

I've been hoping to have something like this for a while now., so this is very much appreciated. (Now we just need an `LU` or `QR` decomposition...).

Heuristics are rather random (usually not implemented) in this package... it's definitely something that could be improved. For cases like this, where there are generic algorithms for arbitrary sizes, I've generally left no heuristics - users must handle with caution. On the other hand, the other O(n^3) algorithm I implemented was matrix multiplication and I spent a long time making heuristics for that (and alternative O(n^2) codegen - that's amount of code generated (using a loop or `@noinline` functions or whatever), not speed), guided by micro-benchmarks. I don't expect this for this PR (unless you want a very simple cutoff - up to you).

bshall:
No problem at all. I needed to use this with cholesky decomposition but yeah, more general decompositions would be great.

Yeah, I looked at the matrix multiplication code and I wasn't sure if the cutoff was chosen based on speed, codegen or some combination of both. Did you have a general rule based on compilation time or lines of code generated? I'm happy to keep this PR simple if you are and the heuristics can always be added at a later stage.

I had one last question. In the signiture of `\` the element type `T` of both `a` and `b` are required to be the same. Is there a particular reason for this rather than letting the promotion rules in the individual `solve` methods handle different types?

andyferris:
> Did you have a general rule based on compilation time or lines of code generated?

Just the benchmarks in the `perf` directory and some compiler limits; the heuristics are tuned to my desktop PC :). I dare say these will have changed in v0.6, but I haven't had a chance to check.

> In the signiture of `\` the element type `T` of both `a` and `b` are required to be the same. Is there a particular reason for this ... ?

Nope... doesn't seem like it.

c42f:
Yes, it's great to have this in the package, thanks!

Regarding using this in conjunction with `chol` - if you're mainly interested in backdivide, would it make more API sense to use a `cholfact`-like interface, in which case we'd return a fixed size version of `Base.LinAlg.Cholesky`.  Thus:

```julia
A = @SMatrix rand(3,3)
Ac = cholfact(A*A')
# solve linear system quickly based on forward/backward substitution
Ac \ SVector(1,2,3)
```

Not necessary for this PR, just a thought.

bshall:
Okay great. So I'll remove the type constraints and add a few more tests for this PR. I'm busy working on a generic `chol` function for larges static matrices so I'd be happy to add a `cholfact`-like api.

test/solve.jl
```diff
@@ -25,3 +25,30 @@
...
+@testset "Solving triangular system" begin
+    for n in (1, 2, 3, 4),
+          (t, uplo) in ((UpperTriangular, :U),
+                        (LowerTriangular, :L)),
+              (m, v, u) in ((SMatrix{n,n}, SVector{n}, SMatrix{n, 2}),
+                            (MMatrix{n,n}, MVector{n}, SMatrix{n, 2})),
+                  eltya in (Float32, Float64, BigFloat, Complex64, Complex128, Complex{BigFloat}, Int),
+                      eltyb in (Float32, Float64, BigFloat, Complex64, Complex128, Complex{BigFloat})
+
+        eval(quote
```
c42f: The `eval` here seems dubious.

You should be able to do this entirely with normal loops, I would expect?

test/solve.jl
```diff
@@ -25,3 +25,30 @@
...
+            A = $t($eltya == Int ? rand(1:7, $n, $n) : convert(Matrix{$eltya}, ($eltya <: Complex ? complex.(randn($n, $n), randn($n, $n)) : randn($n, $n)) |> z -> chol(z'z) |> z -> $(uplo == :U) ? z : ctranspose(z)))
+            b = convert(Matrix{$eltyb}, $eltya <: Complex ? real(A)*ones($n, 2) : A*ones($n, 2))
+            SA = $t($m(A.data))
+            Sx = SA \ $v(b[:, 1])
+            x = A \ b[:, 1]
+            @test typeof(Sx) <: StaticVector # test not falling back to Base
```
c42f: This is probably better written as 

```
@test Sx isa StaticVector
```

bshall:
So I initially tried to write the test without the `eval` but I ran into this strange error. If you run
```julia
for n in (1,)
    SVector{n}(ones(n))
end
```
You get the following warning:

WARNING: An error occurred during inference. Type inference is now partially disabled.
TypeError(func=:apply_type, context="Vararg count", expected=Int64, got=Any)

Followed by a longish stack trace and finally a `MethodError` saying no method matching `string(::Expr)`. Interestingly, the following works fine
```julia
for n in (1,)
    SMatrix{n,n}(ones(n,n))
end
```
@andyferris and @c42f should I report this as an issue?

c42f:
This sound reminiscent of some problems Andy had previously.

If I run your example in the repl I can't reproduce the issue in julia-0.6-rc2.  Which version are you using?

bshall:
Sorry, it seems I was still on a slightly outdated nightly build. It works perfectly on julia-0.6-rc2 and on julia-0.5.2. Thanks for the feedback, I've made the changes you suggested.

c42f:
Nice, this looks great, thanks.

PR #195, opened 2017-05-29
Uncomment more convert tests

kshyatt:
I left two commented out bc they error and I'm not super sure why


andyferris:
Probably could fix those two with some extra methods on `SArray` and `MArray`, but the trick is to get this working for all `StaticArray`. The converts and constructors are tricky (particularly with single-element arrays of arrays allowed).

Issue #191, opened 2017-05-28
Broadcasted Operations Stall Julia


ChrisRackauckas:
When broadcasting some operations, Julia will get stuck. Example:

```julia
using StaticArrays
u0 = @SMatrix rand(4,2)
reltol = 0.003
uprev = u0
u = u0
max.(abs.(u),abs.(uprev)) .* reltol
```

This is tested on the latest release (on v0.6-rc2) and on StaticArrays master.

c42f:
Yikes, I'm not sure what to make of this one yet.  More minimal test case:

```julia
using StaticArrays
x = broadcast((a,b,c)->0, SVector(1,1), 0, 0)
```

andyferris:
Oh dear. These should be our implementation of broadcast, I suppose.

bshall:
I think the issue is in the `broadcast_sizes` function. If you manually define the sizes @c42f example works
```julia
s = (Size{(2,)}(), Size(), Size())
StaticArrays._broadcast((a,b,c)->0, s, SVector(1,1), 0, 0)
```
But calling `StaticArrays.broadcast_sizes(SVector(1,1), 0, 0)` hangs.

bshall:
It seems like this could be a julia bug. A test case is
```julia
f(t::Tuple) = t
f(t::Tuple, a::Int, as...) = f((t..., a), as...)
f(t::Tuple, a::Float64, as...) = f((t..., a), as...)

f((), 1, 1, 1) # works fine
f((), 1.0, 1.0, 1.0) # works fine
f((), 1, 1.0, 1) # hangs
```
What do you guys think @andyferris @c42f is this a bug or is there something wrong with the recursion? As a workaround for `StaticArrays` you don't really need to define `broadcast_sizes` recursively, you could just loop over the `as` to build up the sizes.

c42f:
Yes, I was just coming to the same conclusion myself, I think this is a julia bug, at least in julia-0.6.0-rc2.

Unfortunately we can't use a simple loop, as we need this to be fully type inferred.  We might be able to get away with a carefully constructed generated function...

bshall:
I just checked on 0.5.1 and it also doesn't work there. Should I go ahead and report it?

> Unfortunately we can't use a simple loop, as we need this to be fully type inferred

Oh right, that makes sense.

andyferris:
This is a bit odd. This is v0.6-rc2?

[...]

PR #196, opened 2017-05-29
Fix broadcast issue

bshall:
Fix for #191

### DIFF
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -1,16 +1,16 @@
-# issue 191
 @testset "Broadcast sizes" begin
-    @test broadcast_sizes(1, 1, 1) === (Size(), Size(), Size())
+    @test @inferred(broadcast_sizes(1, 1, 1)) === (Size(), Size(), Size())
     for t in (SVector{2}, MVector{2}, SMatrix{2, 2}, MMatrix{2, 2})
-        @test broadcast_sizes(ones(t), ones(t), ones(t)) === (Size(t), Size(t), Size(t))
-        @test broadcast_sizes(ones(t), 1, ones(t)) === (Size(t), Size(), Size(t))
-        @test broadcast_sizes(1, ones(t), ones(t)) === (Size(), Size(t), Size(t))
-        @test broadcast_sizes(ones(t), ones(t), 1) === (Size(t), Size(t), Size())
-        @test broadcast_sizes(1, ones(t), 1) === (Size(), Size(t), Size())
-        @test broadcast_sizes(ones(t), 1, 1) === (Size(t), Size(), Size())
-        @test broadcast_sizes(1, 1, ones(t)) === (Size(), Size(), Size(t))
+        @test @inferred(broadcast_sizes(ones(t), ones(t), ones(t))) === (Size(t), Size(t), Size(t))
+        @test @inferred(broadcast_sizes(ones(t), 1, ones(t))) === (Size(t), Size(), Size(t))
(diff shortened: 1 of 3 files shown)


c42f:
I think we can get away with something a bit simpler than this.  Here's what I had:

```julia
@inline broadcast_sizes(a::StaticArray, as...) = (Size(a), broadcast_sizes(as...)...)
@inline broadcast_sizes(a::Number, as...) = (Size(), broadcast_sizes(as...)...)
@inline broadcast_sizes() = ()
```

bshall:
Yeah, that is better.

c42f:
The failed test here isn't your fault, I restarted it.

Do you want to finish this one off by cleaning up a little and adding tests?

bshall:
Yeah, I'm  happy to do that. I'll make your change and add some tests.

c42f:
Awesome, thanks a lot! I had started messing with tests, but hadn't done much.

bshall:
Okay, so I've added a bunch of tests and found some issues/inconsistencies with broadcast from Base. I've listed those here #197, #198, #199 and #200 and I've added (commented out) test cases for them. Let me know if you'd like me to add anything else.

Oh, I also made a small change to the `broadcast!` function so it returns `dest` as well (just to be consistent).

test/broadcast.jl
```diff
@@ -0,0 +1,157 @@
...
+    for t in (SVector{2}, MVector{2}, SMatrix{2, 2}, MMatrix{2, 2})
+        @test broadcast_sizes(ones(t), ones(t), ones(t)) === (Size(t), Size(t), Size(t))
+        @test broadcast_sizes(ones(t), 1, ones(t)) === (Size(t), Size(), Size(t))
+        @test broadcast_sizes(1, ones(t), ones(t)) === (Size(), Size(t), Size(t))
+        @test broadcast_sizes(ones(t), ones(t), 1) === (Size(t), Size(t), Size())
+        @test broadcast_sizes(1, ones(t), 1) === (Size(), Size(t), Size())
+        @test broadcast_sizes(ones(t), 1, 1) === (Size(t), Size(), Size())
+        @test broadcast_sizes(1, 1, ones(t)) === (Size(), Size(), Size(t))
+    end
+    @test broadcast((a,b,c)->0, SVector(1,1), 0, 0) == SVector(0, 0)
```
c42f: Could do with a comment `See #191` perhaps?

test/broadcast.jl
```diff
@@ -0,0 +1,157 @@
...
+        @test @inferred(m .- v) === @SMatrix [0 1; -1 0]
+        @test @inferred(v .- m) === @SMatrix [0 -1; 1 0]
+        @test @inferred(m .^ v) === @SMatrix [1 2; 81 256]
+        @test @inferred(v .^ m) === @SMatrix [1 1; 64 256]
+    end
+
+    @testset "2x2 StaticMatrix with 1x2 StaticMatrix" begin
+        m1 = @SMatrix [1 2; 3 4]
+        m2 = @SMatrix [1 4]
+        # @test @inferred(broadcast(+, m1, m2)) === @SMatrix [2 6; 4 8] #197
```
c42f: All the places where tests are currently commented out (because they're broken) you can use `@test_broken` instead.  This runs the test and verifies that it is, indeed, broken - so we won't forget to add them back in later when these issues are fixed.

bshall:
Great. Thanks for the feedback. I think I've made all the changes you suggested.

PR #203, opened 2017-05-30
RFC: fix construction of FSA types with heterogeneous types

### LINKED ISSUE #202: Missing constructor for FixedSizeArrays with heterogeneous inputs
moved from https://github.com/JuliaGeometry/GeometryTypes.jl/issues/88

There seems to be a missing constructor for heterogeneous inputs when constructing a FixedSizeArray type. For example:

```julia
julia> using StaticArrays.FixedSizeArrays

julia> Vec(0.0, 0, 0)
1-element StaticArrays.FixedSizeArrays.Vec{1,Tuple{Float64,Int64,Int64}}:
 (0.0, 0, 0)
```

where I would expect to instead get a `Vec{3, Float64}`. This happens because there's no matching constructor, so it falls back to `@inline (::Type{SA})(x...) where {SA <: StaticArray} = SA(x)` from `convert.jl`. 

Should we add that missing constructor to `StaticArrays.FixedSizeArrays`?

rdeits:
fixes #202 

(I am assuming that the current behavior is actually a bug; please let me know if it's not)

### DIFF
--- a/test/fixed_size_arrays.jl
+++ b/test/fixed_size_arrays.jl
@@ -204,6 +204,18 @@ for N = (1, 10)
         end
     end
 end
+
+@testset "heterogeneous construction" begin
+    @test typeof(@inferred(Vec(0.0, 0))) == Vec{2, Float64}
+    @test typeof(@inferred(Vec(0, 0.0))) == Vec{2, Float64}
+    @test typeof(@inferred(Point(0.0, 0))) == Point{2, Float64}
+    @test typeof(@inferred(Point(0, 0.0))) == Point{2, Float64}
+    @test typeof(@inferred(Vec(0.0, 0, 0))) == Vec{3, Float64}
+    @test typeof(@inferred(Vec(0, 0.0, 0))) == Vec{3, Float64}
+    @test typeof(@inferred(Point(0.0, 0, 0))) == Point{3, Float64}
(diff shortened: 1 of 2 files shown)


test/fixed_size_arrays.jl
```diff
@@ -204,6 +204,18 @@ for N = (1, 10)
...
+
+@testset "heterogeneous construction" begin
+    @test typeof(@inferred(Vec(0.0, 0))) == Vec{2, Float64}
+    @test typeof(@inferred(Vec(0, 0.0))) == Vec{2, Float64}
+    @test typeof(@inferred(Point(0.0, 0))) == Point{2, Float64}
+    @test typeof(@inferred(Point(0, 0.0))) == Point{2, Float64}
+    @test typeof(@inferred(Vec(0.0, 0, 0))) == Vec{3, Float64}
+    @test typeof(@inferred(Vec(0, 0.0, 0))) == Vec{3, Float64}
+    @test typeof(@inferred(Point(0.0, 0, 0))) == Point{3, Float64}
+    @test typeof(@inferred(Point(0, 0.0, 0))) == Point{3, Float64}
```
c42f: Could also be written as 

```julia
@test @inferred(Point(0, 0.0, 0)) isa Point{3,Float64}
```

rdeits: yeah, that's a bit nicer. Done!

c42f:
Looks good, thanks

PR #210, opened 2017-06-02
inline sum(f::Callable, a::StaticArray)

### LINKED ISSUE #209: `sum(f, vs)` is slow
I'm on 0.6.0-rc2.0, and the master version of `StaticArrays`. It seems that `sum(::Function, ::StaticArray)` calls the generic `mapreduce` instead of the `StaticArrays` one:

```
vs = SVector(1.0, 2.0, 3.0);

julia> @btime sum($vs);
  1.457 ns (0 allocations: 0 bytes)

julia> @btime sum(abs2, $vs);
  38.537 ns (2 allocations: 48 bytes)

julia> @btime mapreduce(abs2, +, $vs);
  1.739 ns (0 allocations: 0 bytes)

julia> @btime sumabs2($vs);
  1.746 ns (0 allocations: 0 bytes)
```

fredrikekre:
Since I derped in #209 i fixed it :) 
fix #209

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -169,6 +169,7 @@ end
 
 # These are all similar in Base but not @inline'd
 @inline sum(a::StaticArray{<:Any, T}) where {T} = reduce(+, zero(T), a)
+@inline sum(f::Base.Callable, a::StaticArray) = mapreduce(f, +, a)
 @inline prod(a::StaticArray{<:Any, T}) where {T} = reduce(*, one(T), a)
 @inline count(a::StaticArray{<:Any, Bool}) = reduce(+, 0, a)
 @inline all(a::StaticArray{<:Any, Bool}) = reduce(&, true, a)  # non-branching versions
--- a/test/mapreduce.jl
+++ b/test/mapreduce.jl
@@ -28,6 +28,7 @@
         @test reduce(+, v1) === 20
         @test reduce(+, 0, v1) === 20
         @test sum(v1) === 20
+        @test sum(abs2, v1) === 120
         @test prod(v1) === 384
         @test mean(v1) === 5.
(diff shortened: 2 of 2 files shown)


c42f:
> Should this work for length zero vectors?

Might be nice. I assume that's a side comment about `mapreduce` rather than something particularly relevant to this PR?

andyferris:
No it's specific to `sum` (`v0= zero(T)`) and `prod` (`v0 = one(T)`). In general, an appropriate `v0` cannot be chosen for `mapreduce`.

fredrikekre:
To be clear, this only inlines the corresponding method from base (https://github.com/JuliaLang/julia/blob/ef9ab602b0307afb9fa21f2323e2abf4b244ad41/base/reduce.jl#L347), so this does not change any behaviour.

andyferris:
Yes, thanks @fredrikekre. These comments are more for the future (sorry my "no" above was ambiguous - was referring to Chris's comment on `mapreduce` specifically, not this PR).

c42f:
Seems relevant here, as we have in base:

```julia
julia> sum(identity, Int[])
0
```

c42f:
Hmm, and yet...

```julia
julia> sum(x->x/2, Int[])
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
 [1] _mapreduce(::##1#2, ::Base.#+, ::IndexLinear, ::Array{Int64,1}) at ./reduce.jl:265
 [2] sum(::Function, ::Array{Int64,1}) at ./reduce.jl:347
```

Which seems rather odd until I looked at the Base implementation of `mapreduce`, where instead of using inference there are a bunch of special cases hardcoded for specific functions (including `identity`).

So I'm just going to merge this :-)

PR #212, opened 2017-06-02
add sqrtm

marius311:
This just adds the sqrtm function and provides a 1x1 and 2x2 implementation. 

One minor note, this is not type-stable, just like Base.sqrtm isn't. This is because depending on the if the matrix is positive definite or not, the result might be a complex matrix even if the input was a real matrix, and this can't be known until runtime. Actually I tend to think this should be handled the way its handled for e.g. sqrt(-1.), that is throwing an error, but maybe that's a discussion for Base, for now this is at least consistent.


src/sqrtm.jl
```diff
@@ -0,0 +1,19 @@
+@inline sqrtm(A::StaticMatrix) = _sqrtm(Size(A),A)
+
+@inline function _sqrtm(::Size{(1,1)}, A::SA) where {SA<:StaticArray}
+    s = sqrt(A[1,1])
```
c42f: Following Base, shouldn't this also be `sqrtm`?

c42f:
Oh, this needs a rebase - there's a conflict.

src/sqrtm.jl
marius311: Ah good catch, thanks.

andyferris:
> One minor note, this is not type-stable, just like Base.sqrtm isn't. This is because depending on the if the matrix is positive definite or not, the result might be a complex matrix even if the input was a real matrix, and this can't be known until runtime. Actually I tend to think this should be handled the way its handled for e.g. sqrt(-1.), that is throwing an error, but maybe that's a discussion for Base, for now this is at least consistent.

So I'm wondering - if it's not type stable, and the idea of this package is to be faster than `Base.Array`, do we gain a lot of speed in practical code here? I'm wondering if this the kind of situation where we can make a slightly different choice to `Base`?

marius311:
Indeed, for me its only about 3x faster than Base, less than alot of the other functions in this package. One solution I think is write a specialized method for SMatrix{<:Number} which uses `sqrt` instead of `sqrtm` which I think should be type-stable. I'm happy to write that up if you think that's a good way to go.

andyferris:
I think Base wants to go in that direction (`sqrt`, `exp` etc for matrices) and it seems "safer" to replace an erroring function than subtly break another, so I'd support that.

PR #213, opened 2017-06-02
More tests for the macros with generators

kshyatt:


test/MArray.jl
```diff
@@ -61,6 +61,7 @@
         @test (ex = macroexpand(:(@MArray fill)); isa(ex, Expr) && ex.head == :error)
         @test (ex = macroexpand(:(@MArray ones)); isa(ex, Expr) && ex.head == :error)
         @test (ex = macroexpand(:(@MArray sin(1:5))); isa(ex, Expr) && ex.head == :error)
+        @test (ex = macroexpand(:(@MArray si[])); isa(ex, Expr) && ex.head == :error)
```
c42f: Why should this be an error?

At the moment this works fine:
```julia
julia> si = Int
Int64

julia> @SVector si[]
0-element SVector{0,Int64}
```

kshyatt: Ah It wasn't working on latest master

kshyatt:
I'll get rid of that last bad commit and then this should be ok

c42f:
Looks good, thanks!

PR #215, opened 2017-06-03
Implement promote_rule for StaticArray to mirror Base.Array

### LINKED ISSUE #176: Promotion rules should follow Vector
The following is inconsistent:
```julia
julia> promote_type(SVector{2,Float64},SVector{2,Complex128})
StaticArrays.SArray{Tuple{2},T,1,2} where T

julia> promote_type(Vector{Float64},Vector{Complex128})
Array{Complex{Float64},1}
```

c42f:
Fixes #176

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -290,3 +290,7 @@ macro MArray(ex)
     end
 
 end
+
+function promote_rule(::Type{<:MArray{S,T,N,L}}, ::Type{<:MArray{S,U,N,L}}) where {S,T,U,N,L}
+    MArray{S,promote_type(T,U),N,L}
+end
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -252,3 +252,7 @@ macro SArray(ex)
         error("Bad input for @SArray")
     end
 end
+
+function promote_rule(::Type{<:SArray{S,T,N,L}}, ::Type{<:SArray{S,U,N,L}}) where {S,T,U,N,L}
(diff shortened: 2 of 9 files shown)


c42f:
I don't think `Base` does anything special with this.  The `promote_rule` declaration is specific to `Array`, and doesn't work between `Array` and `UnitRange`, for instance.

The left hand side rule is consistent with the other places where we face similar problems.  `map`, for instance.

andyferris:
> The promote_rule declaration is specific to Array, and doesn't work between Array and UnitRange, for instance.

OK, in that case I'm changing my mind - let's implement this only for `SArray`, `MArray`, and `SizedArray` (individually).

We shouldn't simply "guess" what combinations of different array types are. Consider `[RotX(1.0), ones(SMatrix{3,3})]` - promoting both to a rotation would be a real problem! (The second isn't even unitary/orthogonal).

OTOH `map` goes through `similar` / `similar_type` which is a little bit safer since they can be set to a more "generic" container, e.g. `similar_type` for rotations are `SMatrix{3,3}`.

c42f:
Ok, I've changed this around as suggested.  Not quite happy about the duplication, but let's go with it for now.

As part of implementing this for `SizedArray`, I found a whole heap of brokenness, so I fixed that as well and added various tests.

c42f:
By the way, you may find the SizedArray changes easier to review by looking at the second commit in isolation.

PR #221, opened 2017-06-07
unrol inv for 4x4 matrix

### LINKED ISSUE #220: inv not unrolled for Mat{4,4}
Is there a reason that https://github.com/SimonDanisch/FixedSizeArrays.jl/blob/f1b3ec2bfbf376345fa28c863a20301498f0e337/src/ops.jl#L234 didn't make it into StaticArrays?

I could open a PR. It's twice as fast and doesn't destroy my immutable only pipeline ;)

SimonDanisch:
fixes #220


c42f:
Looks good, but needs tests?

c42f:
Actually, this method seems to perform really poorly for some badly conditioned matrices.  For example:

```julia
julia> x = 1:4
1:4

julia> A = x.+x' + 1e-5*rand(4,4)
4×4 Array{Float64,2}:
 2.00001  3.00001  4.00001  5.00001
 3.0      4.0      5.00001  6.00001
 4.00001  5.00001  6.00001  7.00001
 5.00001  6.00001  7.0      8.0

julia> SA = SMatrix{4,4}(A)
4×4 StaticArrays.SArray{Tuple{4,4},Float64,2,16}:
 2.00001  3.00001  4.00001  5.00001
 3.0      4.0      5.00001  6.00001
 4.00001  5.00001  6.00001  7.00001
 5.00001  6.00001  7.0      8.0

julia> A*inv(A) - eye(4)
4×4 Array{Float64,2}:
 1.27949e-10  -9.74902e-10  2.05975e-9  -1.28684e-9
 1.80956e-10  -1.84883e-9   3.82326e-9  -1.87593e-9
 2.65428e-10  -1.44447e-9   3.19285e-9  -1.73033e-9
 2.52043e-10  -2.10835e-9   4.36524e-9  -2.12032e-9

julia> SA*inv(SA) - eye(4)
4×4 StaticArrays.SArray{Tuple{4,4},Float64,2,16}:
 -0.00332528  -0.00309416  0.00149557  -0.00157627
 -0.0045737   -0.00442967  0.00152296  -0.00157071
 -0.00582214  -0.00576519  0.00155034  -0.00156514
 -0.00707057  -0.00710071  0.00157773  -0.00155958
```

I assume there's some disastrous cancellation going on here.

Whether or not this is a problem depends on the application.

SimonDanisch:
Wouldn't surprise me.. Well for my use cases it seems like it was working fine up to now..
I checked and there are already tests for SMatrix{4,4} & inv, which this PR passes ;)

KristofferC:
I think the unrolled versions have numerical stability for all sizes, ref: https://github.com/JuliaLang/julia/pull/12454

c42f:
@KristofferC thanks for the link :+1:  Isn't the 3x3 method implemented there just like our 4x4 version here?  Our 4x4 version here seems pretty unstable, despite being completely unrolled.

I implemented the following on the bus this morning, based on a partitioning method.  A recursive partitioning method might be just the thing for `StaticArrays` and let us have specialized versions up to 10x10, say.

I'm not sure how this compares to LU (which we haven't implemented yet in StaticArrays).  Basing this all off a good unrolled LU might be a better option.

```julia
@noinline function inv4x4(a::StaticMatrix{4,4})
    # See http://www.freevec.org/function/inverse_matrix_4x4_using_partitioning
    # Partition
    i12 = SVector(1,2)
    i34 = SVector(3,4)
    @inbounds P = a[i12,i12]
    @inbounds Q = a[i12,i34]
    @inbounds R = a[i34,i12]
    @inbounds S = a[i34,i34]
    invP = inv(P)
    invP_Q = invP*Q
    S2 = inv(S - R*invP_Q)
    R2 = -S2*(R*invP)
    Q2 = -invP_Q*S2
    P2 = invP - invP_Q*R2
    [[P2 Q2];
     [R2 S2]]
end
```

c42f:
By the way, this block version seems insanely faster and more stable for the one test I've checked it with... unless I just botched my benchmarks :-)  No idea why, it's possible that the 4x4 version here still has some bounds checks or something?

SimonDanisch:
That's indeed fascinating.
It's 24ns on my system, which is almost nothing for the amount of work (360 for my version).
Seems to also yield the right answer, though! :)
Turns out that `det` is the culprit here, taking about 270ns.

Lol, how the hell is this generating an SMatrix:
```Julia
[[P2 Q2];
     [R2 S2]]
```
Did you just hack this into vcat/hcat? :) :+1:

andyferris:
`vcat` and `hcat` are already defined - awesome syntax hack, however! 

LLVM does a nice job at optimising. 24ns is amazing!

Nice, Chris :)

SimonDanisch:
Sweet! Well, I guess I just close this and let @c42f  create his version as a PR ;)

PR #222, opened 2017-06-08
Add specialised *_mul_* and *_ldiv_* methods for triangular matrices

bshall:


bshall:
It looks like Travis stopped because the tests took more than 10 minutes to run. Should I reduce the number of tests?

andyferris:
Try removing the outer testset, [here](https://github.com/JuliaArrays/StaticArrays.jl/blob/master/test/runtests.jl#L4), so that it makes some output now-and-then.

andyferris:
How many minutes does just `test/triangular.jl` take locally on your machine?

bshall:
It takes about 5 minutes to run them locally.

bshall:
Looks like 1 of the jobs still aborts.

andyferris:
5 minutes for just `test/triangular.jl` seems too long to me. I like to run the unit tests regularly during development, so where we can we try to keep the tests lean.

We should have a separate place for fuzz testing (a variety of sizes and types). My opinion is the unit tests should mostly be designed to cover the codepaths in the code you wrote, with maybe two eltypes to test promotion and so-on, rather than proving compatibility with all combinations of `Float32, Float64, BigFloat, Complex64, Complex128, Complex{BigFloat}, Int`. As you have it, this test will have to compile `4*7*2*7 = 392` times. For instance, try `n=1:4`, `eltype = (Int, Float64)`, and both upper/lower triangular.

andyferris:
BTW, how does speed compare for these vs. what we had before vs. standard approach in `Base`?

bshall:
Yeah, that makes sense. I'll remove some of the combinations and make the tests quicker. 

As for the performance I've run benchmarks for a couple of the operations listed below. Here `A` is a n by n `UpperTriangular` matrix and `B = A*ones(n,n)`. The improvement in performance of `A_ldiv_B` seen for larger values of `n` is primarily due to a better array access pattern (I hadn't noticed before because I only benchmarked for `n <= 4`). There are other specialised operations not included in the benchmark (e.g. if `A` is a `StaticMatrix` and `B` is a `AbstractTriangular`) which should give similar improvements. If you'd like me to run more of these benchmarks let me know. Also, the * after some of the timings denotes operations which fall back to `Base`. Finally, by wrapped I mean that the benchmarks were run for the matrix `A` wrapped as an `UpperTriangular` and for unwrapped I mean `A` is an ordinary `SMatrix` with the lower triangle set to zeros. The benchmark script is [here](https://gist.github.com/bshall/bd3006e170ce8c0c6137b27defb684bd) if anyone wants to reproduce the results.

For `n = 3`:

| Operation | Base | Master(wrapped) | Master(unwrapped) | This PR |
| --- | --- | --- | --- | --- |
| `A_mul_B` | 203.312 ns  | 64.403 ns* | 8.311 ns | 5.412 ns |
| `Ac_mul_B` | 199.209 ns | 84.116 ns* | 8.346 ns | 5.187 ns |
| `At_mul_B` | 199.209 ns | 85.315 ns* | 8.347 ns | 5.681 ns |
| `A_ldiv_B` | 253.738 ns | 53.368 ns | 76.033 ns | 54.009 ns |
| `Ac_ldiv_B` | 304.408 ns | 190.983 ns* | 79.244 ns | 52.825 ns |
| `At_ldiv_B` | 218.962 ns | 186.916 ns* | 81.335 ns | 54.253 ns |

For `n = 6`:

| Operation | Base | Master(wrapped) | Master(unwrapped) | This PR |
| --- | --- | --- | --- | --- |
| `A_mul_B` | 229.176 ns  | 215.951 ns* | 72.396 ns | 44.307 ns |
| `Ac_mul_B` | 232.043 ns | 255.097 ns* | 70.674 ns | 45.489 ns |
| `At_mul_B` | 235.068 ns | 249.166 ns* | 70.679 ns | 44.176 ns |
| `A_ldiv_B` | 384.966 ns | 327.304 ns | 366.127 ns | 234.468 ns |
| `Ac_ldiv_B` | 409.290 ns | 513.073 ns* |  411.910 ns* | 216.000 ns |
| `At_ldiv_B` | 407.085 ns | 512.879 ns* | 401.250 ns* | 221.994 ns |

For `n = 10`:

| Operation | Base | Master(wrapped) | Master(unwrapped) | This PR |
| --- | --- | --- | --- | --- |
| `A_mul_B` | 1.302 μs  | 685.338 ns* | 345.414 ns | 185.126 ns |
| `Ac_mul_B` | 1.277 μs | 756.514 ns* | 395.595 ns | 185.291 ns |
| `At_mul_B` | 1.285 μs | 758.555 ns* | 405.090 ns | 185.385 ns |
| `A_ldiv_B` | 1.597 μs | 1.328 μs | 1.388 μs | 686.338 ns |
| `Ac_ldiv_B` | 1.627 μs | 1.223 μs* | 1.555 μs* | 638.475 ns |
| `At_ldiv_B` | 1.615 μs | 1.250 μs* | 1.520 μs* | 639.413 ns |

bshall:
I've removed the extra test cases and the triangular tests now run in slightly under a minute on my local machine. I left the `Complex128` type just to test that the conjugation operations (`Ac_mul_B`, `Ac_ldiv_B`, etc.) are working properly. If you think that the tests still take too long I can remove it though.

andyferris:
So we've had a test failure due to the `rand` methods making (almost) singular matrices.

You can fix this by setting a random seed with `srand`, or by constructing non-singular matrices (e.g. `m = rand(n,n); eye(n,n) + m'*m`, but you probably want non-symmetric and even degenerate examples).

However, I tend to just put in *one* calculation I can do analytically (say, with integers) and confirm it gets the correct result.

Also, the test seems to be using `Base`'s LU factorization - is that expected?

```
Solving linear system: Error During Test
  Test threw an exception of type Base.LinAlg.SingularException
  Expression: (StaticArrays.SArray{Tuple{2,2},T,2,L} where L where T)(A) \ (StaticArrays.SArray{Tuple{2},T,1,2} where T)(b) ≈ A \ b
  Base.LinAlg.SingularException(2)
  Stacktrace:
   [1] A_ldiv_B! at ./linalg/lu.jl:238 [inlined]
   [2] \(::Base.LinAlg.LU{Float64,Array{Float64,2}}, ::Array{Float64,1}) at ./linalg/factorization.jl:48
   [3] \(::Array{Float64,2}, ::Array{Float64,1}) at ./linalg/generic.jl:817
   [4] eval(::Module, ::Any) at ./boot.jl:235
   [5] macro expansion at /home/travis/.julia/v0.6/StaticArrays/test/solve.jl:12 [inlined]
   [6] macro expansion at ./test.jl:856 [inlined]
   [7] anonymous at ./<missing>:?
```

src/triangular.jl
```diff
@@ -0,0 +1,553 @@
...
+A_mul_Bt(A::Base.LinAlg.AbstractTriangular{<:Any,<:StaticMatrix}, B::StaticMatrix) = (*)(A, transpose(B))
+Ac_mul_Bc(A::Base.LinAlg.AbstractTriangular{<:Any,<:StaticMatrix}, B::StaticMatrix) = Ac_mul_B(A, B')
+Ac_mul_Bc(A::StaticMatrix, B::Base.LinAlg.AbstractTriangular{<:Any,<:StaticMatrix}) = A_mul_Bc(A', B)
+At_mul_Bt(A::Base.LinAlg.AbstractTriangular{<:Any,<:StaticMatrix}, B::StaticMatrix) = At_mul_B(A, B.')
+At_mul_Bt(A::StaticMatrix, B::Base.LinAlg.AbstractTriangular{<:Any,<:StaticMatrix}) = A_mul_Bt(A.', B)
+
+@inline \(A::Union{UpperTriangular{<:Any,<:StaticMatrix},LowerTriangular{<:Any,<:StaticMatrix}}, B::StaticVecOrMat) = _A_ldiv_B(Size(A), Size(B), A, B)
+@inline Ac_ldiv_B(A::Union{UpperTriangular{<:Any,<:StaticMatrix},LowerTriangular{<:Any,<:StaticMatrix}}, B::StaticVecOrMat) = _Ac_ldiv_B(Size(A), Size(B), A, B)
+@inline At_ldiv_B(A::Union{UpperTriangular{<:Any,<:StaticMatrix},LowerTriangular{<:Any,<:StaticMatrix}}, B::StaticVecOrMat) = _At_ldiv_B(Size(A), Size(B), A, B)
+
```
c42f: This is an impressive amount of functionality!

My only worry is that there's a lot of almost-duplication here between the different methods. Did you try just transposing / conjugating the matrices before passing to a couple of unified `_A_mul_B` functions? While this would be a very bad idea for large matrices, for tiny ones I'd guess that it's almost as efficient, and may be exactly the same when all loops are completely unrolled and operations inlined.

If so, we could have the same functionality here, but with a *lot* less code.

test/triangular.jl
```diff
@@ -0,0 +1,65 @@
+@testset "Triangular-matrix multiplication" begin
```
c42f: Worth adding an `srand` here, to keep testing consistent and predictable.

Generally I'd agree with @andyferris that hard coded test cases are better for basic unit tests (simply from a clarity point of view in being extremely explicit when something goes wrong).  However, you've got a lot of cases here, so hard coding them all doesn't seem practical.  (Generally I'd also aim to cover each code path only once or so for each type of input, just to keep testing time under control.  We do need a better way to do better comprehensive testing vs fast unit tests, but that's a problem for the package as a whole - I'd say we go with what's here for now :-) )

test/runtests.jl
```diff
@@ -1,36 +1,35 @@
...
-    include("MVector.jl")
-    include("SMatrix.jl")
-    include("MMatrix.jl")
-    include("SArray.jl")
-    include("MArray.jl")
-    include("FieldVector.jl")
-    include("Scalar.jl")
-    include("SUnitRange.jl")
-    include("SizedArray.jl")
-    include("custom_types.jl")
```
bshall: With the triangular tests included the tests take a little longer than 10 minutes to run and if Travis doesn't see any output for longer than 10 minutes it terminates. @andyferris suggested removing the outer testset to prevent this.

src/triangular.jl
bshall: Yeah, the methods are rather similar and I can definitely see the value of reducing the amount of code to maintain. In the Transpose column above I've included some benchmarks of just transposing / conjugating the matrices before the `_A_mul_B` function is called. For `n <= 4` the decrease in performance is pretty negligible but after this point it seems to make a "practical" difference. For example in the `n = 6` case you get a 1.4x speedup.

I know this package is geared towards small matrices so if you guys think it's not worth the extra maintenance burden I'd be happy to replace the extra methods.

bshall:
@andyferris I don't think that failure is from the triangular tests. I moved all the tests for this PR into `test\triangular.jl` so I think the failure is for the general solve method already in `StaticArrays` (which explains why `Base`'s LU factorization is called.

test/triangular.jl
bshall: Cool, I can definitely add an `srand`. Also, I'm happy to spend some time writing explicit test cases if you'd prefer that. Another way of reducing test time is to test only for a single value of `n`. When there are specific methods for each size they definitely need to be tested independently but with a general method I'm not sure how much you gain by testing multiple sizes.

andyferris:
Sorry, I merged Katie's test and seem to have made a conflict in `test/solve.jl`...

test/runtests.jl
c42f: Ah, good to know, thanks.

src/triangular.jl
c42f: 1.4x definitely seems worthwhile to me, and I'd certainly like `StaticArrays` to be solidly implemented for matrices larger than 4x4. Would a reasonable alternative be to add a bit more logic inside the code generators to deal with the different ordering for both transposed and non-transposed?  Looking through the multiplication code, it looks like the entire method is almost identical for each case, with just a bit of the inner loop being different.

I think we'd at least like `StaticArrays` to work up until the limit where unrolling and stack allocation ceases to help.  Based on previous experience this seems *very* roughly to be a bit more than 10x10 matrices and less than 20x20.  This depends on the architecture and particular algorithm.  Ideally it would degrade gracefully for sizes larger than that.

andyferris: > Ideally it would degrade gracefully for sizes larger than that.

Yes. To give a rough rule, N^3-sized code expansions should bail-out to code which is no more than N^2 for some value of N around 8 (look at matrix_multiply.jl) because they tend to be faster, and we usually have serious problems when N > 14 (where 14 is a magical "it makes the compiler's head hurt" number). To make algorithms with less code you can introduce a loop or some `@noinline` functions.

(@c42f also note I have quite significant amounts of code duplication throughout the package... not that this is good!)

c42f: Hah, I know.  So in some sense this is an unfair standard to ask for, but in another sense we shouldn't have more of a bad thing ;-) Refactoring code in this package is already too much of an epic task...

@bshall is doing a great job so far, so I was hoping it's not too much to ask for a bit more code refactoring in the effort to have some code length reduction, provided it doesn't make things too hard to read :-)

bshall: Thanks for the feedback! I'll have some time this weekend to do the looping versions for larger values of N. As for the refactoring, I think the best option might be a `_A_mul_B` function with a BLAS like signature, by which I mean adding two extra arguments of type `Type{Val{X}}` where `X` is one of the characters "N", "T" or "C" specifying whether `A` or `B` should be transposed (or conjugate transposed). What are your thoughts on this idea? I'm happy to go ahead with it if you guys are satisfied.

I'm not sure if lazy transpose or conjugate wrappers are on the cards anytime in the near future but there's also the possibility of waiting for that.

c42f: That sounds good to me.  You'd need to use a `Val{Symbol}` to get the `Val` thing work.

My impression is that the special lowering of transpose/conjugate multiplication will go away when Base gets transpose/conjugate wrappers, so we'd need to revisit all this again at that stage.  For now I think we ignore it.  @andyferris knows the state of this better than me, I suspect.

andyferris:
Sorry we've been a bit quiet lately - busy with JuliaCon and work. This is an impressive amount of work, which will be important to the package - thanks.

I was wondering if we could benchmark putting `sub2ind` inside the generators. At the very least, it might make compilation faster (Julia will currently inline `sub2ind` for each `getindex` which would take longer for the compiler as compared to just calling it from the generator). Also, linear indexing is easier deal with and for LLVM to introspect and make optimizations, so you may (or may not) see speed improvements. Anyway, worth a try for just one of your methods to see what happens.

The large combinations of transposes/etc is a bit annoying. Not sure if something generic can be done, but I have experience doing what you have. I really want to get a transposed matrix wrapper into Julia Base and that might help...

bshall:
No problem at all.

Yeah, had just finished changing the indexing to use `sub2ind` and it significantly reduces the compilation time (espessially for larger matrices) so I've used it throughout now.

I've also compressed the code a bit using some loops and `@eval`. I can go a lot further than this (by reducing all the branches to ternary operators that manipulate the expressions based on whether things should be transposed or conjugated) but I found that I was pretty unreadable and probably more of a maintenance nightmare than just repeating stuff a bit. I think it's already a fair bit harder to read than the original commit so if you'd like me to go back to that let me know.

Also, would you mind restarting the tests for this. I forgot to resolve the conflict before I make the last commit.

bshall:
I also had a loop version for larger matrices but it's quite disappointing since any savings you get by avoiding multiplications with the lower triangular elements are eaten up by all the if statements in the k loop. I couldn't think of a better alternative though. Perhaps you guys have some ideas.

```julia
@generated function _A_mul_B_loop(::Size{sa}, ::Size{sb}, A::UpperTriangular{TA,<:StaticMatrix}, B::StaticArrays.StaticVecOrMat{TB}) where {sa,sb,TA,TB}
    m = sb[1]
    n = length(sb) > 1 ? sb[2] : 1
    if m != sa[1]
        throw(DimensionMismatch("right hand side B needs first dimension of size $(sa[1]), has size $m"))
    end

    TAB = promote_op(matprod, TA, TB)
    X = [Symbol("X_$(i)_$(j)") for i = 1:m, j = 1:n]
    init = [:($(X[i,j]) = A.data[$(sub2ind(sa,i,i))]*B[$(sub2ind(sb,i,j))]) for i = 1:m, j = 1:n]

    code = quote end
    for j = 1:n
        for i = 1:m
            push!(code.args, :(if k > $i; $(X[i,j]) += A.data[$i,k]*B[k,$j]; end))
        end
    end

    return quote
        @_inline_meta
        @inbounds $(Expr(:block, init...))
        for k = 2:$m
            @inbounds $code
        end
        return similar_type(B, $TAB)(tuple($(X...)))
    end
end
```

c42f:
This PR has been languishing (my fault, not yours).  Let's get it finished.

I agree the current version isn't really readable.  I started to have a think about how to make this better and I've got some ideas, but I didn't have enough time to get a proper demo going.  (The main idea is to implement the unrolling via symbolic evaluation of the operations knowing the sparsity pattern, rather than ad-hoc code generation, followed by dumping the explicit symbolic representation.)

But honestly, I think we should just merge this and fuss with making the implementation cleaner as a next step.  To get it merged, let's do the following:

* Resolve the minor conflicts in runtests.jl
* Revert the implementation back to the simpler more explicit version - this will make refactoring simpler later, I think.

How does that sound to you?  I hope it's ok.

andyferris:
I agree, @c42f. We can always improve, refactor, etc later.

mschauer:
I think hard encoded sparsity combined with generated non-fill-in cholesky decompositions is very interesting.

bshall:
No problem at all. I've reverted back to the simpler version and fixed the conflicts. I'm pretty interested to hear more about @c42f 's idea.

c42f:
Great, thanks a lot.

c42f:
I added a call to `srand` in the tests before merging.

c42f:
@bshall here's my idea - use something like the following embarrassingly simple symbolic algebra system inside a generated function, to allow the use of base linear algebra to generate expressions for StaticArrays without so many ugly code generators:

https://github.com/c42f/SimpleSymbolic.jl

So far, this approach seems to work far better than should be possible for a "CAS" written in 67 lines of julia :-P

PR #225, opened 2017-06-12
Another try at 4x4 matrix inverses

c42f:
Potentially replaces #221 

This is an alternative to #221 and works very nicely for well conditioned problems, but it still has its problems.  I wrote some code to fuzz test against a variety of poorly conditioned matrices, and it's not hard to find cases where the implemented version in this PR is significantly worse than `inv(::Matrix)` in Base, as measured by how close `inv(A)*A` is to the identity.

Generally speaking, I'd prefer the `StaticArrays` matrix algorithms to be somewhat more numerically stable than the code I have here.  Still could be that an unrolled LU would be a better idea.

Oh, this code generalizes very easily (just remove the 4x4 restriction), but the larger sizes seem to suffer from more numerical problems in the partitioning.


SimonDanisch:
Can we actually hook into @fastmath?
Would be nice to have some infrastructure to select fast/imprecise slow/precise ;)

src/SUnitRange.jl
```diff
@@ -19,7 +19,7 @@ end
 @pure SUnitRange(a::Int, b::Int) = SUnitRange{a, max(0, b - a + 1)}()
 
 @pure @propagate_inbounds function getindex(x::SUnitRange{Start, L}, i::Int) where {Start, L}
-    @boundscheck if i < Start || i >= (Start + L)
+    @boundscheck if i < 1 || i > L
```
andyferris: Thanks... this thing is half-baked.

c42f:
> more numerically stable than #221?

#221 was quite unstable in one test, but I didn't fuzz test it comprehensively, or even the version here, to be quite honest. I'll try to have a better go at that before merging.

Hmm, `Base.Test` would be a lot more useful if it had a way to group tests into "normal unit tests" vs more extensive randomized tests, and to run individual test files / testsets by regex match of the name, etc. As it is, I've just got the fuzz testing commented out here, which isn't great.

src/SUnitRange.jl
c42f: Now with extra testing :-)  I only used this because I wanted the general recursive version to work... but I got hung up on numerical stability issues.

andyferris:
> Hmm, Base.Test would be a lot more useful if it had a way to group tests into "normal unit tests" vs more extensive randomized tests, and to run individual test files / testsets by regex match of the name, etc. As it is, I've just got the fuzz testing commented out here, which isn't great.

Agreed. As for running subsets of tests by name, I believe that's on the roadmap - to make `Pkg.test` a bit more like `Base.runtests`, so that's great. I'm guessing that with this feature alone, we could have slow, comprehensive tests yet rapid development by targetting just the most relevant tests to the changes made. OTOH having multiple "layers" of testing (unit, fuzz, integration) run at the appropriate times would be wonderful. Here the Julia language uses completely separate things like @nanosoldier for different "styles" of tests (in that case, performance testing).

c42f:
Performance tests might be sufficiently different that `Base.Test` might not be a good tool for the job? Then again, integration between `@testset` and `@benchmark` might work out ok...

c42f:
I hope `@nanosoldier` didn't mind you taking its name in vain there :-P

c42f:
@SimonDanisch hooking into `fastmath` certainly seems doable for a lot of operations.  It's a little limited though, because there's a blessed set of operators supported by the macro, for instance there's no fastmath version of `\`.  It also seems a bit of a heavy handed tool, perhaps?  Still it would be great to have something along these lines!

c42f:
Sorry to be slow on this one.  Here's a comparison of the block decomposition code here vs `inv(::Array)` for 4x4 matrices, with various almost-rank deficiencies.

Looks ok for Rank 4,3,2, but falls apart for rank 1 where the top left 2x2 block becomes rank deficient.  The blob in the bottom left corner shows that for well conditioned matrices this method is perfectly fine.  

It's also insanely faster and no worse than our specialized 3x3 code, so we could just merge it as an improvement for now; if people need the slower more stable version they can always convert to a normal `Array`?

4x4_accuracy

andyferris:
Thanks Chris for the numerical stability analysis.

My opinion is to merge first and find more stable algorithms later (and possibly keep this for `fastmath`). The rank-1 problems seem slightly annoying, but oh well.

andyferris:
yay! :)

c42f:
By the way - if it's not clear, the diagram above was generated by uncommenting the code at the bottom of `test/inv.jl`.

andyferris:
Nice to have included it!

(Funny that github can't handle Julia's "advanced" nested commenting...)

c42f:
Oh.  Err yes, I left in that commented out junk. It was used for printing matrices to make the test cases.

PR #226, opened 2017-06-12
Implement diag() with Val as second argument

wsshin:
This PR implements a type-stable `diag` function for `SMatrix`.  Specifically, for `A::SMatrix`, 

- `diag(A)` returns the main diagonal of `A` as `SVector`.
- `diag(A, Val{k})` returns the `k`th diagonal of `A` as `SVector`.  (The contents of the result are the same as `diag(A, k)`, but the resulting type is `SVector` instead of `Vector`.)

The PR also adds simple tests on the newly implemented `diag`.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -8,7 +8,7 @@ import Base: getindex, setindex!, size, similar, vec, show,
              length, convert, promote_op, promote_rule, map, map!, reduce, reducedim, mapreducedim,
              mapreduce, broadcast, broadcast!, conj, transpose, ctranspose,
              hcat, vcat, ones, zeros, eye, one, cross, vecdot, reshape, fill,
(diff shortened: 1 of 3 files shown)


wsshin:
One question: is `@inferred` used in the test code to check type stability?  I just mimicked the other part of the test code without knowing exactly what the macro does.

andyferris:
Yes - the `@inferred` macro will throw an error if `Core.Inference.return_type()` isn't the same as the return type of the function. It's particularly useful for testing performance-sensitive code.

c42f:
Exactly, nice work here.  I've been wondering whether we should have a shortcut for the repeated occurrence of the pattern `@test @inferred(a) == b` all throughout our tests. Perhaps something simple like

```julia
@test_inferred a == b
```

could turn into

```julia
@test @inferred(a) == @inferred(b)
```

or some such.

PR #227, opened 2017-06-13
Couple missing branches for sqrtm and solve

kshyatt:


andyferris:
Thanks :)

andyferris:
Interesting inference regression on nightly...

PR #229, opened 2017-06-15
Test for fix to impure promote_tuple_eltype

### LINKED ISSUE #228: Calling promote_tuple_eltype() causes promote_type to change behavior
Yeah, this is a weird one. Consider the following:

```julia
module X

using StaticArrays: SVector

struct Foo
    x::Int
end

struct Bar{S}
    x::Int
end

Base.convert(::Type{Foo}, b::Bar) = Foo(b.x)
Base.promote_rule(::Type{Bar{S1}}, ::Type{Bar{S2}}) where {S1, S2} = Foo

end

b1 = X.Bar{:x}(1)
b2 = X.Bar{:y}(2)

@show promote_type(typeof(b1), typeof(b2))
```

shows:

```julia
promote_type(typeof(b1), typeof(b2)) = X.Foo
```

Now try the same thing, but call the SVector constructor before looking at the result of promote_type:

```julia
module X

using StaticArrays: SVector

struct Foo
    x::Int
end

struct Bar{S}
    x::Int
end

Base.convert(::Type{Foo}, b::Bar) = Foo(b.x)
Base.promote_rule(::Type{Bar{S1}}, ::Type{Bar{S2}}) where {S1, S2} = Foo

end

b1 = X.Bar{:x}(1)
b2 = X.Bar{:y}(2)

StaticArrays.SVector(b1, b2)
@show promote_type(typeof(b1), typeof(b2))
```

shows:

```julia
promote_type(typeof(b1), typeof(b2)) = X.Bar
```

In the second example, `promote_type` gives the wrong result (Bar instead of Foo). 

```julia
julia> versioninfo()
Julia Version 0.6.0-rc3.0
Commit ad290e93e4 (2017-06-07 11:53 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin13.4.0)
  CPU: Intel(R) Core(TM) i7-2860QM CPU @ 2.50GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, sandybridge)
```

This seems like it might be a Julia bug, or it might be that StaticArrays is doing something bad with `@pure`. I'm using StaticArrays at 9ccda4e

rdeits:
~~This fixes #228 by changing `promote_tuple_eltype` and moving the call to `promote_type` into the generated expression instead of the generator body. The generated function now just exists to set up the call to promote_type rather than execute it.~~

~~I've also added a simple test for this issue. I *think* that this should work just as well as the old code, since everything still seems to infer just fine, but I could be wrong.~~

Edit: the fix came in with c089124 so this just adds a regression test

### DIFF
--- a/src/util.jl
+++ b/src/util.jl
@@ -14,7 +14,7 @@ end
 
 # Base gives up on tuples for promote_eltype... (TODO can we improve Base?)
 @generated function promote_tuple_eltype{T <: Tuple}(::Union{T,Type{T}})
-    types = []
+    t = Union{}
     for i = 1:length(T.parameters)
         tmp = T.parameters[i]
         if tmp <: Vararg
@@ -27,7 +27,6 @@ end
         $t
     end
 end
-
 
 # The ::Tuple variants exist to make sure that anything that calls with a tuple
 # instead of a Tuple gets through to the constructor, so the user gets a nice
--- a/src/util.jl
+++ b/src/util.jl
@@ -14,19 +14,20 @@ end
 
 # Base gives up on tuples for promote_eltype... (TODO can we improve Base?)
 @generated function promote_tuple_eltype{T <: Tuple}(::Union{T,Type{T}})
-    t = Union{}
+    types = []
     for i = 1:length(T.parameters)
         tmp = T.parameters[i]
         if tmp <: Vararg
             tmp = tmp.parameters[1]
         end
-        t = promote_type(t, tmp)
+        push!(types, tmp)
     end
     return quote
         $(Expr(:meta,:pure))
-        $t
+        $(Expr(:call, :promote_type, types...))
     end
 end
+
 
 # The ::Tuple variants exist to make sure that anything that calls with a tuple
 # instead of a Tuple gets through to the constructor, so the user gets a nice
--- a/test/promotion.jl
+++ b/test/promotion.jl
@@ -1,0 +1,19 @@
+# Test for https://github.com/JuliaArrays/StaticArrays.jl/issues/228
+module PromotionTest
+    struct Foo
+        x::Int
+    end
+
+    struct Bar{S}
+        x::Int
+    end
+
+    Base.promote_rule(::Type{Bar{S1}}, ::Type{Bar{S2}}) where {S1, S2} = Foo
+end
+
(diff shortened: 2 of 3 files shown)


src/util.jl
```diff
@@ -14,20 +14,21 @@ end
...
     for i = 1:length(T.parameters)
         tmp = T.parameters[i]
         if tmp <: Vararg
             tmp = tmp.parameters[1]
         end
-        t = promote_type(t, tmp)
+        push!(types, tmp)
     end
     return quote
         $(Expr(:meta,:pure))
```
rdeits: NB: I don't actually know what `Expr(:meta, :pure))` does here, so I left it alone

andyferris: It's what the `@pure` macro does. `@pure function f(x); ...; end` transforms to `function f(x); $(Expr(:meta, :pure)); ...; end` (similarly for `@inline`, `@noinline` and `@propagate_inbounds`). A better style is to use a `@_pure_meta` macro call here (which just inserts `$(Expr(:meta, :pure))`).

However, I've replaced this with `@_inline_meta` here since pure is the wrong concept for this.

andyferris:
Sorry I've updated this on master in parallel. I expand the `promote_type` into nested calls so it works better for very large tuples.

Not sure if the tests here are worthwhile to keep? Its kind of esoteric, OTOH it is good practice to have regression tests. (I can see this regressing)

rdeits:
I think the cost of having the test is pretty low, and it might catch a regression someday. I'll update this PR to just include the test, since this is fixed in master.

c42f:
@andyferris I'm not sure about the background on this one. Should we merge this or close it?

c42f:
Test failures look to be julia-0.7 failures with the nightlies, btw.

c42f:
Needs a rebase, I guess (or reset and cherry pick of 20da08c, perhaps).

rdeits:
I suspect this isn't worth the effort to rebase and merge, so I'll close it.

PR #237, opened 2017-06-25
Add complex matrix exponential for 2x2

mschauer:
Extends @Godisemo 's code to the complex case.

### DIFF
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -40,7 +40,27 @@ end
     (newtype)((m11, m21, m12, m22))
 end
 
-# TODO add complex valued expm
+@inline function _expm(::Size{(2,2)}, A::StaticMatrix{<:Any,<:Any,<:Complex})
+    T = typeof(exp(zero(eltype(A))))
+    newtype = similar_type(A,T)
+
+    @inbounds a = A[1]
+    @inbounds c = A[2]
+    @inbounds b = A[3]
+    @inbounds d = A[4]
+
+    z = sqrt((a-d)*(a-d) + 4.0*b*c )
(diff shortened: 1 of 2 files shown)


andyferris:
Cool - good work.

Just a question - is there a reason for having a separate `Real` codepath? For reals, do you know if one method faster or more accurate than the other? (If we can just use one "complex safe" method I would be quite happy :) )

Godisemo:
My real version works is based on a version that works for complex, it is a matter of interchanging sinh and sin and adding an absolute value and some more detail. Conclusion is that it is a lot slower than the specialized real version.

mschauer:
It looks like it is worthwhile to keep a specialized real version.

src/expm.jl
```diff
@@ -40,7 +40,28 @@ end
...
+    newtype = similar_type(A,T)
+
+    @inbounds a = A[1]
+    @inbounds c = A[2]
+    @inbounds b = A[3]
+    @inbounds d = A[4]
+
+    z = sqrt((a-d)*(a-d) + 4.0*b*c )
+    e = exp(a/2.0 + d/2.0 - z/2.0)
+    f = exp(a/2.0 + d/2.0 + z/2.0)
```
c42f: A few operations can be saved here by factoring the division by 2.0.

Also, it's probably better to avoid `Float64` literals to make promotion work more sensibly - so set `Tr = real(T)` at the top and use `Tr(2)` rather than `2.0` here and elsewhere in this function.

mschauer: I replaced it with multiplication/division by integer which preserve floating point accuracy

Issue #235, opened 2017-06-22
a Diagonal of a SVector doesn't keep its type 


yakir12:
First seen [here](https://discourse.julialang.org/t/how-to-scale-coordinates-in-coordinatetransformations-jl/4394?u=yakir12):
```julia
julia> m = SVector(1,2,3)
3-element SVector{3,Int64}:
 1
 2
 3

julia> x = Diagonal(m)
3×3 Diagonal{Int64}:
 1  ⋅  ⋅
 ⋅  2  ⋅
 ⋅  ⋅  3

julia> x.diag
3-element Array{Int64,1}:
 1
 2
 3
```

fredrikekre:
Thats by definition of [`Diagonal` in `Base`](https://github.com/JuliaLang/julia/blob/8915d9b3384d5ffbafeb83974c267d983f4cb07d/base/linalg/diagonal.jl#L5-L7), it is always a `Vector`.

yakir12:
I see.

fredrikekre:
Could perhaps define an unexported `Diagonal` in `StaticArrays`. Something like:
```julia
struct Diagonal{T, N}
    diag::SVector{N, T}
end
Base.Diagonal(v::SVector) = Diagonal(v)
```

```julia
julia> m = SVector(1,2,3)
3-element SVector{3,Int64}:
 1
 2
 3

julia> Diagonal(m)
StaticArrays.Diagonal{Int64,3}([1, 2, 3])
```
Perhaps too many additional methods are needed to make this usable though.

c42f:
Oh, thanks for pointing this out.  It's another of many places in `Base.LinAlg` where things don't work quite right if you're using `StaticArrays`.

Obviously it's best to fix Base for 0.7, though I guess as a rather nasty hack, we could overload `Diagonal(::StaticArray)` to return our own special `StaticArrays.Diagonal`.  Hmm.

c42f:
Or we could just define `SDiagonal` for now.

yakir12:
The only objective I'd care about, as a user, is the performance and ease of use. So aesthetics aside, `SDiagonal` sounds great.

c42f:
[...]

PR #240, opened 2017-07-02
Porting SDiagonal from Bridge.jl

mschauer:
This was originally written for FixedSizeArrays and ported to StaticArrays by @Getzdan. As I think we both still aren't very familiar with the internals of StaticArrays please review carefully.

Tag #235


mschauer:
I add tests the following days if nobody beats me.

mschauer:
I put some more work into this, this is probably halfway done, help is most welcome.

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+Base.@propagate_inbounds function getindex{N,T}(D::SDiagonal{N,T}, i::Int, j::Int)  
+    @boundscheck checkbounds(D, i, j)
+    if i == j
+        @inbounds return D.diag[i]
+    else
+        zero(T)
+    end
+end
+
+# avoid linear indexing?
```
andyferris: Would be nice. Unfortunately most the static arrays internal code assumes a linear indexing style - we should change this, but it won't be easy...

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+@generated function SDiagonal{N,T}(a::SMatrix{N,N,T})
+    expr = [:(a[$i,$i]) for i=1:N]
+    :(SDiagonal{N,T}($(expr...)))
+end
+
+
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal{N,T}) = D
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal) = SDiagonal{N,T}(convert(SVector{N,T}, D.diag))
+
+size{N}(D::SDiagonal{N}) = (N,N)
```
andyferris: Is this necessary? There should be a fallback already.

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+@inline (::Type{SDiagonal}){N,T}(a::SVector{N,T}) = SDiagonal{N,T}(a) 
+
+@generated function SDiagonal{N,T}(a::SMatrix{N,N,T})
+    expr = [:(a[$i,$i]) for i=1:N]
+    :(SDiagonal{N,T}($(expr...)))
+end
+
+
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal{N,T}) = D
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal) = SDiagonal{N,T}(convert(SVector{N,T}, D.diag))
```
andyferris: `D::SDiagonal{N}` maybe?

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
+# Originally contributed by D. Getz (https://github.com/getzdan), M. Schauer
+# at https://github.com/mschauer/Bridge.jl under MIT License
+
+import Base: getindex,setindex!,==,-,+,*,/,\,transpose,ctranspose,convert, size, abs, real, imag, conj, eye, inv
```
andyferris: merge these imports with the global ones?

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+import Base.LinAlg: ishermitian, issymmetric, isposdef, factorize, diag, trace, det, logdet, expm, logm, sqrtm
+
+@generated function scalem{T, M, N}(a::SMatrix{M,N, T}, b::SVector{N, T})
```
andyferris: use `StaticMatrix` and `StaticVector` not `SMatrix` and `SVector` in the signature, please. Same with the method below.

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+    expr = vec([:(a[$j,$i]*b[$i]) for j=1:M, i=1:N])
+    :(SMatrix{M,N,T}($(expr...)))
```
andyferris: use the `similar_type` interface here.

src/SDiagonal.jl
andyferris: Also, we don't need to assume the eltype is the same.

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+struct SDiagonal{N,T} <: StaticMatrix{N, N, T}
+    diag::SVector{N,T}
+    SDiagonal{N,T}(diag::SVector{N,T}) where {N,T} = new(diag)
+end    
+diagtype{N,T}(::Type{SDiagonal{N,T}}) = SVector{N,T}
+diagtype{N}(::Type{SDiagonal{N}}) = SVector{N}
+diagtype(::Type{SDiagonal}) = SVector
+
+# this is to deal with convert.jl
+@inline (::Type{SD})(a::AbstractVector) where {SD <: SDiagonal} = SDiagonal(diagtype(SD)(a))
```
andyferris: I guess this should be `SDiagonal(convert(diagtype(SD), a))`

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+diagtype{N,T}(::Type{SDiagonal{N,T}}) = SVector{N,T}
+diagtype{N}(::Type{SDiagonal{N}}) = SVector{N}
+diagtype(::Type{SDiagonal}) = SVector
+
+# this is to deal with convert.jl
+@inline (::Type{SD})(a::AbstractVector) where {SD <: SDiagonal} = SDiagonal(diagtype(SD)(a)) 
+@inline (::Type{SD})(a::Tuple) where {SD <: SDiagonal} = SDiagonal(diagtype(SD)(a)) 
+@inline (::Type{SDiagonal}){N,T}(a::SVector{N,T}) = SDiagonal{N,T}(a) 
+
+@generated function SDiagonal{N,T}(a::SMatrix{N,N,T})
```
andyferris: This seems like a `convert` method to me

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+    :(SDiagonal{N,T}($(expr...)))
+end
+
+
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal{N,T}) = D
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal) = SDiagonal{N,T}(convert(SVector{N,T}, D.diag))
+
+size{N}(D::SDiagonal{N}) = (N,N)
+
+function size{N}(D::SDiagonal{N},d::Int)
```
andyferris: There is a global `StaticArray` fallback for this also.

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+    SDiagonal{N,T}(diag::SVector{N,T}) where {N,T} = new(diag)
+end    
+diagtype{N,T}(::Type{SDiagonal{N,T}}) = SVector{N,T}
+diagtype{N}(::Type{SDiagonal{N}}) = SVector{N}
+diagtype(::Type{SDiagonal}) = SVector
+
+# this is to deal with convert.jl
+@inline (::Type{SD})(a::AbstractVector) where {SD <: SDiagonal} = SDiagonal(diagtype(SD)(a)) 
+@inline (::Type{SD})(a::Tuple) where {SD <: SDiagonal} = SDiagonal(diagtype(SD)(a)) 
+@inline (::Type{SDiagonal}){N,T}(a::SVector{N,T}) = SDiagonal{N,T}(a)
```
andyferris: `a::StaticVector{N,T}`

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+-(A::SDiagonal) = SDiagonal(-A.diag)
++(Da::SDiagonal, Db::SDiagonal) = SDiagonal(Da.diag + Db.diag)
+-(Da::SDiagonal, Db::SDiagonal) = SDiagonal(Da.diag - Db.diag)
+-(A::SDiagonal, B::SMatrix) = eye(typeof(B))*A - B
+
+*{T<:Number}(x::T, D::SDiagonal) = SDiagonal(x * D.diag)
+*{T<:Number}(D::SDiagonal, x::T) = SDiagonal(D.diag * x)
+/{T<:Number}(D::SDiagonal, x::T) = SDiagonal(D.diag / x)
+*(Da::SDiagonal, Db::SDiagonal) = SDiagonal(Da.diag .* Db.diag)
+*(D::SDiagonal, V::SVector) = D.diag .* V
```
andyferris: `V::AbstractVector`

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+*(V::SVector, D::SDiagonal) = D.diag .* V
```
andyferris: Should `V` be a `RowVector`?

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+*(A::SMatrix, D::SDiagonal) = scalem(A,D.diag)
```
andyferris: `A::AbstractMatrix` (and below)

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+*(D::SDiagonal, A::SMatrix) = scalem(D.diag,A)
+\(D::SDiagonal, b::SVector) = D.diag .\ b
```
andyferris: Can we make this work for matrices as well?

also `/(vector, diagonal)`?

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+    -pi<imag(x)<pi ? x : real(x)+(mod2pi(imag(x)+pi)-pi)*im
+end
+
+eye{N,T}(::Type{SDiagonal{N,T}}) = SDiagonal(ones(SVector{N,T}))
+
+expm(D::SDiagonal) = SDiagonal(exp.(D.diag))
+logm(D::SDiagonal) = SDiagonal(log.(D.diag))
+sqrtm(D::SDiagonal) = SDiagonal(sqrt.(D.diag))
+
+\(D::SDiagonal, B::SMatrix) = scalem(1 ./ D.diag, B)
```
andyferris: `B::AbstractMatrix` (and below)

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+/(B::SMatrix, D::SDiagonal) = scalem(1 ./ D.diag, B)
+\(Da::SDiagonal, Db::SDiagonal) = SDiagonal(Db.diag ./ Da.diag)
+/(Da::SDiagonal, Db::SDiagonal) = SDiagonal(Da.diag ./ Db.diag )
+
+function inv{N,T}(D::SDiagonal{N,T})
+    for i = 1:N
```
andyferris: Possibly `@inbounds` ?

andyferris: Also, we generally unroll such loops elsewhere... 

It's also slightly annoying that the code to check if we should throw a singular exception will be slower than taking the inverse itself... but I guess `Base.Diagonal` does this too?

mschauer: Yes, but I guess the divisions will still be the most expensive operation here.

src/SDiagonal.jl
mschauer: yes

src/SDiagonal.jl
mschauer: ok

src/SDiagonal.jl
mschauer: would this work? this calls the inner constructor

src/SDiagonal.jl
mschauer: In diagonal.jl this is a constructor.

src/SDiagonal.jl
mschauer: ok

src/SDiagonal.jl
mschauer: ok

src/SDiagonal.jl
```diff
@@ -0,0 +1,117 @@
...
+function size{N}(D::SDiagonal{N},d::Int)
+    if d<1
+        throw(ArgumentError("dimension must be ≥ 1, got $d"))
+    end
+    return d<=2 ? N : 1
+end
+
+Base.@propagate_inbounds function getindex{N,T}(D::SDiagonal{N,T}, i::Int, j::Int)  
+    @boundscheck checkbounds(D, i, j)
+    if i == j
```
mschauer: also looks nicer

src/SDiagonal.jl
mschauer: or rather: this is already a fallback https://github.com/JuliaLang/julia/blob/master/base/linalg/rowvector.jl#L181 which looks like it is no overhead

mschauer:
Not at all, thank you for the careful reading!

src/SDiagonal.jl
```diff
@@ -0,0 +1,109 @@
+# Originally contributed by D. Getz (https://github.com/getzdan), M. Schauer
+# at https://github.com/mschauer/Bridge.jl under MIT License
+
+import Base: getindex,setindex!,==,-,+,*,/,\,transpose,ctranspose,convert, size, abs, real, imag, conj, eye, inv
+import Base.LinAlg: ishermitian, issymmetric, isposdef, factorize, diag, trace, det, logdet, expm, logm, sqrtm
+
+@generated function scalem{M, N}(a::StaticMatrix{M,N}, b::StaticVector{N})
+    expr = vec([:(a[$j,$i]*b[$i]) for j=1:M, i=1:N])
+    :(let val1 = ($(expr[1])); similar_type(SMatrix{M,N},typeof(val1))(val1, $(expr[2:end]...)); end)
```
andyferris: Perhaps these should be inlined?

`:(@_inline_meta; let ... )`

src/SDiagonal.jl
```diff
@@ -0,0 +1,109 @@
...
+
+@generated function SDiagonal{N,T}(a::StaticMatrix{N,N,T})
+    expr = [:(a[$i,$i]) for i=1:N]
+    :(SDiagonal{N,T}($(expr...)))
+end
+
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal{N,T}) = D
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal{N}) = SDiagonal{N,T}(convert(SVector{N,T}, D.diag))
+
+Base.@propagate_inbounds function getindex{N,T}(D::SDiagonal{N,T}, i::Int, j::Int)
```
andyferris: `@propagate_inbounds` is needed if you want to defer bounds checking to an inner function. This should be simply `@inline` since you are doing your own `@boundscheck` and `@inbounds` here.

src/SDiagonal.jl
andyferris: Still this

andyferris:
A final style note is that we've been moving to using `where` syntax everywhere, e.g instead of `f{T}(x::T) = ...` we use `f(x::T) where {T} = ...`.

src/SDiagonal.jl
```diff
@@ -0,0 +1,109 @@
...
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal{N,T}) = D
+convert{N,T}(::Type{SDiagonal{N,T}}, D::SDiagonal{N}) = SDiagonal{N,T}(convert(SVector{N,T}, D.diag))
+
+Base.@propagate_inbounds function getindex{N,T}(D::SDiagonal{N,T}, i::Int, j::Int)  
+    @boundscheck checkbounds(D, i, j)
+    @inbounds return ifelse(i == j, D.diag[i], zero(T))
+end
+
+# avoid linear indexing?
+Base.@propagate_inbounds function getindex{N,T}(D::SDiagonal{N,T}, k::Int)
```
andyferris: This is a good use of `@propagate_inbounds`.

andyferris: However, it has already been imported so you don't need the `Base.`

src/SDiagonal.jl
mschauer: ok

src/SDiagonal.jl
mschauer: ok

src/SDiagonal.jl
mschauer: ok

src/SDiagonal.jl
mschauer: ok

mschauer:
@andyferris I hope I addressed all your comments.

andyferris:
> @andyferris I hope I addressed all your comments.

It's looking good. There are some more imports we could move to the main package file, that's about it that I see.

mschauer:
I left them there because they are imported locally elsewhere: https://github.com/JuliaArrays/StaticArrays.jl/search?utf8=✓&q=import&type= We could centralize these imports independently.

andyferris:
Fine - no worries then :)

andyferris:
Thanks again

src/SDiagonal.jl
```diff
@@ -0,0 +1,106 @@
+# Originally contributed by D. Getz (https://github.com/getzdan), M. Schauer
```
c42f: By the way, if you're porting someone else's code, and you want to attribute them, it's also quite possible to make them the author in git using something like `git commit --author="Some Body <some.body@gmail.com>"`

c42f: When I do this kind of thing, I'd be inclined to add the original chunk of code (in probably non-working form) under the other author's name as a single commit (`@mention`ing them to make sure they're happy with that). Then add any necessary changes under your own name in further commits.

The nice thing about doing it this way is you avoid baking authorship into comments which people feel they can't remove (like the one above), but you also get to do proper attribution which I feel is quite important.

Just some thoughts, I'm happy this was merged already.

fredrikekre: When you port entire files (like in this case) it is also possible to preserve the full git history http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/

fredrikekre:
Ref https://github.com/JuliaLang/julia/pull/22718 for a potential future rewrite to utilize `Base.Diagonal` for later Julia versions.

test/SDiagonal.jl
```diff
@@ -0,0 +1,104 @@
...
+        @test SDiagonal{1,Int64}((1,)).diag === SVector{1,Int64}((1,))
+        @test SDiagonal{1,Float64}((1,)).diag === SVector{1,Float64}((1,))
+     
+        @test SDiagonal{4,Float64}((1, 1.0, 1, 1)).diag.data === (1.0, 1.0, 1.0, 1.0)
+        @test SDiagonal{4}((1, 1.0, 1, 1)).diag.data === (1.0, 1.0, 1.0, 1.0)
+        @test SDiagonal((1, 1.0, 1, 1)).diag.data === (1.0, 1.0, 1.0, 1.0)
+
+        # Bad input
+        @test_throws Exception SMatrix{1,Int}()
+        @test_throws Exception SMatrix{2,Int}((1,))
```
tkelman: `@test_throws Exception` is generally a very bad idea. It will pass if the code has a typo.

PR #241, opened 2017-07-05
Integrate StaticArrays with julia 0.6 broadcast

### LINKED ISSUE #239: element-wise `isequal` on SArray of non-numeric entries does not produce SArray of Bool.
When `isequal` is applied element-wise to `SArray`s of numeric entries, we get `SArray`s of `Bool`
```julia
julia> VERSION
v"0.6.1-pre.0"

julia> v_int = @SVector([1,2,3])
3-element SVector{3,Int64}:
 1
 2
 3

julia> v_int .== 1
3-element SVector{3,Bool}:
  true
 false
 false
```

However, when `isequal` is applied element-wise to `SArray`s of non-numeric entries, we do not get `SArray`s of `Bool`, as shown in the following example with enumeration:
```julia
julia> @enum Axis X Y Z

julia> v_enum = @SVector([X,Y,Z])
3-element SVector{3,Axis}:
 X
 Y
 Z

julia> v_enum .== X
3-element BitArray{1}:
  true
 false
 false
```

Another example with a custom type:
```julia
julia> immutable MyType
           n::Int
       end

julia> m1 = MyType(1); m2 = MyType(2); m3 = MyType(3);

julia> v_mytype = @SVector([m1,m2,m3])
3-element SVector{3,MyType}:
 MyType(1)
 MyType(2)
 MyType(3)

julia> v_mytype .== m1
3-element BitArray{1}:
  true
 false
 false
```

Is this fixable?  If it is, where?  Probably in `src/broadcast.jl`?

c42f:
Generalize broadcast to follow the julia-0.6 rules for defining which
values are scalars.

I haven't bothered to implement `promote_containertype` for a combination of `Tuple` and `StaticArray` yet because it's a bit more work and it seems a little questionable that people would even use this combination rather than just using `StaticVector`.

Fixes #239

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -2,18 +2,38 @@
 ## broadcast! ##
 ################
 
-# TODO: bad codegen for `broadcast(-, SVector(1,2,3))`
+import Base.Broadcast:
+    _containertype, promote_containertype, broadcast_indices,
+    broadcast_c, broadcast_c!
 
-@propagate_inbounds function broadcast(f, a::Union{Number, StaticArray}...)
-    _broadcast(f, broadcast_sizes(a...), a...)
-end
+# Add StaticArray as a new output type in Base.Broadcast promotion machinery.
+# This isn't the precise output type, just a placeholder to return from
+# promote_containertype, which will control dispatch to our broadcast_c.
+_containertype(::Type{<:StaticArray}) = StaticArray
 
-@propagate_inbounds function broadcast{T}(f::Function, a::Type{T}, x::StaticArray)
-    _broadcast(f, (Size(), Size(x)), T, x)
+# With the above, the default promote_containertype gives reasonable defaults:
+#   StaticArray, StaticArray -> StaticArray
+#   Array, StaticArray       -> Array
+#
+# We could be more precise about the latter, but this isn't really possible
+# without using Array{N} rather than Array in Base's promote_containertype.
+#
+# Base also has broadcast with tuple + Array, but while implementing this would
+# be consistent with Base, it's not exactly clear it's a good idea when you can
+# just use an SVector instead?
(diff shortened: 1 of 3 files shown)


andyferris:
Cool!

Quick question - does `Base` have `@inline` for the methods we get routed through?

c42f:
Yes, we get routed through exactly one method - `broadcast` or `broadcast!`, and these are `@inline`.  From there we already end up in our implementation of `broadcast_c`.

Only potential caveat vs what we already have is that they don't have `propagate_inbounds`.  Generated code for normal `SVector` looks good however.

c42f:
Test failures were a minor clash of variable names with the test `@enum`. Fixed now.

src/broadcast.jl
```diff
@@ -2,18 +2,38 @@
...
+
+
+# Override for when output type is deduced to be a StaticArray.
+@inline function broadcast_c(f, ::Type{StaticArray}, as...)
+    _broadcast(f, broadcast_sizes(as...), as...)
 end
 
 @inline broadcast_sizes(a::StaticArray, as...) = (Size(a), broadcast_sizes(as...)...)
-@inline broadcast_sizes(a::Number, as...) = (Size(), broadcast_sizes(as...)...)
+@inline broadcast_sizes(a, as...) = (Size(), broadcast_sizes(as...)...)
```
andyferris: This is a nice side-effect :)

test/broadcast.jl
```diff
@@ -139,4 +143,13 @@ end
             @test eltype(a) == Float32
         end
     end
+
+    @testset "" begin
```
andyferris: testset name?

c42f: Oops, fixed.

Issue #194, opened 2017-05-28
Return `UpperTriangular` from `chol`


bshall:
To be consistent with the methods from Base I was wondering if the cholesky decomposition of a static array should return an `UpperTriangular` type? I know the `AbstractTriangular` types have `setindex!` and other mutating methods defined for them which may be an issue?

andyferris:
Yeah - consistency with `Base` is definitely a primary consideration of this package.

I did have an idea once to implement our own triangular types, with the idea of keeping only `(n^2 + n)/2` elements on the stack, as a micro-optimization. Not sure if that would ever by worthwhile. As it is, I'm not sure if the triangular type is simply slower than an `SMatrix` with extra zeros - probably solvable using `ifelse` inside `getindex`.

OTOH we can add more specializations to methods on `UpperTriangular{<:Any, <:StaticMatrix}` and `LowerTriangular{<:Any, <:StaticMatrix}` as necessary.

c42f:
Personally I wouldn't bother worrying about the extra elements.  There's a good chance that for really small matrices they're never even stored on the stack anyway, just passed in registers.

So wrapping in an `UpperTriangular` type makes a lot of sense to me.

Xref regarding cholfact - https://github.com/JuliaArrays/StaticArrays.jl/pull/192#issuecomment-304598119

andyferris:
> So wrapping in an `UpperTriangular` type makes a lot of sense to me.

OK - let's do this. 

One way to optimize around branching and `sub2ind` for immutable arrays is to fill in the zeros correctly upon construction, and mainly use the `UpperTriangular` tag for dispatch purposes (and allow fast, direct linear `getindex` for speed).

c42f:
> use the UpperTriangular tag for dispatch purposes and allow fast, direct linear getindex for speed

Sounds reasonable to me.

KristofferC:
FWIW we store only the lower triangular part and use non linear indexing (with a branch) at Tensors.jl (https://github.com/KristofferC/Tensors.jl/blob/master/src/indexing.jl#L5).  For the generated code, I have never seen a branch though, so the whole thing gets optimized away nicely.

c42f:
@KristofferC Right it makes sense that the branches disappear if the whole thing is unrolled.

Ok, a good reason to have special storage would be if people want to store a lot of triangular static arrays inside an array.  In that case it's a big win to only store the upper/lower part.

c42f:
[...]

PR #243, opened 2017-07-08
Improve cholesky factorization

bshall:
This PR extends the `chol` to higher dimensions, adds a `cholfact` method and a `\` based on the triangular solvers in #222. Also, `chol` will now return an `UpperTriangular` fixing #194.

The `Cholesky` struct in Base changed in v0.7 so I've targeted this later version with this PR. I'm not sure how to get this working on v0.6 and v0.7 simultaneously.

I've kept the old tests for the time being (just to show that everything is consistent) but these should be removed when everything is ready to be merged.

### DIFF
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -1,54 +1,118 @@
-# Generic Cholesky decomposition for fixed-size matrices, mostly unrolled
-@inline function Base.chol(A::StaticMatrix)
+import Base: chol, cholfact, \
+
+@inline _chol(A::StaticMatrix{<:Any,<:Any,T}, ::Type{UpperTriangular}) where {T} = _chol(Size(A), A, UpperTriangular)
+@inline _chol(A::StaticMatrix{<:Any,<:Any,T}, ::Type{LowerTriangular}) where {T} = _chol(Size(A), A, LowerTriangular)
+
+@generated function _chol(::Size{s}, A::StaticMatrix{<:Any,<:Any,T}, ::Type{UpperTriangular}) where {s, T}
+    if s[1] != s[2]
+        error("matrix must be square")
+    end
+    n = s[1]
+    TX = promote_type(typeof(chol(one(T), UpperTriangular)), Float32)
+
+    X = [Symbol("X_$(i)_$(j)") for i = 1:n, j = 1:n]
+    init = [:($(X[i,j]) = A[$(sub2ind(s,i,j))]) for i = 1:n, j = 1:n]
+
+    code = quote end
+    for k = 1:n
+        ex = :($(X[k,k]))
+        for i = 1:k-1
+            ex = :($ex - $(X[i,k])'*$(X[i,k]))
+        end
+        push!(code.args, quote $(X[k,k]), info = _chol($ex, UpperTriangular) end)
+        push!(code.args, :(info == 0 || return UpperTriangular(similar_type(A, $TX)(tuple($(X...)))), info))
+        if k < n
+            push!(code.args, :(XkkInv = inv($(X[k,k])')))
+        end
+        for j = k + 1:n
+            ex = :($(X[k,j]))
+            for i = 1:k-1
+                ex = :($ex - $(X[i,k])'*$(X[i,j]))
+            end
+            push!(code.args, :($(X[k,j]) = XkkInv*$ex))
+        end
+    end
+
+    quote
+        @_inline_meta
+        @inbounds $(Expr(:block, init...))
+        @inbounds $code
+        @inbounds return UpperTriangular(similar_type(A, $TX)(tuple($(X...)))), convert(Base.LinAlg.BlasInt, 0)
+    end
+end
+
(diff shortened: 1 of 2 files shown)


bshall:
I'm also not sure if the `cholfact` and `\` should have an `@inline`.

andyferris:
> I'm not sure how to get this working on v0.6 and v0.7 simultaneously.

Would using our own struct be suitable, in your opinion? (You can use `@static if VERSION < v"0.7"` to make this v0.6 only, if that is better).

> I'm also not sure if the `cholfact` and `\` should have an `@inline`.

I think that might be helpful. Benchmarking over a range of matrix sizes is the way to answer this question. (There's some pretty unclear code in the `perf` directory which can be extended if you want to play with this).

andyferris:
Awesome PR, btw. I haven't looked at the code in detail yet, but thanks :)

src/cholesky.jl
```diff
@@ -1,54 +1,118 @@
...
-    ishermitian(A) || Base.LinAlg.non_hermitian_error("chol")
-    _chol(Size(A), A)
-end
+import Base: chol, cholfact, \
 
-@inline function Base.chol(A::Base.LinAlg.RealHermSymComplexHerm{<:Real, <:StaticMatrix})
-    _chol(Size(A), A.data)
-end
+@inline _chol(A::StaticMatrix{<:Any,<:Any,T}, ::Type{UpperTriangular}) where {T} = _chol(Size(A), A, UpperTriangular)
+@inline _chol(A::StaticMatrix{<:Any,<:Any,T}, ::Type{LowerTriangular}) where {T} = _chol(Size(A), A, LowerTriangular)
```
fredrikekre: Seems like these two methods could simply be:
```julia
@inline _chol(A::StaticMatrix, ::Type{UL}) where {UL} = _chol(Size(A), A, UL)
```

src/cholesky.jl
```diff
@@ -1,54 +1,118 @@
...
 
-@generated function _chol(::Size{(1,1)}, A::StaticMatrix)
-    @assert size(A) == (1,1)
-    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
-    newtype = similar_type(A,T)
+@generated function _chol(::Size{s}, A::StaticMatrix{<:Any,<:Any,T}, ::Type{UpperTriangular}) where {s, T}
+    if s[1] != s[2]
+        error("matrix must be square")
```
fredrikekre: This check is not necessary; the `ishermitian` check does this before we enter this function.

src/cholesky.jl
```diff
@@ -1,54 +1,118 @@
...
+    end
+    n = s[1]
+    TX = promote_type(typeof(chol(one(T), UpperTriangular)), Float32)
```
fredrikekre:
```julia
TX = typeof(chol(one(T)))
```
ref https://github.com/JuliaLang/julia/pull/22245#discussion_r120967908

src/cholesky.jl
```diff
@@ -1,54 +1,118 @@
...
     end
 end
 
-@generated function _chol(::Size{(2,2)}, A::StaticMatrix)
-    @assert size(A) == (2,2)
-    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
-    newtype = similar_type(A,T)
+@generated function _chol(::Size{s}, A::StaticMatrix{<:Any, <:Any, T}, ::Type{LowerTriangular}) where {s, T}
+    if s[1] != s[2]
+        error("matrix must be square")
```
fredrikekre: Same comment as above.

src/cholesky.jl
```diff
@@ -1,54 +1,118 @@
...
+    end
+
+    n = s[1]
+    TX = promote_type(typeof(chol(one(T), LowerTriangular)), Float32)
```
fredrikekre: Same as above.

src/cholesky.jl
```diff
@@ -1,54 +1,118 @@
...
+        @inbounds return LowerTriangular(similar_type(A, $TX)(tuple($(X...)))), convert(Base.LinAlg.BlasInt, 0)
     end
 end
 
-@generated function _chol(::Size{(3,3)}, A::StaticMatrix)
-    @assert size(A) == (3,3)
-    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
-    newtype = similar_type(A,T)
+## Numbers
+_chol(x::Number, uplo) = Base.LinAlg._chol!(x, uplo)
```
fredrikekre: Accidentally left this in from `Base`?

bshall:
> Would using our own struct be suitable, in your opinion?
Yeah, I think this is probably the best way to go. I'll get the done and make the changes suggested by @fredrikekre. Thanks for the feedback by the way.

andyferris:
Where is this at?

bshall:
Sorry about the delay, I've been away for a bit. I'll hopefully finish this up by the end of the week.

c42f:
There's a few conflicts but it appears there's still a lot of potential in this PR if it was modernized.  We should have some benchmarks to decide when to stop unrolling.

PR #245, opened 2017-07-19
Add logdet

mschauer:
The infrastructure can easily be extended for ``logdetabs`` and others. 
Please check that I am doing the double interpolation ``$($f)`` correctly.

### DIFF
--- a/src/det.jl
+++ b/src/det.jl
@@ -1,4 +1,5 @@
 @inline det(A::StaticMatrix) = _det(Size(A), A)
+@inline logdet(A::StaticMatrix) = _logdet(Size(A), A)
 
 @inline _det(::Size{(1,1)}, A::StaticMatrix) = @inbounds return A[1]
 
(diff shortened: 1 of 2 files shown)


src/det.jl
```diff
@@ -24,18 +25,25 @@ end
...
+        @generated function $symb{S}(::Size{S}, A::StaticMatrix)
+            if S[1] != S[2]
+                throw(DimensionMismatch("matrix is not square"))
+            end
+            return quote # Implementation from Base
+                @_inline_meta
+                T = eltype(A)
+                T2 = typeof((one(T)*zero(T) + zero(T))/one(T))
+                if istriu(A) || istril(A)
+                    return convert(T2, $($f)(UpperTriangular(A))) # Is this a Julia bug that a convert is not type stable??
```
andyferris: `convert` is not treated differently to any other method, so in theory it doesn't have to return `T2`, and inference must infer that it does.

mschauer: The comment is old, i just wrapped it in the ``for - eval`` loop to extend it.

andyferris:
Cool, @mschauer - I can't look in detail right now, but thanks :)

PR #253, opened 2017-07-24
Add @boundscheck annotation to bounds check in convert method

tkoolen:
Benchmark code:
```julia
@benchmark convert(SMatrix{3, 3, Float64, 9}, x) setup = (x = rand(3, 3))
```
With `julia -O3 --check-bounds=no`, minimum benchmark time goes from  5.456 ns to 2.302 ns.


c42f:
Looks good to me.

tkoolen:
Thanks!

PR #254, opened 2017-07-26
immutable -> struct, type -> mutable struct

tkoolen:


src/FieldVector.jl
```diff
@@ -8,7 +8,7 @@ while a mutable `FieldVector` will behave similarly to an `MVector`.
 
 For example:
 
-    immutable/type Point3D <: FieldVector{3, Float64}
+    [mutable] struct Point3D <: FieldVector{3, Float64}
```
c42f: Perhaps we don't need the `[mutable]` here, as it's just an example?

mschauer: Yes, better add an example later on where mutable makes more sense.

tkoolen: How's this?

src/FieldVector.jl
```diff
@@ -8,11 +8,18 @@ while a mutable `FieldVector` will behave similarly to an `MVector`.
...
         y::Float64
         z::Float64
     end
+
+or
+
+    mutable struct Point2D{T} <: FieldVector{2, T}
+        x::T
+        y::T
+    end
```
c42f: I think, to some extent, this might just encourage users down the wrong path?  Personally I've never found a practical use for a mutable point like this.

tkoolen: Fine by me.

c42f:
Thanks a lot :-)

PR #255, opened 2017-07-26
Support Val{k} as second argument in diagm

wsshin:
This PR implements `diagm(::SVector, ::Type{Val{k}})`, which type-stably creates a matrix with a specified `k`th super-or sub-diagonal.  Previously there was only `diagm(::SVector)` without the second argument.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -172,11 +172,18 @@ end
     end
 end
 
-@inline diagm(v::StaticVector) = _diagm(Size(v), v)
-@generated function _diagm(::Size{S}, v::StaticVector) where {S}
-    Snew = (S[1], S[1])
+@inline diagm(v::StaticVector, k::Type{Val{D}}=Val{0}) where {D} = _diagm(Size(v), v, k)
+@generated function _diagm(::Size{S}, v::StaticVector, ::Type{Val{D}}) where {S,D}
+    S1 = S[1]
+    Snew1 = S1+abs(D)
+    Snew = (Snew1, Snew1)
+    Lnew = Snew1 * Snew1
     T = eltype(v)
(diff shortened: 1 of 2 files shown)


src/linalg.jl
```diff
@@ -172,11 +172,18 @@ end
...
-    Snew = (S[1], S[1])
+@inline diagm(v::StaticVector, k::Type{Val{D}}=Val{0}) where {D} = _diagm(Size(v), v, k)
+@generated function _diagm(::Size{S}, v::StaticVector, ::Type{Val{D}}) where {S,D}
+    S1 = S[1]
+    Snew1 = S1+abs(D)
+    Snew = (Snew1, Snew1)
+    Lnew = Snew1 * Snew1
     T = eltype(v)
-    exprs = [i == j ? :(v[$i]) : zero(T) for i = 1:S[1], j = 1:S[1]]
+    ind = diagind(Snew1, Snew1, D)
```
c42f: Nice.  I didn't know about `diagind`.

test/linalg.jl
```diff
@@ -40,6 +40,8 @@ using StaticArrays, Base.Test
 
     @testset "diagm()" begin
         @test @inferred(diagm(SVector(1,2))) === @SMatrix [1 0; 0 2]
+        @test @inferred(diagm(SVector(1,2,3), Val{2})) == diagm([1,2,3], 2)
```
c42f: A minor thing - might be worth adding a type assert here `::SMatrix` on the left hand side outside the `@inferred` to assert that these come out as static arrays.

wsshin: Good idea.  Just made changes accordingly.

Issue #256, opened 2017-07-27
kron(A,B) implementation?


aggrushin:
I was trying to use kron(A,B) but seems that this is not implemented.
Any plans on doing so?

Thank you,

aggrushin:
Thanks! Is it merged to the main branch?

PR #257, opened 2017-07-27
Implement kron

c42f:
Addresses #256


mschauer:
Should ``kron`` give a SMatrix? Already kron(4x4,4x4) has 256 elements, the rule of thumb being
>  A very rough rule of thumb is that you should consider using a normal Array for arrays larger than 100 elements

c42f:
We should certainly produce an `SMatrix` for small sizes, for example, if the user take `kron` of a 2x2 and a 3x3 `SMatrix` they should very much expect to get out an `SMatrix`.

We could consider falling back to a `SizedMatrix` for larger sizes and calling the `Base` version of `kron`, but we don't systematically do that anywhere else in `StaticArrays` - we just leave it up to the user to decide by getting them to convert to a `Base.Array`.

This is not completely different from `broadcast`, which can quickly produce arrays with a lot of elements.  But it's true that `broadcast` only does this sometimes, whereas `kron` is guaranteed to generate a tensor product worth of extra elements...

mschauer:
We have several instances where a function returns SizedArray instead of SArray for bigger arguments. At first I was irritated, but then I figured out that as long as the return type can be inferred from the arguments, this does not make many problems.

c42f:
> as long as the return type can be inferred from the arguments, this does not make many problems

That's true.  Ok, I can make this change to prevent things exploding in the user's face for large output sizes.  It's a bit of extra infrastructure to build, but sounds worthwhile.

c42f:
Well, deciding on a reasonable implementation here took a lot longer than I intended, but here's another go.  There's some extra infrastructure here now to allow us to do this kind of thing a bit more easily.

test/SizedArray.jl
```diff
@@ -68,6 +68,8 @@
         @test convert(Vector, SizedArray{Tuple{4}, Int, 1}(collect(3:6))) == collect(3:6)
         @test Matrix(SMatrix{2,2}((1,2,3,4))) == [1 3; 2 4]
         @test convert(Matrix, SMatrix{2,2}((1,2,3,4))) == [1 3; 2 4]
+        # Conversion after reshaping
+        @test_broken Array(SizedMatrix{2,2}([1,2,3,4])) == [1 3; 2 4]
```
c42f: @andyferris I discovered this while writing `drop_sdims` and it seems quite problematic - any thoughts on how to fix it?  AFAICT we'll have to change the way that the SizedArray reshaping support works completely, and probably remove it from the normal constructors.

Actually the test here doesn't show the worst of it.  The bigger problem is that `SizedMatrix{2,3}` could wrap an Array of size `(3,2)`.

andyferris: Hmm... yeah we should fix that...

src/StaticArrays.jl
```diff
@@ -28,6 +28,7 @@ export @MVector, @MMatrix, @MArray
 
 export similar_type
 export push, pop, shift, unshift, insert, deleteat, setindex
+export drop_sdims
```
andyferris: This seems an optional export

c42f: Definitely.  I thought it might be useful, but we could make it internal until we figure out whether we really want it to be part of the public API.

andyferris: Hmm... let's just do that for now.

src/util.jl
```diff
@@ -59,3 +59,34 @@ end
...
+"""
+    drop_sdims(a)
+
+Return an `AbstractArray` with the same elements as `a`, but with static
+dimensions removed (ie, not a `StaticArray`).
+
+This is useful if you want to override dispatch to call the `Base` version of
+operations such as `kron` instead of the implementation in `StaticArrays`.
+Normally you shouldn't need to do this, but it can be more efficient for
+certian algorithms where the number of elements of the output is a lot larger
```
mschauer: certain

c42f:
I think this is good to go, I've removed `drop_sdims` from the export list for now.  We can see how it pans out and consider whether we want it to be an exported public function later.

Issue #250, opened 2017-07-24
Bug in inverse of 4x4 matrix returning NaN


TsurHerman:
This came up when working with coordinate transformation matrices in openGL

MWE
```
julia> using StaticArrays

julia> A =@SMatrix [1 0 0 0;0 0 1 0;0 1 0 0;0 0 0 1]
4×4 StaticArrays.SArray{Tuple{4,4},Int64,2,16}:
 1  0  0  0
 0  0  1  0
 0  1  0  0
 0  0  0  1

julia> inv(A)
4×4 StaticArrays.SArray{Tuple{4,4},Float64,2,16}:
 NaN  NaN  NaN  NaN
 NaN  NaN  NaN  NaN
 NaN  NaN  NaN  NaN
 NaN  NaN  NaN  NaN
```

tkoolen:
Looks like the fact that the method doesn't work for the case when then upper left 2x2 block is singular (as it is in the @TsurHerman's example) is acknowledged in http://www.freevec.org/function/inverse_matrix_4x4_using_partitioning, referenced in the code:

> Well, in that particular case, we have to call a generic -but slower- function

TsurHerman:
The following code borrowed from @SimonDanisch FixedSizeArrays,and slightly modified to shave off another 15% benchmarks at 50ns on my laptop whereas the current implementation benchmarks at 30ns.
```

@inline det4x4(A) = @inbounds return (
    A[13] * A[10]  * A[7]  * A[4]  - A[9] * A[14] * A[7]  * A[4]   -
    A[13] * A[6]   * A[11] * A[4]  + A[5] * A[14] * A[11] * A[4]   +
    A[9]  * A[6]   * A[15] * A[4]  - A[5] * A[10] * A[15] * A[4]   -
    A[13] * A[10]  * A[3]  * A[8]  + A[9] * A[14] * A[3]  * A[8]   +
    A[13] * A[2]   * A[11] * A[8]  - A[1] * A[14] * A[11] * A[8]   -
    A[9]  * A[2]   * A[15] * A[8]  + A[1] * A[10] * A[15] * A[8]   +
    A[13] * A[6]   * A[3]  * A[12] - A[5] * A[14] * A[3]  * A[12]  -
    A[13] * A[2]   * A[7]  * A[12] + A[1] * A[14] * A[7]  * A[12]  +
[...]

PR #260, opened 2017-07-29
replace inv and det for 4x4 matrices,

TsurHerman:
Added inv and det for 4x4 matrix size , as a stopgap until a more compact representation is found.


TsurHerman:
Failure is in 32bit systems due to fact that implicit conversion to float is to Float32.
This affects accuracy of badly conditioned 4x4 matrices on 32 bit systems.
I think this test can be relaxed.

What do you recommend?

mschauer:
>  implicit conversion to float is to Float32
Can be more specific what is going on? There is no Float, only Float32 and Float64 (unlike Int, Int32, Int64) because also 32 bit systems have Float64.

TsurHerman:
inv. and det for 4x4 do not specify type explicitly but instead rely on the system automatic promotion
in expressions such as:
`A[13] * A[10]  * A[7]  * A[4]  - A[9] * A[14] * A[7]  * A[4] `

for x64 bit systems the test for ill conditioned matrix of size 4x4 passes
`norm(Matrix(inv(sm)*sm - eye(4))) < 10*norm(inv(m)*m - eye(4))
`
on x32 bit systems it fails, by not much by the way.
Best course of action would be to relax the test.

mschauer:
I think I am specifically curious which promotions go different on 32 systems.

TsurHerman:
I may have a 32 bit system behind my facsimile sitting on top of my laser-disc ... no its not there anymore. :-)

I agree it is interesting(it might have implications on embedded systems), but I can't help with that.

fredrikekre:
The test is with a `Float64` matrix even on 32bit, so should not be promotion that is the problem.

mschauer:
Yes, that is what I thought, but wasn't sure about. But it is not a catastrophic loss of accuracy anyway, replacing 10* by 12* would be a simple solution.

ChrisRackauckas:
Just chiming in since I was pointed here by @mschauer. Yes, AppVeyor 32-bit seems to apply different optimizations than 64-bit. We found this by very carefully comparing a whole bunch of different algorithms, and found some very small differences even when some macros were removed.

https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/20
https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/22

As @fredrikekre points out:

>The test is with a Float64 matrix even on 32bit, so should not be promotion that is the problem.

Both 32-bit and 64-bit use `Float64`, bit only refers to integer size. So it really is likely due to differences beyond your control, and not due to any type promotions.

test/inv.jl
```diff
@@ -66,8 +66,8 @@ end
          3.42945636322714076e-01 3.29329508494837497e-01 2.25635541033857107e-01 6.03912636987153917e-02
          4.77828437366344727e-01 4.86406974015710591e-01 1.95415684569693188e-01 6.74775080892497797e-02]
     sm = SMatrix{4,4}(m)
-    @test_broken norm(Matrix(inv(sm)*sm - eye(4))) < 10*norm(inv(m)*m - eye(4))
-    @test_broken norm(Matrix(sm*inv(sm) - eye(4))) < 10*norm(m*inv(m) - eye(4))
+    @test norm(Matrix(inv(sm)*sm - eye(4))) < 12*norm(inv(m)*m - eye(4))
+    @test norm(Matrix(sm*inv(sm) - eye(4))) < 12*norm(m*inv(m) - eye(4))
```
c42f: These tests were particularly chosen as examples where the previous implementation had trouble.  So they're probably not that relevant to the new implementation.

Presumably we'd be able to find some equivalent problems by rerunning the code at the bottom of this file.

mschauer: This is still a difficult case (as evidenced by the bounds being quite tight). But running the problem generator would be a good idea.

test/inv.jl
```diff
@@ -55,8 +55,8 @@ end
         1.63327582123091175e-01 1.70552365412100865e-01 4.55720450934200327e-01 1.23299968650232419e-01
     ]
     sm = SMatrix{4,4}(m)
-    @test norm(Matrix(sm*inv(sm) - eye(4))) < 10*norm(m*inv(m) - eye(4))
-    @test norm(Matrix(inv(sm)*sm - eye(4))) < 10*norm(inv(m)*m - eye(4))
+    @test norm(Matrix(sm*inv(sm) - eye(4))) < 12*norm(m*inv(m) - eye(4))
+    @test norm(Matrix(inv(sm)*sm - eye(4))) < 12*norm(inv(m)*m - eye(4))
```
c42f: This change is ok I think. There's nothing special about the factor of `10` which you've replaced with 12 here.  I'd say some difference is to be expected with a switch in algorithms.

src/inv.jl
```diff
@@ -30,32 +30,29 @@ end
...
-    R2 = -S2*(R*invP)
-    Q2 = -invP_Q*S2
-    P2 = invP - invP_Q*R2
-    [[P2 Q2];
-     [R2 S2]]
+        (A[1,4]*A[2,3]*A[3,2] - A[1,3]*A[2,4]*A[3,2] - A[1,4]*A[2,2]*A[3,3] + A[1,2]*A[2,4]*A[3,3] + A[1,3]*A[2,2]*A[3,4] - A[1,2]*A[2,3]*A[3,4]) * idet
+        (A[1,3]*A[2,4]*A[3,1] - A[1,4]*A[2,3]*A[3,1] + A[1,4]*A[2,1]*A[3,3] - A[1,1]*A[2,4]*A[3,3] - A[1,3]*A[2,1]*A[3,4] + A[1,1]*A[2,3]*A[3,4]) * idet
+        (A[1,4]*A[2,2]*A[3,1] - A[1,2]*A[2,4]*A[3,1] - A[1,4]*A[2,1]*A[3,2] + A[1,1]*A[2,4]*A[3,2] + A[1,2]*A[2,1]*A[3,4] - A[1,1]*A[2,2]*A[3,4]) * idet
+        (A[1,2]*A[2,3]*A[3,1] - A[1,3]*A[2,2]*A[3,1] + A[1,3]*A[2,1]*A[3,2] - A[1,1]*A[2,3]*A[3,2] - A[1,2]*A[2,1]*A[3,3] + A[1,1]*A[2,2]*A[3,3]) * idet]
+        return SMatrix{4,4,eltype(B),16}(B)
```
c42f: The return type here should arguably be deduced using `similar_type` rather than always returning `SMatrix`.  See, for example, the implementations of `_inv` for the smaller sizes.

The idea here is that users can preserve the container type of their static matrix if they so please.  I'm not sure this is used a lot in practice, but best for consistency to keep the same pattern.

TsurHerman: Agreed , waiting for my pull request regarding det.jl to be merged into master, since there are additional small fixes,and then I will merge back master here and change it to be consistent (adding the similar_type instead of SMatrix)

src/det.jl
```diff
@@ -21,6 +21,23 @@ end
     return vecdot(x0, cross(x1, x2))
 end
 
+@inline function _det(::Size{(4,4)}, A::StaticMatrix,S::Type)
+    A = similar_type(A,S)(A)
```
andyferris: There used to be compiler problems with variable names that change types. Has this been fixed? (I.e. is code optimal here when `eltype(A) != S`?)

TsurHerman: yes there is a difference , it is not big (about 2 ns difference) , but the code produced if 
we avoid variable name changing type looks better .it looks like what you would expect  hand optimized assembly code would look like, a long list of vector intrinsics.

TsurHerman: Any preferred style?

I though of changing each function that this line appears as the first line in this way:

```
@inline function _det(::Size{(3,3)}, A::StaticMatrix, S::Type)
    A = similar_type(A,S)(A)
```

would become:

```
@inline function _det(::Size{(3,3)}, _A::StaticMatrix, S::Type)
    A = similar_type(_A,S)(_A)
```

TsurHerman: Or move the promoting to the dispatching function all together and dropping the propagation of the promoted type.
Better in my opinion

Both will work , what style do you prefer?

andyferris: By now, I am used to the "SSA" form so I do tend to write `a2 = f(a); a3 = g(a2)` etc. In this case something like `A_S` would look clear. 

Dispatch might be a very clean solution - the line won't even appear when `S` is already correct, which may be important for e.g. heap allocated static arrays or if `similar_type` always changes types or if the compiler doesn't optimiza away the construction for some reason. 

So let's try dispatch!

src/inv.jl
TsurHerman: Updated everywhere to return similar_type and not SMatrix

TsurHerman:
@c42f your requested changes are in place, `similar_type` instead of `Smatrix`
besides that moved the type promotion to the dispatching function. 
So if you first merge this PR and then yours about logdet .. then I think this covers the everything.

det inv and logdet give correct result in a speedy manner for matrices of small size up to size 4 with any type that Base supports including integer and unsigned , and calls Base on large matrices

src/inv.jl
```diff
@@ -1,24 +1,24 @@
-@inline inv(A::StaticMatrix) = _inv(Size(A), A)
-
-@inline _inv(::Size{(1,1)}, A) = similar_type(typeof(A), typeof(inv(one(eltype(A)))))(inv(A[1]))
-
-@inline function _inv(::Size{(2,2)}, A)
+@inline function inv(A::StaticMatrix)
     T = eltype(A)
     S = typeof((one(T)*zero(T) + zero(T))/one(T))
-    newtype = similar_type(A, S)
+    A_S = similar_type(A,S)(A)
```
andyferris: `A_S = convert(similar_type(A, S), A)`

src/det.jl
```diff
@@ -1,26 +1,41 @@
 @inline function det(A::StaticMatrix)
     T = eltype(A)
     S = typeof((one(T)*zero(T) + zero(T))/one(T))
-    _det(Size(A),A,S)
+    A_S = similar_type(A,S)(A)
```
andyferris: `A_S = convert(similar_type(A, S), A)`

andyferris: Can potentially remove a new allocation of e.g. a `MMatrix`.

TsurHerman: I dont see that changes anything , but I am updating anyway

TsurHerman: And the biggest problem is that inv for `MMatrix` returns `SMatrix`

TsurHerman:
Hi guys, any chance that this is getting in soon?

c42f:
Ah, this one slipped off my radar.  Looks good now thanks.

@mschauer I did rerun the problem generator, and it produced the following graphs which compare the two different methods of computing the 4x4 inverse with the Base implementation:

Partitioning method (previous)
inv_partition

Adjugate method (changes here)
inv_adjugate

So clearly the new method behaves more robustly for well conditioned matrices (the dark blue blob is tight). If nothing else, it's clear we need to do this well, so I think this is a reasonable tradeoff for now.

The new method does have some has some systematic problems for approximately low ranked matrices, where if fares a lot worse than the previous method.  We'll just have to put up with that for now.

c42f:
Oh, we should really add the test case which kicked all this off - the 4x4 permutation matrix from #250.  I'll do this now.

c42f:
Ok, I added that test case. Let's merge this as soon as CI has passed.

PR #262, opened 2017-07-29
More  consistent treatment of unsigned types in det and inv

TsurHerman:
for small matrices size 2x2 3x3 4x4 and types UIn8 and UInt16 UInt32
with high enough values , the result in StaticArrays differ from that of Base in the sense
that it throws an Inexact error.

with this PR this codes follow through:

```
A = @SMatrix rand(UInt16,3,3)
isapprox(det(A), det(Array(A)))
isapprox(A * inv(A),eye(size(A)...))
```

whereas on master it usually fails on InExact error.


src/det.jl
```diff
@@ -7,25 +7,26 @@
...
+@inline function _det(::Size{(2,2)}, A::StaticMatrix{2, 2, <:Unsigned})
+    @inbounds return det(SMatrix{2,2,Int64}(A))
+end
+
+@inline function _det(::Size{(3,3)}, A::StaticMatrix{3, 3, <:Unsigned})
+    @inbounds return det(SMatrix{3,3,Int64}(A))
+end
+
+@inline function _det(::Size{(4,4)}, A::StaticMatrix{4, 4, <:Unsigned})
+    @inbounds return det(SMatrix{4,4,Int64}(A))
```
c42f: I think you can get away with a generic solution for the `Unsigned` issue here instead of these several versions:

```julia
@inline det(A::StaticMatrix{<:Any,<:Any,<:Unsigned}) = det(map(signed, A))
```

Or possibly the equivalent thing with `_det` if it makes more sense.

Depending on which way we want to go, we can use either `map(Signed, A)` or `map(signed, A)` here, for checked and unchecked conversion, respectively.

TsurHerman: `det(map(signed, A))`
is not sufficient because it might return non-sensical results since it just reinterprets 
A as signed.

I added a more general and consistent version please review.

src/det.jl
```diff
@@ -7,25 +7,22 @@
...
 end
 
-@inline function _det(::Size{(3,3)}, A::StaticMatrix{<:Any, <:Any, <:Unsigned})
-    @inbounds x0 = SVector(Signed(A[1]), Signed(A[2]), Signed(A[3]))
-    @inbounds x1 = SVector(Signed(A[4]), Signed(A[5]), Signed(A[6]))
-    @inbounds x2 = SVector(Signed(A[7]), Signed(A[8]), Signed(A[9]))
-    return vecdot(x0, cross(x1, x2))
+@generated function det(A::StaticMatrix{<:Any,<:Any,T}) where T <:Unsigned
+    T1 = promote_type(T,typeof(signed(one(T))))
+    (T1 <: Unsigned) && (T1 = Float64) #if still unsigned cast to Float64
```
andyferris: please use `if`.

Does base call `float` or something in this case?

TsurHerman: Base calls `T2 = typeof((one(T)*zero(T) + zero(T))/one(T))` to determine the type to convert to.
Which translates to Float64 across to board.

Which led to test following test showing that actually **_det_** is broken in StaticArrays for probably any integer type for which the mathematical operations overflows

try this code:
```
A = @SMatrix rand(Int8,2,2) 
det(A)
det(Array(A))
```

andyferris: Yeah... not surprised about overflows as we've tended to optimize for speed rather than correctness in this package. Not sure if this is a good choice or not?

TsurHerman: I added a simplification of det.jl covers all cases, conforms to Base and without loss of performance or accuracy(in Int8 Int16 Int32)

TsurHerman:
Added a simplification of det.jl which covers all cases, conforms to Base and without loss of performance or accuracy(in Int8 Int16 Int32)

Basically moved the promoted type calculation to the dispatching function. The compiler 
effectively eliminates that , so no generated function is needed which leads o a more "symmetric" code.

For Float64 Float32 and Float16 , this results in a no-op since no converting is performed.

For Integer and unsigned Integer types , performance will be slightly less , because extra converting is being performed, but the upside is that there is no overflow. 
In the current state of StaticArrays det can return completely wrong result for integer and unsigned integer types.

For matrices of sizes greater than 3 , for which there currently no specific solution in StaticArrays,
It reverts to converting to Array and calling Base.det. Which is identical to the current state in `master`

src/det.jl
```diff
@@ -1,49 +1,26 @@
...
     return vecdot(x0, cross(x1, x2))
 end
 
-@inline function _det(::Size{(3,3)}, A::StaticMatrix{<:Any, <:Any, <:Unsigned})
-    @inbounds x0 = SVector(Signed(A[1]), Signed(A[2]), Signed(A[3]))
-    @inbounds x1 = SVector(Signed(A[4]), Signed(A[5]), Signed(A[6]))
-    @inbounds x2 = SVector(Signed(A[7]), Signed(A[8]), Signed(A[9]))
-    return vecdot(x0, cross(x1, x2))
+@inline function _det(::Size, A::StaticMatrix,::Type)
+    return det(Matrix(A))
```
c42f: Here you should be able to use a recent function I added - `drop_sdims(A)`, rather than allocating a `Matrix` with `Matrix(A)`.  Presumably this will be significantly faster.

src/det.jl
```diff
@@ -1,49 +1,26 @@
...
+    _det(Size(A),A,S)
+end
 
-@inline _det(::Size{(1,1)}, A::StaticMatrix) = @inbounds return A[1]
+@inline logdet(A::StaticMatrix) = log(det(A))
 
-@inline function _det(::Size{(2,2)}, A::StaticMatrix)
-    @inbounds return A[1]*A[4] - A[3]*A[2]
-end
+@inline _det(::Size{(1,1)}, A::StaticMatrix,::Type) = @inbounds return A[1]
```
c42f: Perhaps this should this be `convert(S, A[1])` for consistency?

src/det.jl
```diff
@@ -1,49 +1,26 @@
...
     @inbounds x0 = SVector(A[1], A[2], A[3])
     @inbounds x1 = SVector(A[4], A[5], A[6])
     @inbounds x2 = SVector(A[7], A[8], A[9])
     return vecdot(x0, cross(x1, x2))
 end
 
-@inline function _det(::Size{(3,3)}, A::StaticMatrix{<:Any, <:Any, <:Unsigned})
-    @inbounds x0 = SVector(Signed(A[1]), Signed(A[2]), Signed(A[3]))
-    @inbounds x1 = SVector(Signed(A[4]), Signed(A[5]), Signed(A[6]))
-    @inbounds x2 = SVector(Signed(A[7]), Signed(A[8]), Signed(A[9]))
```
c42f: It's great to see this stuff go away :-)

c42f:
That's true, I did wonder whether `logdet()` might be better done as `log(det(A))` for sizes where we specialize `det`, and `logdet(drop_sdims(A))` for larger sizes.

mschauer:
Using e.g. the LU factorization of ``[a b; c d]`` gives
```
log(a) + log(d - (b c)/a)
```
Can this be more stable than ``log(ad - cd)``? I don't think so
```julia
julia> x = [1000 10; 10 0.1+eps(0.1)] 
2×2 Array{Float64,2}:
 1000.0  10.0
   10.0   0.1

julia> logdet2(x...) - logdet2(big.(x)...)
-3.364722366212109806115106416143722318025029732753681638886454186235485918784308e-01

julia> logdet1(x...) - logdet1(big.(x)...)
-3.127557100038973759369466656096524976106084420253681638886454297360758737735636e-01
```

src/det.jl
TsurHerman: drop_sdims increases allocation and is slightly less performant .. I don't know what to do..

TsurHerman:
@c42f I added the convert method to array of size (1,1) as it is more consistent to Base this way, 
drop_sdims is not more performant on my machine so I chose to leave it this way.

Even if it is not perfect this PR stops a few holes in the previous implementation , and together with 
my other PR it at-least delivers what StaticArrays is promising, fast and correct operations on small static matrices and conformance to Base on anything else.

Lets move fwd and merge this PR, someone can later issue another PR which improve on that.

src/det.jl
c42f: Ah, that's unfortunate. Good to know though, I'll have to chase this up later.

PR #263, opened 2017-07-31
Add more functions implemented with [map]reduce and [map]reducedim

wsshin:
This PR adds more functions implemented with `[map]reduce` and `[map]reducedim` to `src/mapreduce.jl`:
-  `iszero`
- directional versions of `sum`, `prod`, `all`, `any`
- mapped version of `prod`, `count`, `all`, `any`, `minimum`, `maximum`
Note that `Base` already has these functions.

Additionally, this PR corrects the following typos:
- `reducedim(op, a::StaticArray, ::Val{D})` to `reducedim(op, a::StaticArray, ::Type{Val{D}})`
- `reducedim(op, a::StaticArray, ::Val{D}, v0)` to `reducedim(op, a::StaticArray, ::Type{Val{D}}, v0)`

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -160,27 +160,76 @@ end
 ## reducedim ##
 ###############
 
-@inline reducedim(op, a::StaticArray, ::Val{D}) where {D} = mapreducedim(identity, op, a, Val{D})
-@inline reducedim(op, a::StaticArray, ::Val{D}, v0) where {D} = mapreducedim(identity, op, a, Val{D}, v0)
+@inline reducedim(op, a::StaticArray, ::Type{Val{D}}) where {D} = mapreducedim(identity, op, a, Val{D})
+@inline reducedim(op, a::StaticArray, ::Type{Val{D}}, v0) where {D} = mapreducedim(identity, op, a, Val{D}, v0)
 
 #######################
 ## related functions ##
 #######################
 
 # These are all similar in Base but not @inline'd
-@inline iszero(a::StaticArray{<:Any, T}) where {T} = reduce((x,y) -> x && (y==zero(T)), true, a)
-@inline sum(a::StaticArray{<:Any, T}) where {T} = reduce(+, zero(T), a)
-@inline sum(f::Base.Callable, a::StaticArray) = mapreduce(f, +, a)
-@inline prod(a::StaticArray{<:Any, T}) where {T} = reduce(*, one(T), a)
-@inline count(a::StaticArray{<:Any, Bool}) = reduce(+, 0, a)
-@inline count(f::Base.Callable, a::StaticArray) = mapreduce(x->f(x)::Bool, +, 0, a)
-@inline all(a::StaticArray{<:Any, Bool}) = reduce(&, true, a)  # non-branching versions
-@inline any(a::StaticArray{<:Any, Bool}) = reduce(|, false, a) # (benchmarking needed)
+#
+# Implementation notes:
+#
+# 1. When providing an initial value v0, note that its location is different in reduce and
+# reducedim: v0 comes earlier than collection in reduce, whereas it is the last argument in
+# reducedim.  The same difference exists between mapreduce and mapreducedim.
+#
+# 2. mapreduce and mapreducedim usually do not take initial value v0, because we don't
+# always know the return type of an arbitrary mapping function f.  (We usually want to use
+# some initial value such as one(T) or zero(T) as v0, where T is the return type of f, but
+# if users provide type-unstable f, its return type cannot be known.)  Therefore, mapped
+# versions of the functions implemented below usually require the collection to have at
+# least two entries.
+#
+# 3. Exceptions are the ones that require Boolean mapping functions.  For example, f in
+# all and any must return Bool, so we know the appropriate v0 is true and false,
+# respectively.  Therefore, all(f, ...) and any(f, ...) are implemented by mapreduce(f, ...)
+# with an initial value v0 = true and false.
+#
+# 4. Some implementations (e.g., count(a, Val{D})) are commented out because corresponding
+# Base functions (e.g., count(a, D)) do not exist yet.
+@inline iszero(a::StaticArray{<:Any,T}) where {T} = reduce((x,y) -> x && (y==zero(T)), true, a)
+
+@inline sum(a::StaticArray{<:Any,T}) where {T} = reduce(+, zero(T), a)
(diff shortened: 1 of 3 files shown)


tkoolen:
Just browsing, but FYI for the `Val` change: the Base convention that methods should take `::Type{Val{T}}` is going to be changed to `::Val{T}` in 1.0 (see [latest docs](https://docs.julialang.org/en/latest/manual/types/#"Value-types"-1) vs. [0.6 docs](https://docs.julialang.org/en/release-0.6/manual/types/#"Value-types"-1) and https://github.com/JuliaLang/julia/pull/22475.

wsshin:
Didn't know about the discussion.  Good to know.  Thanks!

src/mapreduce.jl
```diff
@@ -160,25 +160,76 @@ end
 ## reducedim ##
 ###############
 
-@inline reducedim(op, a::StaticArray, ::Val{D}) where {D} = mapreducedim(identity, op, a, Val{D})
-@inline reducedim(op, a::StaticArray, ::Val{D}, v0) where {D} = mapreducedim(identity, op, a, Val{D}, v0)
+@inline reducedim(op, a::StaticArray, ::Type{Val{D}}) where {D} = mapreducedim(identity, op, a, Val{D})
```
andyferris: FYI: This possibly wasn't a typo. I had been thinking about [this PR](https://github.com/JuliaLang/julia/pull/22475) for a long time :)

However, we should probably add a `if VERSION < v"0.7"` or something (or simply support both?)

src/mapreduce.jl
```diff
@@ -160,25 +160,76 @@ end
...
+# versions of the functions implemented below usually require the collection to have at
+# least two entries.
+#
+# 3. Exceptions are the ones that require Boolean mapping functions.  For example, f in
+# all and any must return Bool, so we know the appropriate v0 is true and false,
+# respectively.  Therefore, all(f, ...) and any(f, ...) are implemented by mapreduce(f, ...)
+# with an initial value v0 = true and false.
+#
+# 4. Some implementations (e.g., count(a, Val{D})) are commented out because corresponding
+# Base functions (e.g., count(a, D)) do not exist yet.
```
andyferris: Yet? Meaning they are mooted for v0.7?

(We are allowed to be a bit more agile and move a bit beyond `Base` wherever it makes sense, so long as we don't break anything that tends to work for `AbstractArray`)

andyferris:
Basically looks good to me. Thanks @wsshin.

I guess we may as well try and support the syntax for v0.6 and v1.0? (Ideally, I'd like to support multiple versions of Julia - in the past we've been hampered by internal compiler changes meaning different approaches are valid/optimal, but so far I am not really expecting this for v0.7/v1.0).

src/mapreduce.jl
wsshin: But having to use `Val(D)` in `reducedim` and `Val{D}` in `mapreducedim` seems confusing and inconsistent.  I don't see the use of `Val(D)` and `::Val{D}` anywhere else in StaticArrays, either.  Wouldn't it be better to handle this in a separate PR collectively?

andyferris: OK, so what I would like is if *StaticArrays* used `Val{D}` everywhere in v0.6 and `Val(D)` everywhere in v0.7. (I suppose it's OK if the v0.6 version also supports `Val{D}()`, partly to avoid this being a breaking change)

If you were to implement that here, that would be awesome (as seems pertinent to the change already done on this line) - however, let me know if you'd prefer to not do that at the moment, and I'll get to it later.

wsshin: When making this change, I would like to have things like 
```julia
default_similar_type{T,S,D}(::Type{T}, s::Size{S}, ::Type{Val{D}})
```
in `abstractarray.jl` change to
```julia
default_similar_type{T,S,D}(::Type{T}, s::Size{S}, ::Val{D})
```
as well.  Because such a change is outside `mapreduce.jl`, I think creating a separate PR dedicated to `::Type{Val{D}}` -> `::Val{D}` would be cleaner...  So yes, I would prefer not to make this change at the moment.

andyferris: Fair enough

wsshin:
Wait.  While trying to uncomment `count(a, Val{D})`, I found some bugs.

andyferris:
OK

wsshin:
The bug I was referring to was ignorance of the possibility of element type change in `mapreducedim`.  Because of this, `count(f::Function, a::StaticArray, ::Type{Val{D}})` was returning an array of `Float64`, rather than `Int64`, when `eltype(a)` was `Float64`.  Now properly promoted element types are used in `mapreducedim` and its friends.

One question, though.  I used `Base.promote_op` to deduce the proper element types after mapping and operation, but in `_map` I find that `Core.Inference.return_type` is used.  I find that `Base.promote_op` wraps `Core.Inference.return_type` (https://github.com/JuliaLang/julia/blob/master/base/promotion.jl#L321), so I guessed that `promote_op` takes care of some difficult situations better than `return_type`, but I'm not sure which one is better.  Any suggestions?

src/mapreduce.jl
wsshin: 5172c39 uncomments these functions!

andyferris:
Feel free to use either. I think there are historic reasons surrounding this... something about one being inferable and the other not (but these days they should both be fine).

BTW, I'd be *most* happy with the idea of just letting the natural types fall out, rather than using inference. At the moment we can't do this everywhere since we'd need a "eltypeless" constructor for every `StaticArray`, but unlike `Base.Array`, we can construct the result before wrapping it in an array so *technically* we don't need eltype prediction (in such a system, inference will still work to make everything fast, but only in the background).

wsshin:
I think I know what you mean by "letting the natural types fall out", but just in case, could you point me to a couple of examples of this practice in StaticArrays, if any?

andyferris:
Constructors are one case, things get promoted to a common type and an output is generated.

I think complete reductions also might do this?

wsshin:
Hmmm...  Not sure if I understood what you meant.  Could you be more specific about how to apply your idea in `_mapreducedim`?

andyferris:
I don't think you can implement this idea (at this stage) for partial reductions because the infrastructure isn't available. I was just expressing an idea of where I'd like to get to.

wsshin:
I see.  Then can we merge this now?

andyferris:
Done!

PR #264, opened 2017-08-01
Remove pre from minimum version

ChrisRackauckas:
CI no longer tests the pre-releases, so it should require v0.6 instead.


c42f:
Looks good, thanks.

Issue #267, opened 2017-08-04
Calling log(det()) instead of logdet()


mschauer:
#262 introduces the regression that StaticArrays' ``logdet`` just calling ``log`` after ``det`` instead of the specialized ``logdet`` in Base.

PR #268, opened 2017-08-05
Sizeclass logdet fix

c42f:
I quickly bashed this out as a potential fix for #267 

@mschauer - any thoughts on where the cutoff should be?
@andyferris - any thoughts on the naming and structure of the new trait here?  The idea here is to break different sizes into categories for dispatch (so `SizeCategory` would also be a potential name).


TsurHerman:
In the case of `logdet` your implementation is not an improvement over log(det(A)).

With your small large trait, for matrices that are defined small (1,2,3) it results in fast computation of `det` for matrices which are defined small .

In the naive implementation of log(det(A)) since det(A) for small matrices defaults to fast computation
the resulting code is just as fast. Moreover once you merge my PR for 4x4 fast `det` , you will have to change small matrices to include size 4x4 . 

In some other algorithms small would mean something else (maybe non rectangular matrices) 

Overall I strongly vote down on this PR, as it is a waste of creativity .. solving a non-issuse.

I think the strategy outlined in det and inv, where the general function dispatches a local size traited
function , with potentially other parameters , like the upcasting type needed to hold the result 
in a conformant way to Base. Is the way to go, and of course the last function for Sizes that are not covered just calls the Base implementation on converted to Matrix or dropped_sdims version(Which I am not sure is better)

Or am I missing something?

mschauer:
I would call ``log(_det`` whenever there is a specialized version ``_det`` for fixed sized arrays. I like the infrastructure actually, this might come in handy later.

TsurHerman:
But det already calls a specialized version, if one exists. The infrastructure is already there.

Once `lufact` is implemented in fast way for small matrices, then specialized version for det and inv 
will become redundant , and their definition can be identical to that of Base.

The rule of thumb ,as I see it, is to overload only the bottom most functions ,  and let the higher abstractions deal with the rest.

mschauer:
Somehow I think you are missing something. Say ``StaticArrays.logdet(a)`` is called where ``a`` is a static small matrix. The point is not about doing ``Base.log(StaticArrays.det(a))`` when ``a`` is small. It is about doing ``Base.logdet(Matrix(a))`` when ``a`` is not so small (but still static), which is different from ``Base.log(Base.det(Matrix(a)))``

TsurHerman:
Arguably different. If you follow code path for log(det(A)) and logdet(A) you will find that they are almost if not completely identical. 

logdet passes through absdet which is det + sign .. anyway in my opinion for this specific function , there is no proof that there is anything to gain .
Over simply log(det)) especially for sizes that are not trivial.

TsurHerman:
But I understand now that this is a Base issue and not StaticArrays issue that logdet and log(det()) are identical. 
So I remove my passionate objection :-) 

As I said a lot of this things would be redundant if lufact was implemented for StaticArrays

mschauer:
> logdet passes through absdet which is det + sign ... 

logdet does not pass through "absdet" but through logabsdet

```
logdet(a) = logabsdet(a)+log(s)
log(det(a)) = log(exp(logabsdet(a))*s)
```

Edit:
I agree that lufact would be really nice.

c42f:
@TsurHerman Yes, you're missing something: In addition to what has already been pointed out, it's a rather common pattern in this package to want to call specialized algorithms for small sizes, and other algorithms, possibly from `Base`, for larger sizes. From this point of view, the small piece of trait infrastructure I've added here should look less like tasteless overengineering, and more like forward planning.

As a side note, I've never wanted `logdet` myself, so I've personally got no skin in this game, I'm simply maintaining the package and responding to @mschauer's quite reasonable concern that the `logdet` implementation had changed. The correct way to avoid a feature you don't like getting into this package is to (1) read and understand what the patch is trying to do (2) clearly and *politely* explain why an alternative might be better. I don't claim to be an expert in the wide field of numerical linear algebra, and I'm happy to learn from people who are.

TsurHerman:
@c42f Sorry if I pass as impolite , the fact the you or others clearly ignore things I am pointing out is perceived by me as arrogance and failure to accept "others". Although I don't feel you do that on purpose.

I take great care when contributing , especially to something that is so meticoulesly maintained as StaticArrays , and so my change to logdet was delibirate and done only after I validated that it is only a cosmetic change , the underlying code is the same. I put great deal of thought and followed several code path to come to that conclusion.
The fact that you all ignore that is perceived as uninviting for further contribution.

Again the engineering layed out in inv and det already accomplished dispatching for small matrices and complying to Base on large matrices. Whoever wrote that in the first place did a good job.
I changed it a bit by moving the up casting type calculation to the dispatching function, after I carefully examined the implementation in Base and after carefully measuring performance and looking at compiled code.
This again is mainly cosmetic , instead of each sub-algorithm calculate the resulting type it gets it as a parameter.

I find the small big trait to be over engeineering at this moment , my experience comes from delivering algorithmic solutions on tight schedules, so yes over engineering is something that pushes my buttons.

Again thanks for maintaining StaticArrays , it is very useful for me (doing lots of 4x4 operations in OpenGL) and I apologize if I pass as impolite , forgive me I am from a different culture.

andyferris:
Thank you @c42f for thinking of nicer ways of organizing size-based dispatch here. @TsurHerman, this package is at peril of becoming difficult to maintain because of a maze of size-specialized behaviors, so engineering some solution here at some stage will probably be necessary. I wouldn't want all our code to look like [what's currently in matrix multiplication](https://github.com/JuliaArrays/StaticArrays.jl/blob/master/src/matrix_multiply.jl#L127-L154), for example (however - in many cases it *should* have such nuanced behavior, we basically won't do well for larger static arrays in many cases at the moment). Also, Julia may one day have a trait-based dispatch system, and it's an interesting exercise to explore how this might be useful.

I guess that there is one "lighter weight" solution, which would be use `Union`s of `Size`s instead of traits, for example to have `UpTo3x3 = Union{Size{(1,1)}, Size{(2,2)}, Size{(3,3)}}` or something.  Even just for the naming, describing the range of sizes at the method definition site could make it a bit clearer to navigate code. Just a thought.

c42f:
@TsurHerman Thanks for the thoughtful reply. I think the problem is simply that it's easy to miscommunicate when working only with text. I certainly didn't mean to ignore your comments, indeed, I thought they'd all been addressed already. Generally I just assumed that there's a good reason that `Base` has `logdet`.

Let's delve into the code in `Base.LinAlg`.
First, you get to
https://github.com/JuliaLang/julia/blob/ae1719872092b594228a9bfeedfc3fa8e91c9529/base/linalg/generic.jl#L1261
which should go here
https://github.com/JuliaLang/julia/blob/ae1719872092b594228a9bfeedfc3fa8e91c9529/base/linalg/generic.jl#L1239
then finally we get to the details of the implementation here
https://github.com/JuliaLang/julia/blob/ae1719872092b594228a9bfeedfc3fa8e91c9529/base/linalg/lu.jl#L296
which (without delving further) looks like it's summing the logs of the diagonal.

Let's contrast this to the code for `det`, which just multiplies the elements of the diagonal:
https://github.com/JuliaLang/julia/blob/ae1719872092b594228a9bfeedfc3fa8e91c9529/base/linalg/lu.jl#L281

Ok, this is interesting - there's no floating point numerical issues in a multiplication chain, other than overflow or underflow:

```julia
julia> log(det(10000*eye(100)))
Inf

julia> logdet(10000*eye(100))
921.034037197617
```

So as far as I can tell, the only reason to prefer `logdet(A)` over `log(det(A))` in the `Base` implementation is for largeish matrices with large or small elements, where overflow/underflow may be an issue.  Other than that, it's actually less efficient, because `log` is called many times.

So having looked at this in more detail, I have to agree that `log(det(A))` is quite a reasonable implementation for `StaticArrays`.  Perhaps you covered this elsewhere, but I missed seeing it in the larger conversation.

c42f:
The `SizeClass` stuff might still be useful, though @andyferris has an interesting alternative.  What I was trying to achieve with this was to replace the current arbitrary cutoffs implemented in various places (such as the matmul code pointed out by @andyferris), and the mechanism I used to implement `kron` with something a bit more uniform.

It would be even better if we could somehow automatically tune the cutoffs, or let them be configured in some simple way per architecture.  I don't yet know how to do this in a nice way.

mschauer:
> So having looked at this in more detail, I have to agree that log(det(A)) is quite a reasonable implementation for StaticArrays. Perhaps you covered this elsewhere, but I missed seeing it in the larger conversation.

But then you would have to hook directly into https://github.com/JuliaLang/julia/blob/ae1719872092b594228a9bfeedfc3fa8e91c9529/base/linalg/lu.jl#L281 instead of calling Base.det.

TsurHerman:
@c42f 
given that:

```
julia> log(det(10000*eye(100)))
Inf

julia> logdet(10000*eye(100))
921.034037197617
```

Although it is unlikely to use StaticArrays for matrices size 100x100. I think this example warrants 
a proper `lodget` specialized for small matrices and `Base.logdet ` on anything else, just for completion.

What I did was a quick test on small enough matrices, obviously I missed something 

@mschauer you are correct, but since you are correct I think for now it would be easier to do 
proper logdet. it would look cleaner I think, and once lufact is there for StaticArrays all the code in det and inv can be changed to use that. 

Again I apologize for the noise, thank you guys for your work.

src/traits.jl
```diff
@@ -127,3 +127,20 @@ end
 # Return the "diagonal size" of a matrix - the minimum of the two dimensions
 diagsize(A::StaticMatrix) = diagsize(Size(A))
 @pure diagsize(::Size{S}) where {S} = min(S...)
+
+
+struct Small ; end
+struct Large ; end
+
+"""
+    SizeClass(a::StaticArray, small_limit::Size)
```
andyferris: if `SizeClass` is a supertype of `Small` and `Large`, then it should be capitalized.

If it is just a function, it should be lower case.

src/det.jl
```diff
@@ -4,7 +4,9 @@
     _det(Size(A),A,S)
 end
 
-@inline logdet(A::StaticMatrix) = log(det(A))
+@inline logdet(a::StaticMatrix) = _logdet(SizeClass(a, Size(3,3)), a)
+@inline _logdet(::Small, a::StaticMatrix) = log(det(a))
```
andyferris: Another random suggestion:
```julia
@inline function logdet(a::StaticMatrix)
    if Length(a) < 10
        ...
    else
        ...
    end
end
```

mschauer:
Ping.

andyferris:
@mschauer see also #301

c42f:
I think the fix in #301 is fine.

Let's close this for now - we can come back to the size class stuff if it seems like a good idea later.

PR #269, opened 2017-08-05
Add error tests

kshyatt:
That `convert` test is dying :( but only on `nightly` I guess


andyferris:
And it's GREEN! :)

src/arraymath.jl
```diff
@@ -111,7 +116,7 @@ end
     end
 end
 
-@inline rand!(rng::AbstractRNG, a::SA) where {SA <: StaticArray} = _rand!(rng, Size(SA), a)
+@inline rand!(rng::AbstractRNG, a::SA) where {SA <: StaticArray} = _rand!(rng, a)
```
andyferris: What happened to the `Size` here?

kshyatt: Not sure! I can put it back, or if someone wants to use the online editor...

andyferris:
Thanks again, Katie :)

PR #271, opened 2017-08-07
Add eigfact in a dumb way

kshyatt:
I got mad that we don't have this, but it's pretty simple to do for the single-matrix case. Hope it's ok.


test/eigen.jl
```diff
@@ -5,10 +5,16 @@
         @test vals === SVector(2.0)
         @test eigvals(m) === vals
         @test vecs === SMatrix{1,1}(1.0)
+        ef = eigfact(m)
+        @test ef[:values] === SVector(2.0)
+        @test ef[:vecs] === SMatrix{1,1}(1.0)
```
andyferris: Tests say `:vecs` doesn't exist as a key

fredrikekre: should be `:vectors` :)

kshyatt:
There's another problem with the symmetric tests I'm trying to fix, stand
by for updates :)

kshyatt:
CI passed! Good to go?

Issue #242, opened 2017-07-08
broadcast between SMatrix{M,N} and SMatrix{M,1} (or SMatrix{1,N}) doesn't work


wsshin:
Broadcasting an operator over a matrix (`SMatrix{M,N}`) and a column-like (`SMatrix{M,1}`) or row-like (`SMatrix{1,N}`) matrix doesn't work and generates a `DimensionMismatch` error is generated.  (Seems to be related to https://github.com/JuliaArrays/StaticArrays.jl/issues/199.)
```julia
julia> VERSION
v"0.6.1-pre.0"

julia> SA = @SMatrix [1 2; 3 4]
2×2 StaticArrays.SArray{Tuple{2,2},Int64,2,4}:
 1  2
 3  4

julia> SC = @SMatrix [2;4]
2×1 StaticArrays.SArray{Tuple{2,1},Int64,2,2}:
 2
 4

julia> SR = @SMatrix [2 4]
1×2 StaticArrays.SArray{Tuple{1,2},Int64,2,2}:
 2  4

julia> broadcast(+, SA, SC)
ERROR: DimensionMismatch("Tried to broadcast on inputs sized Tuple{Int64,Int64}[(2, 2), (2, 1)]")
Stacktrace:
 [1] _broadcast(...) at /Users/wsshin/.julia/v0.6/StaticArrays/src/broadcast.jl:52
 [2] broadcast(::Function, ::StaticArrays.SArray{Tuple{2,2},Int64,2,4}, ::StaticArrays.SArray{Tuple{2,1},Int64,2,2}) at /Users/wsshin/.julia/v0.6/StaticArrays/src/broadcast.jl:8

julia> broadcast(+, SA, SR)
ERROR: DimensionMismatch("Tried to broadcast on inputs sized Tuple{Int64,Int64}[(2, 2), (1, 2)]")
Stacktrace:
 [1] _broadcast(...) at /Users/wsshin/.julia/v0.6/StaticArrays/src/broadcast.jl:52
 [2] broadcast(::Function, ::StaticArrays.SArray{Tuple{2,2},Int64,2,4}, ::StaticArrays.SArray{Tuple{1,2},Int64,2,2}) at /Users/wsshin/.julia/v0.6/StaticArrays/src/broadcast.jl:8
```
Is this difficult to fix?

If an actual vector is used instead of a column-like matrix, broadcasting works.  Therefore, currently I am working around this as follows:
```julia
[...]

Issue #200, opened 2017-05-29
Broadcast should handle transpose of `StaticVector`


bshall:
If `x` is a `SVector` then `broadcast(f, x', ...)` should return a `StaticArray`. Currently the broadcasting falls back to Base except for broadcasts with a scalar. For example
```julia
x = SVector(1, 1)
1 .+ x'
```
returns a `RowVector{T, SVector{2, T}}`.

andyferris:
The behavior in the OP is intended.

We can't have that all statically-sized arrays are `<: StaticArray`. The `RowVector` code in `Base` is designed to be compatible with packages like `StaticArrays` (to wrap `RowVector` on the outside). Trait-based dispatch would help to express static-sizeness better...

However, the problem is that not all operations work on `RowVector{<:StaticArray}`. e.g.

```julia
julia 0.6> x = SVector(2,3)
2-element SVector{2,Int64}:
 2
 3

julia 0.6> x' .+ x
2×2 Array{Int64,2}:
 4  5
 5  6
```

We should look into this.

wsshin:
@bshall, this has been also resolved by https://github.com/JuliaArrays/StaticArrays.jl/pull/274.  Could you close this issue as well?

Issue #199, opened 2017-05-29
Another broadcast inconsistency with Base


bshall:
In Base you get
```julia
broadcast(+, [1], []) == [] # true
broadcast(+, [], [1]) == [] # true
```
but in `StaticArrays` you get
```julia
broadcast(+, SVector(1), SVector()) === SVector() # true
broadcast(+, SVector(), SVector(1)) === SVector() # DimesionMismatch error
```

wsshin:
@bshall, this has been fixed by https://github.com/JuliaArrays/StaticArrays.jl/pull/274.  Could you close this issue?

c42f:
Sorry, I forgot to close this off.  Github should have been auto-closed this, but perhaps it needs something like `fixes #197, fixes #199 ...`?  Anyway, closing now.

Issue #198, opened 2017-05-29
Type weirdness in division of array by scalar (issue #14725 in julia)


bshall:
After the fix from JuliaLang/julia#14725 the element type of `a = Number[2, 2.0, 4//2, 2+0im] / 2` is `Number` but in `StaticArrays` you get 
```julia
eltype(SVector{4, Number}(2, 2.0, 4//2, 2+0im) / 2) == Any
```
I guess it's just some promotion rules.

PR #274, opened 2017-08-08
Fix several broadcast issues (fixes #197, #199, #200, #242)

### LINKED ISSUE #197: Broadcast inconsistency with Base
I was writing a few tests for `broadcast` and I noticed an inconstancy. In Base you can broadcast a 2x2 array with a 1x2 array as follows
```julia
broadcast(+, eye(2), [1  4])
```
But the equivalent in `StaticArrays` gives a `DimensionMismatch` error
```julia
broadcast(+, eye(SMatrix{2, 2}), @SMatrix([1  4]))
```

wsshin:
This PR fixes several issues raised for `broadcast` on `SArray`.  Specifically, 
- it broadcasts properly between `SArray` and `RowVec{<:Any,SVector}` to generate an `Sarray` (fixing #197, #200, #242), and
- it also produces a result consistent with `Base`'s for zero-length `SArary` (fixing #199).
- Additionally, it replaces `Base.promote_op` with `Core.Inference.returnt_type`, because the former infers a wrong return type of an operation when the input arguments of the operation are abstract (see https://discourse.julialang.org/t/alternatives-to-base-promote-op-op-type-for-op-type/5289/6).

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -235,7 +235,7 @@ end
 @generated function _diff(::Size{S}, a::StaticArray, ::Type{Val{D}}) where {S,D}
     N = length(S)
     Snew = ([n==D ? S[n]-1 : S[n] for n = 1:N]...)
-    T = Core.Inference.return_type(-, Tuple{eltype(a),eltype(a)})
+    T = typeof(one(eltype(a)) - one(eltype(a)))
 
     exprs = Array{Expr}(Snew)
     itr = [1:n for n = Snew]
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -17,7 +17,7 @@ end
         exprs[i] = :(f($(tmp...)))
     end
     eltypes = [eltype(a[j]) for j ∈ 1:length(a)] # presumably, `eltype` is "hyperpure"?
-    newT = :(Core.Inference.return_type(f, Tuple{$(eltypes...)}))
+    newT = :(Base.promote_op(f, $(eltypes...)))
     return quote
         @_inline_meta
         @inbounds return similar_type(typeof(_first(a...)), $newT, Size(S))(tuple($(exprs...)))
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -17,7 +17,7 @@ end
         exprs[i] = :(f($(tmp...)))
     end
     eltypes = [eltype(a[j]) for j ∈ 1:length(a)] # presumably, `eltype` is "hyperpure"?
-    newT = :(Base.promote_op(f, $(eltypes...)))
+    newT = :(Core.Inference.return_type(f, Tuple{$(eltypes...)}))
     return quote
         @_inline_meta
         @inbounds return similar_type(typeof(_first(a...)), $newT, Size(S))(tuple($(exprs...)))
@@ -105,7 +105,7 @@ end
     N = length(S)
     Snew = ([n==D ? 1 : S[n] for n = 1:N]...)
     T0 = eltype(a)
-    T = :((T1 = Base.promote_op(f, $T0); Base.promote_op(op, T1, T1)))
+    T = :((T1 = Core.Inference.return_type(f, Tuple{$T0}); Core.Inference.return_type(op, Tuple{T1,T1})))
 
     exprs = Array{Expr}(Snew)
     itr = [1:n for n ∈ Snew]
@@ -235,7 +235,7 @@ end
 @generated function _diff(::Size{S}, a::StaticArray, ::Type{Val{D}}) where {S,D}
     N = length(S)
     Snew = ([n==D ? S[n]-1 : S[n] for n = 1:N]...)
-    T = Base.promote_op(-, eltype(a), eltype(a))
+    T = Core.Inference.return_type(-, Tuple{eltype(a),eltype(a)})
 
     exprs = Array{Expr}(Snew)
     itr = [1:n for n = Snew]
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -17,7 +17,7 @@ end
         exprs[i] = :(f($(tmp...)))
     end
     eltypes = [eltype(a[j]) for j ∈ 1:length(a)] # presumably, `eltype` is "hyperpure"?
-    newT = :(Base.promote_op(f, $(eltypes...)))
+    newT = :(Core.Inference.return_type(f, Tuple{$(eltypes...)}))
     return quote
         @_inline_meta
         @inbounds return similar_type(typeof(_first(a...)), $newT, Size(S))(tuple($(exprs...)))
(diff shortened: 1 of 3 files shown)


wsshin:
Sorry about the duplicate commits.  I pulled from master and rebased on master after creating this PR, and ended up having several duplicate commits.

wsshin:
A part of the reason I pushed several commits is that I thought I could address #198 as well by using `Base.promote_op` instead of `Core.Inference.return_type`, because the former seemed to infer a function's return type more specifically when the function's input arguments are abstract.  This turned out to be an illusion; see the discussion in https://discourse.julialang.org/t/alternatives-to-base-promote-op-op-type-for-op-type/5289/6.  So, I decided to use `Core.Inference.return_type` and give up on addressing #198.

Ultimately, I think we need to promote types element-by-element, as discussed in https://github.com/JuliaLang/julia/issues/19669.  Julia 0.6 seems to support this already (even though the mentioned issue is still open), but I haven't looked into how it does in detail.

andyferris:
Very nice - thanks :)

Does this preserve `RowVector`ness? We should reproduce behavior similar to this in `Base`:
```
julia 0.6> 1 .+ [1,2]'
1×2 RowVector{Int64,Array{Int64,1}}:
 2  3

julia 0.6> [1,2]' .* [2,3]'
1×2 RowVector{Int64,Array{Int64,1}}:
 2  6
```
(Sorry, don't have a REPL to check this out right now)

wsshin:
Yes it does:
```
julia> 1 .+ SVector(1,2)'
1×2 RowVector{Int64,SVector{2,Int64}}:
 2  3

julia> SVector(1,2)' .* SVector(2,3)'
1×2 RowVector{Int64,SVector{2,Int64}}:
 2  6
```
However, I didn't intend this behavior to be honest.

wsshin:
I think now I know why my code does not disturb `RowVector`ness.

Whenever some operation is broadcasted between `Union{Number,RowVector}` such that the output is expected to be a `RowVector`, the following function defined [here](https://github.com/JuliaLang/julia/blob/master/base/linalg/rowvector.jl#L152-L153) is called first:
```julia
@inline broadcast(f, rowvecs::Union{Number,RowVector}...) =
    RowVector(broadcast(transpose∘f, to_vecs(rowvecs...)...))
```
This basically strips all the `RowVector{...}` from the arguments, broadcasts the operation, and then puts `RowVector{...}` back to the result.  Therefore in this case `broadcast_c` of the StaticArrays package does not see `RowVector`s at all and only sees `StaticArray`s as its arguments.  In other words, the following code I added
```julia
_containertype(::Type{<:RowVector{<:Any,<:SVector}}) = StaticArray
@inline broadcast_sizes(a::RowVector{<:Any,<:SVector}, as...) = (Size(a), broadcast_sizes(as...)...)
```
does not play any role in this procedure and preserves `RowVector`ness.

However, the added code plays role when an operation is broadcasted between `RowVector`s and `StaticArray`s such that the output is expected to be a `StaticArray` rather than a `RowVector`.  In this case, because not all the arguments are `Union{Number,RowVector}`, `broadcast` is directly applied without striping `RowVector{...}`.  Then `broadcast_c` of the StaticArrays package sees `RowVector`s together with `StaticArrays` as its arguments.  The code I added treats these `RowVector`s as `StaticArray`s and perform `broadcast` to produce a `StaticArray` output.

test/broadcast.jl
```diff
@@ -45,32 +45,32 @@ end
     @testset "2x2 StaticMatrix with 1x2 StaticMatrix" begin
         m1 = @SMatrix [1 2; 3 4]
         m2 = @SMatrix [1 4]
-        @test_broken @inferred(broadcast(+, m1, m2)) === @SMatrix [2 6; 4 8] #197
-        @test_broken @inferred(m1 .+ m2) === @SMatrix [2 6; 4 8] #197
+        @test @inferred(broadcast(+, m1, m2)) === @SMatrix [2 6; 4 8] #197
+        @test @inferred(m1 .+ m2) === @SMatrix [2 6; 4 8] #197
```
c42f: Perhaps it makes sense to remove the issue reference from these individually, and instead describe the behaviour which is being tested in a short comment.  (That comment could include the issue crossref instead.)

In my opinion the source code (including comments) should be the source of truth about design decisions and the reasons for testing given behaviour: the source is a living document, but issues become abandoned as soon as they're closed.

src/mapreduce.jl
```diff
@@ -235,7 +235,7 @@ end
 @generated function _diff(::Size{S}, a::StaticArray, ::Type{Val{D}}) where {S,D}
     N = length(S)
     Snew = ([n==D ? S[n]-1 : S[n] for n = 1:N]...)
-    T = Base.promote_op(-, eltype(a), eltype(a))
+    T = Core.Inference.return_type(-, Tuple{eltype(a),eltype(a)})
```
c42f: IIRC calling `Inference.return_type` is considered a bit of a dirty trick by the core devs ;-) If used, it can change the behaviour of code depending on optimizations to the julia inference engine.  (Is this _the_ reason to avoid it?  Perhaps there's more and someone can enlighten me?).

I know we do it in another couple of places in `StaticArrays`, but I think we can avoid it here by doing something like

```julia
T = typeof(one(eltype(a)) - one(eltype(a)))
```

or some similar such trick?

The version for `_mapreducedim` further up is more tricky, because it takes an arbitrary `op` in the reduction. Unlike `diff`, it's not clear that the data will be numeric or that calling `one(eltype(a))` makes sense in that case.

wsshin:
I agree with @c42f on both points, and made changes accordingly.  Thanks for the comments!

c42f:
Very nice, thanks :-)

wsshin:
Thanks for merging!  I thought this merge would automatically close the Issues mentioned in the PR title, but it closed only the first one.  I closed the last one that I opened.  Could you close the middle two?

andyferris:
> In other words, the following code I added ... does not play any role in this procedure and preserves `RowVector`ness.

That's great news. :) Thanks for this.

wsshin:
It seems that this PR hasn't been added to a release yet.  I don't know how PRs to release are selected in general, but I'm curious if there are any reasons that prevent releasing this PR.  One of my packages has several parts that can be simplified using the features implemented in this PR.

andyferris:
Wow, it's been a while then!

I guess we can do a release? The v0.7 compatibility stuff is only half done and has made the code uglier in some places, but the tests all pass on v0.6.

wsshin:
That will be great for me.

Also, is there an easy way to check if some PRs have been released or not?  I just searched through the release descriptions for the PR numbers, but wonder if there is an easier way.  

According to my search, my other merged PRs, #263 and #255, don't seem released yet, either...

mschauer:
I checked my dependency https://github.com/mschauer/Bridge.jl against master and all is fine as well.

wsshin:
If there is anything I could help in order to create a release with this PR, please let me know.

andyferris:
Yes, sorry, I'll do it now. @wsshin it's easiest to do a "release" via github interface and let [attobot](https://github.com/attobot/attobot) do the rest.

PR #275, opened 2017-08-09
Fix reshape bug and add tests

kshyatt:


kshyatt:
Test failure appears to be originating elsewhere.

andyferris:
Nice catch

PR #279, opened 2017-08-09
Test last convert method

kshyatt:


test/core.jl
```diff
@@ -135,4 +135,6 @@
     @test Tuple{2, 3, 5} != StaticArrays.Size{(2, 3, 4)}
     @test StaticArrays.Size{(2, 3, 4)} != Tuple{2, 3, 5}
     @test StaticArrays.check_length(2) == nothing
+
+    @test convert(Tuple, @SVector [1]) == (1,)
```
mschauer: Better, ``@test convert(Tuple, @SVector [2]) === (2,)``
because e.g.
```
julia> Size(@SVector [1]) == (1,)
true
```

kshyatt: Great idea, updated!

test/core.jl
```diff
@@ -135,4 +135,6 @@
     @test Tuple{2, 3, 5} != StaticArrays.Size{(2, 3, 4)}
     @test StaticArrays.Size{(2, 3, 4)} != Tuple{2, 3, 5}
     @test StaticArrays.check_length(2) == nothing
+
+    @test convert(Tuple, @SVector [2]) == (2,)
```
andyferris: I tend to use `===` for isbits immutables.

mschauer: Cue my comment ;-)

PR #280, opened 2017-08-09
Last test for deque?

kshyatt:


kshyatt:
Hmm, this doesn't seem to be testing that line, I wonder why.

mschauer:
The method
```julia
setindex(a::StaticArray, x, inds::Int...) = setindex(a, x, sub2ind(size(typeof(a)), inds...))
```
in https://github.com/JuliaArrays/StaticArrays.jl/blob/master/src/deque.jl#L88 is an immutable version of `setindex!` and needs to be called specifally.

mschauer:
@kshyatt I think this is not what you wanted to do.

kshyatt:
Yeah, I think you're right. Haven't had time to go back and fix it if
someone else wants to.

mschauer:
Nah - if at some point coverage is 99.9%, we *then* now where to look...

PR #289, opened 2017-08-15
Specialise chol for SDiagonal and support static block matrices

mschauer:
The second part is done by hooking into `Base.LinAlg._chol!`, which is imho the right way to do it but makes it necessary to keep track on base in this regard.

### DIFF
--- a/src/SDiagonal.jl
+++ b/src/SDiagonal.jl
@@ -82,10 +82,16 @@ function logdet(D::SDiagonal{N,T}) where {N,T<:Complex} #Make sure branch cut is
 end
 
 eye(::Type{SDiagonal{N,T}}) where {N,T} = SDiagonal(ones(SVector{N,T}))
+one(::Type{SDiagonal{N,T}}) where {N,T} = SDiagonal(ones(SVector{N,T}))
+one(::SDiagonal{N,T}) where {N,T} = SDiagonal(ones(SVector{N,T}))
+Base.zero(::SDiagonal{N,T}) where {N,T} = SDiagonal(zeros(SVector{N,T}))
 
 expm(D::SDiagonal) = SDiagonal(exp.(D.diag))
 logm(D::SDiagonal) = SDiagonal(log.(D.diag))
 sqrtm(D::SDiagonal) = SDiagonal(sqrt.(D.diag))
+Base.chol(D::SDiagonal) = SDiagonal(Base.chol.(D.diag))
+Base.LinAlg._chol!(D::SDiagonal, ::Type{UpperTriangular}) = chol(D)
+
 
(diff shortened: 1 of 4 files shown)


src/SDiagonal.jl
```diff
@@ -82,10 +82,16 @@ function logdet(D::SDiagonal{N,T}) where {N,T<:Complex} #Make sure branch cut is
 end
 
 eye(::Type{SDiagonal{N,T}}) where {N,T} = SDiagonal(ones(SVector{N,T}))
+one(::Type{SDiagonal{N,T}}) where {N,T} = SDiagonal(ones(SVector{N,T}))
+one(::SDiagonal{N,T}) where {N,T} = SDiagonal(ones(SVector{N,T}))
```
mschauer: If I read Coverall correctly, `one(::SDiagonal{N,T})` is not needed for recursive cholesky. Should I leave this?

andyferris: This looks useful in any case.

src/cholesky.jl
```diff
@@ -7,6 +7,8 @@ end
 @inline function Base.chol(A::Base.LinAlg.RealHermSymComplexHerm{<:Real, <:StaticMatrix})
     _chol(Size(A), A.data)
 end
+@inline Base.LinAlg._chol!(A::StaticMatrix, ::Type{UpperTriangular}) = chol(A)
```
andyferris: Is this function meant to mutate `A`?

andyferris:
If we extend `Base` methods, IMO they need to follow the same mutation / aliasing semantics. Some `StaticArray`s are mutable, so `_chol!` should mutate them.

(This is a general problem with hooking into `Base.LinAlg` - I'd like to hook into it as much as possible but it typically makes assumptions of mutability and that calling `similar` is a good thing)

src/cholesky.jl
mschauer: Convention in Base for this functions is that the `!` indicates that `_chol!` has the licence to modify, but does not need to. Compare https://github.com/JuliaLang/julia/blob/9dbaed2fcfd0cc209e4c4a3bad231361b6aa29be/base/linalg/cholesky.jl#L118

mschauer:
I agree, `_chol!` should actually modify a mutable StaticMatrix and create a new SMatrix, depending on the argument.

mschauer:
As long as there is no mutating chol for MMatrices this is fine, or?

mschauer:
Ping.

andyferris:
Sorry, yes... so am I right in thinking that the `Base.LinAlg._chol!` function will be used primarily for block matrices like `chol(::AbstractMatrix{<:StaticMatrix})`? If so, then this seems pretty sensible.

PR #294, opened 2017-08-20
Even more eigfact fun

kshyatt:
The 1x1 complex matrix case wasn't being handled right ... this eigenvalue is very easy to compute :p.

Added in more tests for an untouched code branch.

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -13,7 +13,7 @@ end
     end
 end
 
-@inline _eigvals(::Size{(1,1)}, a, permute, scale) = @inbounds return SVector(a.data[1])
+@inline _eigvals(::Size{(1,1)}, a, permute, scale) = @inbounds return SVector(Tuple(a))
 @inline _eigvals(::Size{(1, 1)}, a::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real} = @inbounds return SVector(real(parent(a).data[1]))
 
 @inline function _eigvals(::Size{(2,2)}, A::Base.LinAlg.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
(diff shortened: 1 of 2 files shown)


src/eigen.jl
```diff
@@ -6,12 +6,14 @@ end
 @inline function eigvals(a::StaticArray; permute::Bool=true, scale::Bool=true)
     if ishermitian(a)
         _eigvals(Size(a), Hermitian(a), permute, scale)
+    elseif Size(a) == Size{(1,1)}()
+        _eigvals(Size(a), a, permute, scale)
     else
         error("Only hermitian matrices are diagonalizable by *StaticArrays*. Non-Hermitian matrices should be converted to `Array` first.")
```
c42f: Hmm, this error seems like an inconsistency with various other `StaticArrays` code (not your fault, obviously, nor something that needs to be fixed here unless you feel the urge.  But we should call `Base` and propagate the static size to the output rather than erroring out).

src/eigen.jl
```diff
@@ -6,12 +6,14 @@ end
...
     end
 end
 
-@inline _eigvals(::Size{(1,1)}, a, permute, scale) = @inbounds return SVector(real(a.data[1]))
+@inline _eigvals(::Size{(1,1)}, a, permute, scale) = @inbounds return SVector(a.data[1])
```
c42f: Rather than `a.data[1]`, I think this should be `Tuple(a)` which is now the interface for obtaining the tuple of elements, regardless of whether `a` is an `SMatrix` or some other type of static array.

src/eigen.jl
andyferris: Agreed, ideally this should fall back to a `Base` method.

kshyatt:
Yeah, you can now find the eigenvalues of a 1x1 matrix no matter its structure. Serious technical contributions only from me ;).

Issue #291, opened 2017-08-16
Error while executing `using StaticArrays` on nightly


tkoolen:
I'm getting an error even before https://github.com/JuliaArrays/StaticArrays.jl/issues/283 now.

```
fatal: error thrown and no exception handler available.
Base.ArgumentError(msg="No StaticArray found in argument list")
rec_backtrace at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/stackwalk.c:86
record_backtrace at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/task.c:246 [inlined]
jl_throw at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/task.c:568
same_size at /Users/twan/code/julia/RigidBodyDynamics/v0.7/StaticArrays/src/traits.jl:117 [inlined]
map at /Users/twan/code/julia/RigidBodyDynamics/v0.7/StaticArrays/src/mapreduce.jl:10
stacktrace at ./stacktraces.jl:151
display_error at ./client.jl:128
display_error at ./client.jl:142
do_call at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/interpreter.c:70
eval at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/interpreter.c:262
eval_body at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/interpreter.c:539
jl_toplevel_eval_body at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/interpreter.c:511
jl_toplevel_eval_flex at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/toplevel.c:602
jl_toplevel_eval_in at /Users/osx/buildbot/slave/package_osx10_9-x64/build/src/builtins.c:505
eval at ./sysimg.jl:48
_start at ./client.jl:429
true_main at /Applications/Julia-0.7.app/Contents/Resources/julia/bin/julia (unknown line)
main at /Applications/Julia-0.7.app/Contents/Resources/julia/bin/julia (unknown line)
ERROR: Failed to precompile StaticArrays to /Users/twan/code/julia/RigidBodyDynamics/lib/v0.7/StaticArrays.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:598
 [2] _require(::Symbol) at ./loading.jl:407
 [3] require(::Symbol) at ./loading.jl:318
```

```
[...]

Issue #282, opened 2017-08-11
map is overwritten for non-StaticArrays.


andreasnoack:
The problematic lines are https://github.com/JuliaArrays/StaticArrays.jl/blob/7276787a0b147135b758dd836b5951236e30558d/src/mapreduce.jl#L9-L11 The new behavior might be https://github.com/JuliaLang/julia/issues/21026. I'm not sure

This is on 
```julia
Julia Version 0.7.0-DEV.1327
Commit 104dce7fcf* (2017-08-11 16:01 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin16.7.0)
  CPU: Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT NO_AFFINITY HASWELL)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, haswell)
Environment:
```

tkoolen:
Ref: https://github.com/JuliaLang/julia/issues/21026#issuecomment-322873794.

andyferris:
Thanks. We'll need to dream up a workaround for this little "hack" to express such a set of types.

For anyone finding this (and issues like this): It's likely that at some point it will become too difficult to target both v0.6 and v0.7. Given Julia changes like https://github.com/JuliaLang/julia/pull/23168 I'm very tempted to just say "StaticArrays should not be considered as suitable for "real work" on v0.7 until much closer to to the release time". I feel `master` should target v0.6 (with "feature" improvements like LU decompositions) and we merge any v0.7 fixes which are compatible with v0.6. I'm hopeful we can find a simple fix for this issue.

c42f:
Hmm, this is unfortunate, though it was always a bit of a hack relying on this peculiarity.

The really nasty thing IIRC was if you try to do this in another way (by directly writing out various function signatures), the number of ambiguities quickly explodes in your face.

Another option might be to refactor the `map` internals to look a bit more like `broadcast`, as `Broadcast.containertype()` was pretty nice to hook into. Obviously this would be changes to `Base`...

martinholters:
Fixed by #296. (Although the new behavior---only intercept `map` if the first argument (after the function) is a `StaticArray`--- still leaves something to be desired, this particular issue is fixed.)

andyferris:
[...]

PR #296, opened 2017-09-08
Fix `map` for Julia v0.7

andyferris:
This works around a redefinition of `map` for all `AbstractArray` that occurs on v0.7 but not v0.6.

The newly-missed optimization opportunities are probably pretty rare in practice.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -6,8 +6,14 @@
 
 # The following type signature for map() matches any list of AbstractArrays,
 # provided at least one is a static array.
-@inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
-    _map(f, same_size(as...), as...)
+if VERSION > v"0.7.0-"
+    @inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
+        _map(f, same_size(as...), as...)
+    end
+else
+    @inline function map(f, as::StaticArray)
+        _map(f, same_size(as...), as...)
+    end
 end
 
 @generated function _map(f, ::Size{S}, a::AbstractArray...) where {S}
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -6,7 +6,7 @@
 
 # The following type signature for map() matches any list of AbstractArrays,
 # provided at least one is a static array.
-if VERSION > v"0.7.0-"
+if VERSION < v"0.7.0-"
     @inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
         _map(f, same_size(as...), as...)
     end
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -11,8 +11,8 @@ if VERSION < v"0.7.0-"
         _map(f, same_size(as...), as...)
     end
 else
-    @inline function map(f, as::StaticArray)
-        _map(f, same_size(as...), as...)
+    @inline function map(f, a1::StaticArray, as::AbstractArray...)
+        _map(f, same_size(a1, as...), a1, as...)
     end
 end
 


andyferris:
Fixes #282. Hopefully fixes #291.

src/mapreduce.jl
```diff
@@ -6,8 +6,14 @@
...
 # The following type signature for map() matches any list of AbstractArrays,
 # provided at least one is a static array.
-@inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
-    _map(f, same_size(as...), as...)
+if VERSION < v"0.7.0-"
+    @inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
+        _map(f, same_size(as...), as...)
+    end
+else
+    @inline function map(f, as::StaticArray)
```
timholy: How about
```julia
map(f, a1::StaticArray, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
    _map(f, same_size(a1, as...), a1, as...)
end
```
And, do you really need `@inline` on 0.7. One can make the argument that now that we have an inliner that's supposed to be smart enough to figure this stuff out on its own (https://github.com/JuliaLang/julia/pull/22210, https://github.com/JuliaLang/julia/pull/22775), if you need `@inline`, then it's a bug. I still find `@noinline` sometimes handy (you can pick and choose where you want GC frames to be allocated), but I'm not convinced I've ever found a case where `@inline ` helpful, and sometimes it's counterproductive (avoiding it can save you some insane compilation times, e.g., `ntuple(identity, Val(1000))`.

c42f: Yes, this would be less surprising than completely ditching the mixed Static vs Abstract array cases, I suppose.

Regarding how this interacts with Base, I was wondering whether we should have some machinery like that used in Broadcast for determining the output container type, which could solve this problem in general. Any thoughts on feasibility?

andyferris: @timholy I have questioned before whether the inliner takes into account the fact that the overhead of the call is proportional to the length of the arguments. My understanding is that in *StaticArrays* we get improvements for inlining operations on large static arrays because of this fact.

Case in point:
```julia
julia> @generated function _map(f, t::Tuple)
           exprs = [:(f(t[$i])) for i = 1:length(t.parameters)]
           return quote tuple($(exprs...)) end
       end
_map (generic function with 1 method)

julia> f(x) = _map(x->2*x, x)

julia> @code_warntype f(ntuple(identity,3))
Variables:
  #self# <optimized out>
  x::Tuple{Int64,Int64,Int64}
  #33 <optimized out>
  #temp#::Tuple{Int64,Int64,Int64}

Body:
  begin
      # meta: location REPL[17] _map 2
      # meta: location REPL[17] @generated body
      #= line 3 =#
      #temp#::Tuple{Int64,Int64,Int64} = (Main.tuple)((Base.mul_int)(2, (Base.getfield)(x::Tuple{Int64,Int64,Int64}, 1, true)::Int64)::Int64, (Base.mul_int)(2, (Base.getfield)(x::Tuple{Int64,Int64,Int64}, 2, true)::Int64)::Int64, (Base.mul_int)(2, (Base.getfield)(x::Tuple{Int64,Int64,Int64}, 3, true)::Int64)::Int64)::Tuple{Int64,Int64,Int64}
      goto 7
      # meta: pop location
      7: 
      # meta: pop location
      return #temp#::Tuple{Int64,Int64,Int64}
  end::Tuple{Int64,Int64,Int64}

julia> @code_warntype f(ntuple(identity,30))
Variables:
  #self# <optimized out>
  x::NTuple{30,Int64}
  #33::getfield(Main, Symbol("##33#34"))

Body:
  begin
      #33::getfield(Main, Symbol("##33#34")) = $(Expr(:new, :(Main.##33#34)))
      return $(Expr(:invoke, MethodInstance for _map(::getfield(Main, Symbol("##33#34")), ::NTuple{30,Int64}), :(Main._map), :(#33), :(x)))::NTuple{30,Int64}
  end::NTuple{30,Int64}
```

Generally, we've been supporting static arrays up to 16x16 or so (tuples of length 256) and I've tried to avoid `invoke` in the cases where the operation is O(N). I guess I should do more benchmarking to support my case here since perhaps I'm worrying about the wrong thing...

timholy: Model is *currently* fixed cost per `:invoke`, but that would not be hard to change. Would need some benchmarks though.

andyferris: Cool, yes I'm happy to work together on this stuff and will make some benchmarks (when I have some time!!). Did you have any test cases set up from your work on this to-date?

FYI - I do expect that we will majorly update StaticArrays for v1.0 so hopefully we can get rid of all those pesky `@inline`s by then :)

ararslan:
Any chance you could tag a new release now that this is merged?

c42f:
@ararslan does this fix the massive breakage you were observing on 0.7?

src/mapreduce.jl
timholy: > Did you have any test cases set up from your work on this to-date?

I just used BaseBenchmarks, nothing specific.

ararslan:
Actually I'm not sure yet. The breakage was with Distributions, but there seems now to be issues with other packages that are keeping Distributions from loading.

timholy:
In conjunction with some other local changes, this is enough to let me build ImageFiltering.jl now (for the first time in ages). :+1:

PR #297, opened 2017-09-08
Remove incorrect `rand` methods

### LINKED ISSUE #277: Surprising behaviour of `rand(SVector(1,2))`
I was surprised by this behaviour:
```
julia> rand(SVector(1,2))
2-element SVector{2,Int64}:
  3088619120667732393
 -1551417525745515306
```
From the documentation of `rand`:

> Pick a random element or array of random elements from the set of values specified by S; S can be
> 
>     •    an indexable collection (for example 1:n or ['x','y','z'])

I had expected the output to be either equal to A):
```
julia> rand([1, 2])
2
```
or perhaps B):
```
julia> rand((1, 2))
1×2 Array{Float64,2}:
 0.788846  0.302817
```
Instead it seems I am getting C):
```
julia> rand(typeof(SVector(1,2)))
2-element SVector{2,Int64}:
 4251809197651799824
 3934010639117562646
```
I would have really liked it to produce a fast random selection, for example `rand(s::SVector{4})` could compile to `s[1+rand(Int8)&3]` which is ~5 times as fast as `rand(1:4)`.

andyferris:
The elements of instances of static arrays should be used as the set of items to draw from, as done in `Base`. Fixes #277.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -56,9 +56,7 @@ end
     end
 end
 
-@inline rand(rng::AbstractRNG, range::AbstractArray, ::SA) where {SA <: StaticArray} = rand(rng, range, SA)
 @inline rand(rng::AbstractRNG, range::AbstractArray, ::Type{SA}) where {SA <: StaticArray} = _rand(rng, range, Size(SA), SA)
-@inline rand(range::AbstractArray, ::SA) where {SA <: StaticArray} = rand(Base.GLOBAL_RNG, range, SA)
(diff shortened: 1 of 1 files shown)


mschauer:
It would be useful to allow `rand!(a::SArray)`, because for `Arrays`s the`randn(Array)` approach does not work, and to introduce something like `randcopy` for `rand!(copy(a))) to allow code sharing.

andyferris:
I don't quite understand. Aren't `rand!` and `randn!` working as expected?

PR #299, opened 2017-09-11
`rand!` for `@SArray`s copies gracefully.

mschauer:
This is a possible way to alleviate the change in #297, using our earlier conclusion that `!` says a methods may modify if it feels like, but does not have to.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -111,6 +111,7 @@ end
     end
 end
 
+@inline rand!(rng::AbstractRNG, ::SA) where {SA <: SArray} = rand(rng, SA)
 @inline rand!(rng::AbstractRNG, a::SA) where {SA <: StaticArray} = _rand!(rng, Size(SA), a)
 @generated function _rand!(rng::AbstractRNG, ::Size{s}, a::SA) where {s, SA <: StaticArray}
     exprs = [:(a[$i] = rand(rng, eltype(SA))) for i = 1:prod(s)]
@@ -124,7 +125,10 @@ end
 # ambiguity with AbstractRNG and non-Float64... possibly an optimized form in Base?
 @inline rand!(rng::MersenneTwister, a::SA) where {SA <: StaticArray{<:Any, Float64}} = _rand!(rng, Size(SA), a)
 @inline rand!(rng::MersenneTwister, a::SA) where {SA <: StaticArray{<:Tuple, Float64, <:Any}} = _rand!(rng, Size(SA), a)
+@inline rand!(rng::MersenneTwister, a::SA) where {SA <: SArray{<:Tuple,Float64}} = _rand(rng, Size(SA), SA)
 
+
+@inline randn!(rng::AbstractRNG, ::SA) where {SA <: SArray} = randn(rng, SA)
 @inline randn!(rng::AbstractRNG, a::SA) where {SA <: StaticArray} = _randn!(rng, Size(SA), a)
 @generated function _randn!(rng::AbstractRNG, ::Size{s}, a::SA) where {s, SA <: StaticArray}
     exprs = [:(a[$i] = randn(rng, eltype(SA))) for i = 1:prod(s)]
--- a/test/arraymath.jl
+++ b/test/arraymath.jl
@@ -54,5 +54,17 @@
         rand!(m, 1:2)
         check = ((m .>= 1) .& (m .<= 2))
         @test all(check)
+
+        m = @SMatrix [0. 0.; 0. 0.]
+        m2 = rand!(m)
+        check = ((m2 .< 1.) .& (m2 .> 0.))
+        @test all(check)
+    end
+
+    @testset "randn!()" begin # smoke test
+        m = @SMatrix [0. 0.; 0. 0.]
+        m2 = randn!(m)
+        @test typeof(m2) == typeof(m)
+        @test m2 != m     
     end
 end


andyferris:
Thank you @mschauer for the contribution - but unfortunately I have some reservations about this one:

> using our earlier conclusion that `!` says a methods may modify if it feels like, but does not have to.

The `rand!` function definitely populates the input array - see e.g. the docstring for it. 

```
help?> rand!
search: rand! randn! randexp! randsubseq! broadcast_setindex! randn rand randexp randperm randjump randcycle randsubseq randstring RandomDevice srand sprandn sprand bitrand transcode readandwrite

  rand!([rng=GLOBAL_RNG], A, [coll])

  Populate the array A with random values. If the indexable collection coll is specified, the values are picked randomly from coll. This is equivalent to copy!(A, rand(rng, coll, size(A))) or
  copy!(A, rand(rng, eltype(A), size(A))) but without allocating a new array.
```
Generally when their is a pair of functions `f` and `f!` they have a different meaning with intended effects. I do understand the shortcut taken by `_chol!` but that seems a little more internal rather than a documented API.

At a deeper level, I'm a bit confused what we are trying to achieve here - e.g. what code becomes easier to read/write because of this?

mschauer:
In general I would like
```
function W!(y, t)
    n = length(y)
    dt = t/n
    y[1] .= 0.0
    for i in 2:n-1
        y[i] .= y[i-1] + sqrt(dt)*randn!(y[i])
    end
    y
end
```
to work for 
```
W!([SVector(0.0, 0.0) for i in 1:4], 1.0)
W!(zeros(4), 1.0)
```
and
```
W!([zeros(2) for i in 1:4], 1.0)
```

c42f:
I can see why you'd want that... but it does break the semantics of `rand!`, in a way that is generally not very useful except in very particular cases like what you've got here.

If it's just this one case in particular and you want to do a performance hack, how about adding a layer of indirection to use something like

```julia
myrand(a) = rand!(a)
myrand(a::SArray) = rand(a)
```

rather than using `rand!` in your function `W!` directly?

andyferris:
Right - that's a good idea Chris. 

It's worth keeping in mind that we are aiming for an extremely high degree of alignment with Base. Sometimes things need to be fixed there, first. If you feel strongly about this, I'd suggest opening an issue on the Julia GitHub page or starting a discussion on their Discourse. And yes, generally we could improve how Julia handles mutable and immutable structures, and how to make generic code which is efficient for both cases.

mschauer:
I would not call it "a performance hack" because that makes it difficult to discuss this without prejudice. I argue that this is a general scheme, which is also used in Base. (Note that `copy` is a no-op on numbers for example). An implementation of a functionality `f` which is able to deal with mutable and immutable arrays in general needs two ingredients

1.) A function `f!(x) -> y` which may modify it's argument.
2.) A function `f(x) -> y` which may not modify it's argument. 

Then `y = f!(x)` is to be called if `x` is not accessed anymore and `y = f(x)` if `x` is further in use. This problem does not turn up in Base so often, because most important objects have only either immutability or mutability, but not both. It may be a good idea to widen the discussion, but it is really something specific to this package.

SimonDanisch:
But `rand!` is specifically for this use case:
```Julia
A = zeros(10, 10)
rand!(A)
do_stuff_with_randomly_filled(A)
```
Which can't be implemented with immutable arrays and will yield subtle bugs if implemented.

That `rand!` returns `A` so this can "accidentally" work *if* people use it as `A = rand!(A)`, is too subtle for my taste to turn it into an official interface.

mschauer:
I agree that with `rand!` it might be a bit too subtle because the result does not depend on the input. For example `reverse!` would be a much anchor for the discussion. I'll think about it and close this.

andyferris:
There's a third class of function:

3.) A function `f!(x) -> nothing` which must modify `x`, like `setindex!` or `copy!`, because the user will do some more work with `x` that relies on this mutation.

There are some `!` functions which are in class 1 - typically these ones are those that might optionally use the input as scratch space. But when they come in pairs like `rand`, `rand!` or `sort`, `sort!`, or `copy`, `copy!`, they are typically classes 2 and 3.

mschauer:
With "class 2" the functions with property  `a = f!(b, args...); assert(a === b)` which are usually used as `f!(b)`?

andyferris:
Sorry, I don't follow?

andyferris:
The array stuff in `Base` normally follows this kind of pattern:

```julia
function f(a) # class 2
    b = similar(a)
    f!(b)
    return b
end

function f!(a) # class 3
    for i = 1:length a
        a[i] = ...
    end
end
```

You can see how we rely on the semantics of `f!` in order for `f` to work.

mschauer:
So 
```
function f!(a) # class 1
    for i = 1:length(a)
        a[i] = ...
    end
    a
end
```
as in `sort!`, `reverse!`, `clamp!`, `fill!` and friends, I was just confused by `sort`, `sort!`, `rand!` as examples for class 3 in your post.

andyferris:
Oh! Right, yes, many of class 3 happen to return something. `setindex!(a, v, i)` returns `v` (not `a`), while many return the object mutated (`a`), and some return nothing or some other product of the function.

But this is not the identifying feature of the classes. The entire *purpose* of a function in class 3 is, like `setindex!`, to mutate the input. (In class 1, a mutation may occur but isn't really the primary reason for calling the function, and class 1 is relatively rarer)

mschauer:
Ah, I know now what is the right concept for me: I'm concerned with *chainable modifying array functions* and what they should to to immutable arrays. Thanks for the input!

Issue #71, opened 2016-11-10
Implement QR matrix factorization


andyferris:
The `qr()` function should work on statically sized arrays. A first attempt would simply convert to `Array` and wrap the output in a `SizedArray`.

Datseris:
+1 for that, please!

ChrisRackauckas:
I thought, hey, why not let the compiler optimize the crap out of the small matrix cases? Hence #318.

PR #302, opened 2017-09-13
Wrap Base.qr

jw3126:
#71


src/qr.jl
```diff
@@ -0,0 +1,22 @@
+_stabilize_type(::Type{T}) where {T} = typeof((one(T)*zero(T) + zero(T))/one(T))
```
c42f: We use this nice trick elsewhere too (det, solve, eigen, lu, inv).  How about tweaking the function to be called "factorization_eltype" - the expected eltype of the result of factorizing a matrix or solving a linear system:

```julia
factorization_eltype(a::AbstractMatrix{T}) where {T} = typeof((one(T)*zero(T) + zero(T))/one(T))
```

test/qr.jl
```diff
@@ -0,0 +1,38 @@
...
+        @test R ≈ R_ref
+        @test p == p_ref
+    end
+
+
+    for arr in [
+            (@MMatrix randn(2,2)),
+            (@SMatrix randn(10, 12)),
+            map(BigFloat, @SMatrix randn(1,1)),
+            (@SMatrix zeros(Int,10,10)),
```
c42f: Nice tests.  A couple of useful additions:

* A complex matrix is usually good to test generic manipulation of eltypes.
* Not really a big deal, but maybe a `@test_throws` for `_thin_must_hold` behavior?

src/qr.jl
mschauer: Ah, that is a good idea. Maybe this deserves to go into Base because of issues like #23268, which can be solved by overloading `factorization_eltype`.

jw3126: Having this as a reusable function is a great idea. But it needs a good name.
`factorization_eltype` is not perfect since this function is also useful outside of factorizations (e.g. `det`). Also I would prefer to not bake `T = eltype(...)` into the function.
 What do you think of the following:
```Julia 
arithmetic_stabilization(T) = typeof((one(T)*zero(T) + zero(T))/one(T))
```

mschauer: How about `arithmetic_closure` - https://en.wikipedia.org/wiki/Closure_(mathematics) ("f that operation on members of the set always produces a member of the same set")

andyferris: In theory users might make a system that is not closed...

jw3126: We could make the function more "clever":
```Julia
arithmetic_closure1(T) = typeof((one(T)*zero(T) + zero(T))/one(T))
function arithmetic_closure(T)
    S = arithmetic_closure1(T)
    T == S ? S : arithmetic_closure(S)
end
```
or argue that its the users responsibility to overload it.

jw3126: Also if a user has such a system, she might want to redesign it anyway since it will cause subtle type stability issues in a lot of places?

c42f: I'd like to keep mention of "stability" out of such a function name as this would be emphasizing an implementation detail rather than high level meaning of the function. This function is more about inferring the output type of certain types of high level operations, like matrix factorizations and other things.

I think the main question is how we'd like to group the high level operations.

For example, `det` and matrix multiplication do not use `/`, so maybe in these cases the output type should be inferred as `typeof(one(T)*zero(T) + zero(T))`, and we'd have a different name for this.

On that basis there could be two quite different ways to name this function:
* Operationally - closure under `*,+` vs closure under `*,+,/`, vs closure under `+` only, etc
* By category

On the whole, defining it operationally might be easiest and most well defined.

c42f: I'm not so worried by the fact that the user might define an inconsistent and non-closed arithmetic. This seems relatively non-useful, and I can't imagine any useful way of automatically doing the right thing.  Providing a function which can be overridden for such crazy types would be perfectly fine by me ;-)

jw3126: It is true, that `det` mathematically makes sense in cases, where `/` is not defined. However efficient implementations often require `/`. So I think `det` should still use the above function. This is also the behavior in Base:
```Julia
julia> det([1 2; 3 4])
-2.0
```

jw3126: Okay maybe it is a weak argument that usage of `/` in the implementation affects the return type.

jw3126: We could do 
```Julia
closure(T) = T
function closure(T, op, ops...)
    t = one(T)
    S = typeof(op(t,t))
    closure(S, ops...)
end
```

c42f: `det` is a tricky one.  It seems that it really needs to be implemented via factorization to avoid inefficiency and numerical problems, and it's prone to overflow even in stable implementations (hence `logdet`, I recently learned). So I think `Base` makes a reasonable and more or less practical choice here which we should follow for the moment.

The `closure` function is cute. I prefer the version you committed for now though. The term `closure` might be a bit problematic due to the conventional meaning being completely different in mathematics and programming...

src/arraymath.jl
```diff
@@ -149,3 +149,4 @@ end
         return a
     end
 end
+arithmetic_closure(T) = typeof((one(T)*zero(T) + zero(T))/one(T))
```
c42f: Could be mysterious for someone in the future.  Might be worth a one line explanation or so.

src/qr.jl
jw3126: I agree with all you said. So this PR can be merged?

src/arraymath.jl
```diff
@@ -149,4 +149,25 @@ end
         return a
     end
 end
+
+"""
+    arithmetic_closure(T)
+
+Return a "bigger" type, which is closed under `+,-,*,/`.
```
mschauer: "Return a type which can represent `T` and is closed under arithmetic operations `+`, `-`, `*` and `/`."

mschauer: (This is is somewhat a simplification, but let's keep it grounded.)

c42f: Return the type which values of type `T` will promote to under a combination of
the arithmetic operations `+`, `-`, `*` and `/`.

?

c42f:
Travis CI is having OSX build problems again it seems.  I gave up waiting for it to become stable and just merged this :-)

PR #303, opened 2017-09-14
Fix most depwarn and errors on 0.7

yuyichao:


andyferris:
Thanks @yuyichao :) I'm sure this will be useful.

I had worked so hard to avoid Compat for so long, lol... but more seriously we've generally expected to maintain different code for different Julia versions since behaviour/performance here is pretty tightly coupled to the compiler details. @c42f opinion? E.g. an alternative is to begin a `julia-0.7` branch a tag in METADATA according?

c42f:
I think the way we handled the 0.5->0.6 transition was worse for users than it should have been, and I'd like if it was smoother next time.  The main problem, I think, was that we had some rather breaking API changes which happened during rewriting the package.  We then ended up with the core SArray API inconsistent between 0.5 and 0.6 which wasn't great.

So far, I don't see a strong need to rewrite the package for 0.7 so I'm a huge fan of using `Compat` and seeing how we go.

PR #307, opened 2017-09-23
Speed up conversion to StaticArray

tkoolen:
Use `@noinline` trick (see e.g. https://github.com/JuliaLang/julia/pull/23088#issuecomment-320279238) to speed up `convert(::Type{SA}, a::AbstractArray)` when bounds checks are turned on.

Benchmark:
```julia
@benchmark SMatrix{4, 4}(a) setup = a = rand(4, 4)
```

On master:
```
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     14.114 ns (0.00% GC)
  median time:      14.392 ns (0.00% GC)
  mean time:        14.547 ns (0.00% GC)
  maximum time:     43.770 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
```
On this branch:
```
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     11.982 ns (0.00% GC)
  median time:      12.057 ns (0.00% GC)
  mean time:        12.222 ns (0.00% GC)
  maximum time:     37.088 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999
```

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -13,9 +13,13 @@
     unroll_tuple(a, Length(a))
 end
 
+@noinline function dimension_mismatch_fail(SA::Type, a::AbstractArray)
+    error("Dimension mismatch. Expected input array of length $(length(SA)), got length $(length(a))")
+end
+
 @inline function convert(::Type{SA}, a::AbstractArray) where {SA <: StaticArray}
     @boundscheck if length(a) != length(SA)
-        error("Dimension mismatch. Expected input array of length $(length(SA)), got length $(length(a))")
+        dimension_mismatch_fail(SA, a)
     end
 
     return SA(unroll_tuple(a, Length(SA)))


timholy:
Actually I should confess that I was the one who introduced this annoying but useful trick to speed up a couple of (long-standing) hindrances to performance that I happened to notice while I was doing my inlining work. In this case it's not the throw that's the problem, it's the GC root that gets created even when that branch doesn't get taken (because of the string interpolation). If the `throw` were the problem, the difference between these versions would be vastly larger. In an ideal world the `@noinline` annotation wouldn't be necessary, but the inlining cost model currently doesn't account for GC roots. (It runs long before the GC pass is taken, and it would be pretty nontrivial to figure this stuff out.)

Indeed, there's a [special case](https://github.com/JuliaLang/julia/blob/2e864155d0498e73c35f92c91efb4b4799affea1/base/inference.jl#L4771-L4776) that gives `throw` statements a cost estimate of 0, since they really shouldn't ever be part of the normal control flow in a function and hence represent a branch that doesn't affect normal execution.

c42f:
Ah I see. I was thinking that - as a special case - any function such as `error` which contains an unconditional `throw` should never be inlined. If nothing else, to save a bit on code size and icache pressure. But it's very interesting that the problem is in the GC root setup which happens in the frame where the message is formatted.

c42f:
I just noticed https://github.com/JuliaLang/julia/pull/24431 - it looks like we won't need these kind of workarounds fairly soon :-)

PR #308, opened 2017-09-26
Fix bug in 1x1 solve

mschauer:
... and I was puzzling what my code does wrong.

### DIFF
--- a/test/solve.jl
+++ b/test/solve.jl
@@ -6,7 +6,7 @@ using StaticArrays, Base.Test
                 elty in (Float64, Int)
 
         A = elty.(rand(-99:2:99,n,n))
-        b = A*ones(elty,n)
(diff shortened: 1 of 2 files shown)


test/solve.jl
```diff
@@ -6,7 +6,7 @@ using StaticArrays, Base.Test
                 elty in (Float64, Int)
 
         A = elty.(rand(-99:2:99,n,n))
-        b = A*ones(elty,n)
+        b = 2*A*ones(elty,n)
```
tkoolen: Oof, this wasn't the best test. I'd frankly still be a little worried about mistakes in indexing into `b` in the various `solve` implementations. Perhaps make it so that the elements are all different?

c42f:
(Fix looks fine btw.)

PR #311, opened 2017-10-04
ctranspose -> adjoint

fredrikekre:
- extend `Base.adjoint` instead of deprecated `ctranspose`
- fix a doc-overwrite warning in tests


src/StaticArrays.jl
```diff
@@ -6,14 +6,21 @@ import Base: @_inline_meta, @_propagate_inbounds_meta, @_pure_meta, @propagate_i
...
              lu, svd, svdvals, svdfact, factorize, ishermitian, issymmetric, isposdef,
              iszero, sum, diff, prod, count, any, all, minimum,
              maximum, extrema, mean, copy, rand, randn, randexp, rand!, randn!,
              randexp!, normalize, normalize!, read, read!, write, Eigen
 
+if VERSION >= v"0.7.0-DEV.1415"
+    import Base.adjoint
+else
+    import Base.ctranspose
+    const adjoint = ctranspose
```
c42f: I assume there's nothing for this in `Compat` yet?

fredrikekre: There is. The alternative would be this:
```julia
import Compat.adjoint
export adjoint
```
and bump to `Compat 0.32` in `REQUIRE`. I thought this was a bit more confusing than the current implementation, since we essentially extend `Compat.adjoint` rather than the base method. Which solution do you prefer?

fredrikekre: On second thought that might bet better after all, since then users can use `adjoint` on v0.6 as well.

mschauer: Doesn't Compat deal with that anyway?

fredrikekre: After an explicit `using Compat`, (i.e. you would have to to `using StaticArrays; using Compat`).

mschauer: That is just fine (and transparent).

fredrikekre: Okay, so I'll remove the `export adjoint`?

andyferris:
Hmmm... this PR should be designed to have zero impact on v0.6 users, I guess. Compat usage should be optional. So long as that's true I'm really happy!

fredrikekre:
Yea, should be fine now.

src/StaticArrays.jl
c42f: Yep, let's use Compat for this, and remove the export. Oh, and I see that you've already done this - nice.

PR #313, opened 2017-10-05
Anonymous FieldVector macros

ChrisRackauckas:
Allows the user to define `FieldVector`s in a much simplified syntax:

```julia
b = @FVector Float64 [a,b,c] [1,2,3]
d = @MFVector Float64 [a,b,c] [1,2,3]
```

`b` is mutable and `d` is not. Allows for an uninitialized version as well.

### DIFF
--- a/src/FieldVector.jl
+++ b/src/FieldVector.jl
@@ -26,3 +26,73 @@ abstract type FieldVector{N, T} <: StaticVector{N, T} end
 Base.cconvert(::Type{<:Ptr}, v::FieldVector) = Base.RefValue(v)
 Base.unsafe_convert(::Type{Ptr{T}}, m::Base.RefValue{FV}) where {N,T,FV<:FieldVector{N,T}} =
     Ptr{T}(Base.unsafe_convert(Ptr{FV}, m))
+
+# Anonymous FieldVector Macros
+
+"""
+    @FVector Type Names
+    @FVector Type Names Values
+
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Can I get a review? Or is this something that should be a separate package?

ChrisRackauckas:
In light of https://github.com/JuliaArrays/StaticArrays.jl/issues/314, I'll be taking these to a separate package so that way they can broadcast correctly.

andyferris:
Sorry Chris, it's been a busy time for me.

TBH I had mixed feelings about this, because I'm currently considering that we reduce the number of macros in *StaticArrays*.

Functionality in a separate package would certainly give everyone what they want. Of course, if you find a system that works really, really well please consider upstreaming it.

c42f:
Yes, sorry this PR didn't get any attention, the general idea seems reasonable.  At a glance I can't see anything fundamental which prevents this macro from working as part of `StaticArrays`.  I'd probably change some details, eg would it make more sense to eval the type definition into the current module at macro expansion time rather than returning it?  That would allow use in more contexts rather than only at top level. Also I wonder whether it's really wise to gensym a new type for every use, or whether having a single type name which is based on the field names would be better.

All that being said, it might be best to have it in a separate package for now. I'd certainly be interested in having a look at what you end up with there.

ChrisRackauckas:
Yeah, I can do a lot more with it in its own package: https://github.com/ChrisRackauckas/LabelledArrays.jl

PR #317, opened 2017-10-06
Move eltype computations into generated function bodies

### LINKED ISSUE #316: Use of `one` in generator causing world age issues for ForwardDiff
ref https://github.com/JuliaDiff/ForwardDiff.jl/issues/277

MWE (expected output is an empty `SVector`):

```julia
julia> using ForwardDiff, StaticArrays

julia> StaticArrays.diff(SVector(ForwardDiff.Dual(1)))
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type ForwardDiff.Dual{Void,Int64,0}
The applicable method may be too new: running in world age 21813, while current world is 21817.
This may have arisen from a call to the constructor ForwardDiff.Dual{Void,Int64,0}(...),
Stacktrace:
 [1] one(::Type{ForwardDiff.Dual{Void,Int64,0}}) at ./number.jl:163
 [2] _diff(...) at /Users/jarrettrevels/.julia/v0.6/StaticArrays/src/mapreduce.jl:245
 [3] diff(::SVector{1,ForwardDiff.Dual{Void,Int64,0}}) at /Users/jarrettrevels/.julia/v0.6/StaticArrays/src/mapreduce.jl:238
```

AFAICT this is caused by the use of `one` (which ForwardDiff overloads for `Dual` numbers) within the generator [here](https://github.com/JuliaArrays/StaticArrays.jl/blob/0c21db3239242fe65bf21a60d43a06c4e979d3d5/src/mapreduce.jl#L241).

c42f:
Having eltype computations involving `one` and `zero` in the generator
code can introduce world age errors for user types which override these
functions after StaticArrays is included.  Instead, move these
computations into the bodies of the functions, and rely on inference to
optimize it away.

Fixes #316.

I'm not quite sure how to test this.  Could make a custom numeric type which overloads `one`, and test `diff` gets executed correctly, but arguably this kind of test with the custom type should be added for all code with generators.  Which is pretty intrusive and a fair amount of work.

### DIFF
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -12,33 +12,29 @@ end
 
 @generated function _chol(::Size{(1,1)}, A::StaticMatrix)
     @assert size(A) == (1,1)
-    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
-    newtype = similar_type(A,T)
 
     quote
         $(Expr(:meta, :inline))
-        ($newtype)((sqrt(A[1]), ))
+        T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
+        similar_type(A,T)((sqrt(A[1]), ))
     end
 end
 
 @generated function _chol(::Size{(2,2)}, A::StaticMatrix)
     @assert size(A) == (2,2)
-    T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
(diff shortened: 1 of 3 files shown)


c42f:
> I thought we did all of this

I think we did - looking at the changes, these are all recent changes where the problem has crept back in again.

PR #318, opened 2017-10-08
WIP: Hardcoded QR

ChrisRackauckas:
It's WIP because I started doing this because I realized I could, but then stopped half way to ask if I should.  The precedent is in `inv`, but...


Datseris:
holy bananas these are some endless expressions....

c42f:
Wow, that's pretty big mountain of code for only the 2x2 case! Unless this can be simplified I think we'd probably better not do this.

I do think we can have a worthwhile small sized QR and LU etc, but I'd guess it's more productive to pursue a specialized fixed size numerical formulation rather than an explicit formula.

Datseris:
@c42f what is the problem with having code with many lines? 

I am asking to gain knowledge.

c42f:
@Datseris If you ask a computer algebra system to generate an explicit formula, it will try to give you a closed form expression without considering what's really important for a numerical implementation:
* Numerical efficiency - an algorithm containing iterations / branches / appropriate factoring can be a lot faster.
* Numerical stability in finite precision - the order of operations often matters, a lot.

There's often a small size limit where a closed form can be very fast and just as stable (eg, 2x2 matrix inverse).  The 2x2 code here looks really complex though.

Datseris:
I see. Yeah both of those things make perfect sense!

c42f:
@ChrisRackauckas well it looks ugly, but how does this actually perform? Is it fast? Accurate?

andyferris:
Ok, yes I see what people are saying. Wow. :)

I imagine just working through the 2x2 case on pen-and-paper should yield a reasonable solution, no? I wonder what literature there is on this stuff, too.

ChrisRackauckas:
I can do this with SymPy instead and run a `FullSimplify` and see what happens. Should help a lot. Right now, this is all of the operations for the 2x2 non-pivoted QR. In this form the compiler can simplify it as well, but I haven't actually tested the speed. It's as accurate as the Base QR, because that's what it is and it has a test that passes.

src/qr.jl
```diff
@@ -19,3 +19,246 @@ function qr(A::StaticMatrix, pivot::Type{Val{false}}; thin::Bool=true)
     RT = similar_type(A, T)
     QT(Q0), RT(R0)
 end
+function qr(A::Size{(1,1)}, pivot::Type{Val{false}})
```
andyferris: This signature can't be right... `A` should be a `StaticMatrix` and you should have a thunk to generate a `Size` and dispatch to a 3-argument e.g. `_qr` function.

mforets:
> I wonder what literature there is on this stuff, too (in terms on numerically stable algorithms).

there is some discussion of the explicit 2x2 case in [pages 77-80 of Numerical Algebra by John Todd](https://books.google.fr/books?id=zLjSBQAAQBAJ&pg=PA77&lpg=PA77&dq=explicit+2x2+qr+decomposition&source=bl&ots=E6ealJWbo-&sig=v8a4gubZWASq2M4w6iAjbczwUms&hl=es-419&sa=X&ved=0ahUKEwiwof34s_jWAhUFb1AKHd0cBPwQ6AEIUDAF#v=onepage&q=explicit%202x2%20qr%20decomposition&f=false). he reduces the problem to finding a fixed point of a 2d system.

ChrisRackauckas:
Superseded by https://github.com/JuliaArrays/StaticArrays.jl/pull/328

PR #319, opened 2017-10-11
Make unit tests deterministic by seeding the RNG

c42f:


mschauer:
We could try one more seed to see if the seed exercises all branches in eigen.

c42f:
True, but the tests in eigen will depend on how many times `rand()` is called before then, so when we add extra tests there's no guarantee this will stay fixed.

I could add `srand()` to the top of each test file to make this less of a problem. I was thinking it would be nice to be able to run individual test files without reseeding, but maybe it's for the best.

Really I'd prefer to avoid random unit tests entirely and have all the test cases hard coded to cover the various branches. But that's time I don't have :)

andyferris:
Agree - in each file is the practical trade-off here.

c42f:
> maybe it's for the best

To clarify, I mean it's probably for the best to have a lot more calls to `srand()`.

c42f:
True, doing it inside runtests.jl is a reasonable idea.

c42f:
Ok, here's a slightly higher effort version. Let's see what coverage says about this.

andyferris:
Cool :)

c42f:
I tried changing the seed here again, but eigen.jl still has a few uncovered lines.  Did we have full coverage before?  I'm a little confused as codecov and coveralls aren't in obvious agreement about whether coverage has gone up or down here.

mschauer:
It's only a hack, lets just merge any version.

c42f:
> It's only a hack

Indeed :-)  Merged now - it should be an improvement at least - and we can continue to work on improving coverage.

PR #320, opened 2017-10-12
WIP: allow block matrix inversion

marius311:
This allows inverse to work if the entries of the SMatrix are themselves matrix-like, and therefore more naturally have inv(A) defined rather than 1/A. 

I'm not totally sure there's anything in Base for which this actually works, but I think its a pretty reasonable thing to have (I have some cases in my code, e.g.). Its pretty easy to create an example type for which this matters from Base.Diagonal if you just give it, 

```
import Base: one, zero
one(::Type{Diagonal{T}}) where {T} = Diagonal(Vector{T}(0))
zero(::Type{Diagonal{T}}) where {T} = Diagonal(Vector{T}(0))
```

then the following works after this PR (but otherwise did not since https://github.com/JuliaArrays/StaticArrays.jl/commit/3fbb8547b8e64604fbc005e13132b4a275ba2477):

```
julia> inv(@SMatrix [Diagonal([2.]) Diagonal([0.]); Diagonal([0.]) Diagonal([2.])])
2×2 StaticArrays.SArray{Tuple{2,2},Diagonal{Float64},2,4}:
 [0.5]   [-0.0]
 [-0.0]  [0.5] 
```

### DIFF
--- a/src/inv.jl
+++ b/src/inv.jl
@@ -9,7 +9,7 @@ end
 
 @inline function _inv(::Size{(2,2)}, A)
     newtype = similar_type(A)
-    idet = 1/det(A)
+    idet = inv(det(A))
     @inbounds return newtype((A[4]*idet, -(A[2]*idet), -(A[3]*idet), A[1]*idet))
 end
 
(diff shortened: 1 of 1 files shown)


mschauer:
Apparently that this only works under additional assumptions, for example when `C` and `D` in 
`
[A B; C D]
`
commute:

```
julia> A, B, C, D = (rand(2,2) for i in 1:4)
Base.Generator{UnitRange{Int64},##3#4}(#3, 1:4)
julia> det([A B; C D])
-0.09891368319244176
julia> det(A*D -B*C)
-0.17211644836301365
```
but 
```
julia> C = diagm(rand(2))
julia> D = diagm(rand(2))
julia> det([A B; C D])
-0.02471368053596748
julia> det(A*D -B*C)
-0.02471368053596748
```

So I am not sure if this should work, because to do it right, one would need the blockmatrix inversion formula https://en.wikipedia.org/wiki/Block_matrix#Block_matrix_inversion to compute the 2x2 inverse.

marius311:
That's a good point, my Diagonal example only gives the right answer since they all commute. I suppose it is more work for proper block inversion, which I could be up for doing. Although to keep this efficient, I guess we need a way to know if * is commutative for a given type, does anyone know if Julia have anyway of specifying this?

mschauer:
No, it is also not specified if you should be able to trust `Number`s to be commuting (see `Quaternion{T<:Real} <: Number` in Quaternions.jl), but taking the block inverse for abstract array entries seems to be sane and the current form for `Numbers` seems to be sane compromise.

andyferris:
If commutativity of the underlying scalars is a big simplification, you are free to write an algorithm that only works for `Real` or for `Union{Complex,Real}` and that would be OK (we can use dispatch or something to catch these cases).

I don't think we should merge this as-is if it doesn't even work for arbitrary real block-matrices with individually invertable blocks (and really, the individual blocks needn't be invertable either).

PR #328, opened 2017-10-29
Add QR decomposition

### LINKED ISSUE #71: Implement QR matrix factorization

denius:
This PR fixes #71. Implemented QR decomposition without pivoting, algorithm derived from `base/linalg/qr.jl`.

There is only `qr()` method, nor `qrfact()` because latest returns struct `QR` which contains `Vector` and is not static. Thus SA needs own `StaticQR` incompatible with Base.LinAlg and that need decision.

~~Type stability problem for `thin` keyword is not solved, but there is exist `@generated` type stable function `qr_householder_unrolled(Size(A), A, Val{false}, Val{false})` to run `thin=false` option.~~
To use `thin=false` option call ~~`StaticArrays._qr(Size(A), A, pivot, Val{false})`~~ `qr(A, pivot, Val{false})`.

The necessity of the pivot option for small matrices is questionable, IMO.


src/qr.jl
```diff
@@ -2,20 +2,200 @@ _thin_must_hold(thin) =
...
+# For original source code see below.
+@generated function qr_householder_unrolled(::Size{sA}, A::StaticMatrix{<:Any, <:Any, TA}, thin::Union{Type{Val{false}},Type{Val{true}}}) where {sA, TA}
+    mQ = nQ = mR = m = sA[1]
+    nR = n = sA[2]
+    # truncate Q and R for thin case
+    m > n && (mR = n)
+    m > n && thin <: Type{Val{true}} && (nQ = n)
+
+    Q = [Symbol("Q_$(i)_$(j)") for i = 1:m, j = 1:m]
+    R = [Symbol("R_$(i)_$(j)") for i = 1:m, j = 1:n]
```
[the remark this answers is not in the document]
c42f: Random thought - it seems like we could get all this from the Base implementation of QR if we had execution tracing.  @andyferris perhaps one day we'll be able to use Cassette.jl to replace many of our code generators...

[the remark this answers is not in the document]
andyferris: Yep :) In fact, I suppose all sorts of things (outside static arrays) could end up with magically better performance using the same tracing/Cassette trick, just as another layer of JIT...

denius:
Thank you for so detailed analysis and answer.

The [benchmark](https://gist.github.com/denius/fcd9e509cbced7455ada65397ed42b76) and its [results](https://gist.github.com/denius/646d0d2300f5d0f5bfe2d2cbf59c938f) was used for comparisons and size cutoff estimation. In results the first line is `@time` for rough measurement of duration of compilation and next two lines are `@btime` for comparison of SA and LAPACK. Compile time significantly increases at 15x15 matrices and from 20x20 there is no big advantage over built-in version of `qr()`.

src/qr.jl
```diff
@@ -2,20 +2,200 @@ _thin_must_hold(thin) =
...
-    Q0, R0 = Base.qr(Matrix(A), pivot)
-    T = arithmetic_closure(eltype(A))
-    QT = similar_type(A, T, Size(diagsize(A), diagsize(A)))
-    RT = similar_type(A, T)
-    QT(Q0), RT(R0)
+    return qr(Size(A), A, pivot, Val{true})
+end
+
+@generated function qr(SA::Size{sA}, A::StaticMatrix{<:Any, <:Any, TA}, pivot::Type{Val{false}}, thin::Union{Type{Val{false}}, Type{Val{true}}}) where {sA, TA}
+    if sA[1] < 17 && sA[2] < 17
```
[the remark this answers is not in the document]
denius: Oh, I see. But there is definitely should be `qr_householder_loop` like `A_mul_B_loop` which resolve compile time explosion and save from built-in `qr()` call.
In any case estimation of number of expressions should be more accurate, I understand.

andyferris:
First, thanks, this is pretty cool. :)

There is a `perf` directory for benchmarks - please throw in your benchmark script and results, as it's really helpful to look a back at things and see "oh that's why we decided on that cutoff last year" or whatever.

> Compile time significantly increases at 15x15 matrices

To me, the benchmark results seem to reflect that compile time is cubic... other than that, nothing too special at 15x15 is there?  It seems we can pick any number between 15 and 20 (depending on compile-time vs run-time tradeoff) and do the LAPACK version from there onwards. (To be honest I was expecting to have this up to 4x4 and that's it, so this is already fantastic!).

c42f:
Thanks for the benchmarks. Could you add them somewhere in the `perf` directory? `perf` is not very organized at all, so you could just add what you have to `perf/qr.jl`.

It's hard to know how to choose the cut in the tradeoff between compile time vs runtime.  One way to compare is to compute the number of QR decompositions you need to perform before the program as a whole (compile + runtime) goes faster.  For 17x17 this is about `8.69 / (20.32e-6 - 4.834e-6)` which amounts to `560 000` QR decompositions.  This seems like rather a lot to me, but it depends on the application.

andyferris:
> For 17x17 this is about `8.69 / (20.32e-6 - 4.834e-6)` which amounts to which amounts to `560 000` QR decompositions. This seems like rather a lot to me, but it depends on the application.

So that's about 1.2GB of data. For 14x14 the tradeoff is around 267 000 matrices which is 400MB of data. It's all in the real of possible usefulness, I guess.

The matrix multiply stuff goes up to 14x14 and then bails out... for simplicity, I vote we just implement the same here.

denius:
For users of interactive application the 2 seconds delayed program start is annoying but for HPC/cluster users where software can starting on half an hour it's negligible. Ideally it should be parameter applied on `using StaticArrays(threshold=14)`.

> The matrix multiply stuff goes up to 14x14 and then bails out... for simplicity, I vote we just implement the same here.

Opposite to QR in the matrix multiply branch cases even greater 14x14 are static, they call `A_mul_B_loop` which is one loop and not fully unrolled but anytime static. I suppose, in the QR method such approach is only possible with recursive functions which will have O(N^2) expressions instead of O(N^3) in current version. Is there exist good sample of recursive `@generated` function on `SMatrix`? And in general, is such recursion possible?

denius:
Some refinements and fixes, as recommended by @c42f, did I do everything right?

I want to note that there is `qr(Size(A), A)` not hided in `_qr()` because it allow to bypass restrictions of base's `qr()` keyworded interface. Thus one may say `thin=false` problem is solved.

andyferris:
IMO I wouldn't worry *too* much about matching the keyword interface from `Base` when it doesn't make sense for when the size is in the type... personally I'm OK to skip it entirely or else assert it were `true`, and throw an error otherwise.

andyferris:
> I want to note that there is `qr(Size(A), A)` not hided in `_qr()` because it allow to bypass restrictions of base's `qr()` keyworded interface. Thus one may say thin=false problem is solved.

I guess I'm confused what problem this solves. As a user, are you going to use this interface? Can you anticipate anyone wanting to use it?

denius:
> personally I'm OK to skip it entirely or else assert it were `true`, and throw an error otherwise (as I see you've done).
> As a user, are you going to use this interface? Can you anticipate anyone wanting to use it?

@andyferris, you are right, the not truncated matrix `Q` is needed only for numerical libraries developers (mainly least squares problem), which always can call internal `StaticArrays._qt(SA,A,p,t)`. In particular, I need only the stable solve and inversion which are on square matrices and `thin` don't matter.

Thus, once the full match in capabilities with `Base.qr()` is not required `qr(Size(A),...)` should be renamed to `_qr()` to reduce the number of exported functions.

src/qr.jl
```diff
@@ -2,20 +2,203 @@ _thin_must_hold(thin) =
...
-    Q0, R0 = Base.qr(Matrix(A), pivot)
-    T = arithmetic_closure(eltype(A))
-    QT = similar_type(A, T, Size(diagsize(A), diagsize(A)))
-    RT = similar_type(A, T)
-    QT(Q0), RT(R0)
+
+
+@generated function _qr(::Size{sA}, A::StaticMatrix{<:Any, <:Any, TA}, pivot::Union{Type{Val{false}}, Type{Val{true}}} = Val{false}, thin::Union{Type{Val{false}}, Type{Val{true}}} = Val{true}) where {sA, TA}
+
+    isthin = thin <: Type{Val{true}}
```
c42f: This can be expressed a little more straightforwardly as `thin isa Val{true}` (same elsewhere)

src/qr.jl
```diff
@@ -2,20 +2,203 @@ _thin_must_hold(thin) =
...
 import Base.qr
 
-function qr(A::StaticMatrix, pivot::Type{Val{true}}; thin::Bool=true)
+
+"""
+    qr(A::StaticMatrix, pivot=Val{false}; thin=true) -> Q, R, [p]
+
+Compute the QR factorization of `A` such that `A = Q*R` or `A[:,p] = Q*R`, see [`qr`](@ref).
+This function is not support `thin=false` keyword option due to type inference instability.
+To use this option call `StaticArrays._qr(Size(A), A, pivot, Val{false})` instead.
```
c42f: Typo: "This function is not" -> "This function does not".

I'd also suggest it's best to remove the recommendation to use `_qr` - it's not a public API, and I don't think we want to support it being stable going forward ;-) Instead, we can leave it as an unsupported hack, or add a new supported version of `qr` with a positional argument for the `thin=Val(false)` option.  If you don't need it, I suggest just removing the reference is fine for now.

denius:
> Instead, we can leave it as an unsupported hack, or add a new supported version of qr with a positional argument for the `thin=Val(false)` option.

The ability to get full, uncutted matrix is important. Thus there is new exported function `qr(A, pivot, thin)` was added.

src/qr.jl
denius: `isa` in `@generated` functions is not worked.

I had try `===` which should work, but test gives an error, because in [line](https://github.com/denius/StaticArrays.jl/blob/qr-wopivoting/src/qr.jl#L29) `if` condition can't rightly recognize `pivot`.

Thus there is remains only `<:` and `==` correct variants. And I prefer `<:` because it deal with types.

c42f: Ah right of course - this is a generated function body so my suggestion to use `isa` is wrong.

I'd say it's quite acceptable and appropriate use `==` here because types are values in julia. In this case `<:` also works because `Val{true}` is a concrete type (ie has no subtypes), but as the subtype operator it suggests different (type partial ordering) semantics.

denius: Ok, I get it, thanks

c42f: Thanks so much for changing all these small things!

I just realized I was a little extra confused here because in julia-0.7-dev the `qr` function has been changed to take a `Val{true}()` instance rather than a `Val{true}` type.  But let's worry about that another time.

andyferris: Yeah, that's my fault...

c42f: Missed one eh? :-)

Datseris:
Oh my god these benchmarks are astonishing! Thanks a lot everybody who worked on this! I will put it to good use asap!

PR #330, opened 2017-11-01
Remove final explicit uses of Inference.return_type

### LINKED ISSUE #198: Type weirdness in division of array by scalar (issue #14725 in julia)

### LINKED ISSUE #329: Type inference failure on `round` broadcast
On `StaticArrays.jl` master with Julia 0.6.1:

```julia
  | | |_| | | | (_| |  |  Version 0.6.1 (2017-10-24 22:15 UTC)
 _/ |\__'_|_|_|\__'_|  |  

julia> using StaticArrays

julia> Core.Inference.return_type(x -> round.(Int, x), (Vector{Float64},))
Array{Int64,1}

julia> Core.Inference.return_type(x -> round.(Int, x), (SVector{4,Float64},))
SVector{4,Any}

julia> f(v) = round.(Int, v)

julia> f(SVector(1.,2.,3.))
3-element SVector{3,Any}:
 1
 2
 3
```

Yet somehow this works fine:
```julia
julia> round.(Int, SVector(1.,2.,3.))
3-element SVector{3,Int64}:
 1
 2
 3
```

Maybe an issue with base Julia, not with StaticArrays?

c42f:
With this approach based on the eltype of tuple of computed elements, we get the correct dynamic result independently of whether inference can infer the return type.

This fixes #198, and partially addresses #329.

With the addition of a workaround hack to override a very specific version of `broadcast` in a very specific way it also fixes #329.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -102,12 +102,10 @@ end
         end
     end
 
-    eltype_exprs = [t <: AbstractArray ? :($(eltype(t))) : :($t) for t ∈ a]
-    newtype_expr = :(Core.Inference.return_type(f, Tuple{$(eltype_exprs...)}))
-
     return quote
         @_inline_meta
-        @inbounds return similar_type($first_staticarray, $newtype_expr, Size($newsize))(tuple($(exprs...)))
+        elements = tuple($(exprs...))
+        @inbounds return similar_type($first_staticarray, eltype(elements), Size($newsize))(elements)
     end
 end
 
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -105,7 +105,7 @@ end
     return quote
         @_inline_meta
         elements = tuple($(exprs...))
-        @inbounds return similar_type($first_staticarray, eltype(elements), Size($newsize))(elements)
+        @inbounds return similar_type($first_staticarray, promote_tuple_eltype(elements), Size($newsize))(elements)
     end
 end
 
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -25,7 +25,7 @@ end
     return quote
         @_inline_meta
         elements = tuple($(exprs...))
-        @inbounds return similar_type(typeof(_first(a...)), eltype(elements), Size(S))(elements)
+        @inbounds return similar_type(typeof(_first(a...)), promote_tuple_eltype(elements), Size(S))(elements)
     end
 end
 
@@ -126,7 +126,7 @@ end
     return quote
         @_inline_meta
         elements = tuple($(exprs...))
-        @inbounds return similar_type(a, eltype(elements), Size($Snew))(elements)
+        @inbounds return similar_type(a, promote_tuple_eltype(elements), Size($Snew))(elements)
     end
 end
 
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -22,11 +22,10 @@ end
         tmp = [:(a[$j][$i]) for j ∈ 1:length(a)]
         exprs[i] = :(f($(tmp...)))
     end
-    eltypes = [eltype(a[j]) for j ∈ 1:length(a)] # presumably, `eltype` is "hyperpure"?
-    newT = :(Core.Inference.return_type(f, Tuple{$(eltypes...)}))
     return quote
         @_inline_meta
-        @inbounds return similar_type(typeof(_first(a...)), $newT, Size(S))(tuple($(exprs...)))
+        elements = tuple($(exprs...))
+        @inbounds return similar_type(typeof(_first(a...)), eltype(elements), Size(S))(elements)
     end
 end
 
@@ -110,8 +109,6 @@ end
 @generated function _mapreducedim(f, op, ::Size{S}, a::StaticArray, ::Type{Val{D}}) where {S,D}
     N = length(S)
     Snew = ([n==D ? 1 : S[n] for n = 1:N]...)
-    T0 = eltype(a)
-    T = :((T1 = Core.Inference.return_type(f, Tuple{$T0}); Core.Inference.return_type(op, Tuple{T1,T1})))
 
     exprs = Array{Expr}(Snew)
     itr = [1:n for n ∈ Snew]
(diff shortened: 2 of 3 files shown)


src/broadcast.jl
```diff
@@ -102,15 +102,17 @@ end
...
     end
 
-    eltype_exprs = [t <: AbstractArray ? :($(eltype(t))) : :($t) for t ∈ a]
-    newtype_expr = :(Core.Inference.return_type(f, Tuple{$(eltype_exprs...)}))
-
     return quote
         @_inline_meta
-        @inbounds return similar_type($first_staticarray, $newtype_expr, Size($newsize))(tuple($(exprs...)))
+        elements = tuple($(exprs...))
+        @inbounds return similar_type($first_staticarray, eltype(elements), Size($newsize))(elements)
```
andyferris: `eltype(::Tuple)` is (or at least, was) a bit pesimised. In `util.jl` there is a function called `promote_tuple_eltype` that I originally wrote for this kind of stuff. I'm not sure what the current status on which one is better (and it's possible that reducing via `typejoin` is a better choice here than `promote_type` anyway??)

c42f: Interesting.  I didn't give type reduction a lot of thought TBH, because I'm expecting all the elements to have the same inferred type (How could they not? They're all copies of the same expression unrolled many times.)

In the known bad cases which required the broadcast workaround, the issue seems to be in inferring the type of the element tuple, rather than in the `typejoin` step.

c42f:
> What does it do for `Union{T,Null}`

An excellent question. On this branch we have (for both 0.7 and 0.6):

```julia
julia> broadcast(+, SVector(1.0,null), SVector(2.0,3.0))
2-element SVector{2,Any}:
 3.0
  null
```

Where as, with current StaticArrays master, we have

```julia
julia> broadcast(+, SVector(1.0,null), SVector(2.0,3.0))
2-element SVector{2,Union{Float64, Nulls.Null}}:
 3.0
  null
```

which is arguably better (though probably neither is useful on 0.6).

I wonder whether it's a problem with the way I've done things on this branch, or a flaw in the "use `Union{Null,T}` for missingness" idea? Basically, by using the actual runtime type of the elements via `eltype`, we can compute a more precise container type for the SVector at runtime, but at the cost of a much less precise type at compile time (thinking that `Union{Null,T}` wil be efficient in 0.7).  This doesn't sound like a win :-(

I guess, if nothing else, what I've done here is to bring these methods in line with the rest of StaticArrays, where we generally infer the static array type from the element tuple.

I'm now not quite sure whether to merge this!

c42f:
If I replace `eltype` with `promote_tuple_eltype`, we have

```julia
julia> broadcast(+, SVector(1.0,null), SVector(2.0,3.0))
2-element SVector{2,Union{Float64, Nulls.Null}}:
 3.0
  null
```

which is an improvement, i think.  It's still not type inferrable though.

What do you think about `Union` Andy?  Is it a problem for another time?  It seems like we'd have to reassess many things in the package if we wanted to allow possible `Null`s to propagate through in a nice way.

c42f:
Ok, I've pushed a change to use `promote_tuple_type` instead.

The broadcast workaround doesn't seem necessary on 0.7, so presumably the associated inference problem has been sorted out in Base.

andyferris:
Yeah, `Null` and `Union` is a pain. I'm not sure what the correct answer is here (sorry).

c42f:
Hmm, looks like I've broken the test case which asserts that StaticArrays shares the Base behaviour when you have an array of `Number`.  Ie,

```julia
julia> Number[2, 2.0, 4//2, 2+0im] .+ 2
4-element Array{Number,1}:
  4
  4.0
 4//1
 4+0im
```

With `promote_tuple_eltype`, the resulting array is a `Complex{Float64}` instead.

Honestly, I'm not sure this really matters, and perhaps I could just remove that test case: having an `SArray` of abstract types like `Number` seems a bit useless.

andyferris:
I have this feeling that `map` and `broadcast` should be widening the `eltype` until all elements match, not promoting them at the end (and therefore potentially changing them). Does that make sense? Maybe we need a `typejoin_tuple_eltype`-kind-of-thing?

c42f:
Yes, I was just about to suggest that we might need our own purpose built `typejoin`-like thing :-)  We can then tweak it to the needs of the users of this package.

Regarding promotion vs typejoinn - we do already use promotion in similar circumstances:

```julia
julia> SVector(1, 1.0)
2-element SVector{2,Float64}:
 1.0
 1.0
```

It's arguably a similar kind of thing, and certainly much more predictable and useful behavior from a performance standpoint.  (We follow Base here, which has the same array construction conventions.)

I do think that it would be nice if arithmetic on `SVector{N,Union{Null,T}}` preserved the union type, which is how things currently work when we rely directly on calling `return_type`.  Thinking julia 0.7 here, the `eltype` based version would be a bit of a performance disaster in this case.  Consider `SVector(1, null) .+ 1`.  When using the `typejoin` version this expands to:

```julia
a = SVector(1,null)
# a::SVector{2,Union{Int,Null}}
elements = (1 + a[1], 1 + a[2])
# elements::Tuple{Union{Int, Null},Union{Int, Null}}
newtype = eltype(elements)
# newtype::Type{S} where S<:Tuple{Union{Int,Null}, Union{Int,Null}}
newvec = similar_type(a, newtype)(elements)
# newvec::StaticArrays.SArray{Tuple{2},_,1,2} where _
```

Unfortunately, inferring `StaticArrays.SArray{Tuple{2},_,1,2} where _` for the output seems pretty impractical as it will completely destroy the performance.  I wonder whether there's a way around this which makes sense (other than what we already do)?

andyferris:
I think the time for this has finally come.

See #495,  #493 and JuliaLang/julia#28981.

I think I'll make an updated version of this PR this soon. Upon reflection on what `Base` does, it appears we should use `eltype` not `promote_tuple_eltype` for the eltype calculation. Basically, unlike literal construction of arrays, operations like `map` and `broadcast` don't tend to `promote` all the elements to a common type. E.g.

```julia
julia> map(-, Union{Float64, Int}[1, 1.0])
2-element Array{Real,1}:
 -1  
 -1.0
```

andyferris:
Implemented by #503

PR #332, opened 2017-11-02
Workaround inference problem in round.(Int,v)

### LINKED ISSUE #329: Type inference failure on `round` broadcast

c42f:
Addresses the same problem as #330 but in a more minimal way.  Fixes #329 .

julia-0.6 doesn't seem to do constant propagation of `Type`s in quite
the way which would be required to get good efficiency for broadcasting
`round`.

Insert a minimal workaround for this particularly common case.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -111,6 +111,12 @@ end
     end
 end
 
+if VERSION < v"0.7.0-DEV"
+# Workaround for #329
+@inline function Base.broadcast(f, ::Type{T}, a::StaticArray) where {T}
+    map(x->f(T,x), a)
+end
+end
 
 ################
 ## broadcast! ##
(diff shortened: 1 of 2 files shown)


c42f:
I'm going to just merge this as it's a minimal fix.  We can decide what to do about #330 later :-)

PR #336, opened 2017-11-12
Ensure static array constructors create copies

### LINKED ISSUE #335: MArray Constructor should make copy
As [discussed on discourse](https://discourse.julialang.org/t/convert-or-copy-to-staticarrays-marray/7001/2) (and following the conclusion in JuliaLang/julia#23107), the `MArray` constructor should return a copy if an `MArray` is passed to it, but it does not in StaticArrays v0.6.4:

```julia
julia> a = @MVector [1,2,3]
3-element MVector{3,Int64}:
 1
 2
 3

julia> MVector(a) === a
true
```

c42f:
I'm not sure whether I've missed anything here, but calling `Tuple` seems like the simplest way to achieve a copying constructor.

Fixes #335 

The second commit here is a quick fix for some broken constructors I found when creating a test case for this.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -71,7 +71,7 @@ end
     end
 end
 
-@inline MArray(a::StaticArray) = MArray{size_tuple(typeof(a))}(Tuple(a))
+@inline MArray(a::StaticArray) = MArray{size_tuple(Size(a))}(Tuple(a))
 
 # Simplified show for the type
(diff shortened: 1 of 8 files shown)


mschauer:
Maybe a test that the `convert(T, x::T)`s are not affected (not copying)?

c42f:
Sounds like a good idea.  I added that test as well.

PR #343, opened 2017-11-17
don't allow FieldVector to be constructed from tuple of the wrong size

### LINKED ISSUE #342: Constructing a FieldVector with too many elements kills Julia
Derived from https://discourse.julialang.org/t/crash-stackoverflow-when-calling-constructor-with-too-many-inputs/7116

Here's a simple example:

```julia
julia> struct Foo{T} <: FieldVector{2, T}
         x::T
         y::T
       end

julia> Foo{Int}(1,2,3)
# Julia crashes here, presumably due to stack overflow
```

I believe there's an easy fix. I'll open a PR.

rdeits:
fixes #342

### DIFF
--- a/src/FieldVector.jl
+++ b/src/FieldVector.jl
@@ -17,7 +17,7 @@ For example:
 abstract type FieldVector{N, T} <: StaticVector{N, T} end
 
 # Is this a good idea?? Should people just define constructors that accept tuples?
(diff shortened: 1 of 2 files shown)


rdeits:
I also added a test to make sure that this wouldn't break the (admittedly strange) use case of having a FieldVector whose field is itself a tuple.

c42f:
This looks good to me, thanks!

PR #345, opened 2017-11-21
Fix qr() type promotion

denius:
Fixes bug similar to [#24589](https://github.com/JuliaLang/julia/issues/24589) in Julia like in corresponding [PR](https://github.com/JuliaLang/julia/pull/24602). This fixes type stability and type promotion for `Rational{}` and so on.


c42f:
Eh, I had a niggling feeling that `arithmetic_closure` wouldn't be quite enough in general and we'd need to approach this on a case by case basis.

Fix looks reasonable to me.

denius:
For most of matrix operations, like `*`, `inv()`, `lu()`, `arithmetic_closure` is enough. The `qr()` along with `svd()`, `eig()` is exception because of `sqrt` in it in `norm.`

PR #346, opened 2017-11-21
Simplify "value type (Val)" usage

expnn:
In 0.7, `Val(N)` is the instance of `Val{N}`, so it does not require `Type{Val{N}}` any more. 
```julia
julia> isa(Val(4), Val{4})
true
```
Even though, this is not true for 0.6 and early versions of julia, all tests are passed till now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ specialized for (potentially immutable) `StaticArray`s. Many of Julia's
 built-in method definitions inherently assume mutability, and further
 performance optimizations may be made when the size of the array is known to the
 compiler. One example of this is by loop unrolling, which has a substantial
-effect on small arrays and tends to automatically triger LLVM's SIMD
+effect on small arrays and tends to automatically trigger LLVM's SIMD
 optimizations. Another way performance is boosted is by providing specialized
 methods for `det`, `inv`, `eig` and `chol` where the algorithm depends on the
 precise dimensions of the input. In combination with intelligent fallbacks to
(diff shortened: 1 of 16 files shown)


c42f:
Looks like a good simplification, thanks. `Compat` defines the `Val(N)` constructor, so you can rely on this being available.

However, this is a rather large breaking change to the public API - I think it deserves a round of deprecations at least, and possibly just continued support for both versions until 0.7 is out. (In fact, I wonder whether constant propagation will be powerful enough in 0.7 that we can remove some of these.)

There's two general cases:

1. Functions where we needed to break compatibility with `Base` to get the output type inferred (eg, `diagm`)
2. Functions where we use exactly the same input types as `Base` (eg, the `pivot` argument to `lu`).

For the first case, it seems we could do a round of deprecations, and just go with `Val(N)` immediately. For the second case, I'd be uncomfortable about breaking consistency with 0.6 until 0.7 is out, and I think we should continue to support both versions until 0.7 has been released.

So, from my point of view, here's a plan to get this merged: (1) Make a `depwarn` version of all functions in class (1). For all functions in class (2) create a version which takes the old arguments and forwards them to your new version.

BTW, the CI tests are currently failing in LU, so I think there's some more work to do there.

Thanks for contributing!

c42f:
Actually on further reflection, perhaps we should postpone merging this until it's clearer what the situation regarding constant propagation will actually be in 0.7?  It could be that we can just remove most of our uses of `Val` entirely which would allow us to be exactly compatible with `Base`.  See for example https://github.com/JuliaLang/julia/issues/24011 and linked issues.

It would be unfortunate to force users of `StaticArrays` through an extra unnecessary deprecation cycle.

c42f:
Seems this was sorted out in #473 and elsewhere. Sorry we dropped the ball on this PR :-/

PR #348, opened 2017-12-07
Use new broadcast API

### LINKED ISSUE #347: ERROR: MethodError: no method matching result_style(::Type{StaticArray})
Hi 

found a new issue today (my code was running without any issues, so it must be something related to julia itself)

Julia Version 0.7.0-DEV.2675
Commit f383276c5b (2017-11-27 20:28 UTC)
Platform Info:
  OS: Linux (x86_64-redhat-linux)
  CPU: Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)

and a fresh clone from this repository.

The error:

```
julia> A = @SMatrix ones(2,2)
 1.0  1.0
 1.0  1.0

julia> A*0.1
ERROR: MethodError: no method matching result_style(::Type{StaticArray})
Closest candidates are:
  result_style(::Any, ::Any) at broadcast.jl:273
  result_style(::Base.Broadcast.BroadcastStyle) at broadcast.jl:270
  result_style(::S<:Base.Broadcast.BroadcastStyle, ::S<:Base.Broadcast.BroadcastStyle) where S<:Base.Broadcast.BroadcastStyle at broadcast.jl:271
Stacktrace:
 [1] combine_styles(::SArray{Tuple{2,2},Float64,2,4}) at ./broadcast.jl:265
 [2] combine_styles(::SArray{Tuple{2,2},Float64,2,4}, ::Float64) at ./broadcast.jl:266
 [3] *(::SArray{Tuple{2,2},Float64,2,4}, ::Float64) at /home/lenz/.julia/v0.7/StaticArrays/src/linalg.jl:26
 [4] top-level scope

```

pablosanjose:
Attempt to fix #347 

(Please review carefully, as I'm still quite new at this.)

I implemented a `StaticArrayStyle{N} <: AbtractArrayStyle{N}` to plug into the new `broadcast` API introduced in julialang/#23939, together with two specialised indirections for `broadcast` and `broadcast!` that call the old `@generated _broadcast` and `@generated _broadcast!` routines. The latter does still seem to be a bit faster and allocate a little less than the Base `broadcast!` fallback. However, the Base fallback also works.

~~I see tight `@code_native` output, similar or identical to the one before julialang/#23939. Timings and allocations also seem identical, although I didn't do exhaustive testing.~~ ~~[EDIT: not always]~~ [EDIT2: fixed now, hopefully]

**Caveat**
The package tests now work again, except for those involving broadcast between `Scalar(...)` and normal `Array`s. I'm afraid I need a bit of help with this one. As an example `Scalar(3) .* [1,2,3]` fails with the same error as `Scalar(3) * 2`, a `BoundsError`. This is the case both before and after julialang/#23939, but for some reason `Scalar(3) .* [1,2,3]` did work with the old broadcast API, and returned `Array{Int}[2, 4, 6]`. What is the expected behavior for this? What should `Scalar(3) * 2` give? How should `Scalar(3) .* [1,2,3]` be fixed with the new broadcast API?

[EDIT: fixed Scalar broadcast. I needed to properly define the correct broadcast promotions using binary methods for `BroadcastStyle`]


pablosanjose:
This still needs some work. The `@code_native` output is not as good as before...

pablosanjose:
Aha. The problem was that specialising `broadcast(f, ::StaticArrayStyle, ::Type{ElType}, inds::Tuple, As...)` was unnecessarily slow, because both `ElType` and `inds` where computed with the generic `combine_eltypes` and `combine_indices` from Base. Those where not used, as the old method relied on the static `StaticArray.broadcast_sizes` methods. Specialising on `broadcast(f, ::StaticArrayStyle, ::Void, ::Void, As...)` instead gets us back to the old timings:

```
julia> @btime f()
  1.382 ns (0 allocations: 0 bytes)
3-element SVector{3,Int64}:
 0
 4
 9
```

src/broadcast.jl
```diff
@@ -2,37 +2,77 @@
...
+
+    # Add a new BroadcastStyle for StaticArrays, derived from AbstractArrayStyle 
+    # A constructor that changes the style parameter N (array dimension) is also required
+    struct StaticArrayStyle{N} <: AbstractArrayStyle{N} end
+    StaticArrayStyle{M}(::Val{N}) where {M,N} = StaticArrayStyle{N}()
+
+    BroadcastStyle(::Type{<:StaticArray{D, T, N}}) where {D, T, N} = StaticArrayStyle{N}()
+    
+    # Fix Precedence: Make StaticArray - Array -> Array
+    BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.VectorStyle) where {N,M} = Broadcast.Unknown()
```
andyferris: `N` is undefined.

src/broadcast.jl
```diff
@@ -2,37 +2,77 @@
...
+    BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.MatrixStyle) where {N,M} = Broadcast.Unknown()
+
+    # Add a broadcast method that calls the old @generated routine
```
andyferris: "old" isn't really necessary here

src/broadcast.jl
pablosanjose: Thanks Andy!

src/broadcast.jl
```diff
@@ -2,37 +2,77 @@
...
+    # Add a new BroadcastStyle for StaticArrays, derived from AbstractArrayStyle 
+    # A constructor that changes the style parameter N (array dimension) is also required
+    struct StaticArrayStyle{N} <: AbstractArrayStyle{N} end
+    StaticArrayStyle{M}(::Val{N}) where {M,N} = StaticArrayStyle{N}()
+
+    BroadcastStyle(::Type{<:StaticArray{D, T, N}}) where {D, T, N} = StaticArrayStyle{N}()
+    
+    # Fix Precedence: Make StaticArray - Array -> Array
+    BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.VectorStyle) where M = Broadcast.Unknown()
+    BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.MatrixStyle) where M = Broadcast.Unknown()
```
timholy: `VectorStyle` and `MatrixStyle` are unfortunate (and hopefully temporary) objects because we don't trust that sparse broadcasting generalizes correctly beyond `Array` (see https://github.com/JuliaLang/julia/pull/23939#discussion_r147042741). However, this suggests we'll have to support them throughout julia 1.x. :cry: 

Long-term the one we want to implement would be with `DefaultArrayStyle`. Instead of returning `Unknown`, it might be best to do something like this (untested):
```julia
BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.DefaultArrayStyle{N}) where {M,N} =
    DefaultArrayStyle(Broadcast._max(Val(M), Val(N))
```

pablosanjose: Hey Tim, thanks for reviewing! Your designs are a pleasure to work with. 

So you suggest including this rule, in addition to the `VectorStyle` and `MatrixStyle` ones, to take over once the latter are removed, yes? Done.

timholy: Right. Also the `DefaultArrayStyle` is necessary even now for dimensions higher than 2.

Thanks for the kind words about the API. I think it's also fair to say that unfortunately they don't help StaticArrays that much; they're more oriented around improving the API for mutable arrays. But at least it's documented, and hopefully that's worth something.

Really grateful to you for tackling this!

krcools:
CI fails because too much output is generated. Update BEAST - I removed the debug output that was inserted in assemblechunk. Hopefully this resolves the problem.

pablosanjose:
Please forgive the supernoob question: what is that failed check from coveralls? Is it just because I need to add some tests? (I'm not sure if this is waiting for me to do something...)

tkoolen:
Note that there have been additional changes to `broadcast!` recently (see https://github.com/JuliaLang/julia/pull/24992), which should enable resolution of your `TODO`. See https://docs.julialang.org/en/latest/manual/interfaces/#extending-in-place-broadcast-1 for the required method signatures.

Re: coveralls: a slight decrease in code coverage is to be expected. Code coverage is determined on Julia 0.6, so the 0.7 code doesn't get run. It doesn't matter (check is not required to pass before merging).

tkoolen:
Actually, here's a PR against your branch that implements the change: https://github.com/pablosanjose/StaticArrays.jl/pull/1.

timholy:
And probably more changes coming...https://github.com/JuliaLang/julia/pull/25377

andyferris:
Sorry, I haven't got around to this in a while. Is this good to go?

pablosanjose:
It is as far as I'm concerned... (At least until JuliaLang/julia#25377 rains hellfire on us! :-P)

pablosanjose:
I just tried this again in the latest Julia master, and I think there are still `broadcast` problems.
```
julia> @SVector[1,2,3]*2
ERROR: UndefVarError: Inference not defined
Stacktrace:
 [1] @generated body at /Users/pablo/.julia/v0.7/StaticArrays/src/broadcast.jl:152 [inlined]
 [2] _broadcast at /Users/pablo/.julia/v0.7/StaticArrays/src/broadcast.jl:94 [inlined]
 [3] broadcast at /Users/pablo/.julia/v0.7/StaticArrays/src/broadcast.jl:62 [inlined]
 [4] broadcast at ./broadcast.jl:615 [inlined]
 [5] *(::SVector{3,Int64}, ::Int64) at /Users/pablo/.julia/v0.7/StaticArrays/src/linalg.jl:26
 [6] top-level scope
```

Not sure if this happened some weeks ago. I'll try to look into it if I can find the time.

andyferris:
FYI @pablosanjose I've been working on a branch which deals with all the v0.7 carnage and have incorporated these changes there (thanks!).

However every day I nearly get the tests passing and then the next day something else breaks. (This time it's `Base.LinAlg` -> `LinearAlgebra` stdlib change. Fun.)

pablosanjose:
You have all my support and empathy. StaticArrays is a central tool for many of us. Keep at it!

Issue #97, opened 2017-01-26
Function like ntuple for constructing StaticArrays


dpsanders:
The following presumably creates a temporary:

```
julia> SVector(ntuple(i->i^2, 3))
3-element StaticArrays.SVector{3,Int64}:
 1
 4
 9
```
It would be great to be able to do
```
julia> SVector(i->i^2, 3)
3-element StaticArrays.SVector{3,Int64}:
 1
 4
 9
```
without creating the tuple temporary.

SimonDanisch:
You won't be able to make it much better than this, since the innermost constructor of SVector will accept a tuple.
However you can do this: `ntuple(i->i^2, Val{3})`, to speed things up a bit ;)

KristofferC:
The tuple "temporary" will just be eliminated in the compiler optimizations though?

SimonDanisch:
exactly!

andyferris:
Agreed. At least when I checked simple cases for this kind of stuff some time ago, `@code_native` showed what appeared to be optimal code (but I could be wrong, and that could have changed, and it could depend on exactly what you do and how).

That constructor seems fine, We could maybe provide a `@pure` version so that we don't need to write the `Val` around the length (the same goes for `ntuple` in Base, really). The package has `Size` trait type which is a bit like `Val` which could fit into here somewhere.

What I really want is `SVector(i^2 for i = 1:3)` (or `SVector(f(i) for i = 1:N)`, where `N` is some `Const` value). AFAIK this will only be optimally fast if the `Generator` is `Const` in inference, which means the constructor for ranges and generators would need to be pure or effect free or have a tfunc (yuck) or something. (When I get into this stuff, this is usually where @vtjnash offers some insight about possibilities and impossibilities - any thoughts, Jameson?)

The final problem with `SVector(f(i) for i = 1:N)` and `SVector(f, N)` is when `f` is a closure (more precisely, a closure that captures at least one variable, or any other kind of non-constant, callable object) then not even a `@pure` constructor will help. The worst part isn't that the values can't be computed at compile-time; it is that the length of the `SVector` won't be known at compile time (one way around that for `SVector(f, N)` is recursive inlining and inference/constant prop, but even that won't work for `SVector(f(i) for i=1:N)` when `f` isn't constant, unless we also keep track of `Const`ness for individual fields of immutables. This all seems rather speculative, so I don't know a good solution for any of this, except to use `Val`/`Size` for the size).

[...]

Issue #24, opened 2016-09-05
Naming for small vectors


c42f:
As discussed at https://github.com/SimonDanisch/FixedSizeArrays.jl/issues/159, I rather liked writing `Vec` and `Mat` for small sized arrays in FixedSizeArrays.  Perhaps we should try to recover these names for StaticArrays, perhaps not?

Note that SIMD.jl has also started using Vec so there is unfortunately already a naming clash with FixedSizeArrays there @eschnett

andyferris:
Also interesting: SIMD.jl is probably closer to `VecElement` which is in Base.

andyferris:
OK, I'll respond with the history of my decision and we can work on the future from there:

 * I wanted to do multidimensional arrays, but `Arr` seems kinda silly. It's nice to have something as close to `typealias SVector{T} SArray{T,1}` as possible but I didn't manage to get there because there are only a small set of operations allowed in a type alias and `SArray` has a complicated parameter structure (if we can address this over time that is good).
 * There is a history of other packages like *ArrayFire.jl* has `AFArray` and *DistributedArrays.jl* has `DArray` (each with `AFVector` and `DVector`, etc) so it's becoming a little idiomatic in Julia to name things this way.
 * Being careful not to have a package race/war to see who can claim `Vec` and `Mat` first (back then I didn't want to clash with *FixedSizeArrays* either.
 * There is more than one prefix for different types of object, like `SArray` and `MArray`. I'm thinking something like `PArray` or `PtrArray` for a pointer-based, fixed-size array.
 * For the sake of being verbose rather than concise (people who know Julia but not *StaticArrays.jl* might guess that `SVector` behaves quite similarly to `Vector`).

eschnett:
The name `VecElement` in Base was chosen because it is shorter than `SIMDVectorElement`, and because it is more descriptive than `VecEl`.

If there is an overall consistent naming scheme, then renaming `SIMD`'s types is not out of the question.

c42f:
Yes, all valid points.

On the subject of verbosity, I feel that `Vec` is something that gets typed rather a lot, so I'd consider the "small names for commonly used things" rule applies in this case.  I also like the "short names for short vectors" pun ;-)

What's the idea behind `PArray` vs `MArray`?

andyferris:
@eschnett `SIMDElement` ?

@c42f Would a typealias for `Vec` be useful? 

> What's the idea behind PArray vs MArray?

[...]

PR #349, opened 2017-12-08
Add short-form macro for SArray construction

cdsousa:


mschauer:
Ref  #24

cdsousa:
Actually I have no preference... I can change the code if people agree on
something.
`@SA` is closer to a type.
`@sa` is closer to something that resembles a "custom literal macro" (as
strings), and is visually lighter.

c42f:
Sorry this PR has languished :-(

However, I think we should close it in favor of a larger rethink of the constructor syntax in the StaticArrays-1.0 timeframe cf. #633 / #636 / #97

PR #350, opened 2017-12-09
Fix Array{T}(uninitialized,...) deprecations

pablosanjose:
With `@compat` decorations.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.6.0
-Compat 0.32
+Compat 0.39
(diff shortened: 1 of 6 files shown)


fredrikekre:
There shouldn't be a need for `@compat`. But you need `Compat 0.39` in REQUIRE

pablosanjose:
Oh! So how does it go then? You just REQUIRE Compat and use the master branch syntax, and it will just work in all versions?

That's cool!

pablosanjose:
Just to be clear, one still needs to do `using Compat`, yes?

fredrikekre:
Yes https://github.com/JuliaArrays/StaticArrays.jl/blob/bb5102170dae514e440c1db5f22ba6e6e2bfc59a/src/StaticArrays.jl#L19 but you might have to add it to the tests too.

fredrikekre:
Actually, it was already there https://github.com/JuliaArrays/StaticArrays.jl/blob/3801f1742bfddec19f1331e5153bae9d08a522fe/test/testutil.jl#L2

PR #355, opened 2018-01-10
Adapt to Base.data_pointer_from_objref removal

tkoolen:
Ref: https://github.com/JuliaLang/julia/pull/25362.

### DIFF
--- a/src/util.jl
+++ b/src/util.jl
@@ -80,10 +80,10 @@ TrivialView(a::AbstractArray{T,N}) where {T,N} = TrivialView{typeof(a),T,N}(a)
 # See https://github.com/JuliaLang/julia/issues/23826
 # """
 #     drop_sdims(a)
-# 
+#
 # Return an `AbstractArray` with the same elements as `a`, but with static
 # dimensions removed (ie, not a `StaticArray`).
-# 
+#
 # This is useful if you want to override dispatch to call the `Base` version of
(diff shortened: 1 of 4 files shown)


c42f:
Looks good to me.

Looking at Base, I'm quite confused by the difference between `data_pointer_from_objectref` vs `pointer_from_objectref`, they seem to be the same all the way back to julia-0.5.

src/util.jl
```diff
@@ -92,3 +92,9 @@ TrivialView(a::AbstractArray{T,N}) where {T,N} = TrivialView{typeof(a),T,N}(a)
 # """
 @inline drop_sdims(a::StaticArray) = TrivialView(a)
 @inline drop_sdims(a) = a
+
+@static if VERSION >= v"0.7.0-DEV.3256"
+    const _pointer_from_objref = pointer_from_objref
+else
+    const _pointer_from_objref = Base.data_pointer_from_objref
+end
```
c42f: With the previous comment in mind... I assume we can just use the public `pointer_from_objectref` in julia 0.6 as well as 0.7?

tkoolen: Yeah I think you're right. I thought the additional type assertion maybe had a function, but judging by https://github.com/JuliaLang/julia/commit/1750a83ce233151d29101961db9de10e74a1f624 it was just redundant.

c42f:
Travis was having an issue with github availability.  I restarted it.

PR #357, opened 2018-01-13
Fix #356, Performance of SMatrix(::StaticMatrix)

### LINKED ISSUE #356: Performance of SMatrix(::StaticMatrix)
```julia
m = rand(SMatrix{3, 3})
@btime SMatrix($m)
```
results in 2.470 μs (33 allocations: 1.94 KiB). This particular use case is not so interesting, but this method also creates an `SMatrix` from  other `StaticMatrix` types, which is occasionally useful. PR coming.

tkoolen:
Fixes https://github.com/JuliaArrays/StaticArrays.jl/issues/356.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -36,6 +36,8 @@ mutable struct MArray{S <: Tuple, T, N, L} <: StaticArray{S, T, N}
     end
 end
 
+convert_similar_type(::Type{<:MArray}, ::Type{T}, S::Size) where {T} = MArray{size_tuple(S), T, length(S), get(Length(S))}
+
 @generated function (::Type{MArray{S,T,N}})(x::Tuple) where {S,T,N}
     return quote
         $(Expr(:meta, :inline))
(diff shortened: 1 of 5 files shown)


c42f:
This fix looks functional, but perhaps we should just be deleting this method entirely, and fixing the equivalent `SArray` constructor instead?

(This has the feeling of a lot of other issues we've had in `StaticArrays` ever since `SMatrix` became a simple type alias of `SArray`.)

tkoolen:
I finally got around to this again. What do you think of the new approach?

It currently requires defining a `convert_similar_type` method for each `StaticArray` subtype that needs to work with these generic `StaticArray` constructor methods. I wanted to just use `similar_type`, but for this purpose it's too inconsistent in whether it returns an `MArray` or `SArray` type. Unfortunately, it's not possible to define `convert_similar_type` generically for all possible subtypes of `StaticArray`, because in general it isn't even known how many type parameters any given concrete `StaticArray` subtype has.

andyferris:
OK, interesting approach. I can see that we need *something* like this. (I wonder if Julia v0.7 includes any implementation for controlling eltypes for `AbstractArray`?)

> Unfortunately, it's not possible to define convert_similar_type generically for all possible subtypes of StaticArray, because in general it isn't even known how many type parameters any given concrete StaticArray subtype has.

We used to have a hilarious implementation for exactly this, that would analyze the type tree to the point where it could replace exactly the correct type parameters as necessary. :)

c42f:
It looks like #356 was closed by taking a much more minimal approach in #358. Should we close this issue?

tkoolen:
Sure.

PR #358, opened 2018-01-13
Add SHermitianCompact

tkoolen:
I was just playing around yesterday with a statically-sized symmetric matrix type that (as opposed to `Base.Symmetric`) only stores the lower triangle.

This is currently lacking all tests and some things still need to be implemented, but what's there now seem to be working in my scrap notebook. Some operations are slightly faster, some slightly slower than plain `SMatrix` depending on the operation, matrix size, and processor architecture. It's exciting to see that multiplying two `SSymmetricCompact{3, 3}`s (using the generic code in `matrix_multiply.jl`!) is slightly faster than multiplying regular `SMatrix{3, 3}`s. Overall I was slightly underwhelmed by the performance, but I still think this could be useful.

I was just wondering:

1. Do you think this belongs in StaticArrays itself? It's kind of like `SDiagonal` (which I used for inspiration), so maybe?
2. What do you think `SSymmetricCompact(::Tuple)` should do? I saw that for `SDiagonal`, it assumes that the `Tuple` is just the diagonal, which I thought was kind of strange given the use of the `Tuple` methods in `convert.jl`. It's also not how the Rotations.jl types do it, and I ended up following those examples.


andyferris:
Cool - I've always wanted to try this and see how it panned out!

There's no reason this *doesn't* belong here, but I guess there is a question as to what the killer application of this beastie is? I guess it could be useful for storing *lots* of little, Hermitian matrices in memory?

Also, we need to keep in mind that `transpose` is recursive.

> for SDiagonal, it assumes that the Tuple is just the diagonal, which I thought was kind of strange given the use of the Tuple methods in convert.jl

Oh dear, we should probably change that, the `StaticArray` interface should take precedence over a convenience method.

KristofferC:
In continuum mechanics, symmetric second order tensors is very common (we have something similar to this in Tensors.jl).

andyferris:
> In continuum mechanics, symmetric second order tensors is very common

So that would be a `SSymmetricCompact{3,3}` representing stress or strain or something at every grid point / finite element? Thus the space savings relative to many `Symmetric(::SMatrix{3,3})`s are substantial?

KristofferC:
Yes. In addition, the derivative of a symmetric second order tensors with another gives symmetric fourth order tensors where the storage saving is more substantial (36 vs 81 elements for symmetric fourth order vs unsymmetric).

tkoolen:
I was personally hoping to use this in RigidBodyDynamics to speed up operations involving moments of inertia.

Made a bit more progress. Next up is `eye`/`one` and tests.

> Also, we need to keep in mind that `transpose` is recursive.

Thanks for the heads up; fixed this. I did notice that this is actually currently not the case for `StaticMatrix`:
https://github.com/JuliaArrays/StaticArrays.jl/blob/a83cac63fd23b3a33fd2b3fe015b091c540e874e/src/linalg.jl#L58-L70

andyferris:
>  did notice that this is actually currently not the case for StaticMatrix

Yeah I think this was a design decision I was trying to ignore when we first wrote static arrays. :) We should fix that up.

I'm slightly worried that result here might be that the eltype is a `Union` because some elements are transposed and others not. (Same for `Symmetric` in `Base`).

tkoolen:
> I'm slightly worried that result here might be that the eltype is a Union because some elements are transposed and others not. (Same for Symmetric in Base).

Maybe I don't follow, but isn't that problem avoided here, since `SSymmetricCompact` only stores the lower triangle?
```julia
@inline transpose(a::SSymmetricCompact) = SSymmetricCompact(transpose.(a.lowertriangle))
```

tkoolen:
Alright, I think this is ready for review now.

Regarding the recursiveness of `transpose`: it seems to me that Base also doesn't do transpose recursively, since this test:
```julia
a = Symmetric([[rand(Complex{Int}) for i = 1 : 2] for row = 1 : 3, col = 1 : 3])
@test transpose(SSymmetricCompact{3}(a)) == transpose(a)
```
passes with the current code, but not with
```julia
transpose(a::SSymmetricCompact) = SSymmetricCompact((map(transpose, a.lowertriangle)))
```

andyferris:
> it seems to me that Base also doesn't do transpose recursively

Is that v0.6 or v0.7?

tkoolen:
0.6 and 0.7 both have (different) issues.

On 0.6.1:
```julia
julia> a = Symmetric([[Complex(row * i, col * i + 1) for i = 1 : 2] for row = 1 : 3, col = 1 : 3])
3×3 Symmetric{Array{Complex{Int64},1},Array{Array{Complex{Int64},1},2}}:
 Complex{Int64}[1+2im, 2+3im]  Complex{Int64}[1+3im, 2+5im]  Complex{Int64}[1+4im, 2+7im]
 Complex{Int64}[1+3im, 2+5im]  Complex{Int64}[2+3im, 4+5im]  Complex{Int64}[2+4im, 4+7im]
 Complex{Int64}[1+4im, 2+7im]  Complex{Int64}[2+4im, 4+7im]  Complex{Int64}[3+4im, 6+7im]

julia> transpose(a)[1]
2-element Array{Complex{Int64},1}:
 1+2im
 2+3im

julia> transpose(Array(a))[1]
ERROR: MethodError: Cannot `convert` an object of type RowVector{Complex{Int64},Array{Complex{Int64},1}} to an object of type Array{Complex{Int64},1}
This may have arisen from a call to the constructor Array{Complex{Int64},1}(...),
Stacktrace:
 [1] transpose_f!(::Base.#transpose, ::Array{Array{Complex{Int64},1},2}, ::Array{Array{Complex{Int64},1},2}) at ./linalg/transpose.jl:54
 [2] transpose(::Array{Array{Complex{Int64},1},2}) at ./linalg/transpose.jl:121
```

on 5-day old master:
```julia
julia> a = Symmetric([[Complex(row * i, col * i + 1) for i = 1 : 2] for row = 1 : 3, col = 1 : 3])
3×3 Symmetric{Array{Complex{Int64},1},Array{Array{Complex{Int64},1},2}}:
 [1+2im, 2+3im]  [1+3im, 2+5im]  [1+4im, 2+7im]
 [1+3im, 2+5im]  [2+3im, 4+5im]  [2+4im, 4+7im]
 [1+4im, 2+7im]  [2+4im, 4+7im]  [3+4im, 6+7im]

julia> transpose(a)[1]
2-element Array{Complex{Int64},1}:
 1 + 2im
 2 + 3im

julia> transpose(Array(a))[1]
1×2 Transpose{Complex{Int64},Array{Complex{Int64},1}}:
 1+2im  2+3im
```

andyferris:
@tkoolen thought I should let you know I'm trying to ignore this until all the v0.7 carnage is dealt with, but I am generally supportive of this addition.

timholy:
Also relevant for https://github.com/JuliaIO/NRRD.jl/issues/30

mschauer:
I think `SSymmetric` might be on the edge of still belonging to `StaticArrays.jl`. For more complicated objects matrices one can think of going down the route of https://github.com/mschauer/StaticLinearMaps.jl/blob/master/src/StaticLinearMaps.jl

tkoolen:
> For more complicated objects matrices one can think of going down the route of https://github.com/mschauer/StaticLinearMaps.jl/blob/master/src/StaticLinearMaps.jl

As in, in a separate package, perhaps. The `SDiagonal` precedent has lost some of its power since I opened this PR. But I don't think StaticLinearMaps itself is the right place, since the `SLinearMap` abstract type is not a `StaticMatrix` subtype. And in the end I do still think this belongs in StaticArrays, because it means not having to pay the overhead of maintaining another package and because of improved discoverability.

mschauer:
? I said (or meant to say) that it belongs here.

tkoolen:
Ah, I misunderstood then.

tkoolen:
I've brought this up to date and fixed some issues with recursive transposes and adjoints. @timholy, could I ask you to review this?

tkoolen:
By the way, if you're wondering about the efficiency of `getindex`, the compiler is really doing some magic here, especially for types where `transpose(x) = x`:

```julia
julia> sa = SSymmetricCompact(rand(SMatrix{5, 5}));

julia> f(sa, i) = @inbounds getindex(sa, i)

julia> @code_native f(sa, 1)
        .text
; ┌ @ REPL[23]:1 within `f'
; │┌ @ SSymmetricCompact.jl:82 within `getindex' @ REPL[23]:1
        shlq    $4, %rsi
; │└
; │┌ @ pair.jl:50 within `getindex'
        movabsq $139903100985504, %rax  # imm = 0x7F3DBAA320A0
; │└
; │┌ @ SSymmetricCompact.jl:83 within `getindex' @ SVector.jl:37 @ tuple.jl:24
        movq    -16(%rsi,%rax), %rax
; ││ @ SSymmetricCompact.jl:84 within `getindex'
        vmovsd  -8(%rdi,%rax,8), %xmm0  # xmm0 = mem[0],zero
; │└
        retq
        nopw    (%rax,%rax)
; └
```

I don't understand the trick the compiler is using yet, but since the native code for various matrix sizes only seems to differ in one constant (e.g. `139903100985504` above), I wonder if the underlying trick could be used to implement a fast `getindex(::Symmetric, i)` and change the `IndexStyle` for `Symmetric` to `IndexLinear`.

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,239 @@
+struct SSymmetricCompact{N, T, L} <: StaticMatrix{N, N, T}
```
timholy: Needs docs about the storage order or `lowertriangle`.

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,239 @@
...
+    lowertriangle::SVector{L, T}
+
+    @inline function SSymmetricCompact{N, T, L}(lowertriangle::SVector{L}) where {N, T, L}
+        check_symmetric_parameters(Val(N), Val(L))
+        new{N, T, L}(lowertriangle)
+    end
+end
+
+@generated function check_symmetric_parameters(::Val{N}, ::Val{L}) where {N, L}
```
timholy: Does this really have to be generated?
```julia
julia> csp(::Val{N}, ::Val{L}) where {N,L} = 2L == N*(N+1)
csp (generic function with 1 method)

julia> csp(vn, vl)
false

julia> @code_llvm csp(vn, vl)

;  @ REPL[6]:1 within `csp'
define i8 @julia_csp_12412() {
top:
  ret i8 0
}
```

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,239 @@
...
+
+@inline (::Type{SSymmetricCompact{N, T}})(lowertriangle::SVector{L}) where {N, T, L} = SSymmetricCompact{N, T, L}(lowertriangle)
+@inline (::Type{SSymmetricCompact{N}})(lowertriangle::SVector{L, T}) where {N, T, L} = SSymmetricCompact{N, T, L}(lowertriangle)
+
+@inline function SSymmetricCompact(lowertriangle::SVector{L, T}) where {T, L}
+    N = triangularroot(L)
+    SSymmetricCompact{N, T, L}(lowertriangle)
+end
+
+@generated function (::Type{SSymmetricCompact{N, T, L}})(a::Tuple) where {N, T, L}
```
timholy: What size should `a` have? Because of your formula for `index`, it seems it's of length N^2? This needs to be checked.

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,239 @@
...
+end
+
+@inline function (::Type{SSymmetricCompact{N, T}})(a::Tuple) where {N, T}
+    L = triangularnumber(N)
+    SSymmetricCompact{N, T, L}(a)
+end
+
+@inline (::Type{SSymmetricCompact{N}})(a::Tuple) where {N} = SSymmetricCompact{N, promote_tuple_eltype(a)}(a)
+@inline (::Type{SSymmetricCompact{N}})(a::NTuple{M, T}) where {N, T, M} = SSymmetricCompact{N, T}(a)
+@inline SSymmetricCompact(a::StaticMatrix{N, N, T}) where {N, T} = SSymmetricCompact{N, T}(a)
```
timholy: Document that only the lower triangle is used.

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,239 @@
...
+            indexmat[col, row][1] => false
+        end
+    end
+    quote
+        Base.@_inline_meta
+        return $(tuple(indexmat...))
+    end
+end
+
+Base.@propagate_inbounds function Base.getindex(a::SSymmetricCompact{N}, i::Int) where {N}
```
timholy: Why implement this rather than `getindex(a, i::Int, j::Int)` and make it `IndexCartesian`?

tkoolen: The main reasons are that `getindex(a, i::Int, j::Int)` would also not be trivial, requiring a solution similar to `_symmetric_compact_indices` to be fast, and a lot of functionality inside of StaticArrays implicitly assumes that all `StaticArray`s are `IndexLinear`, e.g.:

https://github.com/JuliaArrays/StaticArrays.jl/blob/407c65fb2ae652a439d09d82edf1a22eaab58dc6/src/mapreduce.jl#L34-L35

src/SSymmetricCompact.jl
tkoolen: Added docs.

src/SSymmetricCompact.jl
tkoolen: You're right, not necessary, fixed.

src/SSymmetricCompact.jl
tkoolen: This is now documented in the `SSymmetricCompact` docstring.

src/SSymmetricCompact.jl
tkoolen: This is now documented in the `SSymmetricCompact` docstring.

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,269 @@
...
+* `N`: matrix dimension;
+* `T`: element type for lower triangle;
+* `L`: length of the `SVector` storing the lower triangular elements.
+
+Note that `L` is always the `N`th [triangular number](https://en.wikipedia.org/wiki/Triangular_number).
+
+An `SSymmetricCompact` may be constructed either:
+
+* from an `AbstractVector` containing the lower triangular elements; or
+* from a `Tuple` containing both upper and lower triangular elements (in column major order); or
```
c42f: So I guess this inconsistency between `Tuple` and `AbstractVector` is due to the rest of `StaticArrays` assuming dense storage? That's unfortunate.

Does this `Tuple` constructor ignoring the top half, combined with the `similar_type` machinery mean that a lot of `StaticArrays` functionality is actually broken with `SSymmetricCompact`? For example, matrix multiply of `SSymmetricCompact` with `SMatrix`?

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,269 @@
...
+@inline (::Type{SSymmetricCompact{N}})(a::Tuple) where {N} = SSymmetricCompact{N, promote_tuple_eltype(a)}(a)
+@inline (::Type{SSymmetricCompact{N}})(a::NTuple{M, T}) where {N, T, M} = SSymmetricCompact{N, T}(a)
+@inline SSymmetricCompact(a::StaticMatrix{N, N, T}) where {N, T} = SSymmetricCompact{N, T}(a)
+
+@inline (::Type{SSC})(a::SSymmetricCompact) where {SSC <: SSymmetricCompact} = SSC(a.lowertriangle)
+@inline (::Type{SSC})(a::SSC) where {SSC <: SSymmetricCompact} = a
+
+@inline (::Type{SSC})(a::AbstractVector) where {SSC <: SSymmetricCompact} = SSC(convert(lowertriangletype(SSC), a))
+
+# TODO: similar_type overload?
```
c42f: I think this might be required, and you'll have to fall back to `SMatrix` for correctness? Unless we introduced a way to handle symmetry-preserving operations.

src/SSymmetricCompact.jl
tkoolen: It's a bit of a pickle (see https://github.com/JuliaArrays/StaticArrays.jl/pull/358#issue-162781070, item 2), but I think the current implementation still makes sense.

Any `StaticArray` definitely needs to be constructable from a `Tuple`, which needs to have the interpretation of the full list of column-major-ordered elements.; that's the basic assumption in `convert.jl`. That also means we can't use `Tuple` as the type of the `lowertriangle` field without causing all kinds of confusion with the constructors. So I chose to use an `SVector` as the storage type for the lower triangular elements and designed the constructors in accordance. The precedent for this was `SDiagonal` (which has since been replaced with `Symmetric{T, SVector{<:Any, T}`; maybe the same will happen for this case in time). 

Note that the `StaticMatrix` subtype constructors / `convert` methods currently also have the job of implementing `reshape`,

https://github.com/JuliaArrays/StaticArrays.jl/blob/407c65fb2ae652a439d09d82edf1a22eaab58dc6/src/convert.jl#L9-L10

which I'm not 100% on board with anyway. `SSymmetricCompact` hijacks this particular use case, but I'd say it's niche enough that it's OK. Alternatively, a new type could be added precisely for this purpose, but I don't think there's much of an advantage in doing so.

> Does this Tuple constructor ignoring the top half, combined with the similar_type machinery mean that a lot of StaticArrays functionality is actually broken with SSymmetricCompact? For example, matrix multiply of SSymmetricCompact with SMatrix?

No. `similar_type` is not overloaded for `SSymmetricCompact`, so the default `SMatrix` is used. So in that sense, `SSymetricCompact` is very similar to e.g. `Rotations.Quat`, and so the only thing that really matters is that `getindex(::StaticArray, i::Int)` works.

src/SSymmetricCompact.jl
tkoolen: `SMatrix` is the default. I should probably just remove this comment instead.

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,239 @@
...
+    quote
+        Base.@_inline_meta
+        return $(tuple(indexmat...))
+    end
+end
+
+Base.@propagate_inbounds function Base.getindex(a::SSymmetricCompact{N}, i::Int) where {N}
+    I = _symmetric_compact_indices(Val(N))
+    j, lower = I[i]
+    @inbounds x′ = a.lowertriangle[j]
```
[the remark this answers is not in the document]
tkoolen: Precedent for this was

https://github.com/JuliaLang/julia/blob/5b637df34396034b0dd353e603ab3d61322369fb/base/broadcast.jl#L823

Suggestions for another name?

src/SSymmetricCompact.jl
timholy: Regarding the nontrivial, doesn't
```julia
i, j = i >= j ? (i, j) : (j, i)
idx = i + (j-1)*N
```
do the job?

If you've not benchmarked it or checked generated code, this seems to be worth doing. But most likely you have and you know that it's all fine.

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,269 @@
...
+    end
+end
+
+LinearAlgebra.ishermitian(a::SSymmetricCompact{N, T}) where {N,T <: Real} = true
+LinearAlgebra.ishermitian(a::SSymmetricCompact) = all(isreal, a.lowertriangle)
+LinearAlgebra.issymmetric(a::SSymmetricCompact) = true
+
+# TODO: factorize?
+
+# TODO: a.lowertriangle == b.lowertriangle is slow (used by SDiagonal). SMatrix etc. actually use AbstractArray fallback (also slow)
```
mschauer: Should/does this have an issue?

src/SSymmetricCompact.jl
```diff
@@ -0,0 +1,269 @@
...
+    end
+end
+
+@generated function LinearAlgebra.transpose(a::SSymmetricCompact{N, T, L}) where {N, T, L}
+    # For M = [A Bᵀ
+    #          B C],
+    # Mᵀ = [Aᵀ Bᵀ
+    #       B  Cᵀ]
+    # In general, we should only recursively transpose the diagonal elements.
+    exprs = Vector{Expr}(undef, L)
```
mschauer: I don't get this comment

src/SSymmetricCompact.jl
tkoolen: No, that's not quite right.

```julia
julia> N = 3; L = StaticArrays.triangularnumber(N)
6

julia> m = SSymmetricCompact(SVector{L}(1 : L))
3×3 SSymmetricCompact{3,Int64,6}:
 1  2  3
 2  4  5
 3  5  6

julia> i, j = 2, 3;

julia> m[i, j]
5

julia> i, j = i >= j ? (i, j) : (j, i)
(3, 2)

julia> idx = i + (j-1)*N
6

julia> m.lowertriangle[idx]
6
```

src/SSymmetricCompact.jl
tkoolen: One might think that you should transpose all of the elements of `a.lowertriangle` (which might e.g. be of type `Matrix{ComplexF64}`) to match the recursive property of `transpose` in the `LinearAlgebra` implementation, as I did initially. But that's not the case. This comment is trying to explain that. Suggestions to make it clearer?

src/SSymmetricCompact.jl
tkoolen: The speed of the `AbstractArray` fallback seems to no longer be an issue. I'll change this accordingly.

src/SSymmetricCompact.jl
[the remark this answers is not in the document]
c42f: Yeah, I know people sometimes use `\prime`. My point is only this made me have to stare and think about what `adjoint` was doing on the left hand side (if only for twenty seconds; it would have been longer the first time).

I think that's an indication we should use another name; `x` would be fine?

src/SSymmetricCompact.jl
c42f: Oh sure, sorry my bad!

src/SSymmetricCompact.jl
c42f: I had to think about this myself when I looked through the code yesterday.
It seems mysterious, but it's only because a symmetric block matrix may not be a symmetric matrix when it's flattened.

```
# Transposition of a symmetric matrix is the identity.
# However, if we happen to be storing a block matrix there may be
# non-symmetric diagonal parts so the blocks on the diagonal need
# transposing:
# [A  Bᵀ] ᵀ  =  [Aᵀ  Bᵀ]  =  [Aᵀ Bᵀ]
# [B  C ]       [Bᵀᵀ Cᵀ]     [B  Cᵀ]
```

src/SSymmetricCompact.jl
timholy: D'oh!  Try
```julia
idx = i - j + 1 + ((j-1)*(2N+2-j)÷2)
```
The point being: your implementation *looks* slow, though perhaps it's not.

tkoolen: That's not quite right either I'm afraid:

```julia
```

As I showed in https://github.com/JuliaArrays/StaticArrays.jl/pull/358#issuecomment-488432426, the compiler does a very good job on the whole `_symmetric_compact_indices` thing. Basically, it computes that tuple at compile time for a given `N`, and then the `getindex` method simply indexes into that constant tuple.

src/SSymmetricCompact.jl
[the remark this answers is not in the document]
tkoolen: I wanted symmetry with `setindex` where `x` is an argument. Changed to `value`.

src/SSymmetricCompact.jl
tkoolen: Done.

src/SSymmetricCompact.jl
tkoolen: Oh wait, I should still add the `i, j = i >= j ? (i, j) : (j, i)`; it works then. I'll benchmark.

chriscoey:
possibly relevant comment: https://github.com/JuliaLang/julia/pull/31836#issuecomment-489093440

src/SSymmetricCompact.jl
tkoolen: Alright, with

```julia
Base.IndexStyle(::Type{<:SSymmetricCompact}) = Base.IndexCartesian()

Base.@propagate_inbounds function Base.getindex(a::SSymmetricCompact{N}, i::Int) where N
    # override StaticArray overload, replace with AbstractArray version.
    # Base._getindex(IndexStyle(a), a, Base.to_indices(a, (i,))...)
    c, r = divrem(i - 1, N)
    c, r = c + 1, r + 1
    a[r, c]
end

Base.@propagate_inbounds function Base.getindex(a::SSymmetricCompact{N}, i::Int, j::Int) where N
    i, j = i >= j ? (i, j) : (j, i)
    idx = i - j + 1 + ((j-1)*(2N+2-j)÷2)
    x = a.lowertriangle[idx]
end
```

we get

```julia
julia> @btime a[i] setup = begin
           a = SSymmetricCompact(rand(SMatrix{6, 6}))
           i = rand(1 : length(a))
       end
  3.272 ns (0 allocations: 0 bytes)
```

whereas before,

```julia
julia> @btime a[i] setup = begin
           a = SSymmetricCompact(rand(SMatrix{6, 6}))
           i = rand(1 : length(a))
       end
  1.783 ns (0 allocations: 0 bytes)
```

Do note that multiplying two `SSymmetricCompact`s results in the same runtime before and after this change, as the indices are known at compile time and so it looks like the compiler is able to constant-propagate everything in either case. But just randomly linearly indexing into an `SSymmetricCompact` is faster with the `_symmetric_compact_indices` approach.

timholy: Right, my proposal is really for `IndexCartesian`, where you never have to call `divrem` (which is insanely slow).

But I just benchmarked with
```julia
Base.@propagate_inbounds function Base.getindex(a::SSymmetricCompact{N}, i::Int, j::Int) where {N}
    i, j, lower = j > i ? (j, i, false) : (i, j, true)
    idx = i - j + 1 + ((j-1)*(2N+2-j))>>UInt(1)
    @inbounds value = a.lowertriangle[idx]
    return lower ? value : transpose(value)
end
```
and to my surprise it's no better. So I'm happy going with yours.

tkoolen: For constant `y` the compiler does a very good job on `div(x, y)` and `divrem(x, y)`, almost to the point where it produces the same code as the manual `>> UInt(1)` optimization:

```julia
julia> f(x) = div(x, 2);

julia> g(x) = x >> UInt(1);

julia> @code_native f(3)
        .text
; ┌ @ REPL[15]:1 within `f'
; │┌ @ REPL[15]:1 within `div'
        movq    %rdi, %rax
        shrq    $63, %rax
        leaq    (%rax,%rdi), %rax
        sarq    %rax
; │└
        retq
        nop
; └

julia> @code_native g(3)
        .text
; ┌ @ REPL[16]:1 within `g'
; │┌ @ REPL[16]:1 within `>>'
        sarq    %rdi
; │└
        movq    %rdi, %rax
        retq
        nopw    (%rax,%rax)
; └
```

Similarly for `divrem(i, N)` for fixed `N`.

tkoolen:
I'd be fine with turning this into `SHermitianCompact`.

tkoolen:
OK, changed to `SHermitianCompact`. I think this is good to go now. I'll leave this open for another day or so and then I'll merge absent additional comments.

tkoolen:
Thanks, all!

PR #362, opened 2018-01-21
Catch up on v0.7

andyferris:
Many miscellaneous changes to make package compatible with v0.7. Includes @pablosanjose's work as a subset (thanks!).

Still WIP.


andyferris:
OK, I get the feeling this will never be finished. At least v0.7 "works" (the tests don't all pass, but, you know). If all the v0.6 tests pass CI I may just merge this and continue getting the v0.7 tests to pass in smaller pieces of work.

andyferris:
Finally found a few minutes to get back to this.

Progress is slow. I think (hope) v0.6 tests will pass.

I really don't think it will be easy to get *everything* working on v0.7 without dropping v0.6 support... :(

PR #365, opened 2018-02-09
Fix indexing bug in `mul_unrolled_chunks!`

martinholters:


martinholters:
Bump.

The bug addressed herein can silently cause wrong results, so would be nice to have it fixed ASAP.

andyferris:
Thanks! Sorry for the delay.

PR #366, opened 2018-02-12
Fixes for mul! on Julia 0.7

martinholters:
The first commit fixes deprecations in `ccall` to BLAS `gemm`. Without it, I'm getting a segfault---might be some problem with the deprecations in Julia?

The second commit imports `LinearAlgebra.mul!` so that the definitions extend it instead of defining a new function.


andyferris:
Thanks!

PR #368, opened 2018-02-16
Non-allocating LU decomposition

martinholters:
This is based on `generic_lufact!` from base, but using recursion instead of a `for` loop. It took me a few iterations of refinement, but the resulting code is satisfyingly clean. (Suggestions for a non-`@generated` alternative to `tailindices` welcome!)

For square matrices, it beats `lu(::Matrix)` up to a size of about 18×18. For sizes 2×2, 3×3, and 4×4, it provides a speedup of about 30, 15, and 9, respectively. At 25×25, it is slower by about a factor of two. (That's on Julia 0.6. Allocation of the input `Matrix` for `lu(::Matrix)` was not included in the benchmark, so comparison to the current implementation should be even a bit more favorable.)

Downside is significant compilation time. The recursion is completely inlined, so the functions can get really huge:
```julia
julia> m=20; n=20; A = SMatrix{m,n}(rand(m,n)); @time lu(A); # first call to lu
 45.902073 seconds (57.71 M allocations: 1.940 GiB, 1.17% gc time)
```

The logical next steps are to now use `lu` in `inv` and `solve`, but as this is my first relevant contribution here, I preferred to submit this as a first (and in itself useful) step to get some feedback.

### DIFF
--- a/src/lu.jl
+++ b/src/lu.jl
@@ -1,29 +1,105 @@
 # LU decomposition
 function lu(A::StaticMatrix, pivot::Union{Type{Val{false}},Type{Val{true}}}=Val{true})
-    L,U,p = _lu(Size(A), A, pivot)
+    L,U,p = _lu(A, pivot)
     (L,U,p)
 end
 
 # For the square version, return explicit lower and upper triangular matrices.
 # We would do this for the rectangular case too, but Base doesn't support that.
 function lu(A::StaticMatrix{N,N}, pivot::Union{Type{Val{false}},Type{Val{true}}}=Val{true}) where {N}
-    L,U,p = _lu(Size(A), A, pivot)
+    L,U,p = _lu(A, pivot)
     (LowerTriangular(L), UpperTriangular(U), p)
 end
 
+_lu(A::StaticMatrix{0,0,T}, ::Type{Val{Pivot}}) where {T,Pivot} =
+    (SMatrix{0,0,typeof(one(T))}(), A, SVector{0,Int}())
 
-@inline function _lu(::Size{S}, A::StaticMatrix, pivot) where {S}
-    # For now, just call through to Base.
-    # TODO: statically sized LU without allocations!
-    f = lufact(Matrix(A), pivot)
-    T = eltype(A)
-    # Trick to get the output eltype - can't rely on the result of f[:L] as
-    # it's not type inferrable.
-    T2 = arithmetic_closure(T)
-    L = similar_type(A, T2, Size(Size(A)[1], diagsize(A)))(f[:L])
-    U = similar_type(A, T2, Size(diagsize(A), Size(A)[2]))(f[:U])
-    p = similar_type(A, Int, Size(Size(A)[1]))(f[:p])
-    (L,U,p)
+_lu(A::StaticMatrix{0,1,T}, ::Type{Val{Pivot}}) where {T,Pivot} =
+    (SMatrix{0,0,typeof(one(T))}(), A, SVector{0,Int}())
+
+_lu(A::StaticMatrix{0,N,T}, ::Type{Val{Pivot}}) where {T,N,Pivot} =
+    (SMatrix{0,0,typeof(one(T))}(), A, SVector{0,Int}())
+
+_lu(A::StaticMatrix{1,0,T}, ::Type{Val{Pivot}}) where {T,Pivot} =
+    (SMatrix{1,0,typeof(one(T))}(), SMatrix{0,0,T}(), SVector{1,Int}(1))
+
+_lu(A::StaticMatrix{M,0,T}, ::Type{Val{Pivot}}) where {T,M,Pivot} =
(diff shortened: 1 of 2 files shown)


martinholters:
This works on Julia 0.7 in principle, but is hitting https://github.com/JuliaLang/julia/issues/26083 😢, so one has to trigger inference of `lu` for increasing sizes before actually using it. Hopefully, that will be resolved in time and we don't have to look for a workaround...

Oh, and once I get a general OK for the implementation here, I'll add some more tests, too.

martinholters:
Adding `@inbounds` helped shaving off a bit off the execution time; less than 25%, but why not. While at it, I've added tests that should exercise all code paths.

andyferris:
Sorry for the lack of response. This is really cool! :)

I feel that we need to put in a cut-off that calls the `Base` method for lu factorization at some size cutoff. We give up when any dimension grows larger than 14 for matrix multiplication; I'd suggest implementing a similar thing here since O(N^3) code is Not A Good Thing (we *tolerate* completely unrolling proportional the number of elements of the object, but beyond that gets silly fast).

After that, let's merge! (The `tailindices` thing seems quite reasonable compared to the rest of this package!)

martinholters:
> I feel that we need to put in a cut-off that calls the `Base` method for lu factorization at some size cutoff. We give up when any dimension grows larger than 14 for matrix multiplication

AFAICT, you do this for `mul!`, but not for `*`, which makes sense because at least `mul!(::MMatrix, ::MMatrix, ::MMatrix)` will stay non-allocating even when calling out to BLAS. OTC, `*` seems not to allocate even for large matrices.

Also, it looks like the code size is not the problem in itself. (A `@noinline` on the recursive function doesn't help). But consider this:
```julia
julia> N=20; t=zeros(N); for n=1:N; A=SMatrix{n,n}(rand(n,n)); t[n] = @elapsed lu(A); end

julia> t
20-element Array{Float64,1}:
 0.0365296
 1.50551  
 0.462231 
 0.498421 
 0.562395 
 0.622763 
 0.708565 
 0.837337 
 0.968672 
 1.15836  
 1.39305  
 1.67774  
 2.02063  
 2.44261  
 2.91981  
 3.54015  
 4.39871  
 5.07938  
 6.17949  
 7.11707  

julia> sum(t)
44.129436268
```
So the time to code-gen the 20-by-20 case directly (see timing above) is approximately equal to the cumulative time for code-gen'ing all recursively called functions. I take this as a strong indicator that the time is dominated by inference.

Now look at this (in a fresh session):
```julia
julia> N=20; t=zeros(N); for n=1:N; ps=StaticArrays.tailindices(Val{n}); A=SMatrix{n,n}(rand(n,n)); t[n] = @elapsed A[ps,ps]; end

julia> t
20-element Array{Float64,1}:
 3.492e-6 
 0.150532 
 0.0683794
 0.0759533
 0.089354 
 0.107017 
 0.139573 
 0.171334 
 0.220912 
 0.284708 
 0.376776 
 0.46346  
 0.58994  
 0.752672 
 0.950159 
 1.18343  
 1.45453  
 1.79486  
 2.1965   
 2.6604   

julia> sum(t)
13.730489121999998
```

So almost 30% of the inference time is spent for `getindex(::SMatrix{N,N}, ::SVector{N-1,Int}, ::SVector{N-1,Int})`. And yes, the timings for `lu` after forcing inference of the `getindex` calls go down by almost that amount. The `vcat` calls similarly seem to contribute another 20%. So if we could make these easier on inference, that could already go a long way.

That all said, limiting the non-allocating LU implementation to say N<15 seems the best short-term solution (and it certainly doesn't make things worse compared to master), so I'll do that for now.

> The tailindices thing seems quite reasonable compared to the rest of this package!

Amusingly, `SVector(Base.tail(ntuple(identity, Val{M})))` works just as well for `M<15`. But I'll leave the generated version in in case it used (here or elsewhere) with larger `M` in the future.

martinholters:
> Perhaps a comment in the code on the reason that tailindices exists would be worthwhile.

Done.

andyferris:
I'm happy!

martinholters:
With this patch, it also doesn't crash on Julia 0.7:
```diff
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -15,7 +15,7 @@ Construct a statically-sized array of dimensions `S` (expressed as a `Tuple{...}
 the data from `a`. The `S` parameter is mandatory since the size of `a` is unknown to the
 compiler (the element type may optionally also be specified).
 """
-struct SArray{S <: Tuple, T, N, L} <: StaticArray{S, T, N}
+struct SArray{S, T, N, L} <: StaticArray{S, T, N}
     data::NTuple{L,T}
 
     function SArray{S, T, N, L}(x::NTuple{L,T}) where {S, T, N, L}
```
However, inference gives up on the recursion. This would be fixed by https://github.com/JuliaLang/julia/pull/26172; if that is rejected, I have two ideas for dirty tricks to work around it, but I hope we can do without them.

PR #370, opened 2018-02-23
overload Base.setindex

jw3126:
There is already a `setindex` function in `Base`, we should overload it.
It would allow to remove [this hack](https://github.com/jw3126/Setfield.jl/blob/41afdcf7e63d2f46a173661387049a31df730f4c/src/Setfield.jl).


jw3126:
I think CI fail is unrelated.

PR #374, opened 2018-03-02
Import `eye` from `LinearAlgebra` on julia 0.7

garrison:
This was necessary for my code to work on julia master, since `eye` has been moved to `LinearAlgebra`.

Of course, it leaves me wondering whether any other imports have been missed.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -35,7 +35,7 @@ else
     import Base: sqrt, exp, log
 
     using LinearAlgebra
-    import LinearAlgebra: transpose, adjoint, vecdot, eig, eigvals, eigfact, lyap, trace,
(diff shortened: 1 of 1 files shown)


andyferris:
Yeah, OK, I certainly did leave `eye` in a mess! We need to do this import and add *deprecated* methods to it in v0.7. Did you have the patience to have a go at this? I think we want to target something like `SMatrix{3,3,Float64}(I)` as the new syntax. (See e.g. the code at https://github.com/JuliaArrays/StaticArrays.jl/blob/master/src/linalg.jl#L161-L175)

garrison:
thanks for the reply!

> Did you have the patience to have a go at this?

unlikely I will get to this anytime soon.

PR #376, opened 2018-03-06
Add `Base.dataids(::SArray)`

martinholters:
An `SArray` can never alias anything, so it is safe to return an empty tuple here and prevent the aliasing detection machinery from doing unnecessary work.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -110,6 +110,10 @@ end
 
 @inline Tuple(v::MArray) = v.data
 
+if isdefined(Base, :dataids) # v0.7-
+    Base.dataids(ma::MArray) = (UInt(pointer(ma)),)
+end
+
 @inline function Base.unsafe_convert(::Type{Ptr{T}}, a::MArray{S,T}) where {S,T}
     Base.unsafe_convert(Ptr{T}, pointer_from_objref(a))
 end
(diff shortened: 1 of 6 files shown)


martinholters:
I'm a bit unsure about a good way to test this.
```julia
if isdefined(Base, :dataids) # v0.7-
    @test Base.dataids(SVector((rand(10)...,))) === ()
end
```
seems a bit tautological. OTOH, `Base.unalias(a, b::SArray) === b` should be true no matter the definition of `dataids`. Opinions?

martinholters:
> What should we do about `MArray` and `SizedArray`?

Good point. The default is to use the `objectid`. For a mutable type (`MArray`) this is its hashed memory address. That's ok because two `MArrays` alias if and only if they are identical, and they cannot alias any other objects, unless these wrap them. In that case, the wrapping object is responsible for including the wrapped objects `dataids` in its own `dataids`. (Although it looks like this is at present only the case for a subset of the wrappers in Base.) However, calling `pointer` directly should work equally well and seems to be slightly faster.

OTOH, `SizedArray` indeed should provide a specialized `dataids`, returning the `dataids` of the internal backing `Array`.

I'll update accordingly and add some tests.

martinholters:
The new tests all passed on 0.7 before later tests caused the known failures.

andyferris:
Thanks :)

PR #377, opened 2018-03-07
RFC: Fix "Indexing empty arrays" tests for Julia 0.7

martinholters:
Creating an empty `SArray` without specifying the element type leads to an element type of `Union{}` on Julia 0.6, but fails on Julia 0.7. Hence, the tests now explicitly define their element type.

Alternatively (hence RFC), we could define constructors for the empty case that default to an element type of `Union{}` to keep the 0.6 behavior on 0.7. Not sure how useful/desirable that is, but it would allow e.g. `SVector(x...)` to work even if `x` happens to be empty.

### DIFF
--- a/test/indexing.jl
+++ b/test/indexing.jl
@@ -154,12 +154,12 @@
     end
 
     @testset "Indexing empty arrays" begin
-        @test size(SVector{0}()[:]) == (0,)
-        @test size(SMatrix{0,0}()[:,:]) == (0,0)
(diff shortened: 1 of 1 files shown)


andyferris:
Let's merge this for now. Since we support `push` we should probably support empty vectors well but if people have to specify the element type (even if it's not strictly necessary to infer the output of `push` or other functions) so be it.

PR #378, opened 2018-03-07
Improve/fix multiplication involving one AbstractTriangular{<:Any,<:StaticMatrix}

martinholters:
Ensure that all tested multiplications involving one `AbstractTriangular{<:Any,<:StaticMatrix}` return an `SArray` (wrapped as necessary for row vectors) and allow non-zero imaginary part in tests, both on Julia 0.6 and 0.7.

For the test to pass on 0.7, https://github.com/JuliaLang/julia/pull/26349 is needed (otherwise computation of the `Array`-based reference results fails). Further, for `Pkg.test` to even reach the relevant testsets, #377 is required. But with those, things are looking good locally.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -56,7 +56,11 @@ end
 # Transpose, conjugate, etc
 @inline conj(a::StaticArray) = map(conj, a)
 @inline transpose(m::StaticMatrix) = _transpose(Size(m), m)
-# note: transpose of StaticVector is a Adjoint, handled by Base
+# note: transpose of StaticVector is a Transpose, handled by Base
+if VERSION >= v"0.7-"
+    @inline transpose(a::Adjoint{<:Any,<:Union{StaticVector,StaticMatrix}}) = conj(a.parent)
+    @inline transpose(a::Adjoint{<:Real,<:Union{StaticVector,StaticMatrix}}) = a.parent
+end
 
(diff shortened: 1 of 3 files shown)


src/linalg.jl
```diff
@@ -56,7 +56,11 @@ end
 # Transpose, conjugate, etc
 @inline conj(a::StaticArray) = map(conj, a)
 @inline transpose(m::StaticMatrix) = _transpose(Size(m), m)
-# note: transpose of StaticVector is a Adjoint, handled by Base
+# note: transpose of StaticVector is a Transpose, handled by Base
+if VERSION >= v"0.7-"
+    @inline transpose(a::Adjoint{<:Any,<:Union{StaticVector,StaticMatrix}}) = conj(a.parent)
```
andyferris: Slightly surprised that this isn't done in `Base`. @mbauman?

martinholters: `conj` is eager, while `transpose` and `adjoint` are generally lazy. (Yes, StaticArrays should probably also switch to supporting lazy `transpose`/`adjoint`, but let's take one step at the time).

andyferris: That’s right, I remember now. I did try to sneak in lazy `conj` but it never happened.

andyferris:
This is ready to go?

martinholters:
Yes from my side.

PR #379, opened 2018-03-12
Replace `uninitialized` with `undef`

martinholters:
Not the most pressing issue for 0.7, but one with a simple solution.


andyferris:
Thanks!

PR #381, opened 2018-03-13
Use `LinearIndices` instead of `sub2ind`

martinholters:
Mostly mechanical changes.

I did some very basic benchmarking of `sub2ind(s, i, j)` vs. `LinearIndices(s)[i,j]` and it looks like the latter on 0.7 is indeed equal to the former on 0.6 performance-wise, but `Compat.LinearIndices` (on 0.6) seems to be a tiny bit slower. Fortunately, these are almost only used in `@generated` function generators where performance is less of a concern. The only exception is `setindex` where I have introduced a `VERSION` dependency to be on the safe (performance) side.

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -86,4 +86,8 @@ import Base: setindex
 end
 
 # TODO proper multidimension boundscheck
-@propagate_inbounds setindex(a::StaticArray, x, inds::Int...) = setindex(a, x, sub2ind(size(typeof(a)), inds...))
+if VERSION < v"0.7-"
+    @propagate_inbounds setindex(a::StaticArray, x, inds::Int...) = setindex(a, x, sub2ind(size(typeof(a)), inds...))
+else
+    @propagate_inbounds setindex(a::StaticArray, x, inds::Int...) = setindex(a, x, LinearIndices(a)[inds...])
+end
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -84,7 +84,13 @@ Length(::Type{SA}) where {SA <: StaticArray} = Length(Size(SA))
 
 @pure Base.prod(::Size{S}) where {S} = prod(S)
 
-@pure @inline Base.sub2ind(::Size{S}, x::Int...) where {S} = sub2ind(S, x...)
+if !isdefined(Base, :LinearIndices) # VERSION < v"0.7-"
+    @pure @inline Base.sub2ind(::Size{S}, x::Int...) where {S} = sub2ind(S, x...)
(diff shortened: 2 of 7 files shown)


src/traits.jl
```diff
@@ -84,7 +84,13 @@ Length(::Type{SA}) where {SA <: StaticArray} = Length(Size(SA))
 
 @pure Base.prod(::Size{S}) where {S} = prod(S)
 
-@pure @inline Base.sub2ind(::Size{S}, x::Int...) where {S} = sub2ind(S, x...)
+if !isdefined(Base, :LinearIndices) # VERSION < v"0.7-"
+    @pure @inline Base.sub2ind(::Size{S}, x::Int...) where {S} = sub2ind(S, x...)
```
martinholters: Why does this need to be `@pure`? The `LinearIndices` alternative obviously isn't, but there seems to be no code relying on this to be evaluated at inference time, or did I miss something?

andyferris: I see... mostly this lives in code generators rather than relying on code inference?

At some stage I had the intention of delegating many more things to `Size` and static indices (this would be: static ranges, static Cartesian ranges, and a static version of `LinearIndices`) and having less generated functions. However I didn't get very far. Feel free to simplify if you can prove to yourself that its not needed.

src/deque.jl
```diff
@@ -86,4 +86,8 @@ import Base: setindex
 end
 
 # TODO proper multidimension boundscheck
-@propagate_inbounds setindex(a::StaticArray, x, inds::Int...) = setindex(a, x, sub2ind(size(typeof(a)), inds...))
+if VERSION < v"0.7-"
+    @propagate_inbounds setindex(a::StaticArray, x, inds::Int...) = setindex(a, x, sub2ind(size(typeof(a)), inds...))
```
andyferris: Is this more performant than the below on v0.6?

andyferris:
Thank's a lot for this. One step closer to v0.7 goodness :)

src/deque.jl
andyferris: Nevermind, you talk about this in the OP

src/traits.jl
martinholters: I think we can just leave this as is for the time being. I was only concerned whether the non-`@pure` version for 0.7 might cause any troubles, but the answer seems to be no.

martinholters:
I hope I can continue to work on this. I'll probably open more of these small PRs as time allows. My idea is to get tests to pass again (and fix deprecations along the way), and worry about adjusting the API to 0.7 later (e.g. lazy `adjoint`).

PR #387, opened 2018-03-22
Fix #385, broadcast! with non-static LHS broken on Julia master 

### LINKED ISSUE #385: broadcast! with non-static LHS broken on Julia master
```julia
julia> using StaticArrays

julia> a = zeros(3, 3)
3×3 Array{Float64,2}:
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> b = ones(SMatrix{3, 3})
 1.0  1.0  1.0
 1.0  1.0  1.0
 1.0  1.0  1.0

julia> a .= b
┌ Warning: Deprecated syntax `using the value of `.=``.
└ @ nothing none:0
ERROR: MethodError: Cannot `convert` an object of type Array{Float64,2} to an object of type Size
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:123
Stacktrace:
 [1] Size(::Array{Float64,2}) at ./deprecated.jl:472
 [2] broadcast! at /Users/twan/code/julia/RigidBodyDynamics/v0.7/StaticArrays/src/broadcast.jl:69 [inlined]
 [3] broadcast!(::typeof(identity), ::Array{Float64,2}, ::SArray{Tuple{3,3},Float64,2,9}) at ./broadcast.jl:444
 [4] top-level scope
```

tkoolen:
Fix #385.

* extract `broadcast_size` function out of `broadcast_sizes`, which returns a single `Size` for a single ~~`AbstractArray`~~ argument
* extract `combine_sizes` function (similar to `Base.Broadcast.combine_indices`) out of `_broadcast`, to determine the appropriate destination `Size` from the right hand side arguments (`as...`)
~~* add `broadcast_dest_size` function, used by `broadcast!`, which selects the correct `Size` and checks that sizes match, given `broadcast_size(dest)` and `combine_sizes(broadcast_sizes(as...))`~~

~~On 0.6, `broadcast!` with a non-static LHS dispatches to generic `broadcast!` code in Base, both before this PR and after, meaning that some of this code won't get coverage on 0.6. I can change that if desired; it would probably result in a speedup as well.~~

Edit: I went ahead and pulled the trigger on this since it resolves a TODO.

### DIFF
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -177,5 +177,8 @@ end
         c = SVector(1, 2, 3)
         a .= c
         @test a == [1 1 1; 2 2 2; 3 3 3]
+
+        d = SVector(1, 2, 3, 4)
+        @test_throws DimensionMismatch a .= d
     end
 end
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -166,4 +166,12 @@ end
         @testinf broadcast!(identity, mv, Ref(aX)) == MVector(aX,aX,aX)
         @test mv == SVector(aX,aX,aX)
     end
+
+    @testset "broadcast! with Array destination" begin
+        # Issue #385
+        a = zeros(3, 3)
+        b = ones(SMatrix{3, 3})
+        a .= b
+        @test a == b
+    end
 end
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -170,8 +170,12 @@ end
     @testset "broadcast! with Array destination" begin
         # Issue #385
         a = zeros(3, 3)
-        b = ones(SMatrix{3, 3})
+        b = @SMatrix [1 2 3; 4 5 6; 7 8 9]
         a .= b
         @test a == b
+
+        c = SVector(1, 2, 3)
+        a .= c
+        @test a == [1 1 1; 2 2 2; 3 3 3]
     end
 end
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -226,7 +226,7 @@ end
 
     return quote
         @_inline_meta
-        sizematch($(Size{newsize}()), dest) || throw(DimensionMismatch("array could not be broadcast to match destination"))
+        @boundscheck sizematch($(Size{newsize}()), dest) || throw(DimensionMismatch("array could not be broadcast to match destination"))
         @inbounds $(Expr(:block, exprs...))
         return dest
     end
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -190,7 +190,8 @@ end
     sizes = tuple(sizes...)
 
     # TODO: this could also be done outside the generated function:
-    sizematch(Size{newsize}(), Size(dest)) || throw(DimensionMismatch("Tried to broadcast to destination sized $newsize from inputs sized $sizes"))
+    sizematch(Size{newsize}(), Size(dest)) ||
+        throw(DimensionMismatch("Tried to broadcast to destination sized $newsize from inputs sized $sizes"))
 
     ndims = 0
     for i = 1:length(sizes)
@@ -225,6 +226,7 @@ end
 
     return quote
         @_inline_meta
+        sizematch($(Size{newsize}()), dest) || throw(DimensionMismatch("array could not be broadcast to match destination"))
         @inbounds $(Expr(:block, exprs...))
         return dest
     end
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -36,9 +36,7 @@
         _broadcast(f, combine_sizes(argsizes), argsizes, as...)
     end
 
-    # TODO: This signature could be relaxed to (::Any, ::Type{StaticArray}, ::Type, ...), though
-    # we'd need to rework how _broadcast!() and broadcast_sizes() interact with normal AbstractArray.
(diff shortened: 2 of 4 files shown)


tkoolen:
@andyferris or @c42f, I'd really appreciate a review; sorry, I know you're busy.

test/broadcast.jl
```diff
@@ -164,4 +164,12 @@ end
         @testinf broadcast!(identity, mv, aX) == MVector(aX,aX,aX)
         @test mv == SVector(aX,aX,aX)
     end
+
+    @testset "broadcast! with Array destination" begin
```
andyferris: Can we add a test case where the compiler doesn't know the size of the LHS of `.=` and we have to "expand" to fill it.

E.g. 
```
a = zeros(3,3)
b = SVector(1,2,3)
a .= b
@test a == [1 1 1; 2 2 2; 3 3 3]`
```
or whatever

andyferris:
Thanks a lot @tkoolen. Overall I'm happy - I'd like to ensure we cover all the corner cases in the tests while we're at it :)

test/broadcast.jl
tkoolen: Darn, yes that's broken, good call.

tkoolen: Alright, in any case I think the output of `broadcast_size` for both the destination and the arguments should be used to determine (probably in `combine_size` and `broadcast_dest_size`) whether to fall back to generic broadcast code if any of the dimensions is not known at compile time (by changing the `BroadcastStyle`).

Currently, the `broadcast_size` of a non-static `AbstractArray` is just `Size()`, meaning that information about the number of dimensions is lost. So `broadcast_size(::AbstractArray)` should be changed to avoid this information loss.

There are a couple of options:

1. Use `-1` as a sentinel in `Size` for unknown dimensions, e.g. `Size{(-1, -1)}()` to signify the dimensions of a matrix whose dimensions are unknown at compile time.
1. Allow something like `Size{(Unknown(), Unknown())}`, i.e. a more explicit version of 1 (strictly better than 1 in my opinion).
1. Leave `Size` as is and create something like a `BroadcastSize` `struct` that implements 2 above, but keeps changes local to the `broadcast` code. Entails some code duplication, but is probably less likely to break other parts of StaticArrays.
1. Have `broadcast_size(::AbstractArray{<:Any, N}) where {N}` return something like `SizeUnknown{N}()` and make `combine_sizes` able to handle these `SizeUnknown`s as well. Probably the simplest to implement, but perhaps not the most elegant.

Which of these (or some other option) has your preference?

andyferris: Sorry, I've been travelling.

I'm happy for you to use your judgement and play around with one of these options. I'm supposing the easiest way out of this is to simply bail out and call the default `broadcast`(`!`) method whenever the static dimensions are singleton. But you might need to implement one of the above even to correctly detect all such cases. :)

andyferris:
(conflicts with #393, sorry)

tkoolen:
Alright, I went with option 2 from https://github.com/JuliaArrays/StaticArrays.jl/pull/387#discussion_r178330083, but I used `Dynamic` instead of `Unknown`. It appears that the changes don't need to be too invasive.

I've made `Length` compatible with `Dynamic` as well, since determining whether the length of the destination is known statically is one way to determine whether to fall back to generic broadcast. I've also added `sizematch` and `dimmatch` functions that can handle `Dynamic`.

Although this is a step towards the more general case of handling mixtures of static and non-static arrays where the destination size can be determined statically, the indexing in the generated functions needs to be generalized to be able to handle this, so I'm still falling back to generic broadcast in this case.

Still to do:

- [ ] check for performance regressions
- [ ] perhaps a dedicated test for `sizematch`
- [ ] check for instances where the widened `Size` and `Length` constructors could lead to hard-to-find bugs
- [ ] consider rewriting `same_size` in terms of `sizematch` (although that could be done in a separate PR as well)

tkoolen:
I've run `julia -O3 perf/benchmark3.jl` (with https://github.com/JuliaArrays/StaticArrays.jl/blob/8e36860a998c9db8452a5b868a43c184095cab5a/perf/benchmark3.jl#L62-L66 commented out because of #395) and verified that there are no significant changes on 0.6.

As for `broadcast!`ing with an `Array` destination:

```julia
using BenchmarkTools, StaticArrays; a = rand(SMatrix{3, 3}); b = rand(SVector{3}); c = rand(3, 3);
@btime $c .= $a .+ $b .+ $b .+ $a
```

on master: 35.404 ns
with this PR: 5.330 ns

I've also made sure that the `code_llvm` with `MMatrix` destination is identical on 0.6 for:

```julia
using BenchmarkTools, StaticArrays; a = rand(SMatrix{3, 3}); b = rand(SVector{3}); c = rand(MMatrix{3, 3});
@code_llvm broadcast!(+, c, a, b, b, a)
```

to ensure that the runtime destination size check in `broadcast!` is elided when the destination is a  `StaticArray`.

tkoolen:
Any more comments?

andyferris:
Nice work, @tkoolen! LGTM.

Interesting that we're moving towards `Eigen` with mixtures of dynamic and static sizes. Potentially in Julia v1.x we'll actually be able to support that :) (performance would benefit from having `Array` implemented in Julia with a backing `Buffer` that we could hook into directly).

PR #393, opened 2018-04-07
add missing BroadcastStyle for scalars

pablosanjose:
This
```
julia> 3 * SVector(1, 2)
2-element Array{Int64,1}:
 3
 6
```
should return an `SVector`.  This wrong result is the consequence of a missing a `BroadcastStyle` declaration for broadcasting over an `StaticArray` and a scalar. This PR adds this missing rule, which fixes also other bugs such as e.g. `SMatrix{2}(1,2,3,4) * 3I` returning a `Matrix` instead of an `SMatrix`.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -69,9 +69,9 @@ else
         _broadcast(f, broadcast_sizes(As...), As...)
     end
 
-    # Add a specialized broadcast! method that overrides the Base fallback and calls the old routine
-    @inline function broadcast!(f, C, ::StaticArrayStyle, As...)
-        _broadcast!(f, Size(C), C, broadcast_sizes(As...), As...)
+    # Add a specialized broadcast! method that overrides the Base fallback
+    @inline function broadcast!(f::Tf, dest, ::StaticArrayStyle, As::Vararg{Any,N}) where {Tf, N}
+        _broadcast!(f, Size(dest), dest, broadcast_sizes(As...), As...)
     end
 end
 
@@ -194,7 +194,7 @@ end
     more = prod(newsize) > 0
     current_ind = ones(Int, max(length(newsize), length.(sizes)...))
     while more
-        exprs_vals = [(!(as[i] <: AbstractArray) ? :(as[$i]) : :(as[$i][$(broadcasted_index(sizes[i], current_ind))])) for i = 1:length(sizes)]
+        exprs_vals = [(!(as[i] <: AbstractArray) ? :(as[$i][]) : :(as[$i][$(broadcasted_index(sizes[i], current_ind))])) for i = 1:length(sizes)]
         exprs[current_ind...] = :(dest[$j] = f($(exprs_vals...)))
 
         # increment current_ind (maybe use CartesianRange?)
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -57,6 +57,8 @@ else
     # Precedence rules
     BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.DefaultArrayStyle{N}) where {M,N} =
         Broadcast.DefaultArrayStyle(Broadcast._max(Val(M), Val(N)))
+    BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.DefaultArrayStyle{0}) where {M} =
+        StaticArrayStyle{M}()
     # FIXME: These two rules should be removed once VectorStyle and MatrixStyle are removed from base/broadcast.jl
     BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.VectorStyle) where M = Broadcast.Unknown()
     BroadcastStyle(::StaticArrayStyle{M}, ::Broadcast.MatrixStyle) where M = Broadcast.Unknown()
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -131,7 +131,7 @@ end
     current_ind = ones(Int, length(newsize))
 
     while more
-        exprs_vals = [(!(a[i] <: AbstractArray) ? :(a[$i]) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))])) for i = 1:length(sizes)]
+        exprs_vals = [(!(a[i] <: AbstractArray) ? :(a[$i][]) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))])) for i = 1:length(sizes)]
         exprs[current_ind...] = :(f($(exprs_vals...)))
 
         # increment current_ind (maybe use CartesianRange?)
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -161,9 +161,9 @@ end
     @testset "broadcast general scalars" begin
         # Issue #239 - broadcast with non-numeric element types
         @eval @enum Axis aX aY aZ
-        @testinf (SVector(aX,aY,aZ) .== aX) == SVector(true,false,false)
+        @testinf (SVector(aX,aY,aZ) .== Ref(aX)) == SVector(true,false,false)
         mv = MVector(aX,aY,aZ)
-        @testinf broadcast!(identity, mv, aX) == MVector(aX,aX,aX)
+        @testinf broadcast!(identity, mv, Ref(aX)) == MVector(aX,aX,aX)
         @test mv == SVector(aX,aX,aX)
     end
 end


KristofferC:
Would be good to test the output types for these operations.

pablosanjose:
The types are already tested in test/broadcast.jl, and were failing. Now they pass, at least on my machine.

KristofferC:
Oh, sorry!

pablosanjose:
Why, no problem! There is still one failure in the broadcast tests, but it seems unrelated (`@testset "broadcast general scalars"`). I'm looking at that now.

pablosanjose:
The first commit in this PR fixed the broadcast-with-a-scalar problem, but simultaneously broke a test that checks the following:
```
@enum Axis aX aY aZ
(SVector(aX,aY,aZ) .== aX) == SVector(true,false,false)
```
Digging into this I think the fact that it passed before was an unintended side effect of the missing `BroadcastStyle`, which made the generic `Base.broadcast` to be invoked, instead of the one implemented in `StaticArrays`. With the new `BroadcastStyle` makes the latter be called instead, unveiling another bug. If I'm not mistaken, `StaticArrays._broadcast` does not assume that the inputs it gets are going to be `broadcastable(a)` instead of `a` (following julialang/#23939), which in the above test is a `Ref`. Hence, the test passes once more simply by indexing into scalars as `a[]` instead of `a` (assume they are wrapped in a Ref). With this second commit, the above test passes once more. (There are still some additional broken tests in broadcasts, related to `eltype`s with non-leaf types).

Note: I have my doubts that this second commit of this PR is the 'good' way to make use of the new broadcast API and the `broadcastable` concept, so I would appreciate a critical review here.

andyferris:
I haven't had time to dig in yet - has anyone checked whether codegen is still good on v0.7 with this PR?

pablosanjose:
I'm totally clueless when it comes to assembler, but the codegen of current master (0.7.0-DEV.4822) with this PR looks the same as with a 1 month old master (0.7.0-DEV.4570, before the problem began) without this PR for something like `SVector(1, 2) + 3`. However the second test `SVector(aX,aY,aZ) .== aX` produces slightly different @code_native, and is much much slower with the current PR, which is not good.

pablosanjose:
Ah, ok, I see what happened here. In latest master using scalars without wrapping them in `Ref` is deprecated. The call `SVector(aX,aY,aZ) .== aX` complains that `aX` should be replaced by `Ref(aX)`. That deprecation warning is what seems to be allocating and making the broadcast slow. If I do `SVector(aX,aY,aZ) .== Ref(aX)` I still get different codegen than in the old master, but now it is non-allocating and just as fast as before

0.7.0-DEV.4570 without PR
```
julia> @btime SVector(aX, aY, aZ) .== aX
  1.344 ns (0 allocations: 0 bytes)
3-element SArray{Tuple{3},Bool,1,3}:
  true
 false
 false

julia> @code_native SVector(aX,aY,aZ) .== aX
	.section	__TEXT,__text,regular,pure_instructions
; Function #832 {
; Location: deprecated.jl:167
	pushq	%rbp
	pushq	%r15
	pushq	%r14
	pushq	%rbx
	subq	$72, %rsp
	movl	%esi, %r14d
	movq	%rdi, %rbx
	vxorps	%xmm0, %xmm0, %xmm0
	vmovups	%xmm0, 8(%rsp)
	movabsq	$jl_get_ptls_states_fast, %rax
	callq	*%rax
	movq	%rax, %rbp
	movq	$2, (%rsp)
	movq	(%rbp), %rax
	movq	%rax, 8(%rsp)
	leaq	(%rsp), %rax
	movq	%rax, (%rbp)
; Location: deprecated.jl:168
; Function string; {
; Location: io.jl:141
; Function print_to_string; {
; Location: io.jl:110
	movabsq	$4584034312, %rax       ## imm = 0x1113AD008
	movq	%rax, 24(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 32(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 40(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 48(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 56(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 64(%rsp)
	movabsq	$"#print_to_string#352", %rax
	movabsq	$jl_system_image_data, %rdi
	leaq	24(%rsp), %r15
	movl	$6, %edx
	movq	%r15, %rsi
	callq	*%rax
	movq	%rax, 16(%rsp)
;}}
	movq	%rax, 24(%rsp)
	movabsq	$4584284720, %rax       ## imm = 0x1113EA230
	movq	%rax, 32(%rsp)
	movabsq	$depwarn, %rax
	movabsq	$jl_system_image_data, %rdi
	movl	$2, %edx
	movq	%r15, %rsi
	callq	*%rax
; Location: deprecated.jl:170
; Function broadcast; {
; Location: broadcast.jl:624
; Function broadcast; {
; Location: broadcast.jl:64
; Function _broadcast; {
; Location: broadcast.jl:96
; Function @generated body; {
; Location: broadcast.jl:154
; Function ==; {
; Location: operators.jl:77
	cmpl	%r14d, (%rbx)
	sete	%al
	cmpl	%r14d, 4(%rbx)
	sete	%dl
	cmpl	%r14d, 8(%rbx)
	sete	%cl
;}}}}}
	movq	8(%rsp), %rsi
	movq	%rsi, (%rbp)
	addq	$72, %rsp
	popq	%rbx
	popq	%r14
	popq	%r15
	popq	%rbp
	retq
	nop
;}
```

0.7.0-DEV.4822 with PR
```
julia> @btime SVector(aX, aY, aZ) .== Ref(aX)
  1.344 ns (0 allocations: 0 bytes)
3-element SArray{Tuple{3},Bool,1,3}:
  true
 false
 false

julia> @code_native SVector(aX, aY, aZ) .== Ref(aX)
	.section	__TEXT,__text,regular,pure_instructions
; Function #836 {
; Location: deprecated.jl:167
	pushq	%r15
	pushq	%r14
	pushq	%r12
	pushq	%rbx
	subq	$72, %rsp
	movq	%rsi, %r14
	movq	%rdi, %r15
	vxorps	%xmm0, %xmm0, %xmm0
	vmovups	%xmm0, 8(%rsp)
	movabsq	$jl_get_ptls_states_fast, %rax
	callq	*%rax
	movq	%rax, %rbx
	movq	$2, (%rsp)
	movq	(%rbx), %rax
	movq	%rax, 8(%rsp)
	leaq	(%rsp), %rax
	movq	%rax, (%rbx)
; Location: deprecated.jl:168
; Function string; {
; Location: io.jl:141
; Function print_to_string; {
; Location: io.jl:110
	movabsq	$4567093256, %rax       ## imm = 0x110385008
	movq	%rax, 24(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 32(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 40(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 48(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 56(%rsp)
	movabsq	$jl_system_image_data, %rax
	movq	%rax, 64(%rsp)
	movabsq	$"#print_to_string#355", %rax
	movabsq	$jl_system_image_data, %rdi
	leaq	24(%rsp), %r12
	movl	$6, %edx
	movq	%r12, %rsi
	callq	*%rax
	movq	%rax, 16(%rsp)
;}}
	movq	%rax, 24(%rsp)
	movabsq	$4567241952, %rax       ## imm = 0x1103A94E0
	movq	%rax, 32(%rsp)
	movabsq	$depwarn, %rax
	movabsq	$jl_system_image_data, %rdi
	movl	$2, %edx
	movq	%r12, %rsi
	callq	*%rax
; Location: deprecated.jl:170
; Function broadcast; {
; Location: broadcast.jl:616
; Function broadcast; {
; Location: broadcast.jl:69
; Function _broadcast; {
; Location: broadcast.jl:101
; Function @generated body; {
; Location: broadcast.jl:159
; Function getindex; {
; Location: refvalue.jl:30
; Function getproperty; {
; Location: sysimg.jl:18
	movl	(%r14), %ecx
;}}
; Function ==; {
; Location: operators.jl:81
	cmpl	%ecx, (%r15)
	sete	%al
	cmpl	%ecx, 4(%r15)
	sete	%dl
	cmpl	%ecx, 8(%r15)
	sete	%cl
;}}}}}
	movq	8(%rsp), %rsi
	movq	%rsi, (%rbx)
	addq	$72, %rsp
	popq	%rbx
	popq	%r12
	popq	%r14
	popq	%r15
	retq
;}
```

pablosanjose:
I pushed two more commits with (1) a change in the `"broadcast general scalars" ` tests to use `Ref(aX)` instead of `aX`, as per the deprecation warning, and (2) the equivalent change to `_broadcast!` that was done for `_broadcast` (missed that before, sorry). This seems to get us back to v0.6.2 performace for `broadcast!` for the last test in test/broadcast.jl
```
@enum Axis aX aY aZ
mv = MVector(aX,aY,aZ)
broadcast!(identity, mv, Ref(aX))
```
This passes also now (it did not with the old master), and does not regress compared to 0.6.2

v0.6.2
```
julia> @btime broadcast!(identity, $mv, $aX)
  1.611 ns (0 allocations: 0 bytes)
3-element MVector{3,Axis}:
 aX
 aX
 aX
```

0.7.0-DEV.4822 with PR
```
julia> @btime broadcast!(identity, $mv, Ref($aX))
  1.346 ns (0 allocations: 0 bytes)
3-element MArray{Tuple{3},Axis,1,3}:
 aX::Axis = 0
 aX::Axis = 0
 aX::Axis = 0
```

pablosanjose:
Right, I think I'm done here for the moment. I leave this alone now for review.

andyferris:
Great. I checked, it even works with strings. :)

```
julia> ["a"] .* "b"
1-element Array{String,1}:
 "ab"

julia> SVector("a") .* "b"
1-element SArray{Tuple{1},String,1,1}:
 "ab"
```

Thanks @pablosanjose

Issue #272, opened 2017-08-08
Unary and binary operators on SArray do not return SArray in Julia 0.7


wsshin:
Not sure what changes in 0.7 caused this, but unary operators on `SArray`s no longer give `SArary`s:

```julia
julia> VERSION
v"0.7.0-DEV.1283"

julia> v = SVector(1,2,3);

julia> -v
3-element Array{Int64,1}:
 -1
 -2
 -3
```

Neither do binary operators:
```julia
julia> v-v
3-element Array{Int64,1}:
 0
 0
 0
```

Element-wise binary operators seem to be returning `SArray`s correctly:
```julia
julia> v.-v
3-element SVector{3,Int64}:
 0
 0
 0
```

andyferris:
Interesting...

andreasnoack:
I think this is now fixed. Was probably something on master. However, https://github.com/JuliaArrays/StaticArrays.jl/blob/7276787a0b147135b758dd836b5951236e30558d/src/eigen.jl#L191-L192 now fails with
```julia
SYSTEM: show(lasterr) caused an error
ArgumentError("No StaticArray found in argument list")

Stacktrace:
 [1] include_relative(::Module, ::String) at ./loading.jl:464
 [2] include at ./sysimg.jl:14 [inlined]
 [3] include(::String) at /Users/andreasnoack/.julia/v0.7/StaticArrays/src/StaticArrays.jl:3
 [4] include_relative(::Module, ::String) at ./loading.jl:464
 [5] _require(::Symbol) at ./loading.jl:401
 [6] require(::Symbol) at ./loading.jl:318
 [7] eval(::Module, ::Expr) at ./repl/REPL.jl:3
 [8] eval_user_input(::Any, ::Base.REPL.REPLBackend) at ./repl/REPL.jl:69
 [9] macro expansion at ./repl/REPL.jl:100 [inlined]
 [10] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
```

andreasnoack:
Hm. https://github.com/JuliaArrays/StaticArrays.jl/blob/7276787a0b147135b758dd836b5951236e30558d/src/mapreduce.jl#L9-L11 breaks printing of errors
```julia
julia> foo(::Type{Matrix{T}}) = T
ERROR: UndefVarError: T not defined
julia> using StaticArrays
[a million warnings]
julia> foo(::Type{Matrix{T}}) = T
[...]

PR #397, opened 2018-04-17
diagm 0.7-style

martinholters:
Add `diagm(k1 => v1, k2 => v2, ...)` style interface, where the diagonals `k` are given as `Val` instances (not types!). Also use this syntax in the tests when calling `Base.diagm` to avoid deprecation warnings on Julia 0.7.

The old interface is left in place, but should be deprecated/deleted eventually.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -200,8 +200,11 @@ end
     end
 end
 
-# old interface, to be deprecated/deleted eventually
-@inline diagm(v::StaticVector, k::Type{Val{D}}=Val{0}) where {D} = diagm(k() => v)
+if VERSION < v"v0.7.0-DEV.2161"
+    @inline diagm(v::StaticVector, k::Type{Val{D}}=Val{0}) where {D} = diagm(k() => v)
+else
+    @deprecate(diagm(v::StaticVector, k::Type{Val{D}}=Val{0}) where {D}, diagm(k() => v))
+end
 
 @inline diag(m::StaticMatrix, k::Type{Val{D}}=Val{0}) where {D} = _diag(Size(m), m, k)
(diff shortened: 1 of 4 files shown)


martinholters:
All relevant tests passed on nightly before hitting the known failure (https://github.com/JuliaArrays/StaticArrays.jl/issues/272#issuecomment-381516243).

andyferris:
I'll note that there was some discussion of re-allowing an implicit `0 => ...` in the single-argument `diagm`, since it seems convenient and not too harmful... I wonder if this likely to proceed for v1.0? (@mbauman?)

src/linalg.jl
```diff
@@ -182,24 +182,27 @@ end
...
     end
     return quote
         $(Expr(:meta, :inline))
-        @inbounds return similar_type($v, Size($Snew))(tuple($(exprs...)))
+        @inbounds return SMatrix{$N,$N,$T}(tuple($(exprs...)))
     end
 end
 
+# old interface, to be deprecated/deleted eventually
+@inline diagm(v::StaticVector, k::Type{Val{D}}=Val{0}) where {D} = diagm(k() => v)
```
andyferris: This method should use `Val` instances on v0.7.

andyferris: Or I guess not... maybe we just add a deprecation immediately...

martinholters: Or define as is on 0.6 and as deprecated on 0.7-? I was unsure what to do here.

andyferris: Yeah I think you are right - if we simply deprecate this method on v0.7 everything should follow `Base` and `LinearAlgerbra` (and leave as-is on v0.6).

martinholters:
OSX failure was a timeout.

andyferris:
Thanks again :)

PR #398, opened 2018-04-20
Make Cholesky factorization compatible with 0.7

martinholters:
Note that the error handling is still a bit different as `_chol!` still throws an error instead of returning a non-zero `info`.


andyferris:
I believe Matt Baumann is currently destroying *everything* so we'll be revisiting this shortly :laugh:

martinholters:
Hehe, I had planned to bump this today, remarking it would probably be due for an overhaul after https://github.com/JuliaLang/julia/pull/27249.

PR #399, opened 2018-04-27
make transpose and ctranspose recursive like Base

marius311:
Asking this question in the form a PR which can maybe be merged if the answer is yes, but shouldn't `transpose` and `ctranspose` be recursive like Base? 

For reference with Base (everything here is 0.6.2):

```julia
julia> [[[1 2; 3 4]] [[1 2; 3 4]]; [[1 2; 3 4]] [[1 2; 3 4]]]'
2×2 Array{Array{Int64,2},2}:
 [1 3; 2 4]  [1 3; 2 4]
 [1 3; 2 4]  [1 3; 2 4]
```
StaticArrays without this PR:
```julia
julia> (@SMatrix [[1 2; 3 4] [1 2; 3 4]; [1 2; 3 4] [1 2; 3 4]])'
2×2 StaticArrays.SArray{Tuple{2,2},Array{Int64,2},2,4}:
 [1 2; 3 4]  [1 2; 3 4]
 [1 2; 3 4]  [1 2; 3 4]                                                                                                        
```
and with it:
```julia
julia> (@SMatrix [[1 2; 3 4] [1 2; 3 4]; [1 2; 3 4] [1 2; 3 4]])'
2×2 StaticArrays.SArray{Tuple{2,2},Array{Int64,2},2,4}:
 [1 3; 2 4]  [1 3; 2 4]
 [1 3; 2 4]  [1 3; 2 4]
```

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -65,7 +65,7 @@ end
 @generated function _transpose(::Size{S}, m::StaticMatrix) where {S}
     Snew = (S[2], S[1])
 
-    exprs = [:(m[$(LinearIndices(S)[j1, j2])]) for j2 = 1:S[2], j1 = 1:S[1]]
+    exprs = [:(transpose(m[$(LinearIndices(S)[j1, j2])])) for j2 = 1:S[2], j1 = 1:S[1]]
 
     return quote
         $(Expr(:meta, :inline))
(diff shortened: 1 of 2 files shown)


andyferris:
@marius311 Would you mind adding a couple of unit tests for this one, please? 

And thanks, yes we should make this work :)

andyferris:
bump

marius311:
Added some tests, I think its ready otherwise.

mschauer:
I recently had to define
```
function conj!(A::Array{StaticArrays.SArray{Tuple{2,2},Float64,2,4},2})
    for i in 1:size(A,1)
        A[i,i] = A[i,i]'
        for j in 1:size(A,2)
            if i != j 
                A[i,j], A[j, i] = A[i,j]', A[j, i]'
            end
        end
    end
    A
end
```
to be able to `A'` matrices of static matrices.
Is this the same issue?

andyferris:
Thanks @marius311 :heart: 

@mschauer Yes, I believe so. If this branch doesn't work as you expect - could you please let us know?

mschauer:
Of course. 👍

PR #407, opened 2018-05-17
Adds more Kronecker product specialisations

zygmuntszpak:
Defines `kron` between two `StaticVector` types as well as between a `StaticVector` and a `StaticMatrix` type.  The definitions also include cases for transposed `StaticVector` types (i.e. a `StaticVector` wrapped by a `RowVector`type).

There are three broken tests which are due to the fact that taking the transpose of a `SizedArray` returns a `SMatrix`. I've made a comment in the relevant part of the test. Not sure of this behaviour of transpose is intentional or not?

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -358,12 +358,135 @@ const _length_limit = Length(200)
     if prod(outsize) > length_limit
         return :( SizedMatrix{$(outsize[1]),$(outsize[2])}( kron(drop_sdims(a), drop_sdims(b)) ) )
     end
-    rows = [:(hcat($([:(a[$(LinearIndices(SA)[i, j])]*b) for j=1:SA[2]]...))) for i=1:SA[1]]
+
+    M = [ :(a[$ia,$ja] * b[$ib,$jb]) for ib in 1:SB[1], ia in 1:SA[1], jb in 1:SB[2], ja in 1:SA[2] ]
+
     return quote
         @_inline_meta
-        @inbounds return vcat($(rows...))
+        @inbounds return  similar_type($a, promote_type(eltype(a),eltype(b)), Size($(outsize)))(tuple($(M...)))
     end
 end
+
+@inline kron(a::StaticVector, b::StaticVector) = _kron_vec_x_vec(_length_limit, Size(a), Size(b), a, b)
+@generated function _kron_vec_x_vec(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
+    outsize = SA .* SB
+    if prod(outsize) > length_limit
+        return :( SizedVector{$(outsize[1])}( kron(drop_sdims(a), drop_sdims(b)) ) )
+    end
+
+    m = [ :(a[$ia] * b[$ib]) for ib in 1:SB[1], ia in 1:SA[1]]
+
+    return quote
+        @_inline_meta
+        @inbounds return similar_type($a, promote_type(eltype(a),eltype(b)), Size($(outsize)))(tuple($(m...)))
+    end
+end
+
+@inline kron(a::RowVector{<:Number,<:StaticArray}, b::RowVector{<:Number,<:StaticArray}) = _kron_tvec_x_tvec(_length_limit, Size(a), Size(b), a, b)
+@generated function _kron_tvec_x_tvec(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
+    outsize = SA .* SB
+    if prod(outsize) > length_limit
+        return :( SizedMatrix{$(outsize[1]),$(outsize[2])}( kron(drop_sdims(a), drop_sdims(b)) ) )
+    end
+
+    m = [ :(a[$ia] * b[$ib]) for ib in 1:SB[2], ia in 1:SA[2]]
+
+    return quote
+        @_inline_meta
+        @inbounds return similar_type($a, promote_type(eltype(a),eltype(b)), Size($(outsize)))(tuple($(m...)))
+    end
+end
+
+@inline kron(a::RowVector{<:Number,<:StaticArray}, b::StaticVector) = _kron_tvec_x_vec(_length_limit, Size(a), Size(b), a, b)
+@generated function _kron_tvec_x_vec(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
+    outsize = (SA[1] * SB[1], SA[2])
+    if prod(outsize) > length_limit
+        return :( SizedMatrix{$(outsize[1]),$(outsize[2])}( kron(drop_sdims(a), drop_sdims(b)) ) )
+    end
+
+    M = [ :(a[$ia] * b[$ib]) for ib in 1:SB[1], ia in 1:SA[2]]
+
+    return quote
+        @_inline_meta
+        @inbounds return similar_type($a, promote_type(eltype(a),eltype(b)), Size($(outsize)))(tuple($(M...)))
+    end
+end
+
+@inline kron(a::StaticVector, b::RowVector{<:Number,<:StaticArray}) = _kron_vec_x_tvec(_length_limit, Size(a), Size(b), a, b)
+@generated function _kron_vec_x_tvec(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
+    outsize = (SA[1] * SB[1], SB[2])
+    if prod(outsize) > length_limit
+        return :( SizedMatrix{$(outsize[1]),$(outsize[2])}( kron(drop_sdims(a), drop_sdims(b)) ) )
+    end
+
+    M = [ :(a[$ia] * b[$ib]) for  ia in 1:SA[1], ib in 1:SB[2]]
+
+    return quote
+        @_inline_meta
+        @inbounds return similar_type($a, Size($(outsize)))(tuple($(M...)))
+    end
+end
+
+@inline kron(a::StaticVector, b::StaticMatrix) = _kron_vec_x_mat(_length_limit, Size(a), Size(b), a, b)
+@generated function _kron_vec_x_mat(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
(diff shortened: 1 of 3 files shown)


src/linalg.jl
```diff
@@ -365,6 +365,68 @@ const _length_limit = Length(200)
...
+    end
+    #rows = [:(vcat($([:(a[$(LinearIndices(SA)[i])]*b) for i=1:SA[1]]...)))]
+    rows = [:(hcat($([:(a[$(LinearIndices(SA)[i, j])]*b) for j=1:SA[2]]...))) for i=1:SA[1]]
+    return quote
+        @_inline_meta
+        @inbounds return vcat($(rows...))
+    end
+end
+
+@inline kron(a::StaticMatrix, b::RowVector{<:Number,<:StaticArray}) = _kron_mat_x_vec(_length_limit, Size(a'), Size(b'), a', b')'
```
andyferris: Probably more direct to reshape `b` into a matrix... but I don't think we have a method for that?

src/linalg.jl
```diff
@@ -365,6 +365,68 @@ const _length_limit = Length(200)
     end
 end
 
+@inline kron(a::StaticVector, b::StaticVector) = _kron_vec_x_vec(_length_limit, Size(a), Size(b), a, b)
+@generated function _kron_vec_x_vec(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
+    outsize = SA .* SB
+    if prod(outsize) > length_limit
+        return :( SizedVector{$(outsize[1])}( kron(drop_sdims(a), drop_sdims(b)) ) )
+    end
+    rows = [:(vcat($([:(a[$(LinearIndices(SA)[i])]*b) for i=1:SA[1]]...)))]
```
andyferris: What's the purpose of the outermost array?

I'm curious about how good (performant) `vcat` is with four or more arguments.

zygmuntszpak: Is `vcat` generally slower than `hcat`? I was trying to imitate what was already done for the `kron(StaticMatrix, StaticMatrix)`.

src/linalg.jl
```diff
@@ -365,6 +365,68 @@ const _length_limit = Length(200)
...
+
+@inline kron(a::RowVector{<:Number,<:StaticArray}, b::StaticMatrix) = _kron_vec_x_mat(_length_limit, Size(a'), Size(b'), a', b')'
+
+@inline kron(a::StaticMatrix, b::StaticVector) = _kron_mat_x_vec(_length_limit, Size(a), Size(b), a, b)
+@generated function _kron_mat_x_vec(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
+    outsize = (SA[1] * SB[1], SA[2])
+    if prod(outsize) > length_limit
+        return :( SizedMatrix{$(outsize[1]),$(outsize[2])}( kron(drop_sdims(a), drop_sdims(b)) ) )
+    end
+    #rows = [:(vcat($([:(a[$(LinearIndices(SA)[i])]*b) for i=1:SA[1]]...)))]
```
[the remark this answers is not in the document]
zygmuntszpak: Ooops

src/linalg.jl
zygmuntszpak: Now that you pointed it out I realize that the outer array serves no purpose. Its an artifact of my attempt at molding the code that was already there for the Kronecker product between two matrices. I've removed it from other parts of the code as well.

src/linalg.jl
zygmuntszpak: Are you suggesting that we could reshape `b` into a `1 x 3` matrix and then dispatch on the `kron` that operates on two static matrices? 

It seems that there is no method currently defined for that. 
For example, with 
```julia
s = @SVector [1, 2,  3];
reshape(s',(1,3))

1×3 Base.ReshapedArray{Int64,2,RowVector{Int64,StaticArrays.SArray{Tuple{3},Int64,1,3}},Tuple{}}:
 1  2  3
```
I get this `Base.ReshapedArray` type instead of the required `StaticMatrix`.

zygmuntszpak:
I've removed the dead code and the superfluous outermost array. Would you like me to try to avoid `vcat` and use `hcat` instead?

zygmuntszpak:
Hi Andy,

I've been trying to figure out what you mean but to no avail. It is safe to assume I know nothing about meta programming, and am still learning aspects of Julia. 

The only way that I can see how one can construct the entire output array of the Kronecker product that will avoid the use `vcat` is with a comprehension. 
For example,

```julia

@inline kron(a::StaticMatrix, b::StaticMatrix) = _kron(_length_limit, Size(a), Size(b), a, b)
@generated function _kron(::Length{length_limit}, ::Size{SA}, ::Size{SB}, a, b) where {length_limit,SA,SB}
    outsize = SA .* SB
    if prod(outsize) > length_limit
        return :( SizedMatrix{$(outsize[1]),$(outsize[2])}( kron(drop_sdims(a), drop_sdims(b)) ) )
    end
    M = [ a[ia,ja] * b[ib,jb] for ia in 1:SA[1],  ja in 1:SA[2],  ib in 1:SB[1],  jb in 1:SB[2]]
    return quote
        @_inline_meta
        @inbounds return similar_type($(M),Size($(outsize)))($(M))  
    end
end
```
However, this give the error 
```julia
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type StaticArrays.SArray{Tuple{2,2},Int64,2,4}
This may have arisen from a call to the constructor StaticArrays.SArray{Tuple{2,2},Int64,2,4}(...),
```
with the following input
```julia
A = @SMatrix [1 2; 3 4]
kron(A,A)
```
On the other hand, if I test the code in my own script (outside of the generated function) it seems to work:
```julia
M = [ A[ia,ja] * A[ib,jb] for ia in 1:2,  ja in 1:2,  ib in 1:2,  jb in 1:2]
N = similar_type(M,Size((4,4)))(M)
4×4 StaticArrays.SArray{Tuple{4,4},Int64,2,16}:
 1   3  2   4
 3   9  6  12
 2   6  4   8
 4  12  8  16
```

I also don't understand how precisely one avoids allocations. If I use a comprehension then will it not be allocating on the heap automatically? On the other hand, the existing code also uses a comprehension of rows which seems not to allocate... I guess I don't understand the design principle behind what makes this package work. 

I'd appreciate if you could give me a more detailed example of how you would like me to modify the current implementation. Also, what is the reason for the error in the generated function but not in my own script?

Best wishes,

Zygmunt

andyferris:
Apologies for brevity, but try something like `@inbounds return similar_type($a, Size($(outsize)))(tuple($(M...)))` (however I suspect you need to swap the comprehension order `ja` and `ib` to get the elements arranged the right way?)

zygmuntszpak:
Yes, I need to rethink the ordering and reshaping of the comprehension but unfortunately I still get an error even after I make your suggested amendment.

```julia
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type StaticArrays.SArray{Tuple{2,2},Int64,2,4}
This may have arisen from a call to the constructor StaticArrays.SArray{Tuple{2,2},Int64,2,4}(...),
Stacktrace:
 [1] getindex(::Type{StaticArrays.SArray{Tuple{2,2},Int64,2,4}}, ::Int64, ::Int64) at ./array.jl:208
 [2] (::StaticArrays.##285#286{DataType,DataType})(::NTuple{4,Int64}) at ./<missing>:0
 [3] collect(::Base.Generator{Base.Iterators.Prod{UnitRange{Int64},Base.Iterators.Prod{UnitRange{Int64},Base.Iterators.Prod2{UnitRange{Int64},UnitRange{Int64}}}},StaticArrays.##285#286{DataType,DataType}}) at ./array.jl:475
 [4] _kron(...) at /home/zygmunt/.julia/v0.6/StaticArrays/src/linalg.jl:364
 [5] kron(::StaticArrays.SArray{Tuple{2,2},Int64,2,4}, ::StaticArrays.SArray{Tuple{2,2},Int64,2,4}) at /home/zygmunt/.julia/v0.6/StaticArrays/src/linalg.jl:355
 [6] macro expansion at /home/zygmunt/.julia/v0.6/Atom/src/repl.jl:118 [inlined]
 [7] anonymous at ./<missing>:?
```

That's with the following return
```julia
    return quote
        @_inline_meta
        @inbounds return similar_type($a, Size($(outsize)))(tuple($(M...))) 
    end
```

The stack trace points to this line (with the current incorrectly ordered comprehension)
```julia
M = [ a[ia,ja] * b[ib,jb] for ia in 1:SA[1],  ja in 1:SA[2],  ib in 1:SB[1],  jb in 1:SB[2]]
```

zygmuntszpak:
The correct order of the comprehension is as follows:
```julia
M = [ A[ia,ja] * B[ib,jb] for ib in 1:SB[1],ia in 1:SA[1],jb in 1:SB[2],ja in 1:SA[2]]
```
if one wants to use `reshape(M,outsize)`. 

A quick test...
```julia
A = @SMatrix [1 2; 3 4]
B = @SMatrix [5 6; 7 8]
P = [1 2; 3 4]
Q = [5 6; 7 8]
M = [ A[ia,ja] * B[ib,jb] for ib in 1:SB[1],ia in 1:SA[1],jb in 1:SB[2],ja in 1:SA[2]]
N = similar_type(M,Size((4,4)))(M)
@test N == kron(P,Q)
Test Passed

A = @SMatrix [5 6; 7 8]
B = @SMatrix [1 2; 3 4]
M = [ A[ia,ja] * B[ib,jb] for ib in 1:SB[1],ia in 1:SA[1],jb in 1:SB[2],ja in 1:SA[2]]
N = similar_type(M,Size((4,4)))(M)
@test N == kron(Q,P)
Test Passed
```

zygmuntszpak:
I've [posted a question ](https://discourse.julialang.org/t/request-help-in-understanding-why-certain-operations-in-a-generated-function-lead-to-a-methoderror-but-are-fine-outside-the-generated-function/10980)  on the Discourse forum to see if anyone has any suggestions for resolving the error.

zygmuntszpak:
As per your request, for each specialization the entire output array is constructed using `similar_type` and a `tuple` with all the computed elements. In so doing, I no longer return the transpose of anything and hence the previous `test_broken` cases are no longer an issue.

PR #408, opened 2018-05-17
Fixes to fallback to generic broadcast!.

tkoolen:
Note: this is a PR against sd/lazybcast, not against master!

The fallback is covered by the following test set:

https://github.com/JuliaArrays/StaticArrays.jl/blob/9d80122dabf420373a0c3a0db09b52a2b8df8089/test/broadcast.jl#L170-L183

which did indeed catch an error due to the fact that the `axes` field was not being copied to the `Broadcasted{DefaultArrayStyle{M}}` (which has already been set to `axes(dest)` in `materialize!`  at this point and is needed for `copyto!` to work correctly). Using the available `convert` method fixes the issue.

I also reduced code duplication a bit and removed the fallback changes in `Base.copy`; this can be done in a separate PR, and with the new broadcast API the commented-out code is not the correct way to do it.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -46,7 +46,7 @@
 else
     ## New Broadcast API ##
     import Base.Broadcast:
-    BroadcastStyle, AbstractArrayStyle, Broadcasted, DefaultArrayStyle
+    BroadcastStyle, AbstractArrayStyle, Broadcasted, DefaultArrayStyle, materialize!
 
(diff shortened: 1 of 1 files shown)


andyferris:
Thanks, @tkoolen.

SimonDanisch:
What are the guidelines for tagging StaticArrays by now? Should we just tag now?

PR #419, opened 2018-05-28
Specialize arithmetic_closure on type to eliminate allocation

cstjean:
On master + Julia 0.6.2, 
```julia
using StaticArrays
const TT = Complex{Float64} 
@allocated StaticArrays.arithmetic_closure(TT)
```
returns 224. This is a significant tax when working with `ForwardDiff.Dual`. It is fixed by this PR.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -165,4 +165,4 @@ julia> arithmetic_closure(BigInt)
 BigFloat
 ```
 """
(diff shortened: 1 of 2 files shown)


mschauer:
Should this regression be reported?

KristofferC:
What regression?

mschauer:
Sorry, I misread "On master + Julia 0.6.2", this is not a regression.

PR #420, opened 2018-05-28
 Rework use of RowVector/ConjVector/Transpose/Adjoint.

### LINKED ISSUE #416: Assessment of the use of `RowVector` / `ConjVector` / `Transpose` / `Adjoint`
Just for context, on 0.6 we have

```julia
julia> for x in ([1 2; 3 4], [1, 2], [im, im])
              @show typeof(x) typeof(transpose(x)) typeof(adjoint(x)) typeof(x')
              println()
          end
typeof(x) = Array{Int64,2}
typeof(transpose(x)) = Array{Int64,2}
typeof(adjoint(x)) = Array{Int64,2}
typeof(x') = Array{Int64,2}

typeof(x) = Array{Int64,1}
typeof(transpose(x)) = RowVector{Int64,Array{Int64,1}}
typeof(adjoint(x)) = RowVector{Int64,Array{Int64,1}}
typeof(x') = RowVector{Int64,Array{Int64,1}}

typeof(x) = Array{Complex{Bool},1}
typeof(transpose(x)) = RowVector{Complex{Bool},Array{Complex{Bool},1}}
typeof(adjoint(x)) = RowVector{Complex{Int64},ConjArray{Complex{Int64},1,Array{Complex{Bool},1}}}
typeof(x') = RowVector{Complex{Int64},ConjArray{Complex{Int64},1,Array{Complex{Bool},1}}}
```

while on 0.7:

```julia
julia> for x in ([1 2; 3 4], [1, 2])
           @show typeof(x) typeof(transpose(x)) typeof(adjoint(x)) typeof(x')
           println()
       end
typeof(x) = Array{Int64,2}
typeof(transpose(x)) = Transpose{Int64,Array{Int64,2}}
typeof(adjoint(x)) = Adjoint{Int64,Array{Int64,2}}
typeof(x') = Adjoint{Int64,Array{Int64,2}}

typeof(x) = Array{Int64,1}
typeof(transpose(x)) = Transpose{Int64,Array{Int64,1}}
typeof(adjoint(x)) = Adjoint{Int64,Array{Int64,1}}
typeof(x') = Adjoint{Int64,Array{Int64,1}}

typeof(x) = Array{Complex{Bool},1}
typeof(transpose(x)) = Transpose{Complex{Bool},Array{Complex{Bool},1}}
typeof(adjoint(x)) = Adjoint{Complex{Int64},Array{Complex{Bool},1}}
typeof(x') = Adjoint{Complex{Int64},Array{Complex{Bool},1}}
```

Right now, StaticArrays defines

https://github.com/JuliaArrays/StaticArrays.jl/blob/96bf39c4b723538f2f92e6ad34ce075623e9e9fb/src/StaticArrays.jl#L27

for 0.7, which, while it resolves a lot of test cases using `'`, is arguably not really the right thing to do.

In addition, there are still usages of `RowVector` in the `kron` code and in `triangular.jl` and `util.jl`. This currently results in test failures on Julia master.

Furthermore, there are a lot of cases where methods are defined only for `Adjoint` or `RowVector`, but arguably they should be defined for both `Adjoint` and `Transpose` (on 0.7 at least). For example in the broadcasting code we have (only):

https://github.com/JuliaArrays/StaticArrays.jl/blob/96bf39c4b723538f2f92e6ad34ce075623e9e9fb/src/broadcast.jl#L56-L58

which means that `Transpose(zeros(SMatrix{2, 2})) .+ zeros(SMatrix{2, 2})` just calls the generic broadcast code.

I'll try to improve the situation in a PR when I have the time, but I just wanted to assess the status quo first.

tkoolen:
Fixes #416.

* get rid of iffy Adjoint definition for 0.6.
* define TransposeVector and AdjointVector.
* just use RowVector instead of Adjoint in 0.6-only code.
* simplify Size code a bit.
* use both adjoint and transpose in tests.
* convert tests to use transpose instead of `.'`.

Passes almost all tests on master now.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -23,8 +23,10 @@ if VERSION < v"0.7-"
                         isposdef, normalize, normalize!, Eigen, det, logdet, cross, diff, qr
     const LinearAlgebra = Base.LinAlg
 
+    const TransposeVector{T, V<:AbstractVector{T}} = RowVector{T, V}
+    const AdjointVector{T, V<:AbstractVector{T}} = RowVector{T, ConjVector{T, V}}
+
     const adjoint = ctranspose
-    const Adjoint = RowVector
     const tr = trace
 else
     using Compat
@@ -40,6 +42,9 @@ else
                           kron, diag, vecnorm, norm, dot, diagm, lu, svd, svdvals, svdfact,
                           factorize, ishermitian, issymmetric, isposdef, normalize,
                           normalize!, Eigen, det, logdet, cross, diff, qr
+
+    const TransposeVector{T, V<:AbstractVector{T}} = Transpose{T, V}
+    const AdjointVector{T, V<:AbstractVector{T}} = Adjoint{T, V}
 end
 
 export StaticScalar, StaticArray, StaticVector, StaticMatrix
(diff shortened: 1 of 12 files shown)


andyferris:
Awesome!

We also need to consider #399. Will this work properly for static vectors of (normal) vectors? (And vice-versa?)

tkoolen:
These changes should be orthogonal to #399. Most things aren't specialized on the scalar type of the array (except where noted above), so I don't see why things shouldn't work for static vectors of vectors after merging both #399 and this PR.

PR #429, opened 2018-06-07
Fix #428, zeros and ones shouldn't use _fill

### LINKED ISSUE #428: zeros and ones shouldn't use _fill
https://github.com/JuliaArrays/StaticArrays.jl/blob/8a2ea499644900651a6cf8901a303526309d0eab/src/arraymath.jl#L12

results in problems with non-isbits types:

```julia
julia> v = zeros(SVector{2, BigInt})
2-element StaticArrays.SArray{Tuple{2},BigInt,1,2}:
 0
 0

julia> v[1] === v[2]
true
```

tkoolen:
Fixes #428. Pattern match `_fill` and `_rand` below.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -1,32 +1,28 @@
 @inline zeros(::SA) where {SA <: StaticArray} = zeros(SA)
-@generated function zeros(::Type{SA}) where {SA <: StaticArray}
+@inline zeros(::Type{SA}) where {SA <: StaticArray} = _zeros(Size(SA), SA)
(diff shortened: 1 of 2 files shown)


tkoolen:
Strange; nightly **passed** in 24 seconds after a fatal error.

https://travis-ci.org/JuliaArrays/StaticArrays.jl/jobs/389450045

tkoolen:
I actually just found out that Base has the same issue: https://github.com/JuliaLang/julia/issues/27483. Maybe doing the same thing as Base (the old behavior) is better until the behavior is changed in Base?

PR #433, opened 2018-06-08
fix FixedSizeArrays tests on v0.7

rdeits:
A couple things here: 

* A lot of the unary op tests in `test/fixed_size_arrays.jl` were silently failing because those functions have been moved to SpecialFunctions.jl, and the test itself is wrapped in a `try`
* I *maybe* fixed the random SArray generation? This is my first attempt at understanding the new way Random works. I didn't touch the `@inline rand(rng::AbstractRNG, range::AbstractArray, ::Type{SA}) where {SA <: StaticArray} = _rand(rng, range, Size(SA), SA)` method, which probably needs to be updated too.

### DIFF
--- a/src/FixedSizeArrays.jl
+++ b/src/FixedSizeArrays.jl
@@ -111,20 +111,20 @@ macro fixed_vector(name, parent)
         # Array constructor
         @inline function (::Type{$(name){S}})(x::AbstractVector{T}) where {S, T}
             @assert S <= length(x)
-            $(name){S, T}(ntuple(i-> x[i], Val{S}))
+            $(name){S, T}(ntuple(i-> x[i], Val{S}()))
         end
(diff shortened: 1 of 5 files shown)


src/FixedSizeArrays.jl
```diff
@@ -111,20 +111,20 @@ macro fixed_vector(name, parent)
         # Array constructor
         @inline function (::Type{$(name){S}})(x::AbstractVector{T}) where {S, T}
             @assert S <= length(x)
-            $(name){S, T}(ntuple(i-> x[i], Val{S}))
```
tkoolen: `Val(S)` is a bit more readable (that constructor is `@pure`, so there should not be any type inference issues).

tkoolen: `Val(S)` is a bit more readable (that constructor is `@pure`, so there should not be any type inference issues).

rdeits: ok done

PR #436, opened 2018-06-09
Reduce test time

### LINKED ISSUE #430: Tests take too long
We're getting timeouts on Travis. Could work around the problem using `travis_wait`, but I think it would be better to just reduce the time it takes to run the tests.

tkoolen:
This reduces total test time from about 16:30 to 6:20 on my machine. It also prints stuff more frequently for tests with a heavy compilation load, so that there's a smaller chance of having test timeouts.

The main win was not testing N = 50 for various tests due to https://github.com/JuliaArrays/StaticArrays.jl/issues/430#issuecomment-395977617. This approximately halved test time.

Fixes #430.


tkoolen:
Looks like Travis test time went from ~40 min to ~15 min.

tkoolen:
Test failure on appveyor: https://github.com/JuliaArrays/StaticArrays.jl/issues/438.

PR #441, opened 2018-06-09
Cast expm Pade approximant coefficients to result eltype, fix #438

### LINKED ISSUE #438: Legitimate test failure for exp on 32 bit windows
E.g. in https://ci.appveyor.com/project/andyferris/staticarrays-jl-b6vmi/build/1.0.725/job/1q1ejo2e46qx0gyk:

```
Matrix exponential: Error During Test
  Test threw an exception of type InexactError
  Expression: expm(@SMatrix([1 2 0; 2 1 0; 0 0 1])) ≈ expm([1 2 0; 2 1 0; 0 0 1])
  InexactError()
  Stacktrace:
   [1] _exp(::StaticArrays.Size{(3, 3)}, ::StaticArrays.SArray{Tuple{3,3},Int32,2,9}) at C:\Users\appveyor\.julia\v0.6\StaticArrays\src\expm.jl:106
   [2] macro expansion at C:\Users\appveyor\.julia\v0.6\StaticArrays\test\expm.jl:7 [inlined]
   [3] macro expansion at .\test.jl:
```

I believe this is due to  #424. @schmrlng, if you have time, could you take a look?

schmrlng:
Previously this code was casting large `Int64`s to `Int32` resulting in an `InexactError` (#438); the fix is to not cast to the input `eltype` `T` but instead to the result type `S = typeof((zero(T)*zero(T) + zero(T)*zero(T))/one(T))`. An alternative choice would have been `S = typeof(exp(zero(T)))` (used above in expm.jl), but I decided to stick with the `eltype` that would be returned by the algorithm at hand.

### DIFF
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -68,9 +68,9 @@ end
 
 # Adapted from implementation in Base; algorithm from
 # Higham, "Functions of Matrices: Theory and Computation", SIAM, 2008
-function _exp(::Size, A::StaticMatrix{<:Any,<:Any,T}) where T
+function _exp(::Size, _A::StaticMatrix{<:Any,<:Any,T}) where T
     S = typeof((zero(T)*zero(T) + zero(T)*zero(T))/one(T))
-    A = S.(A)
+    A = S.(_A)
     # omitted: matrix balancing, i.e., LAPACK.gebal!
     nA = maximum(sum(abs.(A), Val{1}))    # marginally more performant than norm(A, 1)
     ## For sufficiently small nA, use lower order Padé-Approximations
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -69,47 +69,49 @@ end
 # Adapted from implementation in Base; algorithm from
 # Higham, "Functions of Matrices: Theory and Computation", SIAM, 2008
 function _exp(::Size, A::StaticMatrix{<:Any,<:Any,T}) where T
+    S = typeof((zero(T)*zero(T) + zero(T)*zero(T))/one(T))
+    A = S.(A)
     # omitted: matrix balancing, i.e., LAPACK.gebal!
     nA = maximum(sum(abs.(A), Val{1}))    # marginally more performant than norm(A, 1)
     ## For sufficiently small nA, use lower order Padé-Approximations
     if (nA <= 2.1)
         A2 = A*A
         if nA > 0.95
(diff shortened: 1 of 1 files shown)


tkoolen:
Thank you!

schmrlng:
Looking more closely at `Int` `expm`, there seem to be allocations for some reason -- I'll try to track these down later today, but at least for now this patch might allow the abbreviated test set PR #436 to proceed.

```
julia> A = @SMatrix([1 2 0; 2 1 0; 0 0 1])
3×3 StaticArrays.SArray{Tuple{3,3},Int64,2,9}:
 1  2  0
 2  1  0
 0  0  1

julia> @benchmark expm($SA)
BenchmarkTools.Trial:
  memory estimate:  8.53 KiB
  allocs estimate:  42
  --------------
  minimum time:     3.365 μs (0.00% GC)
  median time:      3.641 μs (0.00% GC)
  mean time:        4.106 μs (6.52% GC)
  maximum time:     128.480 μs (86.49% GC)
  --------------
  samples:          10000
  evals/sample:     8
```

schmrlng:
The above issue was with type instability in this code (`A` could end up as type `S` or stay `Int`, depending on the conditional):
```
         if s > 0
             si = ceil(Int,s)
             A = A / S(2^si)
         end
```
Fixed by converting `A` up front to `S.(A)`.

src/expm.jl
```diff
@@ -69,47 +69,49 @@ end
 # Adapted from implementation in Base; algorithm from
 # Higham, "Functions of Matrices: Theory and Computation", SIAM, 2008
 function _exp(::Size, A::StaticMatrix{<:Any,<:Any,T}) where T
+    S = typeof((zero(T)*zero(T) + zero(T)*zero(T))/one(T))
+    A = S.(A)
```
andyferris: You'll need to rename this variable to make this fast on v0.6 (this inference "bug" has been fixed in v0.7, yay).

SimonDanisch:
Hope it's okay that I pushed @andyferris changes directly to this branch :)

SimonDanisch:
Is this ready to merge when test pass?

schmrlng:
Yup, I think this is ready to merge. From benchmarking on 0.6 I couldn't actually find a case where the variable rename made a difference (I tried `SMatrix`, `MMatrix` with eltypes `Int` and `Float64`), but I guess it can't hurt!

PR #442, opened 2018-06-12
scalar indexing broadcast fix

SimonDanisch:


src/broadcast.jl
```diff
@@ -135,6 +135,10 @@ end
     end
 end
 
+scalar_getindex(x) = x
+scalar_getindex(x::Ref) = x[]
+scalar_getindex(x::Tuple{<: Any}) = x[1]
```
andyferris: Is the `<:` necessary in a covariant parameter?

martinholters: No, shouldn't be, as `Tuple{<:Any} == Tuple{Any}`.

SimonDanisch: True, I just made a habbit out of always using this pattern...

PR #448, opened 2018-06-23
Activate helpful error message for undefined-Size conversions

timholy:
Adding `S<:Tuple` ensured a method would be higher in the methodtable.

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -68,7 +68,7 @@ end
 
 Base.show(io::IO, ::Size{S}) where {S} = print(io, "Size", S)
 
-#= There seems to be a subtyping/specialization bug...
+Size(a::T) where {T<:AbstractArray} = Size(T)
 function Size(::Type{SA}) where {SA <: StaticArray} # A nice, default error message for when S not defined
     error("""
         The size of type `$SA` is not known.
@@ -82,9 +82,8 @@ function Size(::Type{SA}) where {SA <: StaticArray} # A nice, default error mess
             SMatrix(m)      # this error
             SMatrix{3,3}(m) # correct - size is inferrable
         """)
-end =#
-Size(a::T) where {T<:AbstractArray} = Size(T)
-Size(::Type{<:StaticArray{S}}) where {S} = Size(S)
+end
+Size(::Type{SA}) where {SA <: StaticArray{S}} where {S<:Tuple} = Size(S)  # S defined as a Tuple
 @pure Size(::Type{<:AbstractArray{<:Any, N}}) where {N} = Size(ntuple(_ -> Dynamic(), N))
 
 struct Length{L}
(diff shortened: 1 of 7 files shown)


timholy:
Femtocleaner has a number of useful suggestions. OK to accept? Or would you rather keep this tiny?

andyferris:
Cool, thanks Tim!

Yes I generally love femtocleaner- bring it on :)

src/traits.jl
```diff
@@ -82,9 +82,8 @@ function Size(::Type{SA}) where {SA <: StaticArray} # A nice, default error mess
             SMatrix(m)      # this error
             SMatrix{3,3}(m) # correct - size is inferrable
         """)
-end =#
-Size(a::T) where {T<:AbstractArray} = Size(T)
-Size(::Type{<:StaticArray{S}}) where {S} = Size(S)
+end
+Size(::Type{SA}) where {SA <: StaticArray{S}} where {S<:Tuple} = Size(S)  # S defined as a Tuple
```
fredrikekre: This gives
```
julia> convert(SVector, [1,2,3])
ERROR: UndefVarError: S not defined
Stacktrace:
 [1] Size(::Type{SArray{Tuple{S},T,1,S} where T where S}) at /Users/Fredrik/.julia/packages/StaticArrays/Jd49/src/traits.jl:86
 [2] length(::Type{SArray{Tuple{S},T,1,S} where T where S}) at /Users/Fredrik/.julia/packages/StaticArrays/Jd49/src/abstractarray.jl:2
 [3] convert(::Type{SArray{Tuple{S},T,1,S} where T where S}, ::Array{Int64,1}) at /Users/Fredrik/.julia/packages/StaticArrays/Jd49/src/convert.jl:22
 [4] top-level scope at none:0
```

timholy: Interesting. We tested this with 0.6, where it works as intended.

 I wonder if this is actually a dispatch bug in 0.7. That line explicitly requires that `S<:Tuple`, and when `S` is undefined I don't see how that statement could be true.

timholy: Hmmm, it *should* work...
```julia
julia> abstract type AType{T} end

julia> foo(::Type{A}) where A<:AType = "parameter missing"

julia> foo(::Type{A}) where A<:AType{T} where T<:Tuple = "parameter present"
foo (generic function with 2 methods)

julia> struct C{T} <: AType{T} end

julia> foo(C)
"parameter missing"

julia> foo(C{Tuple{1,2}})
"parameter present"
```

martinholters: I don't what the exact problem is here, but note that, continuing with your definitions:
```julia
julia> foo(C{Union{T,Tuple}} where T <: Tuple)
"parameter present"

julia> foo(::Type{A}) where A<:AType{T} where T<:Tuple = "parameter present: $T"
foo (generic function with 2 methods)

julia> foo(C{Union{T,Tuple}} where T <: Tuple)
ERROR: UndefVarError: T not defined
```
So it's always safer to add an `@isdefined T` conditional.

fredrikekre: Opened #452 to track this.

PR #451, opened 2018-06-28
fixes for 0.7

fredrikekre:
Sorry for the big diff...

Anyhow, most things work without any deprecations.


src/linalg.jl
```diff
@@ -353,7 +342,7 @@ end
     end
 end
 
-const _length_limit = Length(200)
```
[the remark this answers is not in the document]
fredrikekre: Aaarghh. This was the reason for the failed inference tests of `kron` that I have been trying to track down for a while...

test/chol.jl
```diff
@@ -1,49 +1,50 @@
...
+        @test cholesky(m).U ≈ cholesky(m_a).U
+        @test cholesky(Hermitian(m)).L ≈ cholesky(m_a).L
+        @test cholesky(Hermitian(m)).U ≈ cholesky(m_a).U
     end
     @testset "static blockmatrix" for i = 1:10
         m_a = randn(3,3)
         m_a = m_a*m_a'
         m = SMatrix{3,3}(m_a)
-        @test chol(reshape([m, 0m, 0m, m], 2, 2)) ==
+        @test_broken cholesky(reshape([m, 0m, 0m, m], 2, 2)) ==
```
fredrikekre: This `cholesky` call causes a call to `hermitian_type(::Type{Any}) which fails. @martinholters you implemented that, any idea why this happens?

andyferris:
Wow, awesome.

@fredrikekre FYI we are planning at least one release that works on both Julia v0.6 and v1.0, to help people migrate there code since this is a pretty foundational package for some people and it will already be painful to migrate code in some instances.

(Note: After that, I couldn't run away from v0.6 support quickly enough!).

fredrikekre:
> one release that works on both Julia v0.6 and v1.0

The changes here are mostly non-functional, and the breaking parts are the same as in Base, I think it is better that they break at the same time instead. Why is is important the there is a single StaticArrays release that works on both?

> since this is a pretty foundational package

This is why I spent time on this PR, StaticArrays beeing broken blocks upgrading of other packages, and I think we need a v0.7 compatible StaticArrays release ASAP to aid other packages in upgrading.

test/chol.jl
martinholters: `cholesky` calls `cholcopy`, which gives a `Matrix{Any}`. Probably `choltype`, responsible for figuring out the element type, should be made recursive. Note that `cholesky(reshape([m_a, 0m_a, 0m_a, m_a], 2, 2))` also fails for the same underlying reason, except that here, `choltype` fails itself. Probably worth an issue in Julia.

src/StaticArrays.jl
```diff
@@ -10,42 +10,17 @@ import Base: getindex, setindex!, size, similar, vec, show, length, convert, pro
...
+import Random: rand, randn, randexp, rand!, randn!, randexp!
+using Core.Compiler: return_type
+import Base: sqrt, exp, log
+using LinearAlgebra
+import LinearAlgebra: transpose, adjoint, eye, vecdot, eigvals, eigen, lyap, tr,
+                      kron, diag, norm, dot, diagm, lu, svd, svdvals, svdfact,
+                      factorize, ishermitian, issymmetric, isposdef, normalize,
+                      normalize!, Eigen, det, logdet, cross, diff, qr
+const TransposeVector{T, V<:AbstractVector{T}} = Transpose{T, V}
+const AdjointVector{T, V<:AbstractVector{T}} = Adjoint{T, V}
```
tkoolen: These were put into place to make it possible to support both 0.6 and 0.7. We should probably remove them for a 0.7-only release.

src/deque.jl
```diff
@@ -43,8 +44,9 @@ end
     end
 end
 
-@inline shift(vec::StaticVector) = _shift(Size(vec), vec)
-@generated function _shift(::Size{s}, vec::StaticVector) where {s}
+@deprecate shirt popfirst
```
tkoolen: `shift`

src/io.jl
```diff
@@ -1,9 +1,9 @@
 
-@inline function read(io::IO, ::Type{SA}) where {SA<:StaticArray}
-    # Copy Base implementation of `read` for primitive types.  This is less
+@inline function read!(io::IO, ::Type{SA}) where {SA<:StaticArray}
+    # Copy Base implementation of `read!` for primitive types.  This is less
     # efficient in 0.6 that we'd like because creating the Ref allocates.
```
tkoolen: 0.6-specific comment can be removed.

src/mapreduce.jl
```diff
@@ -4,24 +4,14 @@
...
 
-if VERSION < v"0.7.0-"
-    # The following type signature for map() matches any list of AbstractArrays,
-    # provided at least one is a static array.
-    @inline function map(f, as::Union{SA,AbstractArray}...) where {SA<:StaticArray}
-        _map(f, same_size(as...), as...)
-    end
-else
-    # In 0.7, the above construction can no longer be used due to https://github.com/JuliaLang/julia/pull/23117.
-    # Instead, only dispatch to StaticArrays._map if one of the first two arguments is a StaticArray.
```
tkoolen: Perhaps keep a version of this comment around.

test/eigen.jl
```diff
@@ -102,7 +104,7 @@
         v = randn(SVector{3,Float64})
         m = v*v'
         vv = sum(abs2, v)
-        vals, vecs = eig(m)::Tuple{SVector,SMatrix}
+        vals, vecs = eigen(m)
```
tkoolen: `::Tuple{SVector,SMatrix}` was removed. How can we be sure this is not calling generic code?

test/lu.jl
```diff
@@ -1,8 +1,9 @@
-using StaticArrays, Compat.Test
+using StaticArrays, Test, LinearAlgebra
 
+@testset "LU decomposition" begin
```
tkoolen: This was removed specifically to have more stuff printed, since we were running into travis timeouts. Other changes were made at the time to reduce test time, so maybe this is OK, but I'm just a little worried about travis flakiness.

test/lu.jl
```diff
@@ -1,8 +1,9 @@
...
 @testset "LU decomposition ($m×$n, pivot=$pivot)" for pivot in (true, false), m in [0:4..., 15], n in [0:4..., 15]
```
tkoolen: Fix indentation if we do make this a nested `testset`.

tkoolen:
Just my opinion: I considered 0.7.2 to already be (almost) the bridging release that supports both 0.6 and 0.7 (with deprecation warnings in some cases). All the errors on 0.7 had already been fixed in that release, except for those in the QR code. 

Perhaps fixing the QR code should be done in a separate PR, merged before this, and one more release could be made with that change in place, as a better bridge release. But having a release that supports both 0.6 and 0.7 without depwarns is probably a lot of work on top of that for little gain.

andyferris:
> Perhaps fixing the QR code should be done in a separate PR, merged before this, and one more release could be made with that change in place, as a better bridge release

Is there something we need to do to address the changes to `eig`, `cholesky` and so-on? (Since the now need to return a `Factorization`)

src/lu.jl
```diff
@@ -1,14 +1,27 @@
+# define our own LU type, since LinearAlgebra.LU requires p::Vector
+struct LU{L,U,p}
+    L::L
+    U::U
+    p::p
+end
```
tkoolen: `L::L`, `U::U`, `p::p` seems like it could get confusing.

src/mapreduce.jl
tkoolen: Has been addressed.

src/qr.jl
```diff
@@ -1,36 +1,54 @@
-_thin_must_hold(thin) =
-    thin || throw(ArgumentError("For the sake of type stability, `thin = true` must hold."))
+# define our own struct since LinearAlgebra.QR are restricted to Matrix
+struct QR{Q,R}
+    Q::Q
+    R::R
+end
```
tkoolen: Same comment as for `LU` here.

test/qr.jl
```diff
@@ -13,57 +13,36 @@ srand(42)
...
-        @test p isa StaticVector
-
-        Q_ref,R_ref, p_ref = qr(Matrix(arr), pivot)
-        @test Q ≈ Q_ref
-        @test R ≈ R_ref
-        @test p == p_ref
+        # Q_ref,R_ref, p_ref = qr(Matrix(arr), pivot)
+        # @test Q ≈ Q_ref
+        # @test R ≈ R_ref
+        # @test p == p_ref
```
tkoolen: Should this be uncommented again?

src/lu.jl
andyferris: Interesting problem. This normally is avoided by the fact that we use upper case types and lower case fields. `LinearAlgebra` is particularly *nasty* at ignoring this advice, leading to this kind of... stuff.

The interface for this factorization from `LinearAlgebra` seems to involve defining properties `L` and `U`, so we can't ignore that. The type parameters for the type needn't suffer terribly either - do we really want `LU{LT, UT, pT}`? `LU{L_t, U_t, p_t}`? Perhaps what's here currently is the least of all evils.

src/svd.jl
```diff
@@ -9,7 +9,19 @@ struct SVD{T,TU,TS,TVt} <: Factorization{T}
 end
 SVD(U::AbstractArray{T}, S::AbstractVector, Vt::AbstractArray{T}) where {T} = SVD{T,typeof(U),typeof(S),typeof(Vt)}(U, S, Vt)
 
-getindex(::SVD, ::Symbol) = error("In order to avoid type instability, StaticArrays.SVD doesn't support indexing the output of svdfact with a symbol.  Instead, you can access the fields of the factorization directly as f.U, f.S, and f.Vt")
+function Base.getproperty(F::SVD, s::Symbol)
```
andyferris: `@inline` to guarantee constant propagation of `s`? (The `'` below could expand to be a large amount of code, preventing inlining, yet we need this function to be type-inferred)

test/chol.jl
andyferris: Let's deal with the universe of block matrices afterwards.

src/lu.jl
andyferris: OTOH our `SVD` type already uses ugly type parameters... ¯\_(ツ)_/¯

andyferris:
I created a tag. Hopefully the hungry hordes will have there release soon. (After METADATA still needs to filter through to Pkg3 registry...)

andyferris:
https://github.com/JuliaLang/METADATA.jl/pull/15400

test/qr.jl
fredrikekre: Yes, I did not get them to work originally, and thought it had something to do with the "not released" comment just above. But I found a bug so will fix that and activate those tests.

tkoolen:
Great to have this done, thanks again @fredrikekre.

On the topic of dropping support for 0.6: I'm quite happy with the transition this time. I think what happened last time was that support for 0.5 was (had to be?) dropped as part of a very complex PR. That PR was merged three months before the official release of 0.6, after which additional fairly fundamental changes were made on the master branch that people (at least me) couldn't adapt to gradually, because so many things were still broken on Julia master. And when finally all the bugs in the language had been ironed out, I seem to remember subtle performance problems that snuck in and required a labor-intensive debugging process that involved investigating more code changes than I would have liked.

So I think what's better this time around is that:

* the upgrade from 0.6 to 0.7 is much more gradual by itself (the whole type system didn't change this time)
* this allowed us to keep up to date with most 0.7 changes in a series of smaller PRs throughout the last months while retaining support for 0.6, as opposed to one big one
* as a result, this PR, while it touches a lot of files, isn't very complex

src/lu.jl
tkoolen: Yeah, it's probably alright. It was surprising to me at least since I hadn't seen this pattern before, but I guess you're right that the param names won't leak out to user code.

fredrikekre: Yea, I always assumed this was not valid, and usually use this pattern if needed
```
struct A{TT}
    T::TT
end
```

PR #453, opened 2018-06-29
Fixes to new StaticArrays.QR

fredrikekre:


src/qr.jl
```diff
@@ -45,12 +59,12 @@ _qreltype(::Type{T}) where T = typeof(zero(T)/sqrt(abs2(one(T))))
     SizeQ = Size( sA[1], diagsize(Size(A)) )
     SizeR = Size( diagsize(Size(A)), sA[2] )
 
-    if pivot == Val(true)
+    if pivot === Val{true}
```
fredrikekre: We are in type-domain here so we never hitted this branch... :)

test/qr.jl
```diff
@@ -27,20 +27,19 @@ srand(42)
...
-        # # pivot=true cases are not released yet
-        # pivot = Val(true)
-        # QRp = @inferred qr(arr, pivot)
-        # @test QRp isa Tuple
-        # @test length(QRp) == 3
-        # Q, R, p = QRp
-        # @test Q isa StaticMatrix
-        # @test R isa StaticMatrix
-        # @test p isa StaticVector
+        # pivot=true cases are not released yet
```
tkoolen: Is this comment true/helpful at this point?

fredrikekre: I don't know what it means -- perhaps that there is not released StaticArray-specific version and that we just call to LAPACK? I can add that instead

tkoolen: I don't know either. Looks like it was added by @denius in https://github.com/JuliaArrays/StaticArrays.jl/pull/328. Probably your interpretation is right.

fredrikekre:
Failures should be fixed by https://github.com/JuliaLang/julia/pull/27862, note that the failures are in the test code where we compare with regular Arrays, and everything still works for StaticArrays.

fredrikekre:
Good to go? Would be good to merge this and retag https://github.com/JuliaLang/METADATA.jl/pull/15400

PR #454, opened 2018-06-29
import mean from Statistics

fredrikekre:
I wonder if we still need this specialized mean now that it requires an extra dependency. This dependency does not cost anything while Statistics comes with Julia, but might be something to consider removing in the future.


fredrikekre:
Also, should wait with mergin this until Pkg3 knows about the Statistics stdlib.

PR #460, opened 2018-07-04
Add general Tuple(::StaticArray) constructor and remove related convert method

### LINKED ISSUE #446: Performance of typeof(x)(x) for x::FieldVector
I don't think this is a dupe of #356, but it seems along the same lines: 
```julia
  | | |_| | | | (_| |  |  Version 0.6.2 (2017-12-13 18:08 UTC)
 _/ |\__'_|_|_|\__'_|  |  

julia> struct Point3D <: FieldVector{3, Float64}
           x::Float64
           y::Float64
           z::Float64
       end

julia> q = Point3D(1.,2.,3.)
3-element Point3D:
 1.0
 2.0
 3.0

julia> f(x) = typeof(x)(x)

julia> @benchmark f($q)
BenchmarkTools.Trial: 
  memory estimate:  240 bytes
  allocs estimate:  7
  --------------
  minimum time:     336.826 ns (0.00% GC)
  median time:      358.016 ns (0.00% GC)
  mean time:        376.296 ns (4.28% GC)
  maximum time:     8.890 μs (93.98% GC)
  --------------
  samples:          10000
  evals/sample:     218

julia> @code_warntype f(q)
Variables:
  #self# <optimized out>
  x::Point3D
  itr::Base.OneTo{Int64}

Body:
  begin 
      $(Expr(:inbounds, false))
      # meta: location /home/schmrlng/.julia/v0.6/StaticArrays/src/convert.jl Type 4
      # meta: location tuple.jl Type 244
      # meta: location abstractarray.jl start 759
      itr::Base.OneTo{Int64} = $(Expr(:new, Base.OneTo{Int64}, :((Base.select_value)((Base.slt_int)(3, 0)::Bool, 0, 3)::Int64)))
      # meta: pop location
      # meta: pop location
      # meta: pop location
      $(Expr(:inbounds, :pop))
      return (Point3D)((Core._apply)(Core.tuple, $(Expr(:invoke, MethodInstance for _collect(::UnitRange{Int64}, ::Base.Iterators.Rest{Point3D,Tuple{Base.OneTo{Int64},Int64}}, ::Base.HasEltype, ::Base.SizeUnknown), :(Base._collect), :($(Expr(:new, UnitRange{Int64}, 1, :((Base.select_value)((Base.sle_int)(1, 1)::Bool, 1, (Base.sub_int)(1, 1)::Int64)::Int64)))), :($(Expr(:new, Base.Iterators.Rest{Point3D,Tuple{Base.OneTo{Int64},Int64}}, :(x), :((Core.tuple)(itr, 1)::Tuple{Base.OneTo{Int64},Int64})))), :($(QuoteNode(Base.HasEltype()))), :($(QuoteNode(Base.SizeUnknown()))))))::Tuple)::Point3D
  end::Point3D
```
where I'm not sure if it's meaningful but the ``::Tuple`` at the end of the last line of `@code_warntype f(q)` is highlighted in red.

schmrlng:
Fixes #446. The issue seems to be that `Tuple(::StaticArray)` does not fall back to `convert` as defined in the lines above this proposed change (see below) so one fix is to make that fallback explicit.
```julia
  | | |_| | | | (_| |  |  Version 0.7.0-beta.125 (2018-07-02 20:55 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit c5bf467361 (1 day old master)

julia> using BenchmarkTools, StaticArrays

julia> struct Point3D <: FieldVector{3, Float64}
           x::Float64
           y::Float64
           z::Float64
       end

julia> p = rand(Point3D)
3-element Point3D:
 0.6628326877139248
 0.14765795872117993
 0.48470441802594144

julia> @benchmark Point3D($p)
BenchmarkTools.Trial:
  memory estimate:  224 bytes
  allocs estimate:  6
  --------------
  minimum time:     342.486 ns (0.00% GC)
  median time:      355.670 ns (0.00% GC)
  mean time:        387.416 ns (7.73% GC)
  maximum time:     144.887 μs (99.70% GC)
  --------------
  samples:          10000
  evals/sample:     212

julia> @which Tuple(p)
(::Type{T})(itr) where T<:Tuple in Base at tuple.jl:242

julia> @inline Tuple(a::StaticArray) = convert(Tuple, a)
Tuple

julia> @benchmark Point3D($p)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.302 ns (0.00% GC)
  median time:      1.312 ns (0.00% GC)
  mean time:        1.312 ns (0.00% GC)
  maximum time:     17.192 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

```

### DIFF
--- a/test/FieldVector.jl
+++ b/test/FieldVector.jl
@@ -18,6 +18,8 @@
         @test length(p) === 3
         @test eltype(p) === Float64
 
+        @testinf Tuple(p) === (1.0, 2.0, 3.0)
+
         @test (p + p) === Point3D(2.0, 4.0, 6.0)
 
         @test (p[1], p[2], p[3]) === (p.x, p.y, p.z)
@@ -58,6 +60,8 @@
         @test size(p) === (2,)
         @test length(p) === 2
         @test eltype(p) === Float64
+
+        @testinf Tuple(p) === (1.0, 2.0)
 
         @test (p[1], p[2]) === (p.x, p.y)
         @test (p[1], p[2]) === (1.0, 2.0)
--- a/test/core.jl
+++ b/test/core.jl
@@ -149,8 +149,6 @@
     @test StaticArrays.check_length(2) == nothing
     @test StaticArrays.check_length(StaticArrays.Dynamic()) == nothing
 
-    @test convert(Tuple, @SVector [2]) == (2,)
-
     @testset "dimmatch" begin
         @test StaticArrays.dimmatch(3, 3)
         @test StaticArrays.dimmatch(3, StaticArrays.Dynamic())
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -9,10 +9,8 @@
 @inline convert(::Type{SA}, sa::SA) where {SA<:StaticArray} = sa
 @inline convert(::Type{SA}, x::Tuple) where {SA<:StaticArray} = SA(x) # convert -> constructor. Hopefully no loops...
 
-# A general way of going back to a tuple
-@inline function convert(::Type{Tuple}, a::StaticArray)
-    unroll_tuple(a, Length(a))
-end
+# Constructing a Tuple from a StaticArray
+@inline Tuple(a::StaticArray) = unroll_tuple(a, Length(a))
 
 @noinline function dimension_mismatch_fail(SA::Type, a::AbstractArray)
     error("Dimension mismatch. Expected input array of length $(length(SA)), got length $(length(a))")
--- a/test/MArray.jl
+++ b/test/MArray.jl
@@ -101,7 +101,7 @@
         @test m[3] === 13
         @test m[4] === 14
 
-        @test Tuple(m) === (11, 12, 13, 14)
+        @testinf Tuple(m) === (11, 12, 13, 14)
 
         @test size(m) === (2, 2)
         @test size(typeof(m)) === (2, 2)
--- a/test/MMatrix.jl
+++ b/test/MMatrix.jl
@@ -76,7 +76,7 @@
         @test m[3] === 13
         @test m[4] === 14
 
-        @test Tuple(m) === (11, 12, 13, 14)
+        @testinf Tuple(m) === (11, 12, 13, 14)
 
         @test size(m) === (2, 2)
         @test size(typeof(m)) === (2, 2)
--- a/test/MVector.jl
+++ b/test/MVector.jl
@@ -56,7 +56,7 @@
         @test v[2] === 12
         @test v[3] === 13
 
-        @test Tuple(v) === (11, 12, 13)
+        @testinf Tuple(v) === (11, 12, 13)
 
         @test size(v) === (3,)
         @test size(typeof(v)) === (3,)
--- a/test/SArray.jl
+++ b/test/SArray.jl
@@ -96,7 +96,7 @@
         @test m[3] === 13
         @test m[4] === 14
 
-        @test Tuple(m) === (11, 12, 13, 14)
+        @testinf Tuple(m) === (11, 12, 13, 14)
 
         @test size(m) === (2, 2)
         @test size(typeof(m)) === (2, 2)
--- a/test/SMatrix.jl
+++ b/test/SMatrix.jl
@@ -74,7 +74,7 @@
         @test m[3] === 13
         @test m[4] === 14
 
-        @test Tuple(m) === (11, 12, 13, 14)
+        @testinf Tuple(m) === (11, 12, 13, 14)
 
         @test size(m) === (2, 2)
         @test size(typeof(m)) === (2, 2)
--- a/test/SVector.jl
+++ b/test/SVector.jl
@@ -55,7 +55,7 @@
         @test v[2] === 12
         @test v[3] === 13
 
-        @test Tuple(v) === (11, 12, 13)
+        @testinf Tuple(v) === (11, 12, 13)
 
         @test size(v) === (3,)
         @test size(typeof(v)) === (3,)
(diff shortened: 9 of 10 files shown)


schmrlng:
Bump - could there be any unintended consequences of adding this method? My impression is that wherever `Tuple(a)` appears, e.g., in lines 4 and 8 [here](https://github.com/JuliaArrays/StaticArrays.jl/blob/5087eca0ade0fa106ff30bf98ac5e2aca8f65994/src/convert.jl), `convert(Tuple, a)` is implicitly meant.

andyferris:
Sorry... I've been stuck wondering if the constructor or `convert` is correct.

Thinking aloud, I think `covert` is meant to make something equivalent, e.g. via `==` (which is true for conversion between `Number`s and between `AbstractArray`s at least). Since `Tuple`s are never equal to `AbstractArray`s it's probably best to avoid `convert(Tuple, static_array)` entirely throughout the code base. Do you agree?

schmrlng:
That seems reasonable. Notably, `convert(::Type{Tuple}, ::AbstractArray)` doesn't exist in Base (currently, we're hitting the `Tuple` constructor [here](https://github.com/JuliaLang/julia/blob/b6a60dd6601a6947e00a0a77decd283248ba8a7f/base/tuple.jl#L243) without this PR).

It seems that converting to `Tuple` is something that's being done as a convenient intermediate in converting arbitrary `StaticArray` subtypes between each other. Maybe that intermediate should be an `SArray` instead (perhaps even if one or both of the input/output types are mutable). `SArray` is of course just a wrapper around an `NTuple`, but the whole point of the wrapper (and this package) is for an `SArray` to behave like/subtype `AbstractArray`, including making `convert` make sense as you argue. Hopefully the compiler can easily get around the extra layer of indirection in converting to an `SArray` intermediate type instead of a bare `Tuple` intermediate type.

Other StaticArray -> StaticArray conversions (e.g., #356) are probably worth thinking about as well here. Always converting to `SArray` as an intermediate looks reminiscent of `convert_similar_type` proposed in #357; alternative names could maybe include `canonical_type` or `sarray_type`. `convert_similar_type` in #357 only covers `SArray` and `MArray`, but I'm not sure if there's a technical reason it can't be extended to arbitrary `StaticArray`s including `FieldVector`s (though you seem to have thought about it, e.g., https://github.com/JuliaArrays/StaticArrays.jl/pull/357#issuecomment-381476500). Naively I was hoping something like
```
canonical_type(::Type{StaticArray{S,T,N}}) where {S,T,N} = SArray{S,T,N}
canonical_type(::Type{SA}) where {SA<:StaticArray} = canonical_type(supertype(SA))
```
would work, but it doesn't.

schmrlng:
Actually something like this seems to do what I was going for in the previous post, but I'm not certain it's actually what we want.
```
@generated canonical_type(::Type{SA}) where {S,T,N,SA<:StaticArray{S,T,N}} = :(SArray{S, T, N, $(StaticArrays.tuple_prod(S))})
```

schmrlng:
Alternatively it seems like just removing the `convert` method and leaving only the `Tuple` constructor is good enough (the codebase already does a good job of sticking with the `Tuple` constructor, avoiding the `convert` method). Also I don't see either #446 or #356 anymore with this change on 0.7 (I think #356 was fixed separately at some point):
```julia
  | | |_| | | | (_| |  |  Version 0.7.0-beta2.6 (2018-07-14 20:16 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit faee1df893 (3 days old master)

julia> m = rand(SMatrix{3, 3})
 0.482697  0.447041  0.161918
 0.409194  0.400836  0.00193097
 0.551338  0.51259   0.886981

julia> @btime SMatrix($m)
  2.594 ns (0 allocations: 0 bytes)
 0.482697  0.447041  0.161918
 0.409194  0.400836  0.00193097
 0.551338  0.51259   0.886981

julia> struct Point3D <: FieldVector{3, Float64}
           x::Float64
           y::Float64
           z::Float64
       end

julia> q = rand(Point3D)
3-element Point3D:
 0.9549383392574675
 0.3309816545097055
 0.8245079859261837

julia> @btime Point3D($q)
  1.562 ns (0 allocations: 0 bytes)
3-element Point3D:
 0.9549383392574675
 0.3309816545097055
 0.8245079859261837
```

tkoolen:
See also https://github.com/JuliaArrays/StaticArrays.jl/pull/357. Any more thoughts on that by the way?

schmrlng:
Without necessitating any rewrites (though the approach in #357 may still be desirable for other reasons) do you still see the performance problem on 0.7? At least the example at the top of #356 seems fixed for me even without this PR - maybe the 0.7 compiler is just that much better at inference.

tkoolen:
Yeah, #356 seems to be fixed. It's still calling the same constructor, so I think it is indeed due to compiler improvements.

test/core.jl
```diff
@@ -149,8 +149,6 @@
     @test StaticArrays.check_length(2) == nothing
     @test StaticArrays.check_length(StaticArrays.Dynamic()) == nothing
 
-    @test convert(Tuple, @SVector [2]) == (2,)
```
andyferris: Is there still a test of the new method?

test/FieldVector.jl
```diff
@@ -18,6 +18,8 @@
         @test length(p) === 3
         @test eltype(p) === Float64
 
+        @testinf Tuple(p) === (1.0, 2.0, 3.0)
+
```
schmrlng: Without this PR, this test and the one below fail.

PR #461, opened 2018-07-04
Fix #388, hcat allocations with Julia master

### LINKED ISSUE #388: hcat allocations with Julia master
```julia
 x = zeros(SVector{3})
@btime hcat($x, $x, $x, $x, $x, $x)
```

On 0.6.2:

```julia
  8.691 ns (0 allocations: 0 bytes)
3×6 StaticArrays.SArray{Tuple{3,6},Float64,2,18}:
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
```

On 8 day old master:

```julia
455.411 ns (10 allocations: 688 bytes)
3×6 SArray{Tuple{3,6},Float64,2,18}:
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
```

3x5 is fine, so must be reaching a compiler limit.

tkoolen:
Apparently this is a bit easier on the compiler.

(also just use `StaticVecOrMat` instead of `Union{StaticVector,StaticMatrix}`)

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -88,9 +88,11 @@ end
     end
 end
 
-@inline vcat(a::Union{StaticVector,StaticMatrix}) = a
-@inline vcat(a::Union{StaticVector, StaticMatrix}, b::Union{StaticVector,StaticMatrix}) = _vcat(Size(a), Size(b), a, b)
-@generated function _vcat(::Size{Sa}, ::Size{Sb}, a::Union{StaticVector, StaticMatrix}, b::Union{StaticVector,StaticMatrix}) where {Sa, Sb}
+@inline vcat(a::StaticVecOrMat) = a
+@inline vcat(a::Union{StaticVector, StaticMatrix}, b::StaticVecOrMat) = _vcat(Size(a), Size(b), a, b)
+@inline vcat(a::StaticVecOrMat, b::StaticVecOrMat, c::StaticVecOrMat...) =
+    vcat(vcat(a,b), vcat(c...))
+@generated function _vcat(::Size{Sa}, ::Size{Sb}, a::StaticVecOrMat, b::StaticVecOrMat) where {Sa, Sb}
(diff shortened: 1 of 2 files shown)


tkoolen:
Note that tests only pass on nightly because of https://github.com/JuliaLang/julia/pull/27398, but I think we were running into a different limit before this change, since even on nightly inference gives up after 5 arguments before this PR.

src/linalg.jl
```diff
@@ -88,9 +88,11 @@ end
     end
 end
 
-@inline vcat(a::Union{StaticVector,StaticMatrix}) = a
-@inline vcat(a::Union{StaticVector, StaticMatrix}, b::Union{StaticVector,StaticMatrix}) = _vcat(Size(a), Size(b), a, b)
-@generated function _vcat(::Size{Sa}, ::Size{Sb}, a::Union{StaticVector, StaticMatrix}, b::Union{StaticVector,StaticMatrix}) where {Sa, Sb}
+@inline vcat(a::StaticVecOrMat) = a
+@inline vcat(a::Union{StaticVector, StaticMatrix}, b::StaticVecOrMat) = _vcat(Size(a), Size(b), a, b)
```
andyferris: Missed one `Union` minification here

andyferris:
I'm ok with merging this - but the METATDATA people mightn't be happy about the failing tests if we want to tag.

src/linalg.jl
tkoolen: Fixed.

tkoolen:
Made the number of arguments in the test version-dependent (v"0.7.0-beta.47" is the first version to include https://github.com/JuliaLang/julia/pull/27398). Femtocleaner can clean that up later.

PR #464, opened 2018-07-07
export qr, and properly import it

### LINKED ISSUE #463: `qr` has to be exported?...
I am having some code, which starts with 
`using StaticArrays` and somewhere later it does
```
QR = qr(thingy)
Q = QR.Q; R = QR.R
```
I can *guarantee* that the `thingy` is always an `SMatrix`. When I run my code I get a huge amount of 
```
WARNING: Base.qr is deprecated: it has been moved to the standard library package `LinearAlgebra`.
Add `using LinearAlgebra` to your imports.
 in module Main
```
even though I do not use the `LinearAlgebra` version of `qr`. Since the `qr` functionality is provided by `StaticArrays`, I don't think it should require me to also be `using LinearAlgebra`.

Datseris:
Hopefully will close #463

### DIFF
--- a/src/qr.jl
+++ b/src/qr.jl
@@ -1,3 +1,7 @@
+using LinearAlgebra
+import LinearAlgebra: qr
+export qr
+
 # define our own struct since LinearAlgebra.QR are restricted to Matrix
 struct QR{Q,R,P}
     Q::Q


Datseris:
So my mistake: `qr` is indeed properly imported in the main file `StaticArrays.jl`. Yet whenever I use your `qr` method I get 
```
WARNING: Base.qr is deprecated: it has been moved to the standard library package `LinearAlgebra`.
Add `using LinearAlgebra` to your imports.
```
So I reallly don't know..

fredrikekre:
It's because you don't have a `using LinearAlgebra` in your code.

Datseris:
Sure, but if I want the method from StaticArrays, shouldn't `using StaticArrays` be enough? (this is what this PR is trying to achieve btw).

Datseris:
> That is bad since changes to LinearAlgebra might break code that only has using StaticArrays.

This is a very strong argument, I totally agree, haven't thought about it.

fredrikekre:
I'm not saying this shouldn't be done, I just think maybe we should have a discussion about it in general, and perhaps something of a julia-wide policy about reexporting. I know lots of packages does it so maybe its not so bad.

andyferris:
Right. In this case it’s totally valid to use static arrays but no linear algebra in your code, just like its valid to use regular arrays but no linear algebra in your code. I don’t think we should export any names normally acquired via `using LinearAlgebra`.

PR #467, opened 2018-07-20
[WIP] README updates for 0.7

schmrlng:
Since this affects the github front page of `StaticArrays.jl` these changes should probably be merged only once 0.7 is actually released.

Regarding badges - do we want to keep the 0.5 and 0.6 passing badges up indefinitely?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,21 +27,29 @@ Full documentation can be found [here](https://JuliaArrays.github.io/StaticArray
 ## Speed
 
 The speed of *small* `SVector`s, `SMatrix`s and `SArray`s is often > 10 × faster
-than `Base.Array`. See this simplified benchmark:
+than `Base.Array`. For example, here's a
+[microbenchmark](perf/README_benchmarks.jl) showing some common operations.
 
 ```
 ============================================
     Benchmarks for 3×3 Float64 matrices
 ============================================
-Matrix multiplication               -> 5.1x speedup
-Matrix multiplication (mutating)    -> 1.6x speedup
-Matrix addition                     -> 14.0x speedup
-Matrix addition (mutating)          -> 2.1x speedup
-Matrix determinant                  -> 119.3x speedup
-Matrix inverse                      -> 65.6x speedup
-Matrix symmetric eigendecomposition -> 24.8x speedup
-Matrix Cholesky decomposition       -> 12.1x speedup
+Matrix multiplication               -> 5.9x speedup
+Matrix multiplication (mutating)    -> 1.8x speedup
+Matrix addition                     -> 33.1x speedup
+Matrix addition (mutating)          -> 2.5x speedup
+Matrix determinant                  -> 112.9x speedup
+Matrix inverse                      -> 67.8x speedup
+Matrix symmetric eigendecomposition -> 25.0x speedup
+Matrix Cholesky decomposition       -> 8.8x speedup
+Matrix LU decomposition             -> 6.1x speedup
+Matrix QR decomposition             -> 65.0x speedup
 ```
+
+These numbers were generated on an Intel i7-7700HQ using Julia-1.2. As with all
+synthetic benchmarks, the speedups you see here should only be taken as very
+roughly indicative of the speedup you may see in real code. When in doubt,
+benchmark your real application!
 
 Note that in the current implementation, working with large `StaticArray`s puts a
 lot of stress on the compiler, and becomes slower than `Base.Array` as the size
--- a/README.md
+++ b/README.md
@@ -5,12 +5,12 @@
 [![Build Status](https://travis-ci.org/JuliaArrays/StaticArrays.jl.svg?branch=master)](https://travis-ci.org/JuliaArrays/StaticArrays.jl)
 [![Build status](https://ci.appveyor.com/api/projects/status/xabgh1yhsjxlp30d?svg=true)](https://ci.appveyor.com/project/JuliaArrays/staticarrays-jl)
 [![Coverage Status](https://coveralls.io/repos/github/JuliaArrays/StaticArrays.jl/badge.svg?branch=master)](https://coveralls.io/github/JuliaArrays/StaticArrays.jl?branch=master)
-[![codecov.io](http://codecov.io/github/JuliaArrays/StaticArrays.jl/coverage.svg?branch=master)](http://codecov.io/github/JuliaArrays/StaticArrays.jl?branch=master)
+[![codecov.io](https://codecov.io/github/JuliaArrays/StaticArrays.jl/branch/master/graph/badge.svg)](http://codecov.io/github/JuliaArrays/StaticArrays.jl/branch/master)
 [![](https://img.shields.io/badge/docs-latest-blue.svg)](https://JuliaArrays.github.io/StaticArrays.jl/latest)
 [![](https://img.shields.io/badge/docs-stable-blue.svg)](https://JuliaArrays.github.io/StaticArrays.jl/stable)
 
 **StaticArrays** provides a framework for implementing statically sized arrays
-in Julia (≥ 0.5), using the abstract type `StaticArray{Size,T,N} <: AbstractArray{T,N}`.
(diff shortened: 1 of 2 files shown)


andyferris:
Thanks. Not sure on when is best timing but we can merge this before JuliaCon.

There's still a v0.6 version that will be in use for a while - we may even fix a bug or two - so I'd like to keep that around along with a v0.7 (or v1.0, soon I hope :) ) badge

andyferris:
Out of curiousity, where is this at?

schmrlng:
I've just added a few more changes -- dropping the v0.5 badge, adding a v0.7 badge (which actually doesn't seem to be working yet: https://pkg.julialang.org/badges/StaticArrays_0.7.svg), and removing the outdated discussion on *ImmutableArrays* and *FixedSizeArrays*. I can squash the two commits if necessary. The "Speed" section still needs to be updated (I'm not sure which benchmark script was used to generate those results) and the reference therein to julia 0.5 removed.

If the plan is to maintain an 0.6 branch (the closest thing at the moment is the [v0.7.2 tag](https://github.com/JuliaArrays/StaticArrays.jl/tree/v0.7.2)), maybe we should do something along the lines of the message at the top of the README here: https://github.com/timholy/Revise.jl.

andyferris:
We should fix this up for v1.0

KristofferC:
Note that SLP is on by default so any reference to `O3` is likely not needed anymore.

andyferris:
Yes, there is a bit of out-of-date cruft in there.

schmrlng:
Currently this PR has 8 badges at the top. These are listed below with a proposal for what to do with each in parentheses. Thoughts/comments (especially on the dual code coverage badges) are welcome.

- pkg.julialang.org 0.6 badge (remove)
- pkg.julialang.org 0.7 badge, broken (remove for now)
- Travis badge (keep)
- Appveyor badge (keep)
- Coveralls badge
- CodeCov badge (keep only one of this or previous?)
- Latest docs (keep)
- Stable docs (keep)

Other notes:
- I've made a new `README_benchmarks.jl` script for the README microbenchmarks.
- I removed the line "For example, the performance crossover point for a matrix multiply microbenchmark seems to be about 11x11 in julia 0.5 with default optimizations." in the **Speed** subsection. I still think some statement to that effect would be nice, but on my computer (Ryzen 1800X, 8C16T) the crossover point is now at 23x23 so I'd like to see if other people get similar results:
```julia
for N in 1:100
    A = rand(N,N)
    SA = SMatrix{N,N}(A)
    speedup = (@belapsed $A*$A)/(@belapsed $SA*$SA)
    println("$(N)x$(N): $speedup")
    speedup < 1 && break
end
```
- I'm not certain that this line in **Quick start** is still true:
```julia
rand(MMatrix{20,20}) * rand(MMatrix{20,20}) # large matrices can use BLAS
```

andyferris:
> These are listed below with a proposal for what to do with each in parentheses.

I like your suggestions here!

andyferris:
Re. the "speed crossover" thing, it's important that people are a bit scared of large static matrices... compile times can be aweful and the amount of generated code means that practical workfloads will suffer compared to microbenchmarks due to instruction cache being tiny on CPUs.

So those new benchmarks are Julia 1.0 on a Ryzen 1800X? I'd be interested to see what results we get from a modern Intel (for curiousity - I don't think the README needs to be "Intel Inside")

KristofferC:
FWIW, I don't think two documenter badges are needed. If you want the dev docs you can always use the dropdown menu.

fredrikekre:
> FWIW, I don't think two documenter badges are needed. If you want the dev docs you can always use the dropdown menu.

Especially since `latest` and `stable` are pretty ambiguous. Why would you not choose the `latest` docs? It is not "scary" enough, and not clear enough that `latest` mean development version.

Here I tried something new: https://github.com/fredrikekre/Literate.jl/blob/master/README.md and put `stable release` instead.

schmrlng:
Looks good, on the next update I'll essentially just ape what you've got going on at the `Literate.jl` README.

Unfortunately the only modern processor I have access to is my work machine (the Ryzen 1800X); I have a 2011 Macbook Pro with an Intel quad-core but I wouldn't exactly call that modern (it's definitely missing some newer SIMD stuff). If anyone reading this has a better option at hand, can you run `julia README_benchmarks.jl` from this PR or equivalently paste this code into a julia (1.0) REPL and post the output?

```julia
using BenchmarkTools
using LinearAlgebra
using StaticArrays

add!(C, A, B) = (C .= A .+ B)

function simple_bench(N, T=Float64)
    A = rand(T,N,N)
    A = A'*A
    B = copy(A)
    SA = SMatrix{N,N}(A)
    MA = MMatrix{N,N}(A)
    MB = copy(MA)

    print("""
============================================
    Benchmarks for $N×$N $T matrices
============================================
""")
    ops = [
        ("Matrix multiplication              ", *, (A, A), (SA, SA)),
        ("Matrix multiplication (mutating)   ", mul!, (B, A, A), (MB, MA, MA)),
        ("Matrix addition                    ", +, (A, A), (SA, SA)),
        ("Matrix addition (mutating)         ", add!, (B, A, A), (MB, MA, MA)),
        ("Matrix determinant                 ", det, A, SA),
        ("Matrix inverse                     ", inv, A, SA),
        ("Matrix symmetric eigendecomposition", eigen, A, SA),
        ("Matrix Cholesky decomposition      ", cholesky, A, SA)
    ]
    for (name, op, Aargs, SAargs) in ops
        if Aargs isa Tuple && length(Aargs) == 2
            speedup = @belapsed($op($Aargs[1], $Aargs[2])) / @belapsed($op($SAargs[1], $SAargs[2]))
        elseif Aargs isa Tuple && length(Aargs) == 3
            speedup = @belapsed($op($Aargs[1], $Aargs[2], $Aargs[3])) / @belapsed($op($SAargs[1], $SAargs[2], $SAargs[3]))
        else
            speedup = @belapsed($op($Aargs)) / @belapsed($op($SAargs))
        end
        println(name*" -> $(round(speedup, digits=1))x speedup")
    end
end

simple_bench(3)
```

andyferris:
If I run your code on a i7-8650U Skylake I get

```
============================================
    Benchmarks for 3×3 Float64 matrices
============================================
Matrix multiplication               -> 5.8x speedup
Matrix multiplication (mutating)    -> 1.6x speedup
Matrix addition                     -> 24.4x speedup
Matrix addition (mutating)          -> 2.5x speedup
Matrix determinant                  -> 81.1x speedup
Matrix inverse                      -> 57.8x speedup
Matrix symmetric eigendecomposition -> 20.4x speedup
Matrix Cholesky decomposition       -> 8.1x speedup
```

However, I just noted you are using *BenchmarkTools.jl* here. I do love BenchmarkTools, but in the past I was getting problems with it somehow doing less allocations for `Array` than could be reasonably expected, so I had some different benchmark code... let me re-run that too.

c42f:
This has been quietly bit-rotting for too long, so I took the liberty of rebasing and addressing Andy's comments — at least I hope — by using `Ref`s to prevent unfair optimizations. (Julia 1.2 certainly optimizes out some of the test cases in the previous version of the benchmark script.)

As with all microbenchmarks, the results are only somewhat indicative. So instead of really working hard on those right now I've just added a caveat in the documentation. The key point is that people should try `StaticArrays` if they care about performance for small matrices, not to claim that we know exactly how much faster it may be in a given application (which is impossible).

KristofferC:
Tangentially related, there are a large number of 3 years old .txt files in https://github.com/JuliaArrays/StaticArrays.jl/tree/master/perf. Should these just be removed?

c42f:
Quite possible, @andyferris would know if anyone does.

Benchmark cleanup is probably a separate PR though. Ideally we'd do it systematically and regularly on a central machine but I'm skeptical that CI workers would work for that. Is there a way to integrate with the Base benchmarking infrastructure for widely used performance critical packages like StaticArrays?

schmrlng:
Thanks for getting this across the finish line!

c42f:
No problem, sorry it took such a long time!

PR #471, opened 2018-08-04
fix a few more v0.7 deprecations

rdeits:


rdeits:
The only ones I noticed from running the tests are that `mapreduce` now wants an `init` keyword argument, so we should probably update the methods in `StaticArrays` to do the same.

andyferris:
Oops... that was me :sheep:

Yes, let's fix that :)

PR #472, opened 2018-08-07
Fixes for new (map)reducedim API

andreasnoack:
These are some quite fundamental changes to for these functions work and the new versions rely more on constant propagation. Since so much else is getting changed, I've also made the change to dispatch on `Val` values instead of `Val` types. @andyferris Maybe we should sit down and discuss this since we'll be in the same room tomorrow.

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -8,6 +8,8 @@ length(a::Type{SA}) where {SA <: StaticArray} = prod(Size(SA))
 end
 @inline size(a::StaticArray) = size(typeof(a))
 @inline size(a::StaticArray, d::Int) = size(typeof(a), d)
+@inline size(a::StaticArray, ::Val{D}) where D = size(a, D)
+@inline size(a::StaticArray, ::Colon) = length(a)
 
 # This seems to confuse Julia a bit in certain circumstances (specifically for trailing 1's)
 @inline function Base.isassigned(a::StaticArray, i::Int...)
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -222,8 +222,7 @@ end
     @inbounds return similar_type(a, typeof(Signed(a[2]*b[3])-Signed(a[3]*b[2])))(((Signed(a[2]*b[3])-Signed(a[3]*b[2]), Signed(a[3]*b[1])-Signed(a[1]*b[3]), Signed(a[1]*b[2])-Signed(a[2]*b[1]))))
 end
 
-@inline dot(a::StaticVector, b::StaticVector) = _vecdot(same_size(a, b), a, b)
-@inline vecdot(a::StaticArray, b::StaticArray) = _vecdot(same_size(a, b), a, b)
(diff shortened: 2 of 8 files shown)


src/abstractarray.jl
```diff
@@ -8,6 +8,8 @@ length(a::Type{SA}) where {SA <: StaticArray} = prod(Size(SA))
 end
 @inline size(a::StaticArray) = size(typeof(a))
 @inline size(a::StaticArray, d::Int) = size(typeof(a), d)
+@inline size(a::StaticArray, ::Val{D}) where D = size(a, D)
+@inline size(a::StaticArray, ::Colon) = length(a)
```
andyferris: This seems surprising to me and doesn't appear to be in `Base`.

src/linalg.jl
```diff
@@ -222,8 +222,7 @@ end
     @inbounds return similar_type(a, typeof(Signed(a[2]*b[3])-Signed(a[3]*b[2])))(((Signed(a[2]*b[3])-Signed(a[3]*b[2]), Signed(a[3]*b[1])-Signed(a[1]*b[3]), Signed(a[1]*b[2])-Signed(a[2]*b[1]))))
 end
 
-@inline dot(a::StaticVector, b::StaticVector) = _vecdot(same_size(a, b), a, b)
-@inline vecdot(a::StaticArray, b::StaticArray) = _vecdot(same_size(a, b), a, b)
+@inline dot(a::StaticArray, b::StaticArray) = _vecdot(same_size(a, b), a, b)
 @generated function _vecdot(::Size{S}, a::StaticArray, b::StaticArray) where {S}
```
andyferris: Happy to rename this to `_dot`

src/mapreduce.jl
```diff
@@ -60,12 +60,8 @@ end
 ## mapreduce ##
 ###############
 
-@inline function mapreduce(f, op, a::StaticArray, b::StaticArray...)
-    _mapreduce(f, op, same_size(a, b...), a, b...)
-end
-
-@inline function mapreduce(f, op, v0, a::StaticArray, b::StaticArray...)
-    _mapreduce(f, op, v0, same_size(a, b...), a, b...)
+@inline function mapreduce(f, op, a::StaticArray, b::StaticArray...; init = nothing)
```
andyferris: There's a trick to remove the `nothing` sentinel, used in `Base`.

Basically, use `; kw...` and we can see if the resulting NamedTuple has a field called `init` or not. A bit yucky but allows users to initiate a reduciton with `nothing` if they want.

andyferris:
Simon won the "make it work on v1.0" race.

@c42f Yeah its pretty fun but I've spent my time finishing my slides and so-on. So looking forward to freedom tomorrow afternoon! :) Wish you were here.

c42f:
Oh my gosh I thought that "work on v1.0" was a foreshadowing, not the official release. I'm heading right to the recorded founders talk and ceremonial 1.0 merging as soon as the kids are in bed. Yeah, wish I was there joining the 1.0 celebration!

PR #474, opened 2018-08-10
export SUnitRange

schmrlng:
I feel like `SUnitRange` is useful enough (especially for slicing other `StaticArray`s) to merit exporting.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -26,7 +26,7 @@ import LinearAlgebra: transpose, adjoint, dot, eigvals, eigen, lyap, tr,
 end
 
 export StaticScalar, StaticArray, StaticVector, StaticMatrix
-export Scalar, SArray, SVector, SMatrix
+export Scalar, SArray, SVector, SMatrix, SUnitRange
 export MArray, MVector, MMatrix
 export FieldVector
 export SizedArray, SizedVector, SizedMatrix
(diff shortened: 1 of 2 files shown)


andyferris:
Yes, this isn't unreasonable... :)

There has been a long-standing plan to support static ranges, LinearIndices and CartesianIndices with StaticArrays. These things would be returned by `axes` and `indices`. I've been waiting for v1.0 for that to happen (because compatibility is hard enough between v0.6 users and v1.0 users already) so it might be time to have a rethink of this soon.

schmrlng:
Sure, if it's a question of API stability/not committing to this exact form for fewer "breaking" changes once more general ranges are worked out I'm happy to keep manually importing this secret functionality.

andyferris:
OK thanks. Please bump this again soon-ish if I don't get to this.

tpapp:
This PR addresses #620. It would be nice to have a docstring though, perhaps with an example in use, and the docs should also mention this.

tpapp:
Friendly bump --- could this be merged please?

hyrodium:
What's the status of this PR? Is this just waiting for resolving the conflicts?

tpapp:
Friendly ping; please reconsider this PR. If anything else needs to be done I am willing to do it or make another PR if that is preferred.

hyrodium:
I think the only problem with `SUnitRange` is https://github.com/JuliaArrays/StaticArrays.jl/issues/978. It would be better to have `SUnitRange <: AbstractUnitRange`, but changing the supertype may be a breaking change because the type is widely used, I guess.

Considering `SUnitRange` as a stable API is not a bad practice, so exporting it would be okay for me.

PR #484, opened 2018-09-02
fixed MMatrix boundschecks, faster getindex

chethega:
As remarked on [https://github.com/JuliaLang/julia/issues/29000](https://github.com/JuliaLang/julia/issues/29000), accessing an `MArray`, `MMatrix`, `MVector` sometimes pulls a stack copy. This patch avoids the stack copy, by using the pointer for `getindex` as well, not just for `setindex`.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -106,8 +106,13 @@ end
 ## MArray methods ##
 ####################
 
-function getindex(v::MArray, i::Int)
-    Base.@_inline_meta
+@propagate_inbounds function getindex(v::MArray, i::Int)
+    if isbitstype(T)
+        @boundscheck if i < 1 || i > length(v)
+            throw(BoundsError())
+        end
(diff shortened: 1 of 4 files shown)


chethega:
Ok, this is fast. How do I fix the following?

```
julia> using StaticArrays
julia> M=MMatrix{2,2}(rand(2,2));
julia> M[3,1]
0.7134474216097795
```

andyferris:
OK what's there looks great now :)

For Cartesian indexing it's the `..._scalar` functions at the start of indexing.jl

chethega:
Ok, I think this can be merged (from my perspective). Can we run a benchmark to see whether the pointer access or the boundscheck confuse the optimizer? As far as I can see all cartesian accesses were effectively `@inbounds` up to now, so that may slow down some relying code :(

andyferris:
Let's roll with this.

PR #494, opened 2018-09-13
Don't force loop unrolling in reductions

KristofferC:
Instead of unrolling the whole loop in the reduction, we can just keep the loop and let LLVM do its job with loop unrolling (since the number of loop iterations is known).

Calling this inner function directly we can see that for larger matrices this can give good speedup:

```jl
f(s) =  StaticArrays._mapreduce(identity, +, Colon(), NamedTuple(), StaticArrays.same_size(s), s)
s = rand(SMatrix{8,8})
```

Before:

After:

```
julia> @btime f($s)
  10.179 ns (0 allocations: 0 bytes)
```

Previously

```
julia> @btime f($s)
  30.229 ns (0 allocations: 0 bytes)
```

The reason for this is that the loop vectorizer has a much easier time with the loop than the SLP with the unrolled code.

Now, why am I calling the inner function directly? Well, for some reason, Julia refuses to inline this new function. Instead, we get:

```
julia> @code_typed sum(s)
CodeInfo(
181 1 ─ %1 = StaticArrays.:+::typeof(+)                                               │╻  #sum#255
    │   %2 = invoke #kw##mapreduce()($(QuoteNode((dims = Colon(),)))::NamedTuple{(:dims,),Tuple{Colon}}, StaticArrays.mapreduce::typeof(mapreduce), StaticArrays.identity::Function, %1::Function, _2::SArray{Tuple{8,8},Float64,2,64})::Float64
    └──      return %2                                                                │
) => Float64
```

and the result is not pretty.

```
julia> @btime sum($s)
  121.209 ns (3 allocations: 1.08 KiB)
```

But if we can fix that inlining problem, this is probably worth doing.


KristofferC:
Test fails because reductions now use SIMD and the tests check `==`.

andyferris:
> Tests fail because reductions now use SIMD and the tests check `==`.

Happy to switch to `isapprox` with appropriate tolerence.

KristofferC:
Inlining is fixed by https://github.com/JuliaLang/julia/pull/29258 so now there is a reason to finish this up :)

KristofferC:
Updated.

Some benchmarks made on https://github.com/JuliaLang/julia/pull/29258

```jl
using BenchmarkTools
using StaticArrays
x = rand(MMatrix{8,8}); s = rand(SMatrix{8,8});
```

Before

```jl
julia> @btime map!(x -> x*2, $x, $s);
  11.056 ns (0 allocations: 0 bytes)

julia> @btime sum($s);
  29.363 ns (0 allocations: 0 bytes)

julia> @btime sum(abs2, $s);
  124.234 ns (3 allocations: 1.08 KiB)

julia> @btime mapreduce(x->x^2, +, $s; init=0.5);
  134.850 ns (4 allocations: 1.09 KiB)
```

After

```jl
julia> @btime map!(x -> x*2, $x, $s);
  12.783 ns (0 allocations: 0 bytes)

julia> @btime sum($s);
  12.906 ns (0 allocations: 0 bytes)

julia> @btime sum(abs2, $s);
  12.283 ns (0 allocations: 0 bytes)

julia> @btime mapreduce(x->x^2, +, $s; init=0.5);
  12.399 ns (0 allocations: 0 bytes)
```

KristofferC:
Arguably the old methods should be left and we should `@static if` based on the VERSION... Thoughts?

src/linalg.jl
```diff
@@ -224,87 +224,65 @@ end
...
-    end
-
-    expr = :(conj(a[1]) * b[1])
-    for j = 2:prod(S)
-        expr = :($expr + conj(a[$j]) * b[$j])
-    end
-
     return quote
         @_inline_meta
-        @inbounds return $expr
```
andyferris: Pity... I was just deleting code which relies on inference...

src/linalg.jl
```diff
@@ -224,87 +224,65 @@ end
...
+        s = zero(promote_op(*, eltype(a), eltype(b)))
```
andyferris: Pity... I was just deleting code which relies on inference...

andyferris:
One interesting thing about this - this optimization goes with the assumption that `StaticArray`s have dense (or at least SIMD-fetchable) layout. I wonder how this works on things like the computed `Rotations` matrices? (Not that you'd ever want to do any of these operations on those!) I really wish we had traits about layout...

Do you have benchmarks for small vectors and matrices? It's important (to me at least) that we don't slow down any sums or norms or dot products of 2-vectors or 3-vectors, or operations involving 2x2 or 3x3 matrices.

src/linalg.jl
KristofferC: I can move this to only be used in the zero length case again.

andyferris: Yes, that would be good I think. Will the SIMD loop vectoriser still work if you peel off the first element in the loop, though? (Or maybe there is a better way?)

By the way, these were the one place where the `Union{}` eltype created errors (because of `zero(Union{})`.

tkoolen:
> Do you have benchmarks for small vectors and matrices? It's important (to me at least) that we don't slow down any sums or norms or dot products of 2-vectors or 3-vectors, or operations involving 2x2 or 3x3 matrices.

I was also wondering about this.

KristofferC:
```

# LinAlg
using BenchmarkTools
using StaticArrays
using LinearAlgebra
for siz in (1,2,3,4,8)
    println("size = $siz x $siz")
    s = rand(SMatrix{siz, siz})
    x = rand(MMatrix{siz, siz})
    @btime map!(x -> x*2, $x, $s);
    @btime sum(abs2, $s);
    @btime sum($s);
    @btime mapreduce(x->x^2, +, $s; init=0.5);
    @btime dot($s, $s)
    @btime norm($s, 1)
    @btime norm($s, 2)
    @btime norm($s, 5)
    @btime det($s)
end
```

```
Master PR
1.503 ns	1.793 ns
1.503 ns	1.528 ns
1.596 ns	1.596 ns
1.505 ns	1.793 ns
1.848 ns	1.503 ns
0.023 ns	0.031 ns
4.155 ns	4.158 ns
59.953 ns	59.977 ns
0.030 ns	0.023 ns
size = 2 x 2	size = 2 x 2
2.032 ns	2.365 ns
1.794 ns	2.044 ns
1.792 ns	1.715 ns
2.219 ns	2.046 ns
1.804 ns	1.798 ns
0.023 ns	2.047 ns
4.159 ns	4.716 ns
124.816 ns	125.834 ns
0.031 ns	0.023 ns
size = 3 x 3	size = 3 x 3
2.044 ns	2.370 ns
2.690 ns	2.372 ns
2.390 ns	2.034 ns
2.985 ns	2.711 ns
2.990 ns	2.833 ns
0.030 ns	3.038 ns
4.163 ns	4.717 ns
246.908 ns	235.005 ns
2.392 ns	2.397 ns
size = 4 x 4	size = 4 x 4
3.036 ns	2.373 ns
5.417 ns	2.712 ns
4.457 ns	2.047 ns
5.636 ns	3.046 ns
6.081 ns	6.077 ns
0.023 ns	2.709 ns
6.744 ns	4.719 ns
382.754 ns	383.897 ns
14.593 ns	14.493 ns
size = 8 x 8	size = 8 x 8
12.004 ns	12.068 ns
93.859 ns (2 allocations: 560 bytes)	12.410 ns
29.360 ns	9.073 ns
100.554 ns (3 allocations: 576 bytes)	12.391 ns
38.960 ns	61.910 ns
0.023 ns	14.052 ns
37.048 ns	15.209 ns
1.416 μs	1.692 μs
355.091 ns	354.120 ns
```

Doesnt seem to be an improvement in all cases, will look at it closer.

KristofferC:
Removed the `map!` since it seems LLVM decided not to unroll the loop even though it perhaps was advantageous to do so.

chriselrod:
> Removed the `map!` loop since it seems LLVM decided not to unroll the loop even though it perhaps was advantageous to do so.

Could you try un-reverting that change and see if it is better now?
Perhaps use `@simd ivdep` to indicate non-aliasing (via the `ivdep`)?

I would like to see this get merged.

PR #495, opened 2018-09-14
Determine `broadcast` element type based on actual values

### LINKED ISSUE #493: A segfault with member functions
Hello. When I tried to execute [example notebook](https://github.com/rdeits/RegionTrees.jl/blob/master/examples/adaptive_distances.ipynb) of RegionTrees.jl package I found a segfault on my system (Ubuntu 18.04). Purifying the crashing code I came up with this one:
```julia
SEGFAULT = true
using StaticArrays

struct TwosArray{N, T, L} <: StaticArray{NTuple{N, 2}, T, N}
    data::NTuple{L, T}
end
Base.getindex(b::TwosArray, i::Int) = b.data[i]

struct SignedDistanceRefinery{F <: Function} # <: AbstractRefinery 
    signed_distance_func::F
end

function refine_data(refinery::SignedDistanceRefinery, v) 
    refinery.signed_distance_func.(v)
end

signed_distance(x) = sqrt(sum(x.^2))
refinery = SignedDistanceRefinery(signed_distance)

vp = TwosArray{1,SArray{Tuple{2},Float64,1,2},2}((SVector(1.,2.),SVector(3.,1.)))

!SEGFAULT || refinery.signed_distance_func.(vp)
refine_data(refinery,vp)
```
Interestingly to cause a segfault both last two lines need to be present which also commute with themselves.

martinholters:
Mimic the behavior for `Array`s and only use `return_type` in the empty case. I wonder, however, whether we shouldn't use `Union{}` in the empty case. We don't have to worry about the type-instability as the size is part of the type, too, and cannot e.g. `push!` to an `SVector`. So there seems to be little harm. Opinions?

Ref. JuliaLang/julia#28981, closes #493. But note that nested broadcasts will still suffer from incomplete inference due to the recursion limiting heuristics.

### DIFF
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -151,22 +151,22 @@ end
     @testset "eltype after broadcast" begin
         # test cases issue #198
         let a = SVector{4, Number}(2, 2.0, 4//2, 2+0im)
-            @test_broken eltype(a + 2) == Number
-            @test_broken eltype(a - 2) == Number
-            @test_broken eltype(a * 2) == Number
-            @test_broken eltype(a / 2) == Number
+            @test eltype(a + 2) == Number
+            @test eltype(a - 2) == Number
+            @test eltype(a * 2) == Number
+            @test eltype(a / 2) == Number
         end
         let a = SVector{3, Real}(2, 2.0, 4//2)
-            @test_broken eltype(a + 2) == Real
-            @test_broken eltype(a - 2) == Real
-            @test_broken eltype(a * 2) == Real
-            @test_broken eltype(a / 2) == Real
+            @test eltype(a + 2) == Real
+            @test eltype(a - 2) == Real
+            @test eltype(a * 2) == Real
+            @test eltype(a / 2) == Real
         end
         let a = SVector{3, Real}(2, 2.0, 4//2)
-            @test_broken eltype(a + 2.0) == Float64
-            @test_broken eltype(a - 2.0) == Float64
-            @test_broken eltype(a * 2.0) == Float64
-            @test_broken eltype(a / 2.0) == Float64
+            @test eltype(a + 2.0) == Float64
+            @test eltype(a - 2.0) == Float64
+            @test eltype(a * 2.0) == Float64
+            @test eltype(a / 2.0) == Float64
         end
         let a = broadcast(Float32, SVector(3, 4, 5))
             @test eltype(a) == Float32
@@ -195,5 +195,11 @@ end
 
         d = SVector(1, 2, 3, 4)
         @test_throws DimensionMismatch a .= d
+    end
+
+    @testset "issue #493" begin
+        X = rand(SVector{3,SVector{2,Float64}})
+        foo493(X) = normalize.(X)
+        @test foo493(X) isa Core.Compiler.return_type(foo493, Tuple{typeof(X)})
     end
 end
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -125,12 +125,17 @@ scalar_getindex(x::Tuple{<: Any}) = x[1]
         end
     end
 
-    eltype_exprs = [t <: Union{AbstractArray, Ref} ? :(eltype($t)) : :($t) for t ∈ a]
-    newtype_expr = :(return_type(f, Tuple{$(eltype_exprs...)}))
+    if isempty(exprs)
+        eltype_exprs = [t <: Union{AbstractArray, Ref} ? :(eltype($t)) : :($t) for t ∈ a]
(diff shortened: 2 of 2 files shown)


test/broadcast.jl
```diff
@@ -151,22 +151,22 @@ end
     @testset "eltype after broadcast" begin
         # test cases issue #198
         let a = SVector{4, Number}(2, 2.0, 4//2, 2+0im)
-            @test_broken eltype(a + 2) == Number
-            @test_broken eltype(a - 2) == Number
-            @test_broken eltype(a * 2) == Number
-            @test_broken eltype(a / 2) == Number
+            @test eltype(a + 2) == Number
```
fredrikekre: Shouldn't these methods be deprecated? They are for general arrays.
```
julia> [1, 2, 3] + 1
ERROR: MethodError: no method matching +(::Array{Int64,1}, ::Int64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:502
  +(::Complex{Bool}, ::Real) at complex.jl:292
  +(::Missing, ::Number) at missing.jl:93
  ...
Stacktrace:
 [1] top-level scope at none:0

julia> @SVector([1, 2, 3]) + 1
3-element SArray{Tuple{3},Int64,1,3}:
 2
 3
 4
```

martinholters: True, but orthogonal to this issue, except that I could use the explicit broadcast in these tests as I'm touching them anyways. (And it's in the broadcast tests.)

fredrikekre: Yea, just wanted to point it out since I noticed it :slightly_smiling_face:

martinholters: Updated the tests locally; will push once I have feedback what to do about the empty case.

andyferris:
> I wonder, however, whether we shouldn't use `Union{}` in the empty case.

It's a **very** interesting question. Here's a splattering of random thoughts.

 * I hate that we have to treat empty containers seperately at all. My belief is that robust systems should be able to gracefully and naturally handle the empty container case. (At the very least, we have to do something reasonable for our *StaticArrays* users, but this is a broader comment about Julia).
 * Parts of the Julia language and implementation (what I call the "covariant" parts) would certainly favor `Union{}` as the output element type. In a fully covariant system it is in-fact rather elegant, and works well with `promote_type` and so-on.
 * On the other hand, invariant type parameters means there's code all over the place which expects certain type parameters. If empty arrays were to come up, they might not match function signatures and so-on. As a small example, if we were to take the "covariant" approach to its extremes, I wonder how much code like `f(::AbstractArray{Float64})` should actually be `f(::AbstractArray{<:Float64})` (which is for all intents and purposes `f(::Union{AbstractArray{Float64}, AbstractArrray{Union{}})`, and the latter case would typically occur only for empty arrays).
 * Using `Union{}` in the empty case for static arrays shouldn't be a problem in practice. But as an example of the tension between covariance and invariance in Julia, if dynamically sized arrays did this then operations like `filter` couldn't possibly be type-stable.
 * I think this question can be resolved seperately to the issue at hand. :)

There is however a very important "covariance vs invariance" discussion involving the work that is in this PR, relating to the whole `DataValue{T}` vs `Union{Missing, T}` debate. Switching to covariance means that broadcast kernels (i.e. `f` in `f.(a)`) that return `Union{Missing, T}` can result in arrays of final element type `Missing`, `T2`, or `Union{Missing, T2}` (where `f(::T) --> T2`). I would be tempted to describe this as the type uncertainty spreading in a non-deterministic fassion, and my staic compilation background makes me prefer that we just get `Union{Missing, T2}`. What actually matters here is (a) what's the performance difference between these two choices and (b) is it likely that this type uncertainty will spread further downstream, or will static analysis of programs with complex data flow tend to remain capped at `Union`s of 3 different types? I'm interested in everyone's thoughts on this.

martinholters:
Ok, let's leave the discussion of the empty case to another time.

So yeah, with this PR `idenitiy.(x::SVector{N,Union{Missing, T}})` can give `SVector{N,Union{Missing, T}}`, `SVector{N,Missing}`, or `SVector{N,T}`, and is inferred as `SVector{N}`. Yes, I agree, that does not sound too attractive.
But without this PR, it gives `SVector{N,R}` where `R` is hopefully (!) some supertype of `Union{Missing, T}`, depending on how inference is feeling today, and is inferred as such. That does not sound too much better to me. And the "hopefully" there (see JuliaLang/julia#28981) makes it far worse, actually. Of course, this usually works fine, even finding the "correct" (tightest) element type. But there is no guarantee, and I don't think it's possible to decide whether we are in good or a bad case.

andyferris:
See new PR #503. It's also worth noting #330.

andyferris:
Done by #503.

Issue #501, opened 2018-09-18
reinterpret of Arrays of static arrays failed


Hong-Xiang:
[`reinterpret` Arrays of static arrays](https://juliaarrays.github.io/StaticArrays.jl/stable/pages/api.html#Arrays-of-static-arrays-1) failed.

steps to reproduce:
```
using StaticArrays
function svectors(x::Matrix{Float64})
    @assert size(x,1) == 3
    reinterpret(SVector{3,Float64}, x, (size(x,2),))
end
m = zeros(3, 10)
svectors(m)
```
outputs:
```
ERROR: MethodError: no method matching reinterpret(::Type{SArray{Tuple{3},Float64,1,3}}, ::Array{Float64,2}, ::Tuple{Int64})
Closest candidates are:
  reinterpret(::Type{T}, ::Base.ReshapedArray, ::Tuple{Vararg{Int64,N}} where N) where T at reshapedarray.jl:191
  reinterpret(::Type{T}, ::A<:AbstractArray{S,N}) where {T, N, S, A<:AbstractArray{S,N}} at reinterpretarray.jl:14
  reinterpret(::Type{T}, ::Any) where T at essentials.jl:370
  ...
Stacktrace:
 [1] svectors(::Array{Float64,2}) at ./REPL[2]:3
 [2] top-level scope at none:0
```

chethega:
@andyferris It appears that this is a documentation bug. Do we need a docfix before merging some variant of [https://github.com/JuliaArrays/StaticArrays.jl/pull/496](https://github.com/JuliaArrays/StaticArrays.jl/pull/496)?

andyferris:
Yes - perhaps clearly documenting the *current* behavior on Julia 1.0 would be the best first step.

dmbates:
Bump.  The old, failing, function `svectors` is still in the documentation as of v0.11.0

PR #502, opened 2018-09-18
docfix reinterpret

chethega:
Partially resolves [https://github.com/JuliaArrays/StaticArrays.jl/issues/501](https://github.com/JuliaArrays/StaticArrays.jl/issues/501).

### DIFF
--- a/docs/src/pages/api.md
+++ b/docs/src/pages/api.md
@@ -241,16 +241,49 @@ Another common way of storing the same data is as a 3×`N` `Matrix{Float64}`.
 Rather conveniently, such types have *exactly* the same binary layout in memory,
 and therefore we can use `reinterpret` to convert between the two formats
 ```julia
-function svectors(x::Matrix{Float64})
-    @assert size(x,1) == 3
-    reinterpret(SVector{3,Float64}, x, (size(x,2),))
+function svectors(x::Matrix{T}, ::Val{N}) where {T,N}
+    size(x,1) == N || throw("sizes mismatch")
+    isbitstype(T) || throw("use for bitstypes only")
+    reinterpret(SVector{N,T}, reshape(x, length(x)))
 end
 ```
-Such a conversion does not copy the data, rather it refers to the *same* memory
-referenced by two different Julia `Array`s. Arguably, a `Vector` of `SVector`s
-is preferable to a `Matrix` because (a) it provides a better abstraction of the
-objects contained in the array and (b) it allows the fast *StaticArrays* methods
-to act on elements.
+Such a conversion does not copy the data, rather it refers to the *same* memory. 
+Arguably, a `Vector` of `SVector`s is often preferable to a `Matrix` because 
+(a) it provides a better abstraction of the objects contained in the array and 
+(b) it allows the fast *StaticArrays* methods to act on elements
+
+However, the resulting object is a Base.ReinterpretArray, not an Array, which may carry some
+runtime penalty on every single access. If you can afford the memory for a copy and can live with
+the non-shared mutation semantics, then it is better to pull a copy by e.g.
+```julia
+function svectorscopy(x::Matrix{T}, ::Val{N}) where {T,N}
+    size(x,1) == N || throw("sizes mismatch")
+    isbitstype(T) || throw("use for bitstypes only")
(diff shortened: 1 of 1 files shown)


docs/src/pages/api.md
```diff
@@ -241,16 +241,59 @@ Another common way of storing the same data is as a 3×`N` `Matrix{Float64}`.
...
+
+ julia> svectorscopy(M, Val{2}())
+3-element Array{SArray{Tuple{2},Int64,1,2},1}:
+ [1, 2]
+ [3, 4]
+ [5, 6]
+```
+If you absolutely insist on obtaining a `Vector{<:SVector}` referencing the same memory 
+and don't mind a maintenance nightmare and data-corrupting tbaa-related bugs appearing in 
+julia 1.1 or later, then you can obtain this by
```
andyferris: :laughing: 

This might seem a little confusing to the uninitiated. I wonder if we should write this in a slightly more factual way?

chethega:
Friendly bump. The CI fail ("The remote server returned an error: (404) Not Found.") has nothing to do with this PR, and I'd guess some docfix should go into 1.0 [https://github.com/JuliaArrays/StaticArrays.jl/issues/532](https://github.com/JuliaArrays/StaticArrays.jl/issues/532).

simonbyrne:
bump

docs/src/pages/api.md
```diff
@@ -241,16 +241,62 @@ Another common way of storing the same data is as a 3×`N` `Matrix{Float64}`.
...
+ [1, 2]
+ [3, 4]
+ [5, 6]
+```
+If applications absolutely insist on obtaining a `Vector{<:SVector}` referencing the same memory,
+then this is possible by directly calling into the runtime. However, starting from 
+Julia 0.7 and 1.0, this violates the language specifications about aliasing 
+(Arrays with different `eltype` must not share memory), and is to be considered a dirty hack.
+This is nevertheless safe on 1.0.* or earlier versions of the Julia compiler, but will violate
+compiler assumptions on Julia 1.1 or later, and cause subtle data-corrupting miscompilation bugs.
```
c42f: If this is so broken, let's not suggest doing it in the docs! (I suggest we just delete this section ;-) )

docs/src/pages/api.md
```diff
@@ -241,16 +241,62 @@ Another common way of storing the same data is as a 3×`N` `Matrix{Float64}`.
 Rather conveniently, such types have *exactly* the same binary layout in memory,
 and therefore we can use `reinterpret` to convert between the two formats
 ```julia
-function svectors(x::Matrix{Float64})
-    @assert size(x,1) == 3
-    reinterpret(SVector{3,Float64}, x, (size(x,2),))
+function svectors(x::Matrix{T}, ::Val{N}) where {T,N}
+    size(x,1) == N || throw("sizes mismatch")
+    isbitstype(T) || throw("use for bitstypes only")
+    reinterpret(SVector{N,T}, reshape(x, length(x)))
```
c42f: We can use `vec(x)` rather than reshape here.

docs/src/pages/api.md
```diff
@@ -241,16 +241,62 @@ Another common way of storing the same data is as a 3×`N` `Matrix{Float64}`.
...
+
+However, the resulting object is a Base.ReinterpretArray, not an Array, which may carry some
+runtime penalty on every single access. If you can afford the memory for a copy and can live with
+the non-shared mutation semantics, then it is better to pull a copy by e.g.
+```julia
+function svectorscopy(x::Matrix{T}, ::Val{N}) where {T,N}
+    size(x,1) == N || throw("sizes mismatch")
+    isbitstype(T) || throw("use for bitstypes only")
+    res = Vector{SVector{N,T}}(undef, size(x,2))
+    GC.@preserve res x ccall(:memcpy, Ptr{Cvoid}, (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t), pointer(res), pointer(x), sizeof(x))
```
c42f: This seems unnecessarily low level for the documentation. How about simply using

`copy(reinterpret(SVector{N,T}, vec(x)))`

If you do think we need this low level optimization perhaps we should be providing some version of `svectors` as part of the official API rather than suggesting users do these low level hacks themselves? (#496 I guess)

PR #507, opened 2018-09-27
 Special case flatten of iterators of static arrays as done for tuples

mschauer:
Before: 
```
julia> X = [@SVector rand(10) for i in 1:10000]
julia> @btime collect(Iterators.flatten(X))
  1.402 ms (18 allocations: 2.00 MiB)
```
after 
```
julia> @btime collect(Iterators.flatten(Y))
  365.198 μs (3 allocations: 781.34 KiB)
```


mschauer:
Ping

PR #509, opened 2018-10-01
Matrix logarithm, Schur decomposition and improvements to matrix square root

mateuszbaran:
I have prepared a few improvements:
- Matrix logarithm based on its implementation in Julia LinearAlgebra. It is about 30us faster than original one, which corresponds to about 2x improvement for general 2x2 and 3x3 matrices.
- Schur decomposition implemented using LAPACK.gees! (like in LinearAlgebra)
- Square root of matrices is fixed for 2x2 matrices with negative determinant (computing it caused an error before). Also, the new implementation for matrices 3x3 and larger is about 8us faster, corresponding to about 2x speedup for 3x3 and 4x4 matrices.

Let me know if there is something that should be improved.


mschauer:
When did `sqrt` change? Here #212 it is still argued that a typestable `sqrt` which errors might be the better choice.

mateuszbaran:
One question though: in the current implementation, `sqrt` uses `LinearAlgebra.sqrt` for matrices 3x3 and larger, and this function is type unstable (returns a complex matrix when the square root is a complex matrix). What should I do in this case? Keep it functionally identical to current behavior or make it behave like `sqrt` for 2x2 matrices?

mateuszbaran:
Thinking about this again, it doesn't seem particularly important. I'll close this PR.

PR #510, opened 2018-10-02
distinguish read and read\! (#505)

### LINKED ISSUE #505: read!(io::IO, ::Type{SA})
Why is [this](https://github.com/JuliaArrays/StaticArrays.jl/blob/86097997bd846f0b9f4331cd55ba8a233375173b/src/io.jl#L2) a method of `read!`, instead of `read`? It does not modify any of its arguments.

jw3126:
fix #505

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,16 +1,22 @@
+@inline function _read_data(io::IO, ::Type{SA}) where {SA <: StaticArray}
+    data = Ref{NTuple{length(SA),eltype(SA)}}()
+    read!(io, data)
+    data[]
+end
 
-@inline function read!(io::IO, ::Type{SA}) where {SA<:StaticArray}
-    elements = Ref{NTuple{length(SA),eltype(SA)}}()
-    read!(io, elements)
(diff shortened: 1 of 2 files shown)


c42f:
I started reviewing this but thought we can be much more minimal in fixing this. My preferred style for the fix here: #515

To explain #515 —
* it leaves `read!` alone so that it remains non-allocating (I *think* what you've got here would be allocating, though I could be wrong with the latest compiler)
* leaves the tests alone, because IMO the very explicit style is easier to debug when there's a test failure.

Bit frustrating that this previously working feature was removed back in the rush to 0.7 and will be good to have it fixed either way!

jw3126:
What I find very confusing in the current version as well as #515 is that `read!(io, myarray)` does not mutate `myarray::MArray`.

jw3126:
I think you are right, lets close this.

c42f:
> read!(io, myarray) does not mutate myarray::MArray

Yikes, I hope not... checking... seems ok to me.

PR #511, opened 2018-10-03
add specialized methods for Triangular * Triangular

### LINKED ISSUE #508: roundtrip SMatrix from cholesky decomposition of SMatrix
I'd like to be able to recover an SMatrix from it's cholesky decomposition, but it doesn't survive the roundtrip:

```
julia> x = @SMatrix [1. .7; .7 .5]
 1.0  0.7
 0.7  0.5

julia> xchol = cholesky(x)
Cholesky{Float64,SArray{Tuple{2,2},Float64,2,4}}
U factor:
2×2 UpperTriangular{Float64,SArray{Tuple{2,2},Float64,2,4}}:
 1.0  0.7
  ⋅   0.1

julia> xchol.U'xchol.U
2×2 Array{Float64,2}:
 1.0  0.7
 0.7  0.5
```

@simonbyrne suggested that adding an `SMatrix` method for `Cholesky{T,SMatrix}` would fix this.

kleinschmidt:
Adds specialized methods so that products of two `AbstractTriangular`s from `SMatrix`s results in another `SMatrix` (or a triangular, if both are upper/lower).

This partially fixes #508 because the `AbstractMatrix` converter for a `Cholesky` falls back on `U'U`, which now returns an `SMatrix` instead of a normal `Matrix`.

### DIFF
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -17,9 +17,12 @@
 @inline Base.:*(A::LinearAlgebra.AbstractTriangular{<:Any,<:StaticMatrix}, B::Transpose{<:Any,<:StaticVecOrMat}) =
     transpose(transpose(B) * transpose(A))
 
+const StaticULT = Union{UpperTriangular{<:Any,<:StaticMatrix},LowerTriangular{<:Any,<:StaticMatrix}}
+
 @inline Base.:*(A::LinearAlgebra.AbstractTriangular{<:Any,<:StaticMatrix}, B::StaticVecOrMat) = _A_mul_B(Size(A), Size(B), A, B)
 @inline Base.:*(A::StaticVecOrMat, B::LinearAlgebra.AbstractTriangular{<:Any,<:StaticMatrix}) = _A_mul_B(Size(A), Size(B), A, B)
-@inline Base.:\(A::Union{UpperTriangular{<:Any,<:StaticMatrix},LowerTriangular{<:Any,<:StaticMatrix}}, B::StaticVecOrMat) = _A_ldiv_B(Size(A), Size(B), A, B)
+@inline Base.:*(A::StaticULT, B::StaticULT) = _A_mul_B(Size(A), Size(B), A, B)
+@inline Base.:\(A::StaticULT, B::StaticVecOrMat) = _A_ldiv_B(Size(A), Size(B), A, B)
 
 
 @generated function _A_mul_B(::Size{sa}, ::Size{sb}, A::UpperTriangular{TA,<:StaticMatrix}, B::StaticVecOrMat{TB}) where {sa,sb,TA,TB}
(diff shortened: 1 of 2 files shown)


kleinschmidt:
Also, when looking at the code it seems like a lot of the specialized `*_mul_*` methods are dead code since they're not exported or used anywhere else in the package as far as I can tell.  AFAIK coverage is still borked right now so there's no way to see if they're actually being tested but might be worth taking them out if they're not used?

kleinschmidt:
I haven't done systematic benchmarking but this gives you the kind of performance gains you'd expect:

```
julia> @btime $U * $U
  299.686 ns (2 allocations: 176 bytes)
3×3 UpperTriangular{Float64,Array{Float64,2}}:
 0.00598455  0.0774304  0.14636   
  ⋅          0.336431   0.521002  
  ⋅           ⋅         0.00217786

julia> @btime $SU * $SU
  3.618 ns (0 allocations: 0 bytes)
3×3 UpperTriangular{Float64,SArray{Tuple{3,3},Float64,2,9}}:
 0.00598455  0.0774304  0.14636   
  ⋅          0.336431   0.521002  
  ⋅           ⋅         0.00217786

julia> @btime $U * $L
  290.427 ns (1 allocation: 160 bytes)
3×3 Array{Float64,2}:
 0.331205   0.402724  0.0182264 
 0.720176   1.04825   0.0387971 
 0.0376679  0.039958  0.00217786

julia> @btime $SU * $SL
  4.327 ns (0 allocations: 0 bytes)
 0.331205   0.402724  0.0182264 
 0.720176   1.04825   0.0387971 
 0.0376679  0.039958  0.00217786
```

src/triangular.jl
```diff
@@ -559,3 +562,129 @@ end
...
+
+@generated function _A_mul_B(::Size{sa}, ::Size{sb}, A::LowerTriangular{<:TA,<:StaticMatrix}, B::UpperTriangular{<:TB,<:StaticMatrix}) where {sa,sb,TA,TB}
+    n = sa[1]
+    if n != sb[1]
+        throw(DimensionMismatch("left and right-hand must have same sizes, got $(n) and $(sb[1])"))
+    end
+
+    X = [Symbol("X_$(i)_$(j)") for i = 1:n, j = 1:n]
+
+    code = quote end
```
andyferris: Do you want a `LineNumberNode`? If not you could just do `code = Expr(:block)` instead.

kleinschmidt: I was just blindly copying the code from the rest of this file; I don't know enough to know whether I should have a preference one way or the other.

andyferris: OK... let's change them all to `Expr(:block)`, then. You'll get a relevant line number from the final quote anyway.

kleinschmidt: I can do that but won't get to it until later today; feel free to make the change yourself in the mean time if you'd rather.

PR #515, opened 2018-10-09
Reinstate read(::IO, ::Type{<:StaticArray})

### LINKED ISSUE #505: read!(io::IO, ::Type{SA})

c42f:
This was removed in the rush to 0.7 compatibility

Fixes #505, replaces #510

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,5 +1,5 @@
 
-@inline function read!(io::IO, ::Type{SA}) where {SA<:StaticArray}
+@inline function read(io::IO, ::Type{SA}) where {SA<:StaticArray}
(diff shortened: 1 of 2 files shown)


jw3126:
Can you add a test, that `read!` is actually mutating on an `MArray`? I think that is broken on current master.

c42f:
Yes it seems fine, but good point about adding another test. I did that now.

Issue #459, opened 2018-07-02
Deprecate ImmutableArrays and FixedSizeArrays


andyferris:
I feel Julia v0.7 might be the correct time for adding deprecations to the `ImmutableArrays` and `FixedSizeArrays` compatibility layers, and we can delete them entirely for our Julia v1.0 releases.

Let me know if this doesn't work for anyone.

PR #503, opened 2018-09-20
Fix all the things

andyferris:
Solve various inference issues. Replaces #495 and #330.
Closes #493.

Note: the eltype of empty arrays may become `Union{}` in
`map`, `broadcast`, and `mapreduce(...; dims = Val{N})`.

Issue #18, opened 2016-08-14
Many ambiguities


timholy:
While it's not _necessary_ to fix these unless they cause errors, this is a bit concerning:

``` julia
julia> using StaticArrays

julia> Base.Test.detect_ambiguities(StaticArrays, Base, Core)
Skipping StaticArrays.MutableFieldVector
Skipping Base.<|
Skipping Base.>:
Skipping Base.cluster_manager
17-element Array{Tuple{Method,Method},1}:
 (+(a::Number, b::StaticArrays.StaticArray) at /home/tim/.julia/v0.6/StaticArrays/src/linalg.jl:19,+(x::Bool, A::AbstractArray{Bool,N<:Any}) at arraymath.jl:91)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
 (-(a::StaticArrays.StaticArray, b::AbstractArray) at /home/tim/.julia/v0.6/StaticArrays/src/linalg.jl:16,-(x::Union{Base.ReshapedArray{T<:Any,1,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N<:Any}}},DenseArray{T<:Any,1},SubArray{T<:Any,1,A<:Union{Base.ReshapedArray{T<:Any,N<:Any,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N<:Any}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N<:Any}},L<:Any}}, y::AbstractSparseArray{Tv<:Any,Ti<:Any,1}) at sparse/sparsevector.jl:1184)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
[...]

PR #521, opened 2018-10-22
Deprecate ImmutableArrays and FixedSizeArrays compatibility interfaces

c42f:
As briefly mentioned in #503 

At this stage, people should use the StaticArrays APIs directly. If there's anything useful left in StaticArrays.FixedSizeArrays we should fold it into the main API, or move it to other packages where it fits more neatly.

CC @andyferris @SimonDanisch 

[edit] Closes https://github.com/JuliaArrays/StaticArrays.jl/issues/459

### DIFF
--- a/src/FixedSizeArrays.jl
+++ b/src/FixedSizeArrays.jl
@@ -6,7 +6,7 @@ etc, using StaticArrays as a backend.
 The type definitions are not "perfect" matches because the type parameters are
 different. However, it should cover common method signatures and constructors.
 """
-module FixedSizeArrays
+module FixedSizeArraysWillBeRemoved
 
 using ..StaticArrays
 
@@ -187,3 +187,6 @@ end
 @fixed_vector Point StaticVector
 
 end
+
+Base.@deprecate_binding FixedSizeArrays FixedSizeArraysWillBeRemoved #=
+    =# false "StaticArrays.FixedSizeArrays is deprecated. Use StaticArrays directly."
--- a/src/ImmutableArrays.jl
+++ b/src/ImmutableArrays.jl
@@ -1,4 +1,4 @@
-module ImmutableArrays
+module ImmutableArraysWillBeRemoved
 
 using ..StaticArrays
 
@@ -34,3 +34,6 @@ export Vector1,   Vector2,   Vector3,   Vector4,
        Matrix4x1, Matrix4x2, Matrix4x3, Matrix4x4
 
 end # module
+
+Base.@deprecate_binding ImmutableArrays ImmutableArraysWillBeRemoved #=
+    =# false "StaticArrays.ImmutableArrays is deprecated. Use StaticArrays directly."


SimonDanisch:
I'm not against this, but we'll need to make extra sure that the transition is smooth.
Almost all my packages rely on this, so messing this up would take out quite a lot of packages^^

SimonDanisch:
I'm not against this, but we'll need to make extra sure that the transition is smooth.
Almost all my packages rely on this, so messing this up would take out quite a lot of packages^^

mschauer:
Data point: I have no dependency on that code since a long time.

timholy:
Deprecating those will help a *lot* with #18. And I'm not sure it's even being used as is; https://github.com/JuliaGeometry/GeometryTypes.jl/blob/65f4f2a65f353968aa0eaf9ee18a1a23352290c3/src/types.jl#L47 suggests `GeometryTypes` is creating its own `Vec` rather than using the one in `StaticArrays.FixedSizeArrays`.

timholy:
Deprecating those will help a *lot* with #18. And I'm not sure it's even being used as is; https://github.com/JuliaGeometry/GeometryTypes.jl/blob/65f4f2a65f353968aa0eaf9ee18a1a23352290c3/src/types.jl#L47 suggests `GeometryTypes` is creating its own `Vec` rather than using the one in `StaticArrays.FixedSizeArrays`.

c42f:
@SimonDanisch yes I don't want to leave you scrambling! Do your packages depend on it via `GeometryTypes`? Perhaps we could move `@fixed_vector` to there?

andyferris:
IIRC there was a GeometryTypes PR to move the functionality there?

c42f:
Ah yes: https://github.com/JuliaGeometry/GeometryTypes.jl/pull/153. Good timing (thanks @aaronang).

See also https://github.com/JuliaGeometry/GeometryTypes.jl/issues/132 and #459 which I didn't see before making this PR :)

mkborregaard:
This broke Plots.

KristofferC:
Minor releases are typically breaking.

mkborregaard:
Also, this https://github.com/JuliaLang/METADATA.jl/pull/19050#issuecomment-432324269

c42f:
Sorry for the breakage, the intention was to provide a proper deprecation warning, and indeed this does so on the surface:

```julia
julia> using StaticArrays.FixedSizeArrays
WARNING: StaticArrays.FixedSizeArrays is deprecatedStaticArrays.FixedSizeArrays is deprecated. Use StaticArrays directly.
 in module StaticArrays

julia> Vec(1,2,3)
3-element Vec{3,Int64}:
 1
 2
 3
```

(botched warning notwithstanding)

So now we need to figure out *why* it broke it, and quickly release a patch update.

fredrikekre:
I am working on a patch. The problem is that before this worked:
```
using StaticArrays.FixedSizeArrays
FixedSizeArrays.Vec
```
since the `using` call brought in `FixedSizeArrays`, but after this PR that does not happen.

c42f:
Oh I see.

PR #524, opened 2018-10-22
(RFC) fix eigen in the diagonal 2x2-case

### LINKED ISSUE #523: Order of eigenvectors of (symmetric) 2x2 matrices
I guess the following is a (quite scary) bug:
```
using LinearAlgebra, StaticArrays, Test
S = SMatrix{2,2}((70., 0., 0., 1/70))
E = eigen(S)
Λs = eigvals(E)
V = eigvecs(E)
@test S * V  ≈ V * Diagonal(eigvals(E))
M = Matrix(S)
F = eigen(M)
Λm = eigvals(F)
@test Λs ≈ Λm
U = eigvecs(F)
@test M * U ≈ U * Diagonal(eigvals(F))
```
I thought one should be able to rely on that `eigvecs(E)[:,1]` is the eigenvector corresponding to `eigvals(E)[1]`, like in the normal matrix case? Or how would one do that otherwise?

dkarrasch:
Fixes #523. I also reduced the amount of `@inbounds` by tearing it to the outside. I think it should apply everywhere in the following block.

Some questions:
* Is `T` equal to `eltype(A)`?
* Is there a noticeable difference between `TA(1)`, `TA(0)` and `one(TA)`, `zero(TA)`?

TODO: add tests for the diagonal 2x2 case.

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -136,19 +136,19 @@ end
     TA = eltype(A)
 
     @inbounds if A.uplo == 'U'
-        if a[3] == 0
+        if a[3] == 0 # A is diagonal
             A11 = a[1]
             A22 = a[4]
             if A11 < A22
                 vals = SVector(A11, A22)
-                vecs = @SMatrix [TA(1) TA(0);
-                                 TA(0) TA(1)]
+                vecs = @SMatrix [one(TA) zero(TA);
+                                 zero(TA) one(TA)]
             else # A22 <= A11
                 vals = SVector(A22, A11)
-                vecs = @SMatrix [TA(0) TA(1);
-                                 TA(1) TA(0)]
+                vecs = @SMatrix [zero(TA) one(TA);
+                                 one(TA) zero(TA)]
             end
-        else
+        else # A is not diagonal
             t_half = real(a[1] + a[4]) / 2
             d = real(a[1] * a[4] - a[3]' * a[3]) # Should be real
 
@@ -171,19 +171,19 @@ end
         end
         return (vals, vecs)
     else # A.uplo == 'L'
-        if a[2] == 0
+        if a[2] == 0 # A is diagonal
             A11 = a[1]
             A22 = a[4]
             if A11 < A22
                 vals = SVector(A11, A22)
-                vecs = @SMatrix [TA(1) TA(0);
-                                 TA(0) TA(1)]
+                vecs = @SMatrix [one(TA) zero(TA);
+                                 zero(TA) one(TA)]
             else # A22 <= A11
                 vals = SVector(A22, A11)
-                vecs = @SMatrix [TA(0) TA(1);
-                                 TA(1) TA(0)]
+                vecs = @SMatrix [zero(TA) one(TA);
+                                 one(TA) zero(TA)]
             end
-        else
+        else # A is not diagonal
             t_half = real(a[1] + a[4]) / 2
             d = real(a[1] * a[4] - a[2]' * a[2]) # Should be real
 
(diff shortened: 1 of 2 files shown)


dkarrasch:
Thanks, @andyferris . I just checked `Base`, one found that
https://github.com/JuliaLang/julia/blob/70be5f25981918af138cce21bace331aa2d16945/base/number.jl#L297
so these don't make a difference, in contrast to the constructors. I'll replace them with `one`/`zero`. The other question is whether the `T` in `A::LinearAlgebra.RealHermSymComplexHerm{T}` always coincides with `eltype(A)`? If that's the case, then I could reduce the amount of types and variables floating around a bit.

fredrikekre:
Unrelated to the bugfix, but the codepath for `'L'` and `'U'` is only different in 2 places, and I think this method could just be
```
@inline function _eig(::Size{(2,2)}, A::LinearAlgebra.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
    a = A.data
    TA = eltype(A)
    lower = A.uplo === 'L'

    @inbounds begin
        A11 = a[1]
        A22 = a[4]
        A2112 = lower ? a[2] : a[3]

        if A2112 == 0 # A is diagonal
            if A11 < A22
                vals = SVector(A11, A22)
                vecs = @SMatrix [one(TA) zero(TA);
                                 zero(TA) one(TA)]
            else # A22 <= A11
                vals = SVector(A22, A11)
                vecs = @SMatrix [zero(TA) one(TA);
                                 one(TA) zero(TA)]
            end
        else # A is not diagonal
            t_half = real(A11 + A22) / 2
            d = real(A11 * A22 - A2112' * A2112) # Should be real

            tmp2 = t_half * t_half - d
            tmp = tmp2 < 0 ? zero(tmp2) : sqrt(tmp2) # Numerically stable for identity matrices, etc.
            vals = SVector(t_half - tmp, t_half + tmp)

            v11 = vals[1] - A22
            n1 = sqrt(v11' * v11 + A2112' * A2112)
            v11 = v11 / n1
            v12 = (lower ? A2112 : A2112') / n1

            v21 = vals[2] - A22
            n2 = sqrt(v21' * v21 + A2112' * A2112)
            v21 = v21 / n2
            v22 = (lower ? A2112 : A2112') / n2

            vecs = @SMatrix [ v11  v21 ;
                              v12  v22 ]
        end
    end
    return (vals, vecs)
end
```
which reduced the code duplication quite a lot.

dkarrasch:
Yes, I was aware of that. I wasn't sure about the original intention to have that code duplication, so I felt it must have been something beyond my understanding of compiler functionality (which is very low). I'll use @fredrikekre's version then.

dkarrasch:
Hm, on my machine (tested on a `Symmetric(SMatrix)`) I get `7.290 ns` on master vs `7.532 ns` on this PR vs. `7.624 ns` with @fredrikekre's suggestion. So it seems that we do pay a price here, anyways. I can reduce runtime slightly to `7.512 ns` by replacing `a[*] == 0` by `a[*] == zero(TA)`!

src/eigen.jl
```diff
@@ -136,7 +136,7 @@ end
     TA = eltype(A)
 
     @inbounds if A.uplo == 'U'
-        if a[3] == 0 # A is diagonal
+        if a[3] == zero(TA) # A is diagonal
```
mschauer: Is this ever needed? You can compare with 0, won't that work fine? For example no need to create a `zero(DualNumber)` if the type is a dual number.

dkarrasch: I guess it depends on what is cheaper: converting `0` to a comparable type or constructing the zero element. I just tested one versus the other, and for `Float` it doesn't make any difference (in contrast to my minor improvement above), but for `Dual` the comparison `== zero(Dual)` is indeed more expensive than `== 0`. I guess I'll revert the last commit then?

Issue #522, opened 2018-10-22
New release 0.9.0


c42f:
Hi guys,

I want the fix from #515 in a release, so I'm planning a new one. Here's the changelog for PRs which affect the code:

#484 - Fixed MMatrix boundshecks, faster getindex
#491 - add propagate_inbounds for SArray getindex
#503 - Fix various inference issues; avoid calling `Core.Compiler.return_type()`
#515 - Reinstate read(::IO, Type{<:StaticArray})
#511 - Optimized methods for Triangular * Triangular
#507 - Improve performance of Iteators.flatten with arrays of static arrays.
#519 - Fix missing size error for constructors

All of these are minor fixes and improvements, with the exception of #503 which is enough of a wildcard that I think we need to bump the version number to v0.9.0 ... hopefully it won't break many things but I guess we're not going to be sure until we try.

Actually perhaps we should go version 1.0 and stop pretending this package is in pre-release ;-)  Then we can start using semver properly.

@andyferris mentioned deprecating some old compatibility shims in #503. For that I've just created #521 which we could deal with before release. ~~Is there anything else?~~

## TODO 0.9
* [x] Deprecate old fixed size interfaces (#512)
* [x] Decide whether #525 goes in here (resolution: no)

andyferris:
Thanks Chris, sounds perfect. I’d love to deprecate those modules before v1. 

We could add depreciations to v0.9.0 and immediately remove them for v1.0.0?

fredrikekre:
I suggest removing all deprecations for this release too.

andyferris:
> all deprecations

Are we happy to drop v0.7 support at this stage?

andyferris:
> Is there anything else?

I think #518 (and potentially #520) will have breaking changes, too. (We can make a release without these, but resolving #518 before v1.0.0 would be nice. I'm thinking of deprecating the one-argument constructors in a v0.9 and get people to use `convert`, which we should fix up to work more robustly (i.e. fix #520)).

c42f:
I added TODO to the issue description for 0.9 and 1.0. Feel free to edit and update as required.

andyferris:
Looks good to me. Let's do v0.9, then.

c42f:
> I'm thinking of deprecating the one-argument constructors in a v0.9 and get people to use convert

[...]

PR #525, opened 2018-10-23
Drop Julia v0.7 support, remove deprecations

andyferris:
We have a solid release for Julia v0.7, which should be good enough for users needing to upgrade to v1.0. Personally, I don't feel it is valuable keeping around the old fluff.

I'm interested in getting this in for #522

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
   - 1.0
   - nightly
 notifications:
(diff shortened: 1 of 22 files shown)


andyferris:
FYI, I kept the deprecations for #515 and #521, since there is no release with those.

andyferris:
> That way people can have consistent functionality in StaticArrays-0.9 in both julia-0.7 and julia-1.0?

Consistent? The idea here was that julia-0.7 users wouldn't be able to install StaticArrays-0.9. StaticArrays-0.8 provides a consistent, upgrade layer.

Is it more that the bugs fixed in StaticArrays-0.9 will be important enough to include for users simply upgrading their Julia-0.6 code to Julia-1.0?

c42f:
Yes, there's bugs fixed in StaticArrays-0.9; they're minor but did break my own code when upgrading from 0.6.

My point is, we loose very little by delaying the removal of these potentially helpful deprecations (and the ability to install on 0.7) by another release. And it could be rather helpful for other package maintainers.

Just a thought, no major worries if it's merged right away.

mschauer:
Can we remove the deprecation warnings without bumping the required version? I suppose the new release will run just fine on 0.7. (I prefer my coworkers using 0.7 for the time being, that makes it easier for me to help them and for them to help themselves using stackoverflow etc.)

c42f:
@mschauer I think that's quite reasonable. That will let users who are on 0.7 continue to get StaticArrays related bugfixes.

The only downside being that the depwarns won't be consistent with `Base` 0.7 related depwarns. But I think that's a reasonable tradeoff.

c42f:
Partially obsoleted by #600 (though I didn't drop 0.7 support there)

mschauer:
My earlier objection is obsolete now, in any case.

PR #527, opened 2018-10-23
Fix bug in module deprecation from 0.9.0 release

c42f:
With the bug, `using StaticArrays.FixedSizeArrays` did not result in a
module name `FixedSizeArrays` in the current namespace, thus breaking
explicitly qualified name lookups like `FixedSizeArrays.Vec(1,2)`.

Provide a binding for this explicitly.

@fredrikekre — here's one possible fix. Thoughts?

### DIFF
--- a/src/FixedSizeArrays.jl
+++ b/src/FixedSizeArrays.jl
@@ -10,6 +10,9 @@ module FixedSizeArraysWillBeRemoved
 
 using ..StaticArrays
 
+const FixedSizeArrays = FixedSizeArraysWillBeRemoved
+export FixedSizeArrays # Ensure deprecated module name is in scope after import
+
 export FixedArray
 export FixedVector
 export FixedMatrix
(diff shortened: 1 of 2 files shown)


c42f:
Oh, I see what you've done there (had a double take reading the diff :-) )

I dunno, I actually kind of prefer not to spam users with a massive pile of depwarns, but just a single one when they first refer to the deprecated binding.

fredrikekre:
Fair enough.

fredrikekre:
Confirmed that Plots-pre-https://github.com/JuliaPlots/Plots.jl/pull/1803 works fine with this PR.

c42f:
Excellent, thanks!

c42f:
I won't have time for a couple of hours to merge & tag once this finishes CI. So if you feel inclined to do so, that would be great. If not, I'll do it later today.

c42f:
Thanks @fredrikekre, I appreciate your help tracking down the real problem and pushing out the release quickly!

Issue #27, opened 2016-09-06
violation of GC safety


vtjnash:
the MVector violates gc safety when it contains non-bitstypes (it lacks a write barrier), resulting in the possibility that the gc will lose the object and delete the objects in the array while still in-use:
https://github.com/JuliaArrays/StaticArrays.jl/blob/1e18069572c15d96c8ee37750e0cfb9c213ba905/src/MVector.jl#L49-L50

andyferris:
Thank you very much Jameson! I was quite worried about this, and was hoping someone more knowledgeable/cleverer than I would take a look someday. (You might be able to tell from the comment).

Can you point me somewhere that explains the correct procedure, or an example of where the correct procedure is implemented? Otherwise I'll have to make that branch an error for now (it's probably not used in the wild).

vtjnash:
I don't think it's possible to implement correctly, maybe @yuyichao can say for certain?

yuyichao:
In additional to gc safety issue, this also violate the immutability since the tuple is not inlined if the type is not a bits type so this is not really fixable.

The GC safety can be worked around (does not apply here due to the immutable issue) by writing llvm code using `llvmcall` that triggers the write barrier manually. (Do note that the llvm code is different on 0.4 and 0.5). This is not recommended for obvious reasons.

andyferris:
> In additional to gc safety issue, this also violate the immutability since the tuple is not inlined if the type is not a bits type so this is not really fixable.

I don't understand this. I'm not sure what you mean by inlined. I *think* I've found the correct bit of memory to write the pointers to. `MVector` is not a mutable type - it's immutable fields can be replaced (wholly, or partially as an optimization, right?). `setindex!` is not defined on `SVector`.

> The GC safety can be worked around (does not apply here due to the immutable issue) by writing llvm code using llvmcall that triggers the write barrier manually. (Do note that the llvm code is different on 0.4 and 0.5). This is not recommended for obvious reasons.

Or calling an appropriate C function, I assume? Maybe there is something I'm missing.

yuyichao:
AFAICT `MVector` is a mutable type and the field cannot be partially replace if not inlined.

> Or calling an appropriate C function

Sure, if you write your own.

yuyichao:
Inlined means the memory of the field is, .... well ...., inlined and the field is not a reference to another piece of memory.

andyferris:
[...]

PR #530, opened 2018-10-24
Add GC.at-preserve for some pointer calls, fix #529.

### LINKED ISSUE #529: MVector reset the value in v0.9.1
Hi guys,

After upgrading to v0.9.1, I started to see a strange behavior when using `MVector`. Take a look at this code:

```
    xs2 = zeros(MVector{N2, T})
    ys2 = zeros(MVector{N2, T})

    for k = 1:N2
        xs2[k] = _zeta(re, zn2[k], z1)/zgdif
        ys2[k] = 1 / tn2[k]
    end

    println("$(ys2[1])")
    yd1 = -tgn2[1] / (t1*t1) * zgdif
    println("$(ys2[1])")
    yd2 = -tgn2[2] / (t2*t2) * zgdif * ( (re+z2)/(re+z1) )^2
```

The output is:

```
0.004568090158377094
4.94e-322
```

I have no idea why the value of `ys2` is changing after this operation. Reverting back to 0.8.3 fix the problem.

fredrikekre:
#484 caused #529 (cc @chethega ) and these `GC.@preserve` fixes it.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -113,21 +113,21 @@ end
 ####################
 
 @propagate_inbounds function getindex(v::MArray, i::Int)
-    @boundscheck checkbounds(v,i) 
+    @boundscheck checkbounds(v,i)
     T = eltype(v)
-    
+
     if isbitstype(T)
-        return unsafe_load(Base.unsafe_convert(Ptr{T}, pointer_from_objref(v)), i)
+        return GC.@preserve v unsafe_load(Base.unsafe_convert(Ptr{T}, pointer_from_objref(v)), i)
     end
     v.data[i]
 end
 
 @inline function setindex!(v::MArray, val, i::Int)
-    @boundscheck checkbounds(v,i) 
+    @boundscheck checkbounds(v,i)
     T = eltype(v)
 
     if isbitstype(T)
-        unsafe_store!(Base.unsafe_convert(Ptr{T}, pointer_from_objref(v)), convert(T, val), i)
+        GC.@preserve v unsafe_store!(Base.unsafe_convert(Ptr{T}, pointer_from_objref(v)), convert(T, val), i)
     else
         # This one is unsafe (#27)
         # unsafe_store!(Base.unsafe_convert(Ptr{Ptr{Nothing}}, pointer_from_objref(v.data)), pointer_from_objref(val), i)


c42f:
This looks good to me. Is there a MWE of #529 we can add as a regression test?

fredrikekre:
I could not find an MWE, it does not trigger in simple cases.

andyferris:
Cool.

Any performance impacts?

ronisbr:
Btw, I have been trying to find a MWE to reproduce what I am seeing in SatelliteToolbox without success. However, it is indeed related to the garbage collector.

c42f:
I had a go trying to make an MWE for this, but the optimizer outsmarted me every time so far.

What we need is something like the following:

```julia
# Use side effects to encourage allocation of v
@noinline do_alloc(x) = (v = MVector((x,)); @show v; v)

function foo(x)
    a = do_alloc(x)
    # Due to the bug, `a` may be considered unused after this point (maybe?)
    # *** We attempt to garbage collect it:
    GC.gc()
    # Now allocate a new object of the same size - likely to reuse the memory for `a`.
    b = do_alloc(x)
    a[1] = 0 # Attempt to clobber memory in `b`
    return b
end
```

However, this doesn't trigger the bug.  At least one reason is that the GC root placement pass is putting the roots for `a` and `b` together into the same GC frame so they will be preserved together:

```asm
julia> @code_native foo(1)
	.text
; Function foo {
; Location: REPL[3]:2
	pushq	%r15
	pushq	%r14
	pushq	%r12
	pushq	%rbx
	subq	$24, %rsp
	movq	%rdi, %r14
	vxorps	%xmm0, %xmm0, %xmm0
	vmovaps	%xmm0, (%rsp)
	movq	$0, 16(%rsp)
	movq	%fs:0, %r15
	movq	$2, (%rsp)   ; * $2  =>  two roots next to each other
	movq	-10920(%r15), %rax
	movq	%rax, 8(%rsp)
	movq	%rsp, %rax
	movq	%rax, -10920(%r15) ; * Push GC frame
	movabsq	$julia_do_alloc_35139, %r12
	callq	*%r12
	movq	%rax, %rbx
	movq	%rbx, 16(%rsp)
...
```

To trigger the bug, we'd need to trick the GC root placement pass into placing these roots separately.

c42f:
Agreed, please go ahead and merge it, the fix looks good to me.

Having tried as far as the above, I think it's quite fair to say the optimizer has defeated me for now :-)

andyferris:
Thanks, @fredrikekre

andyferris:
A tag would be nice for this one, I think.

andyferris:
I just realized... GC preservation should let us do the non-isbits types safely as pointers, like we used to. Right?

ronisbr:
> A tag would be nice for this one, I think.

Hi guys! Thanks for the quick fix! I would really appreciate if a new tag is released, since SatelliteToolbox will not pass the tests withou it.

fredrikekre:
https://github.com/JuliaLang/METADATA.jl/pull/19108

c42f:
> I just realized... GC preservation should let us do the non-isbits types safely as pointers, like we used to. Right?

No I don't think so. Reading back over #27 reminds me that you need a GC write barrier as well as the pointer update... at the very least (`jl_gc_wb` in the C code). There's also the issue of inlining the tuple vs not inlining.

ronisbr:
Just for the records, the tests are now passing without any problems! Thanks everyone :)

I tried really hard to reproduce it. I even copied all the related functions and executed them outside SatelliteToolbox.jl and I was not able to reproduce it. Furthermore, I think it will be an overkill to add the entire NRLMSISE-00 model into a test to StaticArrays to get this particular bug.

c42f:
Yeah, it might not be as simple as I was imagining above, it could also depend on particular rearrangements which the LLVM optimizer is allowed to do if you don't explicitly preserve the root. I'm not sure.

PR #534, opened 2018-10-29
Add `SOneTo` as `AbstractUnitRange` to represent `axes(::StaticArray)`

### LINKED ISSUE #476: Should axes return a tuple of `StaticArray`s?
At the moment the dimensions of a static array are lost in `axes`, which may be inconvenient in certain settings. Maybe a `StaticOneTo{N, T}` could be added?

andyferris:
This has been on the to-do list for 2 years! :tada: 

Now `axes(::StaticArray)` will return a tuple of "static unit ranges / slices". The `SOneTo` type lives under `AbstractUnitRange` so it can work properly with `Base`, and is therefore not a `StaticArray` subtype.

One interesting side-effect is what happens to `collect`:

```julia
julia> v = SVector(1,2,3)
3-element SArray{Tuple{3},Int64,1,3}:
 1
 2
 3

julia> [x for x in v]
3-element SizedArray{Tuple{3},Int64,1,1}:
 1
 2
 3
```
This matches the *OffsetArrays* package... basically if you want to mess with `axes` your behavior is automatically used by `collect`, and in particular for offsets (`Slice`s) you kind of need to rely on this to work out properly. I guess we should roll with this for now. If we can get generator/comprehension syntax working fast (constructing `SVector`s with optimal code) that could be pretty neat.

I believe the benefits of this will eventually be far-reaching. In particular the `axes` and `keys` of an array can now be directly used as an indication of it's *static-sizedness*. This is more-or-less a trait that is already plumbed into many systems (like `collect`). Hopefully this will make dealing with "wrapped up" types like `Diagonal` with `SVector` more natural.

Closes #476

### DIFF
--- a/src/SOneTo.jl
+++ b/src/SOneTo.jl
@@ -1,0 +1,60 @@
+"""
+    SOneTo(n)
+
+Return a statically-sized `AbstractUnitRange` starting at `1`, functioning as the `axes` of
+a `StaticArray`.
+"""
+struct SOneTo{n} <: AbstractUnitRange{Int}
+end
+
+SOneTo(n::Int) = SOneTo{n}()
+
+Base.axes(s::SOneTo) = (s,)
+Base.size(s::SOneTo) = (length(s),)
+Base.length(s::SOneTo{n}) where {n} = n
+
+function Base.getindex(s::SOneTo, i::Int) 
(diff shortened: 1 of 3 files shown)


andyferris:
Ready or not, here it comes...

PR #543, opened 2018-11-16
Return DimensionMismatch rather than ErrorException

timholy:
Seems reasonable to use the "official" exception type here. We don't seem to have a specific test for this (it's probably a `@test_throws Exception ...` test); I can make it more specific if desired.

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -13,7 +13,7 @@
 @inline Tuple(a::StaticArray) = unroll_tuple(a, Length(a))
 
 @noinline function dimension_mismatch_fail(SA::Type, a::AbstractArray)
-    error("Dimension mismatch. Expected input array of length $(length(SA)), got length $(length(a))")
+    throw(DimensionMismatch("expected input array of length $(length(SA)), got length $(length(a))"))
 end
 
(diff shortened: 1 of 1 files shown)


timholy:
Want me to make the test more specific to lock this in place? Say the word and I'll do it.

Will want this tagged at some point for https://github.com/JuliaImages/ImageTransformations.jl/pull/63

andyferris:
Yes - you’re of course very welcome to help out with this stuff! ;)

(To expand on my earlier response, I find that I often use return types in my code to handle the problems that I want to handle, and generally expect exceptions to exit. I’m frequently left with no idea what exception type to return because I can’t imagine the case I would catch these kinds of exceptions - ones like this tend to result from code bugs rather than input data, or interaction with live systems, or whatever).

timholy:
95% of the time I use the exception type only for tests, to make sure it's failing the way I intend it to rather than for a far stupider reason. One of my greatest coding skilz is to introduce a typo in an error branch and getting a `MethodError` or `UndefVarError` while en route to what I hoped would throw a `DimensionMismatch`. Oops.

PR #544, opened 2018-11-17
Increase exception-type specificity for a few errors and tests

timholy:
The stunning sequel to the blockbuster movie #543.


timholy:
In case others tag this before me: tag needs to be 0.11.0 since changing the exception type is a breaking change.

Issue #540, opened 2018-11-10
MArray is slower than Array in reduction


YingboMa:
MWE:
```julia
julia> using BenchmarkTools

julia> a = @MArray rand(4);

julia> @btime sum( $(Array(a)) )
  3.896 ns (0 allocations: 0 bytes)
1.5713808992800153

julia> @btime sum( $a )
  30.021 ns (1 allocation: 16 bytes)
1.5713808992800153
```

c42f:
This is still the case on the latest julia.  Ultimately it may be due to keyword functions (in this case `mapreduce`) not getting inlined.

tkoolen:
Yeah, in some of the Base reduce/mapreduce code, `NamedTuple`s are sometimes passed around as positional argument instead of using keyword arguments, probably for this reason.

https://github.com/JuliaLang/julia/blob/cce5a6b3d7303a4ca74a7f630b452d29b06e5903/base/reduce.jl#L39

mateuszbaran:
Unfortunately, passing keyword arguments around is relatively slow in Julia (it requires an `invoke`). The easiest solution is to define a bunch of methods like
```
@inline sum(a::StaticArray{<:Tuple,T}) where {T} = reduce(+, a)
```
to avoid passing keyword arguments. I can make a patch if you want.

c42f:
> I can make a patch if you want.

Yes please!

Did you read the code @tkoolen pointed to above? It seemed to me that might be a way to avoid some of the cost.

mateuszbaran:
> Did you read the code @tkoolen pointed to above? It seemed to me that might be a way to avoid some of the cost.

Yes, I'm trying to invent a less tedious solution.

c42f:
Fixed in #659 (thanks to @mateuszbaran :-) )

Issue #537, opened 2018-11-02
Certain `SArray` * `Adjoint(SArray)` produces `Array{Float64,2}` instead of `SArray`


zygmuntszpak:
A product of two StaticArrays results in an allocated array. 
Here is a minimal working example:
```julia
A = @SMatrix rand(9,1)
B = SArray{Tuple{9},Float64,1,9}(1,2,3,4,5,6,7,8,9)
C = adjoint(B)
D = A*C
typeof(D) # Array{Float64,2} instead of a StaticArray. 
```

zygmuntszpak:
I think the only time this occurs is when you have one entity that happens to be a column vector (single column of matrix), and the other a row vector, and you are computing the outer product of these two vectors. This arose when I was implementing a particular cost function. I've provided a small example below. I think I've also figured out which additional dispatch rules to add in order to address this. However, I've stopped short of a pull-request because I'm baffled by the fact that when I am summing the problematic term in my cost function (for which my proposed fix now returns a StaticArray) I am encountering a lot of inexplicable allocations. 
I'm not sure if this is a different bug, or of it is due to the fact that I haven't done enough in my proposed fix.

The proposed fix involves adding these two lines to `matrix_multiply.jl`
```julia
@inline *(A::StaticMatrix, B::Adjoint{<:Any, <:StaticVector}) = *(reshape(A, Size(Size(A)[1],)), B) 
@inline mul!(dest::StaticVecOrMat, A::StaticMatrix, B::Adjoint{<:Any, <:StaticVector}) = mul!(dest, reshape(A, Size(Size(A)[1],)), B) 
```

Here is an example that appears to solve the problem, but doesn't quite work because of a lot of inexplicable allocations. 

```julia
using StaticArrays, BenchmarkTools, LinearAlgebra

function hom(v::SVector)
    push(v,1)
end

function T(𝛉::AbstractArray, 𝒞::Tuple{AbstractArray, Vararg{AbstractArray}}, 𝒟::Tuple{AbstractArray, Vararg{AbstractArray}})
    ⊗ = kron
    l = length(𝛉)
    𝐈ₗ = SMatrix{l,l}(1.0I)
    𝐈ₘ =  SMatrix{1,1}(1.0I)
    𝐓 = @SMatrix zeros(l,l)
    N = length(𝒟[1])
    ℳ, ℳʹ = 𝒟
    Λ₁, Λ₂ = 𝒞
[...]

Issue #485, opened 2018-09-03
broadcasting of Tuples and SVectors is broken


dlfivefifty:
```julia
julia> (1,2) .+ [1,2]
2-element Array{Int64,1}:
 2
 4

julia> (1,2) .+ SVector(1,2) # I expected SVector(2,4)
ERROR: MethodError: no method matching +(::Tuple{Int64,Int64}, ::Int64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:502
  +(::Complex{Bool}, ::Real) at complex.jl:292
  +(::Missing, ::Number) at missing.jl:93
  ...
Stacktrace:
 [1] macro expansion at /Users/sheehanolver/.julia/packages/StaticArrays/Ze5H3/src/broadcast.jl:133 [inlined]
 [2] _broadcast at /Users/sheehanolver/.julia/packages/StaticArrays/Ze5H3/src/broadcast.jl:94 [inlined]
 [3] copy at /Users/sheehanolver/.julia/packages/StaticArrays/Ze5H3/src/broadcast.jl:24 [inlined]
 [4] materialize(::Base.Broadcast.Broadcasted{StaticArrays.StaticArrayStyle{1},Nothing,typeof(+),Tuple{Tuple{Int64,Int64},SArray{Tuple{2},Int64,1,2}}}) at ./broadcast.jl:724
 [5] top-level scope at none:0
```

PR #547, opened 2018-11-18
Fix inference on axes for broadcasting

### LINKED ISSUE #546: Non-inferrability in broadcasting
Noticed in https://github.com/JuliaMath/Interpolations.jl/pull/269:

```julia
using StaticArrays, Test
h = SMatrix{1,1}([1.0])
steps = SVector(1.0)
testinf(h, steps) = h ./ (steps .* steps')

julia> testinf(h, steps)
1×1 SArray{Tuple{1,1},Float64,2,1}:
 1.0

julia> @inferred testinf(h, steps)
ERROR: return type SArray{Tuple{1,1},Float64,2,1} does not match inferred return type Any
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at none:0
```

timholy:
Fixes #546

### DIFF
--- a/src/SOneTo.jl
+++ b/src/SOneTo.jl
@@ -58,3 +58,6 @@ end
 Base.@pure function Base.checkindex(::Type{Bool}, ::SOneTo{n1}, ::SOneTo{n2}) where {n1, n2}
     return n1::Int >= n2::Int
 end
+
+Base.promote_rule(a::Type{Base.OneTo{T}}, ::Type{SOneTo{n}}) where {T,n} =
+    Base.OneTo{promote_type(T, Int)}
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -13,6 +13,8 @@ Base.axes(s::StaticArray) = _axes(Size(s))
 @pure function _axes(::Size{sizes}) where {sizes}
     map(SOneTo, sizes)
 end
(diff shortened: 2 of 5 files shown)


ChrisRackauckas:
Does this fix https://github.com/JuliaArrays/StaticArrays.jl/issues/540 ?

timholy:
Nope. Also doesn't fix #537 though it would seem that things here should at least make that easier. Finally, doesn't fix #485.

timholy:
See also https://github.com/JuliaLang/julia/pull/30074.

src/SOneTo.jl
```diff
@@ -58,3 +58,6 @@ end
 Base.@pure function Base.checkindex(::Type{Bool}, ::SOneTo{n1}, ::SOneTo{n2}) where {n1, n2}
     return n1::Int >= n2::Int
 end
+
+Base.promote_rule(a::Type{Base.OneTo{T}}, ::Type{SOneTo{n}}) where {T,n} =
+    Base.OneTo{promote_type(T, Int)}
```
timholy: Hmm, this is not strictly necessary for the fix (it's just the first thing I tried). I could get rid of this if you have reservations about it. (One also has to be rid of the tests this adds to `core.jl`.)

andyferris: I think this is good.

PR #550, opened 2018-11-19
No tuple creation for SizedArrays

BeastyBlacksmith:
Tuple creation is unnecessary for SizedArrays and causes massive slowdown for big arrays.

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -158,7 +158,7 @@ reshape(a::Array, s::Size{S}) where {S} = s(a)
 @inline vec(a::StaticArray) = reshape(a, Size(prod(Size(typeof(a)))))
 
 @inline copy(a::StaticArray) = typeof(a)(Tuple(a))
-@inline copy(a::SizedArray) = typeof(a)(a)
+@inline copy(a::SizedArray) = typeof(a)(copy(a.data))
 
 @inline reverse(v::StaticVector) = typeof(v)(reverse(Tuple(v)))
 
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -154,6 +154,7 @@ reshape(a::Array, s::Size{S}) where {S} = s(a)
 @inline vec(a::StaticArray) = reshape(a, Size(prod(Size(typeof(a)))))
 
 @inline copy(a::StaticArray) = typeof(a)(Tuple(a))
+@inline copy(a::SizedArray) = typeof(a)(a)
 
 # TODO permutedims?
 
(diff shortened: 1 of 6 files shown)


src/abstractarray.jl
```diff
@@ -154,6 +154,7 @@ reshape(a::Array, s::Size{S}) where {S} = s(a)
 @inline vec(a::StaticArray) = reshape(a, Size(prod(Size(typeof(a)))))
 
 @inline copy(a::StaticArray) = typeof(a)(Tuple(a))
+@inline copy(a::SizedArray) = typeof(a)(a)
```
andyferris: Does this really create a copy?

BeastyBlacksmith: As far as I can tell:
```julia
julia> copy(a::SizedArray) = typeof(a)(a)
copy (generic function with 88 methods)

julia> a = Size(100)(rand(100));

julia> b = copy(a);

julia> b[1] = 2
2

julia> a[1], b[1]
(0.3688743314251053, 2.0)
```

BeastyBlacksmith: Another option would be to make `SizedArray` subtype of `AbstractArray` rather than `StaticArray`.

BeastyBlacksmith: @andyferris Another option would be to make `SizedArray` subtype of `AbstractArray` rather than `StaticArray`.

c42f: Uh oh. You're right @BeastyBlacksmith, this is creating a copy:

```

julia> v = SizedVector{2}(zeros(2))
2-element SizedArray{Tuple{2},Float64,1,1}:
 0.0
 0.0

julia> v2 = SizedVector{2}(v)
2-element SizedArray{Tuple{2},Float64,1,1}:
 0.0
 0.0

julia> @which SizedVector{2}(v)
(::Type{SA})(a::StaticArray) where SA<:StaticArray in StaticArrays at ~/.julia/dev/StaticArrays/src/convert.jl:5

julia> @test v2.data === v.data
Test Failed at REPL[16]:1
  Expression: v2.data === v.data
   Evaluated: [0.0, 0.0] === [0.0, 0.0]
ERROR: There was an error during testing
```

But I think this is a bug; I believe the intention of `SizedArray` was to wrap the underlying data. Indeed, we have

```julia
julia> @test SizedVector{2}(data).data === data
Test Passed
```

which seems rather inconsistent as the normal constructor simply wraps the data rather than copies it.

I think the correct solution here is to add a `copy(a)` to your implementation, and to specialize the constructor of `SizedArray` (or, possibly better, `convert`?) so that it wraps rather than copies the data.

BeastyBlacksmith:
This should do the trick?

andyferris:
Thanks @BeastyBlacksmith

BeastyBlacksmith:
You are welcome :)

PR #552, opened 2018-11-20
added static bitsets

chethega:
This PR adds bitwise operations for SVectors and a `<:AbstractSet{Int}` view of `SVector{N, UInt64} where {N}`. 

This set view is mainly interesting as an iterator over the set bits, as well as nice pretty printing. I am intentionally not storing the length, so that conversion from SVector is a no-op.

Tests are included, docs are missing. Also, is this the right way of polyfilling `Base._blsr`?

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -1,19 +1,22 @@
-import Base: +, -, *, /, \
+import Base: +, -, *, /, \, &, |, xor, ~
 
 # TODO: more operators, like AbstractArray
 
 # Unary ops
 @inline -(a::StaticArray) = map(-, a)
+@inline ~(a::StaticArray) = map(~, a)
 
 # Binary ops
 # Between arrays
-@inline +(a::StaticArray, b::StaticArray) = map(+, a, b)
-@inline +(a::AbstractArray, b::StaticArray) = map(+, a, b)
-@inline +(a::StaticArray, b::AbstractArray) = map(+, a, b)
-
-@inline -(a::StaticArray, b::StaticArray) = map(-, a, b)
-@inline -(a::AbstractArray, b::StaticArray) = map(-, a, b)
-@inline -(a::StaticArray, b::AbstractArray) = map(-, a, b)
+let
+    ops = [:+, :-, :&, :|, :xor]
+    type_combinations = [(StaticArray, StaticArray), (StaticArray, AbstractArray), (AbstractArray, StaticArray)]
+    for op in ops
+        for (_L, _R) in type_combinations
+            @eval @inline $op(a::$(_L), b::$(_R) )=map($op, a, b) 
+        end
+    end
+end
 
 # Scalar-array
 @inline +(a::Number, b::StaticArray) = broadcast(+, a, b)
--- a/src/SBitSet.jl
+++ b/src/SBitSet.jl
@@ -1,0 +1,38 @@
+struct SBitSet{N}<:AbstractSet{Int64}
+    chunks::SVector{N,UInt64}
+end
+
+@inline function Base.iterate(B::SBitSet{N}) where N
+    N==0 && return nothing
+    return iterate(B, (1, @inbounds B.chunks[1]))
+end
+
+@inline function Base.iterate(B::SBitSet{N}, s) where N
+    N==0 && return nothing
+    i1, c = s
+    while c==0
+        i1 % UInt >= N % UInt && return nothing
+        i1 += 1
+        @inbounds c = B.chunks[i1]
+    end
+    tz = trailing_zeros(c) + 1
+    c = _blsr(c)
+    return ((i1-1)<<6 + tz, (i1, c))
+end
+
+@inline Base.isempty(B::SBitSet) = iszero(B.chunks)
+@inline Base.length(B::SBitSet) = count_ones(B.chunks)
+@inline Base.union(B::SBitSet, C::SBitSet) = SBitSet(B.chunks | C.chunks)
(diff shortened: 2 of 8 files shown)


chethega:
>I'm prototyping a generic dictionary/array type thing privately; it will support interesting static-key dictionaries and sets and such things with fixed storage and so-on. Doesn't help you right now, of course...

Not a problem, I can use my code regardless of whether it gets merged, and you can feel free to cherry-pick whatever you like :)

Really the only nontrivial thing that `SBitSet` provides is iteration over nonzero bits. But since implementation of the immutable parts of the `AbstractSet` is only a couple of lines, well, why not.

Regarding bitwise OPs: Integers do have well-defined storage format. Would you be happier if the bitwise operations were restricted to pairs of SVectors of 1-16 byte (U)Ints? 


Broadcasted `.&` works, but kinda defies the convenient syntax point of not just using `NTuple`. When expressing logical conditions / circuits as a bitwise operation with 4+ inputs then legibility and syntax matter: They make the difference between a self-documenting one-liner and needing a comment to explain the mess of dots and parentheses.

andyferris:
> Regarding bitwise OPs: Integers do have well-defined storage format. Would you be happier if the bitwise operations were restricted to pairs of SVectors of 1-16 byte (U)Ints?

I think it may be non-crazy to have bitwise operations act on a `SVector`  which is an `isbitstype`, since the bit layout and therefore operation is predictable.

You could of course argue the same about `Complex{Int64}`, say.

```julia
julia> ~(1)
-2

julia> ~(1+im)
ERROR: MethodError: no method matching ~(::Complex{Int64})
Closest candidates are:
  ~(::BigInt) at gmp.jl:470
  ~(::Missing) at missing.jl:79
  ~(::Bool) at bool.jl:39
  ...
Stacktrace:
 [1] top-level scope at none:0
```

I'm not sure which way to go, to be honest.

chethega:
In some sense the `Base` precedent is against support of bitwise ops:
```
julia> [1%UInt]+[2%UInt]
1-element Array{UInt64,1}:
 0x0000000000000003

julia> [1%UInt]|[2%UInt]
ERROR: MethodError: no method matching |(::Array{UInt64,1}, ::Array{UInt64,1})
```
However, if people want to store bits packed in `UInt64` (which `Base` does!) then `[:~, :&, :|, :xor]` should be supported (with elementwise semantics and without dots) for `SVector{N,UInt64} where N` (`SVector{N,Bool} where N` might want to support this as well). 

`UInt64` appears to be julia's default for "untyped aligned quadword", and it stands to reason that bitwise operations make sense for `SVectors` of that. I don't think a lot of people want to do AD of boolean circuits, so I'm happy with this very restrictive dispatch until people complain.

src/linalg.jl
```diff
@@ -1,19 +1,22 @@
...
 # Between arrays
-@inline +(a::StaticArray, b::StaticArray) = map(+, a, b)
-@inline +(a::AbstractArray, b::StaticArray) = map(+, a, b)
-@inline +(a::StaticArray, b::AbstractArray) = map(+, a, b)
-
-@inline -(a::StaticArray, b::StaticArray) = map(-, a, b)
-@inline -(a::AbstractArray, b::StaticArray) = map(-, a, b)
-@inline -(a::StaticArray, b::AbstractArray) = map(-, a, b)
+let
+    ops = [:+, :-, :&, :|, :xor]
```
c42f: I think we should probably not do these bitwise ops at all unless they're also part of normal `AbstractArray`.

chethega:
Ok, let's revive and rewrite this. What I'm currently thinking is:

1. Bitwise ops for SVector go out. @c42f and @andyferris are both against, and I am the only person in favor. 
2. SBitSet stays <: AbstractSet{Int} and implements the set interface.
3. Additionally, SBitSet implements `getindex` and bitwise ops `&, |, xor, ~`. 
4. Additionally to `SBitSet`, we get a mutable `MBitSet` that has `setindex!`.
5. Because of 1, there is no gain in using SVector in the representation. I guess `NTuple{N, VecElement{UInt64}}` is the correct storage type?
6. For performance reasons, we intentionally stretch the set interface: `SBitSet{N}` only compare equal to other `SBitSet{N}` (or `MBitSet{N}`). Then we implement a fast hash. 
7. Optional extra goodies: fast big endian comparison operator, fast reverse iteration, fast "count numbers of bits in 1:k", fast "give me the set 1:k as subset of  1:64N"
8. The meaning of the type parameter `N` stays, and denotes the number of UInt64, i.e. `SBitSet{N}` describes subsets of `1:64N`. We provide no way of expressing "subset of 1:23", with the associated complement operator `~` (i.e. complements are relative to the world of `1:64N`). Such users need to have a mask `1:23` stored somewhere and then write `~x & mask`.

Here, (6) is possibly surprising to some users. Speed of hashing for SVectors of floats is irrelevant, but SBitSet are combinatorial objects for algorithms where `BitArray` or `BitSet` are too slow. I am using a lot of dictionaries where keys are SBitSet. I think this kind of use is more common than users comparing an SBitSet to a Set (and staticarrays is for users that prioritize speed over convenience, anyways). 

(8) may also be surprising and inconvenient for some users. If we wanted complements to work intuitively instead of leaking implementation details (64 bit word size), then we would need to make the actual size of the world part of the type. In my application, many sizes are computed at runtime and range to about 500. I don't want to compile 500 specializations, but compiling 8 specializations is completely fine. I can use my code regardless of what version gets merged here, but I think this consideration applies to more people than just me.

chethega:
Superseded by https://github.com/JuliaArrays/StaticArrays.jl/pull/552

PR #553, opened 2018-11-20
Wrapper for Matrix/Vector of SVector

chethega:
This is my attempt at a julian way of mixing Vector{SVector} and Matrix code. Especially important if one wants to mutate single entries, which is otherwise very annoying. 

Use:
```
julia> r=rand(3, 1000);
julia> s=SVecMat(r);
julia> s[:,678]
3-element SArray{Tuple{3},Float64,1,3}:
 0.274480584707264 
 0.8911180705798702
 0.7139000141428269
julia> s[1,678]=0.1;
julia> s[:,678]
3-element SArray{Tuple{3},Float64,1,3}:
 0.1               
 0.8911180705798702
 0.7139000141428269
```
There is no dirty memory punning involved. This constructor is obviously type-unstable, so one needs a function boundary or use the `Val{N}` constructor. The `getbit` and `setbit` for `UInt64` matrices was the original point of the implementation and combines well with the static bitsets. 

But I think that the API (encode `ncols` in the type, overload `getindex(a,::Colon, j)` and `setindex`) works pretty nice for a lot of `SVector` situations.

Needs tests, needs docs, may need some bugfixes. I think the boundscheck elision works.

What do you think about this kind of convenience wrapper? Offer something like this in `StaticArrays` or not?

### DIFF
--- a/src/SVecMat.jl
+++ b/src/SVecMat.jl
@@ -1,0 +1,62 @@
+struct SVecMat{N, T, Mtyp<:AbstractMatrix{T}} <: AbstractMatrix{T}
+    parent::Mtyp
+end
+
+Base.@propagate_inbounds function SVecMat(::Val{N}, mat) where N 
+    @boundscheck begin 
+        size(mat, 1) != N && throw(DimensionMismatch("Cannot construct N=$N dimensional SvecMat from s=$(size(mat,1)) columns"))
+    end
+    return SVecMat{N}(mat)
+end
+
+@inline SVecMat(mat::AbstractArray{T}) where T = SVecMat{size(mat,1), T, typeof(mat)}(mat)
+
+Base.@propagate_inbounds Base.getindex(M::SVecMat, idxs...) = getindex(M.parent, idxs...)
+Base.@propagate_inbounds Base.setindex!(M::SVecMat, v, idxs...) = setindex!(M.parent, v, idxs...)
+@inline Base.length(M::SVecMat) = length(M.parent)
+@inline Base.size(M::SVecMat{N}) where N = (N, size(M.parent, 2) )
+Base.IndexStyle(::SVecMat{N, T, Mtyp}) where {N,T,Mtyp} = IndexStyle(Mtyp)
+Base.IndexStyle(::Type{SVecMat{N, T, Mtyp}}) where {N,T,Mtyp} = IndexStyle(Mtyp)
+Base.eltype(M::SVecMat) = eltype(M.parent)
+
+
+@inline function Base.getindex(M::SVecMat{N}, ::Colon, j::Integer) where N
+    @boundscheck checkbounds(M, 1, j)
+    return SVector(ntuple(i-> (@inbounds M.parent[i,j]) , N))
+end
+
+@inline function Base.setindex!(M::SVecMat{N,T}, v::SVector{N,T}, ::Colon, j::Integer) where {N,T}
+    @boundscheck checkbounds(M, 1, j)
+    for i=1:N
+        @inbounds M[i,j] = v[i]
+    end
+    v
+end
+
+
+
+Base.@propagate_inbounds function getbit(B::SVecMat{N,UInt64}, i) where N
+    i1,i2 = Base.get_chunks_id(i)
+    return !iszero(B.parent[i1] & (1<<i2))
+end
+
+Base.@propagate_inbounds function getbit(B::SVecMat{N,UInt64}, i, j) where N
+    i1,i2 = Base.get_chunks_id(i)
+    return !iszero(B.parent[i1,j] & (1<<i2))
+end
+
+Base.@propagate_inbounds function setbit!(B::SVecMat{N,UInt64}, b::Bool, i) where N
+    i1, i2 = Base.get_chunks_id(i)
+    u = UInt64(1) << i2
+    c = B.parent[i1]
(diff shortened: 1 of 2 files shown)


andyferris:
Yes, we'd definitely like to offer arrays where some dimensions are statically sized and other dimensions are dynamically sized! In my opinion, the thing which unlocks this for arbitrary dimensions without ridiculous amounts of effort is the recent introduction of `SOneTo`.

Where I'd possibly go with this something like `SizedArray`. We should have a wrapper for an arbitrary `AbstractArray` (of any dimension) where some of the `axes` are overwritten by a `SOneTo`. Any operation using `axes` and `similar` should **already** then create `StaticArray` where possible (and we should of course override `similar` for "partially" static arrays to construct this container; and do the same thing for broadcasting).

(I really am not sure if we want to be returning `SArray`s or `MArray`s at that point - people seem to have different opinions about this. @ChrisRackauckas would the indexing behavior here, where a mutable array is indexed and returns an immutable array, be "a bad thing" for your kind of code? How could we describe such a thing in a generic abstract array trait system? Please note that I still feel that any "functional" operation on an array (mutable or otherwise) should not be presumed to return a mutable array... if you require a mutable result, the caller should *always* use `similar` and then the `!` version of the functional operation).

chethega:
Regarding the return type: This construction really makes no sense at all if we don't return an SVector:
```
julia> using StaticArrays
julia> a=rand(Float32, 4, 100);
julia> svm = SVecMat(a);
julia> f(x,args...)=@inbounds x[args...];
julia> @code_native f(svm, :, 17)
	.text
; Function f {
	movq	(%rsi), %rax
	addq	$-1, %rdx
	imulq	24(%rax), %rdx
	movq	(%rax), %rax
	vmovups	(%rax,%rdx,4), %xmm0
	vmovups	%xmm0, (%rdi)
	movq	%rdi, %rax
	retq
	nopl	(%rax)
;}
julia> svm2 = [svm[:,i] for i=1:size(svm,2)];
julia> @code_native f(svm2, 17)
	.text
; Function f {
	shlq	$4, %rdx
	movq	(%rsi), %rax
	vmovups	-16(%rax,%rdx), %xmm0
	vmovups	%xmm0, (%rdi)
	movq	%rdi, %rax
	retq
	nopw	%cs:(%rax,%rax)
;}

```
Apart from the extra indirection (that hopefully gets hoisted because `Matrix` cannot get resized or reallocated) and function call boiler-plate (that gets inlined away), we really emit only a single multiply and a single SIMD instruction (`vmovups (%rax,%rdx,4), %xmm0`). 

This means we are almost as good as a `Vector{SVector}` (plus the compiler knows that we cannot resize, allowing more loop hoisting; minus we lack the flexibility of resizing; minus, the getindex for `Vector{SVector}` knows the number of columns, so it could have used a `leaq` instead of `imulq`, so I should probably use linear indexing to access the parent). 

So should I close this and instead try to make this work for partially-static sized arrays? Or wait until more feedback arrives? Or go on and refine this (docs, tests, respect parent's `IndexStyle`, maybe look at other traits to get good interaction with views)?

c42f:
I think this is worth doing generically if it's worth doing at all (at least within the `StaticArrays` package), so my preference is to close this PR and pursue the option Andy has outlined above.

You could even potentially make this work within the existing `SizedArray` interface, having the size `Tuple` contain `StaticArrays.Dynamic()` in dynamically sized slots. I think this would make `SizedArray` far more useful than it currently is. But it would also mean that we can't have `SizedArray <: StaticArray` anymore which would be breaking. Nevertheless, we have already observed many situations where `MyArray` is statically sized for some type `MyArray`, but where we can't have `MyArray <: StaticArray`. The recent change to use `Diagonal` rather than `SDiagonal` being just one example.

c42f:
Actually on second thoughts we can still have `SizedArray <: StaticArray`, but we'd have to allow the `S` type parameter to have `StaticArrays.Dynamic()` inside it. I'm currently unsure about whether this will just break everything (ie, most assumptions of working with `StaticArray` subtypes) or not!

andyferris:
Right!

What I've been leaning towards is if array functions generally dispatched with their axes/keys (whatever they will iterate over) then we would be all good for these kinds of patterns without `StaticArray`. For example `map!(f, out, a, b...) = _map!(f, eachindex(out, a, b...), out, a, b...)`. Or ultimately `map!(f, out, a, b) = foreach((i -> @inbounds out[i] = f(a[i], b[i])), eachindex(out, a, b...))` if we never had to worry about closure performance issues. 

Speculatively, sparse arrays and distributed arrays and static arrays and so-on would have something in their keys which describes how they may be iterated in certain circumstances. If you have a new array type you just specialize on `keys` and `foreach` of those keys, and boom, you've got fast `map`, `reduce`, etc for free. (Indexing, broadcasting and even matrix multiplication are essentially interesting types of iteration problems).

Basically - what I'm saying is that maybe the *primary* trait of an array is how you iterate over its indices. Which is of course the whole `IndexStyle` thing but instead of worrying about `IndexStyle` directly this could just be encoded in `keys`.

CC @timholy @mbauman regarding my rampant speculation.

(The pattern could be replicated for `AbstractDict`... I always wanted to make a kick-ass statically-keyed dictionary).

c42f:
The underlying idea of this is very good, so I opened #573 as a place that we can continue discussing a more general version. In the interests of cleaning up I'll close this particular PR for now.

PR #555, opened 2018-11-29
Adds dispatch rules for column vector times transpose(row vector)

zygmuntszpak:
Fixes issue #537

### DIFF
--- a/src/matrix_multiply.jl
+++ b/src/matrix_multiply.jl
@@ -10,14 +10,16 @@ import LinearAlgebra: BlasFloat, matprod, mul!
 @inline *(A::StaticVector, B::StaticMatrix) = *(reshape(A, Size(Size(A)[1], 1)), B)
 @inline *(A::StaticVector, B::Transpose{<:Any, <:StaticVector}) = _mul(Size(A), Size(B), A, B)
 @inline *(A::StaticVector, B::Adjoint{<:Any, <:StaticVector}) = _mul(Size(A), Size(B), A, B)
+@inline *(A::StaticArray{Tuple{N,1},<:Any,2}, B::Adjoint{<:Any,<:StaticVector}) where {N} = vec(A) * B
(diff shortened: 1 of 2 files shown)


src/matrix_multiply.jl
```diff
@@ -10,16 +10,18 @@ import LinearAlgebra: BlasFloat, matprod, mul!
 @inline *(A::StaticVector, B::StaticMatrix) = *(reshape(A, Size(Size(A)[1], 1)), B)
 @inline *(A::StaticVector, B::Transpose{<:Any, <:StaticVector}) = _mul(Size(A), Size(B), A, B)
 @inline *(A::StaticVector, B::Adjoint{<:Any, <:StaticVector}) = _mul(Size(A), Size(B), A, B)
+@inline *(A::StaticMatrix, B::Adjoint{<:Any, <:StaticVector}) = *(reshape(A, Size(Size(A)[1],)), B)
```
andyferris: * Wouldn't `_mul(Size(A), Size(B), A, B)` work?
 * Need the same thing for `Transpose`.

Evizero: what happens if `size(A)` is `(4,2)`, does `reshape` fail then? is a `DimensionMismatch` thrown. I wonder if it wouldn't be cleaner to dispatch on the fact the dimensions 2 of `A` is 1 (similar to the example here) https://github.com/JuliaArrays/StaticArrays.jl/issues/537#issuecomment-442004593 . In that case a matrix `A` thats not a column vector would still go to the fallback method which would ultimately throw a `DimensionMismatch`.

zygmuntszpak: It does throw a `DimensionMismatch` but your solution is definitely neater. I have revised the pull-request as per your suggestion and have included a dispatch rule for `transpose` as well.

Evizero:
I don't know if it is covered by other tests already, but it would be nice to test explicitly if a `DimensionMismatch` is thrown in case the matrix is not a column vector

zygmuntszpak:
> I don't know if it is covered by other tests already, but it would be nice to test explicitly if a `DimensionMismatch` is thrown in case the matrix is not a column vector

Done.

PR #563, opened 2018-12-08
Fix setindex with vector indices

### LINKED ISSUE #562: setindex with arrays of indices
Is there any reason why `inds` is being ignored in the two specializations of `_setindex!` here, but not in the more generic version?

https://github.com/JuliaArrays/StaticArrays.jl/blob/master/src/indexing.jl#L151-L179

You'll get behaviour like

    julia> a = MVector{4}([0.0, 0.0, 0.0, 0.0])
    julia> b = MVector{2}([1, 4])
    julia> a[b] = [1.0, 2.0]
    julia> a
    4-element MArray{Tuple{4},Float64,1,4}:
     1.0
     2.0
     0.0
     0.0

Expected: `[1.0, 0.0, 0.0, 2.0]`

TheBB:
Fixes #562.

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -157,7 +157,7 @@ end
 end
 
 @generated function _setindex!(a::StaticArray, v::AbstractArray, ::Length{L}, inds::StaticVector{<:Any, Int}) where {L}
-    exprs = [:(a[$i] = v[$i]) for i = 1:L]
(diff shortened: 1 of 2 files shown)


andyferris:
Haha, oops.

andyferris:
Thanks for fixing this.

PR #564, opened 2018-12-08
Use `propagate_inbounds` to ensure bounds check elision

YingboMa:


andyferris:
I think the current `@inline`, `@boundscheck` and `@inbounds` pattern does this? `@propagate_inbounds` is for nesting these calls, not the "final" one.

vchuravy:
`@propagate_inbounds` is the stronger statement. And `@inbounds` seems to only have this effect one layer into the callchain:

```
Base.@propagate_inbounds function g1()
           @boundscheck begin
              error("Boundscheck")
           end
           return 1
       end
@inline function g2()
           @boundscheck begin
              error("Boundscheck")
           end
           return 1
       end
Base.@propagate_inbounds f1() = g1()
@inline f2() = g2()
h1() = @inbounds f1()
h2() = @inbounds f2()
```


Notice the difference in `goto if not false` and `goto if not true`
```
julia> @code_typed h1()
CodeInfo(
1 1 ─      goto #3 if not false                                                                                 │╻╷ f1
  2 ─      invoke Main.error("Boundscheck"::String)::Union{}                                                    ││┃  g1
  └──      $(Expr(:unreachable))::Union{}                                                                       │││
  3 ─      goto #4                                                                                              │││
  4 ─      goto #5                                                                                              ││ 
  5 ─ %6 = π (1, Const(1, false))                                                                               │  
  └──      return %6                                                                                            │  
) => Int64

julia> @code_typed h2()
CodeInfo(
1 1 ─      goto #3 if not true                                                                                  │╻╷ f2
  2 ─      invoke Main.error("Boundscheck"::String)::Union{}                                                    ││┃  g2
  └──      $(Expr(:unreachable))::Union{}                                                                       │││
  3 ─      goto #4                                                                                              │││
  4 ─      goto #5                                                                                              ││ 
  5 ─ %6 = π (1, Const(1, false))                                                                               │  
  └──      return %6                                                                                            │  
) => Int64
```

@YingboMa I think the actual problem we saw was from that the callchain of https://github.com/JuliaArrays/StaticArrays.jl/blob/3ccbd4159ee9ecbd86c3ab6d896d64ff47a8020f/src/convert.jl#L6 not propagating the inbounds statement down to the convert

andyferris:
Right - we can put it there.

tkoolen:
Seems uncontroversial. Thanks for the PR!

PR #565, opened 2018-12-10
Introduce StaticArrayLike

### LINKED ISSUE #561: SDiagonal {h,v,hv}cat return type is SparseMatrix when SDiagonal is first argument
With StaticArrays 0.9.2:

```julia
julia> using StaticArrays

julia> [SDiagonal(1, 2) SVector(1, 2)]
2×3 SArray{Tuple{2,3},Int64,2,6}:
 1  0  1
 0  2  2
```

With StaticArrays 0.10.0:

```julia
julia> using StaticArrays

julia> [SDiagonal(1, 2) SVector(1, 2)]
2×3 SparseArrays.SparseMatrixCSC{Int64,Int64} with 4 stored entries:
  [1, 1]  =  1
  [2, 2]  =  2
  [1, 3]  =  1
  [2, 3]  =  2
```

and also

```julia
julia> [SVector(1, 2) SDiagonal(1, 2)]
2×3 Array{Int64,2}:
 1  1  0
 2  0  2
```

This is because it's just falling back to the Base implementation, which uses `similar` on the first argument to determine the output type (also a somewhat questionable choice, but that's another issue).

tkoolen:
Now that we have wrapper types in Base that can be used in conjunction
with StaticArrays, we should discern between actual `StaticArray`s and
non-`StaticArray` subtypes for which we still know a static `Size`. To
this effect, this PR adds the type aliases `StaticallySizedMatrix`,
`StaticallySizedVecOrMat`, and `StaticallySizedArray`, and uses them to
widen various type signatures.

Fixes #561.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -78,6 +78,19 @@ const StaticVector{N, T} = StaticArray{Tuple{N}, T, 1}
 const StaticMatrix{N, M, T} = StaticArray{Tuple{N, M}, T, 2}
 const StaticVecOrMat{T} = Union{StaticVector{<:Any, T}, StaticMatrix{<:Any, <:Any, T}}
 
+# Being a member of StaticallySizedMatrix, StaticallySizedVecOrMat, or StaticallySizedArray implies that Size(A)
+# returns a static Size instance. The converse may not be true.
+const StaticallySizedMatrix{T} = Union{
+    StaticMatrix{<:Any, <:Any, T},
+    Transpose{T, <:StaticVecOrMat{T}},
+    Adjoint{T, <:StaticVecOrMat{T}},
+    Symmetric{T, <:StaticMatrix{T}},
+    Hermitian{T, <:StaticMatrix{T}},
+    Diagonal{T, <:StaticVector{<:Any, T}}
+}
+const StaticallySizedVecOrMat{T} = Union{StaticVector{<:Any, T}, StaticallySizedMatrix{T}}
+const StaticallySizedArray{T} = Union{StaticallySizedVecOrMat{T}, StaticArray{<:Any, T}}
+
 const AbstractScalar{T} = AbstractArray{T, 0} # not exported, but useful none-the-less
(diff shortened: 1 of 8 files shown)


src/StaticArrays.jl
```diff
@@ -78,6 +78,19 @@ const StaticVector{N, T} = StaticArray{Tuple{N}, T, 1}
 const StaticMatrix{N, M, T} = StaticArray{Tuple{N, M}, T, 2}
 const StaticVecOrMat{T} = Union{StaticVector{<:Any, T}, StaticMatrix{<:Any, <:Any, T}}
 
+# Being a member of StaticallySizedMatrix, StaticallySizedVecOrMat, or StaticallySizedArray implies that Size(A)
+# returns a static Size instance. The converse may not be true.
+const StaticallySizedMatrix{T} = Union{
```
c42f: There's an interesting (possibly less confusing) alternative name `StaticArrayLike` (/ `StaticMatrixLike`) at https://github.com/JuliaArrays/StaticArrays.jl/issues/537#issuecomment-442036409

andyferris:
Well, I guess this may be an improvement on the status quo but I have to admit that I do find it a bit unsatisfying - I find things like `StridedArray` to be a bit of an anti-pattern.

 >  like `Symmetric{Diagonal{X}}`

Does this ever get constructed? It seems a little unnecessary.

tkoolen:
> I find things like StridedArray to be a bit of an anti-pattern.

Fully agree. But I think it's just necessary given the current approach in Base.

> Does this ever get constructed? It seems a little unnecessary.

No, no. Just a made-up example.

c42f:
> Well, if it is helpful, I guess we should do it

My thoughts exactly.

tkoolen:
Did the rename and added a comment referring to `StridedArray`.

timholy:
Dang this is just crying out for a new version of [Traitor.jl](https://github.com/andyferris/Traitor.jl). Maybe a Christmas-break project...

PR #567, opened 2018-12-12
Add reverse(::SVector), with tests.

### LINKED ISSUE #566: reverse(::SVector)
Currently it uses the generic fallback in `Base`, so
```julia
julia> using StaticArrays

julia> reverse(SVector(1, 2))
2-element MArray{Tuple{2},Int64,1,2}:
 2
 1
```

Would getting an `SVector` be feasible?

tpapp:
Fixes #566.

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -161,7 +161,7 @@ reshape(a::Array, s::Size{S}) where {S} = s(a)
 
 @inline copy(a::StaticArray) = typeof(a)(Tuple(a))
 
-@inline reverse(v::SVector) = SVector(reverse(Tuple(v)))
+@inline reverse(v::StaticVector) = typeof(v)(reverse(Tuple(v)))
 
 # TODO permutedims?
 
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -161,6 +161,8 @@ reshape(a::Array, s::Size{S}) where {S} = s(a)
 
 @inline copy(a::StaticArray) = typeof(a)(Tuple(a))
 
+@inline reverse(v::SVector) = SVector(reverse(Tuple(v)))
+
 # TODO permutedims?
 
 # full deprecated in Base
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -6,7 +6,7 @@ import Base: getindex, setindex!, size, similar, vec, show, length, convert, pro
              promote_rule, map, map!, reduce, mapreduce, broadcast,
              broadcast!, conj, hcat, vcat, ones, zeros, one, reshape, fill, fill!, inv,
              iszero, sum, prod, count, any, all, minimum, maximum, extrema,
-             copy, read, read!, write
+             copy, read, read!, write, reverse
 
 import Statistics: mean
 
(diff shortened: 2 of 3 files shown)


mschauer:
Up to 30 and a bit.

src/abstractarray.jl
```diff
@@ -161,6 +161,8 @@ reshape(a::Array, s::Size{S}) where {S} = s(a)
 
 @inline copy(a::StaticArray) = typeof(a)(Tuple(a))
 
+@inline reverse(v::SVector) = SVector(reverse(Tuple(v)))
```
andyferris: Ideally. this should be generic with respect to type and dimensionality.

Maybe like `@inline reverse(a::StaticArray) = typeof(a)(reverse(Tuple(a)))` (or else via `similar_type`).

tpapp: Not sure why you want this extended to all `StaticArrays` with dimension above 1, since `reverse` does not work (without keyword arguments) for matrices etc. Eg
```julia
julia> reverse(ones(2, 2))
ERROR: UndefKeywordError: keyword argument dims not assigned
Stacktrace:
 [1] reverse(::Array{Float64,2}) at ./arraymath.jl:62
 [2] top-level scope at none:0
```

andyferris: Sometimes `Base` gets it wrong? ¯\_(ツ)_/¯

`@inline reverse(a::StativVector) = typeof(a)(reverse(Tuple(a)))` is OK if you feel that is safer.

andyferris: `v = MVector(1, 2, 3)`. The result should be an `MVector` rather than `SVector`.

tpapp: But `reverse(::MVector)` works fine already. I am just adding a method `reverse(::SVector)`.

tpapp:
Unless there are further suggestions, can you please merge this?

src/abstractarray.jl
tkoolen: Consider also other non-`SVector` immutable `StaticVector`s:

```julia
julia> struct Point3D <: FieldVector{3, Float64}
           x::Float64
           y::Float64
           z::Float64
       end

julia> reverse(Point3D(1, 2, 3))
3-element MArray{Tuple{3},Float64,1,3}:
 3.0
 2.0
 1.0
```

Also, widening to `StaticVector` improves performance for `MVector`:

```julia
julia> @btime reverse(x) setup = x = MVector(1, 2, 3)
  9.667 ns (1 allocation: 32 bytes)

julia> @inline Base.reverse(v::StaticVector) = typeof(v)(reverse(Tuple(v)))

julia> @btime reverse(x) setup = x = MVector(1, 2, 3);
  5.777 ns (1 allocation: 32 bytes)
```

Lastly, for consistency, the only other place where `SVector` is special-cased like this is

https://github.com/JuliaArrays/StaticArrays.jl/blob/ae146942c54bac73859dca828278b356130e3f8f/src/SDiagonal.jl#L7

and I consider that to be an oversight.

tpapp: All those things would be nice I guess, but this is a PR for fixing #566. I don't understand why you are asking me to do other things outside the scope of this PR in a code review.

tkoolen:
Thanks!

src/abstractarray.jl
c42f: To understand, you need to think like a maintainer for a moment:

Every line of code is a tradeoff between
* The functionality it brings (valuable and visible to everyone)
* The maintenance burden it introduces (largely invisible to users, except that difficult to maintain packages have more bugs)

When you introduce code which only works in one specific case, this tradeoff doesn't look so good to from the maintainers point of view, and it might be rejected. This depends a lot on how bad the problem is and how hard it is to fix in general. (In this case, not so bad, and very easy to fix, respectively.)

PR #569, opened 2018-12-14
Alternate approach to fix #556

### LINKED ISSUE #556: Slicing with colon and range fails

mbauman:
This restructures the similar method table along the lines that I suggested in https://github.com/JuliaArrays/StaticArrays.jl/pull/568#discussion_r241561280. I also used a bit of internals magic to match how the Base `OneTo` operates -- this is a little fragile but I think it is a good way to replicate how `OneTo` works.

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -107,6 +107,7 @@ similar(::Type{SA}) where {SA<:StaticArray} = similar(SA,eltype(SA))
 similar(::SA,::Type{T}) where {SA<:StaticArray,T} = similar(SA,T,Size(SA))
 similar(::Type{SA},::Type{T}) where {SA<:StaticArray,T} = similar(SA,T,Size(SA))
 
+# Cases where a Size is given as the dimensions
 similar(::A,s::Size{S}) where {A<:AbstractArray,S} = similar(A,eltype(A),s)
 similar(::Type{A},s::Size{S}) where {A<:AbstractArray,S} = similar(A,eltype(A),s)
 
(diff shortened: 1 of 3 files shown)


src/abstractarray.jl
```diff
@@ -119,22 +120,19 @@ similar(::Type{A},::Type{T},s::Size{S}) where {A<:AbstractArray,T,S} = mutable_s
...
-# We should be able to deal with SOneTo axes
-similar(::A, shape::Tuple{SOneTo, Vararg{SOneTo}}) where {A<:AbstractArray} = similar(A, eltype(A), shape)
-similar(::Type{A}, shape::Tuple{SOneTo, Vararg{SOneTo}}) where {A<:AbstractArray} = similar(A, eltype(A), shape)
-
-similar(::A,::Type{T}, shape::Tuple{SOneTo, Vararg{SOneTo}}) where {A<:AbstractArray,T} = similar(A, T, Size(last.(shape)))
-similar(::Type{A},::Type{T}, shape::Tuple{SOneTo, Vararg{SOneTo}}) where {A<:AbstractArray,T} = similar(A, T, Size(last.(shape)))
+# Support tuples of mixtures of `SOneTo`s alongside the normal `Integer` and `OneTo` options
+# by simply converting them to either a tuple of Ints or a Size, re-dispatching to either one
+# of the above methods (in the case of Size) or a base fallback (in the case of Ints).
+const HeterogeneousShape = Union{Integer, Base.OneTo, SOneTo}
```
pablosanjose: Are you sure you need Base.OneTo here? Base converts OneTo to Ints. Also, you can get IdentityUnitRange{<:SOneTo} here, shouldn't it be included in HeterogeneousShape? (This is what I did in #568)

src/abstractarray.jl
```diff
@@ -119,22 +120,19 @@ similar(::Type{A},::Type{T},s::Size{S}) where {A<:AbstractArray,T,S} = mutable_s
...
+similar(::Type{A}, shape::Tuple{HeterogeneousShape, Vararg{HeterogeneousShape}}) where {A<:AbstractArray} = similar(A, homogenize_shape(shape))
+# Use an Array for StaticArrays if we don't have a statically-known size
+similar(::Type{A}, shape::Tuple{Int, Vararg{Int}}) where {A<:StaticArray} = Array{eltype(A)}(undef, shape)
 
-# Handle mixtures of SOneTo and other ranges (probably should make Base more robust here)
-similar(::Type{A}, shape::Tuple{AbstractUnitRange, Vararg{AbstractUnitRange}}) where {A<:AbstractArray} = similar(A, length.(shape)) # Jumps back to 2-argument form in Base
-similar(::Type{A},::Type{T}, shape::Tuple{AbstractUnitRange, Vararg{AbstractUnitRange}}) where {A<:AbstractArray,T} = similar(A, length.(shape))
+homogenize_shape(::Tuple{}) = ()
+homogenize_shape(shape::Tuple{Vararg{SOneTo}}) = Size(map(last, shape))
+homogenize_shape(shape::Tuple{Vararg{HeterogeneousShape}}) = map(last, shape)
```
pablosanjose: Nice. I now see that we should *only* do the `static_similar` when *all* elements of `shape` are statically known (i.e. `s[:, 1:2]` should produce a `Matrix` not an `MMatrix`). However the static cases include `SOneTo` and also `IdentityUnitRange{<:SOneTo}`, that gets produced in base when you do `s[:,1:2]`, right?

src/abstractarray.jl
```diff
@@ -119,22 +120,19 @@ similar(::Type{A},::Type{T},s::Size{S}) where {A<:AbstractArray,T,S} = mutable_s
...
 
-const SOneToLike{n} = Union{SOneTo{n}, Base.Slice{SOneTo{n}}}
-deslice(ax::SOneTo) = ax
-deslice(ax::Base.Slice) = ax.indices
-similar(::A,::Type{T}, shape::Tuple{SOneToLike, Vararg{SOneToLike}}) where {A<:AbstractArray,T} = similar(A, T, Size(last.(deslice.(shape))))
-similar(::Type{A},::Type{T}, shape::Tuple{SOneToLike, Vararg{SOneToLike}}) where {A<:AbstractArray,T} = similar(A, T, Size(last.(deslice.(shape))))
+similar(A::AbstractArray, ::Type{T}, shape::Tuple{HeterogeneousShape, Vararg{HeterogeneousShape}}) where {T} = similar(A, T, homogenize_shape(shape))
+similar(::Type{A}, shape::Tuple{HeterogeneousShape, Vararg{HeterogeneousShape}}) where {A<:AbstractArray} = similar(A, homogenize_shape(shape))
+# Use an Array for StaticArrays if we don't have a statically-known size
+similar(::Type{A}, shape::Tuple{Int, Vararg{Int}}) where {A<:StaticArray} = Array{eltype(A)}(undef, shape)
```
pablosanjose: Is this one necessary? We have`similar(a::AbstractArray, ::Type{T}, dims::Dims{N}) where {T,N}    = Array{T,N}(undef, dims)` in base

src/abstractarray.jl
pablosanjose: Aaah,  I see what you've done here. You've overridden the `similar(a::AbstractArray{T}, dims::Tuple)` in base by including `Base.OneTo`. Sorry, carry on!

src/abstractarray.jl
pablosanjose: As above, your definition of `HeterogenousShape` skips the creation of `IdentityUnitRange`, please ignore this

src/abstractarray.jl
pablosanjose: (I can confirm that removing this definition doesn't make any test fail locally under v1.1)

mbauman: Ah, yes, I hit a failure at the REPL at one point with a higher-level function, but I think it was due to an intermediate state of the other methods.

I'm not sure if there's any examples where a higher-level function would fail without this, but Base supports a `similar(T, (2,3))` API where `T` is the type of an AbstractArray.  It falls back to just calling `T(undef, (2, 3))`, which will of course fail for a `StaticArray`.  I can add a specific test for this line.

pablosanjose:
I absolutely agree with the above. I realised this after doing some benchmarking of those kind of cases. The `MArray` version resulted in  microsecond runtimes (dynamic dispatch, I assume), while returning an `Array` was 30-60 ns, the typical times for such things. So I think your solution is the correct one. Thanks!

JeffBezanson:
Bump. Can this be merged?

pablosanjose:
+1 to merge

c42f:
Sorry for the delay, this is one of those odd occasions where the author of the PR is more qualified to comment on its validity than most of the maintainers. Or certainly than myself ;-)

From what I can see quickly this looks completely reasonable, I'll fix the conflict, rerun CI and merge if it passes. I assume people will also want a new release including this fix.

PR #570, opened 2018-12-18
implement solve with mixed StaticArray/DynamicArray types

timholy:
Because of [this depot function](https://github.com/JuliaLang/julia/blob/eddcbd7585152c1e404ed0481ffe44c61d504e77/stdlib/LinearAlgebra/src/generic.jl#L891-L908) we need to handle a variety of different cases for `\`.

This doesn't work yet for a nominally-unrelated indexing reason: the `B[F.p, :]` fails. Out of time for tonight.


c42f:
I've restarted CI here as the problem of #572 is sorted out, I think. Let's see what happens.

timholy:
This one should pass. Presumably the `Diagonal` case I added should return a dynamic array?

PR #571, opened 2018-12-19
Clean up indexing and dispatch

timholy:
Several aspects of indexing were not ideal:

- `getindex(a::StaticArray, inds::Int...)` violates the [AbstractArray interface](https://docs.julialang.org/en/latest/manual/interfaces/#man-interface-array-1)
- non-scalar static indexing of `AbstractArrays` only worked up to 4 dimensions
- there were a lot of `getindex` and `setindex!` ambiguities

The first commit here fixes all of these problems while also reducing `detect_ambiguities(Base, StaticArrays)` from 35 pairs to 9. Serious progress on #18 :fireworks: .

The second is boring but addresses https://github.com/JuliaLang/julia/issues/6383#issuecomment-448642771. A `sed` script to fix this is

```sh
sed -i "s/StaticArray{<:Any/StaticArray{<:Tuple/g" *.jl
```

if this ever creeps back in.

### DIFF
--- a/src/MMatrix.jl
+++ b/src/MMatrix.jl
@@ -57,7 +57,7 @@ end
     end
 end
 
-@inline convert(::Type{MMatrix{S1,S2}}, a::StaticArray{<:Any, T}) where {S1,S2,T} = MMatrix{S1,S2,T}(Tuple(a))
+@inline convert(::Type{MMatrix{S1,S2}}, a::StaticArray{<:Tuple, T}) where {S1,S2,T} = MMatrix{S1,S2,T}(Tuple(a))
 @inline MMatrix(a::StaticMatrix) = MMatrix{size(typeof(a),1),size(typeof(a),2)}(Tuple(a))
 
 # Simplified show for the type
--- a/src/SMatrix.jl
+++ b/src/SMatrix.jl
@@ -52,7 +52,7 @@ end
     end
 end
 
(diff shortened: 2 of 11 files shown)


timholy:
CC @mbauman, if he has time to consult about whether my corruption of `to_indices` is just too evil to be tolerated.

timholy:
Hmm, looks like we never tested `v[1,1]` when `v` is an `SVector`, and the above broke that. Given the vector methods, it seems easiest to go back to the old `getindex(a, i::Int...)`. Bummer. The 3rd commit implements this "reversion" and also tests the problematic case.

Also, it looks like the 2nd commit got rid of a couple more ambiguities, so now we're down to 7. The third commit fortunately did not introduce new ambiguities.

c42f:
I've rerun CI to check whether this interacts with #569. Looks like it's going to be fine across the board, so I'm inclined just to merge this and get the release out.

timholy:
TBH I fear the indexing tests are a bit on the thin side, so messing with the indexing rules is a bit scarier than I initially realized. Especially since there are close to 150 packages in JuliaRegistries/General that depend on this one.

So, I locally merged this and #572. I'm running tests on all my installed packages that depend on StaticArrays (only 14 packages, though). I'll let you know how it goes.

timholy:
OK, I think this is fine.

c42f:
Great, I think I'll tag and release then.

c42f:
Ok, I see you've already done so, thanks!

PR #572, opened 2018-12-19
Fix mixed indexing like `a[v, :]` where `v::SVector` and `a` is dynamic

timholy:
Ref #570


c42f:
There's a conflict now, but I think this problem was fixed by the restructuring @mbauman did in #569.

We could keep the tests though?

timholy:
Superseded by #569, though we may discuss changing it to this implementation. In the meantime let's not leave this open.

PR #581, opened 2019-01-05
Allow empty indexing of zero-dimensional MArrays

### LINKED ISSUE #580: Indexing of zero-dimensional MArrays
Zero-dimensional mutable arrays must implement an "empty" indexing method to work with some other functions, e.g. `show`:

    julia> v = MArray{Tuple{},Int,0,1}(1)
    0-dimensional MArray{Tuple{},Int64,0,1}:
    Error showing value of type MArray{Tuple{},Int64,0,1}:
    ERROR: ArgumentError: invalid index: () of type Tuple{}

This works for `SArrays` because they are specialized as "scalars", and this method is defined for them, but `MArrays` aren't.

TheBB:
This is one way to fix #580.

### DIFF
--- a/src/Scalar.jl
+++ b/src/Scalar.jl
@@ -14,7 +14,6 @@ const Scalar{T} = SArray{Tuple{},T,0,1}
 end
 @inline convert(::Type{SA}, sa::SA) where {SA <: Scalar} = sa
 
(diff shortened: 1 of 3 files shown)


c42f:
Thanks, this looks reasonable to me.

PR #584, opened 2019-01-13
Increase Performance of \ (and /) for small square matrix

judober:
Concerning [this issue](https://github.com/JuliaArrays/StaticArrays.jl/issues/582#issue-398546390).
My proposal for increased performance of A/B if A is 2x2 or 3x3 and B is 2or3xn.
There is an [ongoing discussion](https://discourse.julialang.org/t/performance-of-matrix-division/19477?u=judober) about this.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -28,15 +28,15 @@ end
 
 for Sa in 2:3  # not needed for Sa = 1;
     @eval begin
-    @inline function solve(::Size{($Sa,$Sa)}, ::Size{Sb}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticMatrix{<:Any, <:Any, Tb}) where {Sb, Ta, Tb}
-        d = det(a)
-        T = typeof((one(Ta)*zero(Tb) + one(Ta)*zero(Tb))/d)
-        c = similar(b, T)
-        for col = 1:Sb[2]
-            @inbounds c[:, col] = solve(Size($Sa,$Sa), Size($Sa,), a, b[:, col])
+        @inline function solve(::Size{($Sa,$Sa)}, ::Size{Sb}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticMatrix{<:Any, <:Any, Tb}) where {Sb, Ta, Tb}
+            d = det(a)
+            T = typeof((one(Ta)*zero(Tb) + one(Ta)*zero(Tb))/d)
+            c = similar(b, T)
+            for col = 1:Sb[2]
+                @inbounds c[:, col] = solve(Size($Sa,$Sa), Size($Sa,), a, b[:, col])
+            end
+            return similar_type(b, T)(c)
         end
-        return similar_type(b, T)(c)
-    end
     end # @eval
 end
 
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -26,14 +26,14 @@ end
             (a[1,1]*a[2,2] - a[1,2]*a[2,1])*b[3]) / d )
 end
 
-for Sa in 2:3  # not needed for Sa = 1;
+for Sa in [(2,2), (3,3)]  # not needed for Sa = 1;
     @eval begin
-        @inline function solve(::Size{($Sa,$Sa)}, ::Size{Sb}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticMatrix{<:Any, <:Any, Tb}) where {Sb, Ta, Tb}
+        @inline function solve(::Size{$Sa}, ::Size{Sb}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticMatrix{<:Any, <:Any, Tb}) where {Sb, Ta, Tb}
             d = det(a)
             T = typeof((one(Ta)*zero(Tb) + one(Ta)*zero(Tb))/d)
             c = similar(b, T)
             for col = 1:Sb[2]
-                @inbounds c[:, col] = solve(Size($Sa,$Sa), Size($Sa,), a, b[:, col])
+                @inbounds c[:, col] = solve(Size($Sa), Size($Sa[1],), a, b[:, col])
             end
             return similar_type(b, T)(c)
         end
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -26,6 +26,22 @@ end
             (a[1,1]*a[2,2] - a[1,2]*a[2,1])*b[3]) / d )
 end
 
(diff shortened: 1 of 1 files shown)


judober:
Speed comparison (other machine than on discourse).
Original version:
```
julia> a = Float64.(@SMatrix rand(3,3)); b = Float64.(@SMatrix rand(3,3));

julia> @benchmark a\b
BenchmarkTools.Trial: 
  memory estimate:  80 bytes
  allocs estimate:  1
  --------------
  minimum time:     149.989 ns (0.00% GC)
  median time:      153.293 ns (0.00% GC)
  mean time:        184.259 ns (7.69% GC)
  maximum time:     75.550 μs (99.69% GC)
  --------------
  samples:          10000
  evals/sample:     818
```

This PR (with the same matrices):
```
julia> @benchmark a\b
BenchmarkTools.Trial: 
  memory estimate:  80 bytes
  allocs estimate:  1
  --------------
  minimum time:     66.859 ns (0.00% GC)
  median time:      70.639 ns (0.00% GC)
  mean time:        98.225 ns (16.09% GC)
  maximum time:     73.637 μs (99.82% GC)
  --------------
  samples:          10000
  evals/sample:     978
```

tkoolen:
Could you run the following benchmark (results below are on my machine with StaticArrays 0.10.2):

```julia

julia> @benchmark a \ b setup = begin
           a = rand(SMatrix{3, 3})
           b = rand(SMatrix{3, 3})
       end
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     71.394 ns (0.00% GC)
  median time:      71.689 ns (0.00% GC)
  mean time:        74.010 ns (0.00% GC)
  maximum time:     113.456 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     974
```

andyferris:
Seems like a reasonable approach to me.

judober:
New test with old version:
```

julia> @benchmark a \ b setup = begin
                  a = rand(SMatrix{3, 3})
                  b = rand(SMatrix{3, 3})
              end
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     104.663 ns (0.00% GC)
  median time:      106.303 ns (0.00% GC)
  mean time:        118.250 ns (0.00% GC)
  maximum time:     1.198 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     937
```

With PR:
```

julia> @benchmark a \ b setup = begin
                  a = rand(SMatrix{3, 3})
                  b = rand(SMatrix{3, 3})
              end
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     30.143 ns (0.00% GC)
  median time:      32.320 ns (0.00% GC)
  mean time:        35.998 ns (0.00% GC)
  maximum time:     1.324 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     994
```

src/solve.jl
```diff
@@ -26,6 +26,22 @@ end
             (a[1,1]*a[2,2] - a[1,2]*a[2,1])*b[3]) / d )
 end
 
+for Sa in 2:3  # not needed for Sa = 1;
```
tkoolen: In the rest of the code `Sa` is typically a `Tuple`. For consistency and ease of reading, could you change this to `for Sa in [(2, 2), (3, 3)]` and update the rest of the code accordingly?

src/solve.jl
```diff
@@ -26,6 +26,22 @@ end
...
+    @eval begin
+    @inline function solve(::Size{($Sa,$Sa)}, ::Size{Sb}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticMatrix{<:Any, <:Any, Tb}) where {Sb, Ta, Tb}
```
tkoolen: Following Base, I think it's better if the code inside `@eval` is indented.

src/solve.jl
```diff
@@ -26,6 +26,22 @@ end
...
+        d = det(a)
+        T = typeof((one(Ta)*zero(Tb) + one(Ta)*zero(Tb))/d)
+        c = similar(b, T)
+        for col = 1:Sb[2]
```
tkoolen: It's possible that using `Base.OneTo(Sb[2])` instead of `1:Sb[2]` will be faster. Could you check?

tkoolen:
Do you think the repeated computation of the determinant of `A` for each of the columns inside the inner `solve` calls is being optimized away? How long does `det(A)` take on your machine?

judober:
Ok, first the performance of det(A):
```
julia> using LinearAlgebra

julia> @benchmark det(a) setup = begin
                  a = rand(SMatrix{3, 3})
              end
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     17.886 ns (0.00% GC)
  median time:      17.888 ns (0.00% GC)
  mean time:        20.050 ns (0.00% GC)
  maximum time:     169.474 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
```
You might be right that this is a point for further improvements. I will try it.

judober:
The results when adding OneTo:
```
julia> @benchmark a \ b setup = begin
                  a = rand(SMatrix{3, 3})
                  b = rand(SMatrix{3, 3})
              end
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     30.143 ns (0.00% GC)
  median time:      30.303 ns (0.00% GC)
  mean time:        34.886 ns (0.00% GC)
  maximum time:     460.771 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     994
```
seems pretty identical.

src/solve.jl
judober: Addressed in the latest version.

src/solve.jl
judober: Also addressed in the latest version.

src/solve.jl
judober: No difference. See last comment.

tkoolen: OK, thanks for trying.

tkoolen:
> Though, when I compare the numbers, it must be optimised away, otherwise the four calculations of det(a) would need alone approx. 60 ns.

Yeah, I agree. That's cool!

No further comments from me, @andyferris, good to go?

c42f:
I think this is fine, let's merge it.

PR #586, opened 2019-01-18
Allow using static arrays as index for views

tribut:
The PR allows using static arras as index when using views, such as

~~~~julia
a = [10,20,30,40]
idx = SVector(2,4)
b = view(a, idx)
~~~~


andyferris:
Thanks - and sorry for the two month delay, @tribut!

PR #594, opened 2019-02-12
Improve inferability of resursive `map` on nested `StaticArray`s

### LINKED ISSUE #593: Nested SVectors not correctly inferred?
It seems that nesting SVectors does not type infer correctly:
```jl
julia> using StaticArrays

julia> v = SVector(SVector(3, 2), SVector(5, 7))
2-element SArray{Tuple{2},SArray{Tuple{2},Int64,1,2},1,2}:
 [3, 2]
 [5, 7]

julia> v + v
2-element SArray{Tuple{2},SArray{Tuple{2},Int64,1,2},1,2}:
 [6, 4]
 [10, 14]

julia> @code_warntype v + v
Body::SArray{Tuple{2},_1,1,2} where _1
1 ─ %1 = StaticArrays.:+::Core.Compiler.Const(+, false)
│   %2 = invoke StaticArrays._map(%1::typeof(+), $(QuoteNode(Size(2,)))::Size{(2,)}, _2::SArray{Tuple{2},SArray{Tuple{2},Int64,1,2},1,2}, _3::SArray{Tuple{2},SArray{Tuple{2},Int64,1,2},1,2})::SArray{Tuple{2},_1,1,2} where _1
└──      return %2

julia> using BenchmarkTools
@btime
julia> @btime $v + $v
  8.706 ns (1 allocation: 48 bytes)
2-element SArray{Tuple{2},SArray{Tuple{2},Int64,1,2},1,2}:
 [6, 4]
 [10, 14]
```
(Note the allocation here.)

martinholters:
Fixes #593. Note that run-time behavior should be unaffected (in cases where inference is unaffected) as the `same_size` call is merely moved.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -8,16 +8,21 @@
 # `map(f, as::Union{<:StaticArray,AbstractArray}...)` which included at least one `StaticArray`
 # this is not the case on 0.7 and we instead hope to find a StaticArray in the first two arguments.
 @inline function map(f, a1::StaticArray, as::AbstractArray...)
-    _map(f, same_size(a1, as...), a1, as...)
+    _map(f, a1, as...)
 end
 @inline function map(f, a1::AbstractArray, a2::StaticArray, as::AbstractArray...)
-    _map(f, same_size(a1, a2, as...), a1, a2, as...)
+    _map(f, a1, a2, as...)
 end
 @inline function map(f, a1::StaticArray, a2::StaticArray, as::AbstractArray...)
-    _map(f, same_size(a1, a2, as...), a1, a2, as...)
+    _map(f, a1, a2, as...)
 end
 
-@generated function _map(f, ::Size{S}, a::AbstractArray...) where {S}
+@generated function _map(f, a::AbstractArray...)
+    i = findfirst(ai -> ai <: StaticArray, a)
+    if i === nothing
+        return :(throw(ArgumentError("No StaticArray found in argument list")))
+    end
+    S = Size(a[i])
     exprs = Vector{Expr}(undef, prod(S))
     for i ∈ 1:prod(S)
         tmp = [:(a[$j][$i]) for j ∈ 1:length(a)]
@@ -26,8 +31,9 @@ end
 
     return quote
         @_inline_meta
+        S = same_size(a...)
         @inbounds elements = tuple($(exprs...))
-        @inbounds return similar_type(typeof(_first(a...)), eltype(elements), Size(S))(elements)
+        @inbounds return similar_type(typeof(_first(a...)), eltype(elements), S)(elements)
     end
 end
 
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -22,6 +22,12 @@ end
     if i === nothing
         return :(throw(ArgumentError("No StaticArray found in argument list")))
     end
+    # Passing the Size as an argument to _map leads to inference issues when
+    # recursively mapping over nested StaticArrays (see issue #593). Calling
+    # Size in the generator here is valid because a[i] is known to be a
+    # StaticArray for which the default Size method is correct. If wrapped
+    # StaticArrays (with a custom Size method) are to be supported, this will
+    # no longer be valid.
     S = Size(a[i])
     exprs = Vector{Expr}(undef, prod(S))
     for i ∈ 1:prod(S)
--- a/test/mapreduce.jl
+++ b/test/mapreduce.jl
@@ -149,4 +149,11 @@ end
 
         @test ((@SVector Int64[]) + (@SVector Int64[])) == (@SVector Int64[])
     end
+    @testset "Nested SVectors" begin
+        # issue #593
+        v = SVector(SVector(3, 2), SVector(5, 7))
+        @test @inferred(v + v) == SVector(SVector(6, 4), SVector(10, 14))
+        v = SVector(SVector(3, 2, 1), SVector(5, 7, 9))
+        @test @inferred(v + v) == SVector(SVector(6, 4, 2), SVector(10, 14, 18))
+    end
 end


src/mapreduce.jl
```diff
@@ -8,16 +8,21 @@
...
+    _map(f, a1, a2, as...)
 end
 
-@generated function _map(f, ::Size{S}, a::AbstractArray...) where {S}
+@generated function _map(f, a::AbstractArray...)
+    i = findfirst(ai -> ai <: StaticArray, a)
+    if i === nothing
+        return :(throw(ArgumentError("No StaticArray found in argument list")))
+    end
+    S = Size(a[i])
```
c42f: Is it ok to call `Size` inside the generator here? User defined wrapper types might add methods to this trait.

Though it looks like the current version of `same_size` wouldn't work for those anyway because it uses `_first_static`. Hmm.

martinholters: I'm afraid I don't quite understand what the issue might be here. Could you elaborate?

c42f: The issue is that generators run in the world as it existed when they were defined, so you can't call functions inside the generator which might have new methods added:

```julia
foo(x) = "default"
@generated function bar(x)
    foo(x) # Function foo is frozen in world where bar is defined
end

foo(::Type{Int}) = "override"
@generated function bar2(x)
    foo(x) # Otherwise identical `bar2` sees both foo methods.
end
```

Thence:
```julia
julia> bar(1)
"default"

julia> bar2(1)
"override"
```

It's quite a restriction :-/ This is why we need the pattern of constructing the traits outside the generators and passing them in as types.

martinholters: Ah i see where you getting at. But we (currently) only dispatch to `_map` if at least one of the first two arguments is a `StaticArray`, and `a[i]` will be the first of those. Further, for a `StaticArray` the implementation from https://github.com/JuliaArrays/StaticArrays.jl/blob/bb82cd5b78bb0945565652ebe5b23116e33eb3ab/src/traits.jl#L88 should always be correct. If a subtype overloads `Size` to give a different size than what it declares in its supertype, that's just ... wrong.

But if we in the future want to dispatch wrapped `StaticArray`s, where the wrapper resides in some other package, to `_map`, that would become a problem indeed. At the moment, I think the only option for that dispatch to happen would be for that other package to provide a `map` that invokes `StaticArrays._map`. Which raises the question: Do such packages exist and should we therefore keep the old `_map` signature for compatibility anyway?

c42f: > But we (currently) only dispatch to _map if at least one of the first two arguments is a StaticArray, and `a[i]` will be the first of those

I think you're right that this is quite unlikely to cause problems with things as they exist right now. Maybe we should just go with it, with a comment clearly noting that calling this in the generator is only ok because of the particular restrictions on which `a[i]` we pick out.

> other package to provide a map that invokes `StaticArrays._map`

I don't know of any packages using `_map`. But if they do see fit to use these internal implementation details, I think the onus is on them to keep up to date with `StaticArrays` :-)

andyferris:
I agree with Chris. I'm hoping this is better overall.

martinholters:
Heh, just realized that the original issue can simply be solved by adding just this method:
```julia
@inline function map(f, a::SArray{S}...) where S
    SArray{S}(map(f, map(ai -> ai.data, a)...))
end
```
Obviously, this won't help for `StaticArray`s other than `SArray`. So what if the wrapper functions convert all arguments to `SArray`s, dispatch to this simple `map`, and convert the result back? IIUC correctly, inlining and SROA should avoid actually constructing the `SArray`s/tuples, so if everything works out ideally, there shouldn't be any run-time overhead for this conversion to `SArray`s and back. Do you have any experience whether that is actually true?

Another catch is that `Base.map` for `Tuple`s has this cutoff at 16 elements where it switches to a non-inferable, allocating algorithm which we'd want to avoid by reproducing the implementation without this cutoff. Taking everything together, this might be an alternative implementation (which doesn't need `@generated` functions anymore):
```julia
@inline function map(f, a1::StaticArray, as::AbstractArray...)
    _map(f, a1, as...)
end
@inline function map(f, a1::AbstractArray, a2::StaticArray, as::AbstractArray...)
    _map(f, a1, a2, as...)
end
@inline function map(f, a1::StaticArray, a2::StaticArray, as::AbstractArray...)
    _map(f, a1, a2, as...)
end

@inline function _map(f, as...)
    S = same_size(as...)
    res = map(f, map(a -> convert(SArray{size_tuple(S)}, a), as)...)
    return convert(similar_type(typeof(_first(as...)), eltype(res), S), res)
end
@inline function _map(f, a::SArray{S}...) where S
    SArray{S}(tuple_map(f, map(ai -> ai.data, a)...))
end

tuple_map(f, ::Tuple{}...) = ()
@inline function tuple_map(f, ts::Tuple...)
    (f(map(t -> t[1], ts)...), tuple_map(f, map(Base.tail, ts)...)...)
end
```
Would that be preferable?

c42f:
That non-`@generated` option is clever. Though it seems hard on the compiler for long arrays and appears to generate bad code for lengths greater than 32 (in julia-1.1).

With the latest compiler we could almost do this with a simple imperative loop and an `MArray`; convert that back into a tuple and put it into `similar_type` at the end like you have here. Inferring the element type for the `MArray` would be annoying though. Maybe impossible without going back to using `return_type`.

tkoolen:
Regarding inferability of `map` for tuples, I opted for this solution based on `ntuple` in another package (can be generalized to any number of arguments):

https://github.com/tkoolen/StaticUnivariatePolynomials.jl/blob/9c2ffd3c8278accc29ea63780297100f7c08464e/src/StaticUnivariatePolynomials.jl#L147-L157

Not sure which would be better in terms of compilation time / inference limits.

The `SArray` -> `MArray` -> `SArray` trick can definitely also work, e.g. in

https://github.com/tkoolen/PlanarConvexHulls.jl/blob/5b8c35a8dda9c1cdf53005bbd372d7974b59bb10/src/hrep.jl#L34-L38

which doesn't allocate and produces efficient code. I did have to rely on `arithmetic_closure`, which in the general case would indeed be `return_type`.

martinholters:
I had tried an `ntuple`-based approach, but that also has problems with nested `map`ping, so doesn't lead to a solution of #593. What might be worth investigating is using a `@generated` function for above `tuple_map`. I'll give that a try next week. If that doesn't do the trick, then I'll add some comments to the PR in its current form and we can defer finding a better (non-`@generated`?) solution to later when it's actually needed.

martinholters:
This is what I got:
* Use a `@generated` `tuple_map` improves codegen for the actual mapping, making it quite usable.
* For non-`SArray` inputs, the overhead from the conversion can only be eliminated for small cases (e.g. SizedArray -> SArray up to twelve elements). 
* The conversion probably also again leads to inference issues when recursively mapping over nested  non-`SArray`s.

So I'll add the requested comment and leave the code in this PR in its current form.

c42f:
That's a very clear comment, thanks. Feel free to merge it when you're ready :-)

PR #595, opened 2019-02-23
Reshape with tuple of SOneTo

devmotion:
This PR fixes
```julia
using StaticArrays

X = @SVector [1, 2.5, 3]
reshape(X, axes(X))
```
which fails currently and causes test failures in LabelledArrays (see https://travis-ci.com/JuliaDiffEq/LabelledArrays.jl/jobs/178867473).


ChrisRackauckas:
Can I release?

c42f:
The list of changes seems small and restricted to bugfixes. A release should be fine I think.

PR #596, opened 2019-03-01
Support for `svd(::StaticMatrix, full=true)`

c42f:
A small update to the `svd` wrapper so that `full=true` is supported. Code occasionally uses this, for example `nullspace`... in fact this is the only place `full=true` occurs in `LinearAlgebra`.

XRef: https://discourse.julialang.org/t/why-julia-and-linear-algebra-module-are-incredibly-slow-compare-to-c-and-eigen/21304/6

### DIFF
--- a/src/svd.jl
+++ b/src/svd.jl
@@ -32,11 +32,26 @@ function svdvals(A::StaticMatrix)
     similar_type(A, T2, Size(diagsize(A)))(sv)
 end
 
-function svd(A::StaticMatrix)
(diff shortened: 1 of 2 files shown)


c42f:
Well, CI passed. If nobody objects I'll go ahead and merge it in a couple of days as it should be uncontroversial. (Having said that, a review is always appreciated.)

andyferris:
Beat you to it :stuck_out_tongue:

PR #601, opened 2019-04-18
Drop 0.7

fredrikekre:
closes #525 by replacing it.


c42f:
This looks like a bunch of good cleanup for stuff which was required during the 0.7 dev cycle.

But do we actually need to drop 0.7 itself here? Is anything significant gained by doing so?

fredrikekre:
> But do we actually need to drop 0.7 itself here? Is anything significant gained by doing so?

Same reason Julia 0.7 and 1.0 are different versions -- when you are on 0.7 you should get a StaticArrays release with deprecations. Thus, the next StaticArrays release without deprecations should not be installable on Julia 0.7.

c42f:
Hmm. I don't think packages are or generally can be consistent about synchronizing their deprecations with julia releases. Though it seems fair to argue that `StaticArrays` is sufficiently similar to `Base` that this provides some value.

@mschauer has expressed concern about dropping 0.7 before, but at this stage `StaticArrays` has been pretty stable for a while and 0.7/1.0 is rapidly receding into the past. What do you think Moritz?

fredrikekre:
> I don't think packages are or generally can be consistent about synchronizing their deprecations with julia releases

I agree in general, but in this case the StaticArrays deprecations match the deprecations for Julia.

c42f:
Cool let's go with this.

PR #607, opened 2019-05-15
Fix A\B where A is nonsquare

### LINKED ISSUE #606: Missing `\` methods
Methods for `\` with non square matrix are missing:
```julia
using StaticArrays
A = @SArray randn(3,2)
v = @SArray randn(3,1)
Array(A) \ Array(v) # works
A \ v
```
```
DimensionMismatch("right hand side B needs first dimension of size 2, has size (3, 1)")
```

c42f:
Fix #606

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -1,17 +1,17 @@
-@inline (\)(a::StaticMatrix, b::StaticVecOrMat) = solve(Size(a), Size(b), a, b)
+@inline (\)(a::StaticMatrix, b::StaticVecOrMat) = _solve(Size(a), Size(b), a, b)
 
-@inline function solve(::Size{(1,1)}, ::Size{(1,)}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticVector{<:Any, Tb}) where {Ta, Tb}
+@inline function _solve(::Size{(1,1)}, ::Size{(1,)}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticVector{<:Any, Tb}) where {Ta, Tb}
     @inbounds return similar_type(b, typeof(a[1] \ b[1]))(a[1] \ b[1])
 end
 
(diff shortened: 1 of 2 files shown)


src/solve.jl
```diff
@@ -1,17 +1,17 @@
-@inline (\)(a::StaticMatrix, b::StaticVecOrMat) = solve(Size(a), Size(b), a, b)
+@inline (\)(a::StaticMatrix, b::StaticVecOrMat) = _solve(Size(a), Size(b), a, b)
```
c42f: By the way, I've renamed `solve` to `_solve` here just to make extra clear that it's an internal function.

c42f:
I think this should be pretty safe so I'll merge it shortly.

However, always happy for a review. Any takers? @jw3126?

src/solve.jl
```diff
@@ -28,12 +28,12 @@ end
...
 for Sa in [(2,2), (3,3)]  # not needed for Sa = (1, 1);
     @eval begin
-        @inline function solve(::Size{$Sa}, ::Size{Sb}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticMatrix{<:Any, <:Any, Tb}) where {Sb, Ta, Tb}
+        @inline function _solve(::Size{$Sa}, ::Size{Sb}, a::StaticMatrix{<:Any, <:Any, Ta}, b::StaticMatrix{<:Any, <:Any, Tb}) where {Sb, Ta, Tb}
             d = det(a)
             T = typeof((one(Ta)*zero(Tb) + one(Ta)*zero(Tb))/d)
             c = similar(b, T)
             for col = 1:Sb[2]
-                @inbounds c[:, col] = solve(Size($Sa), Size($Sa[1],), a, b[:, col])
+                @inbounds c[:, col] = _solve(Size($Sa), Size($Sa[1],), a, b[:, col])
```
jw3126: Not sure here. Does `b[: col]` allocate if `b` is a `MArray`?

c42f: No, this is one of the luxuries of the julia-1.0 compiler: a temporary `MArray` which doesn't escape the function doesn't result in any allocations. I still can't quite believe it myself :-)

jw3126: Nice! LGTM

PR #611, opened 2019-06-01
Adopt examples in README to current Julia version

goggle:
I have noticed, that some of the code examples in the `Quick start` section of the README don't work because they were designed for an older version of Julia. This PR modifies these examples to make them work with Julia v1.1.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ julia 0.5 with default optimizations.
 ```julia
 Pkg.add("StaticArrays")  # or Pkg.clone("https://github.com/JuliaArrays/StaticArrays.jl")
 using StaticArrays
+using LinearAlgebra
 
 # Create an SVector using various forms, using constructors, functions or macros
 v1 = SVector(1, 2, 3)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -102,14 +103,14 @@ typeof(v1[[1,2,3]]) <: Vector # Can't determine size from the type of [1,2,3]
 
 # Is (partially) hooked into BLAS, LAPACK, etc:
 rand(MMatrix{20,20}) * rand(MMatrix{20,20}) # large matrices can use BLAS
-eig(m3) # eig(), etc uses specialized algorithms up to 3×3, or else LAPACK
+eigvecs(m3) # eigvecs(), etc uses specialized algorithms up to 3×3, or else LAPACK
```
andyferris: FYI the equivalent thing is `eigen`. (It probably doesn't matter).

goggle: Ah, I didn't find this, I only found `eigvecs` and `eigvals`. I will change it to `eigen`.

goggle:
Let me know If I need to change anything else.
I can also squash the commits if you would like.

Issue #532, opened 2018-10-27
Release 1.0.0


c42f:
As discussed over at #522, it's almost time for a 1.0.0 release. Personally I think there should be no great breaking changes here and we should try to make it compatible with the status quo of the last year or so. However, we can take the opportunity to make it as consistent with the final state of julia-1.0 Base as possible via a few targeted deprecations (with proper depwarns).

Here's a TODO list (moved from prior discussion at #522).  Feel free to add or suggest items.

* [x] Triage past issues & add any v1.0 milestones
* [x] Triage current PRs & add any v1.0 milestones
* [x] Deprecations
  * [x] Remove v0.9 deprecations (#525)
  * [x] Deprecate `SDiagonal` in favor of `Base.Diagonal` #531
  * [x] Remove deprecated FixedSizeArrays module
* [x] A way forward for constructors
  * [x] Make sure we have a strategy for #518 (probably #633 or #636)
  * [x] ~~Deprecate (most of?) the zoo of constructor macros (replace with #633 ?) but keep them around in the 1.0 release cycle.~~ Can be soft-deprecated later
  * [x] ~~Constructors with generators or ntuple like syntax #97~~ Can be done later
  * [x] Fix #520
  * [x] Fix #557
* [ ] ~~Revisit `axes`; `SUnitRange` etc to consider whether these design changes should go into 1.0~~ Too big for 1.0
  * [ ] ~~Tidy up usage of `SOneTo` (follow up of #534; see also #620)~~
  * [ ] ~~Export of `SUnitRange` #474~~
  * [ ] ~~Consider modernizing & merging #150~~
* [x] Resolve fallout from #503 where possible (eltype becoming `Union{}`) cf. #528
* [x] Consider whether we can avoid promoting MArray to SArray quite so often #327

musm:
Presumably the FixedSizedArrays compatibility layer can also be removed?

mschauer:
Yes, it is deprecated and was vendorised in the only dependence: https://github.com/JuliaGeometry/GeometryTypes.jl/pull/153

c42f:
Good idea, see #600.

chakravala:
Will release v1.0.0 have any breaking changes?

c42f:
Probably a few more deprecations to come during cleanup but nothing specifically on my mind. Currently it seems that bumping major versions of widely used packages is a bit more disruptive than would be ideal, so it would seem best to wait until we have a batch of breaking changes.

[...]

PR #613, opened 2019-06-04
Support \ for SVD factorization

timholy:


timholy:
Seems like time for a new release. While we wait for #532, given that some return types have changed I propose 0.11. 

(If there's a good reason to delay a new release, no worries, there's nothing critical about a new release as far as I'm concerned. Just seems like a lot has changed.)

c42f:
Agreed, I've just been thinking in the past few days that we should have a new release.

PR #617, opened 2019-06-06
Add julia compat.

fredrikekre:


c42f:
We no longer test on 0.7, as of d20d707 so we probably should either
* Reinstate CI, or
* Only advertise compatibility with 1.x

c42f:
Thanks, let's try this.

PR #618, opened 2019-06-10
Change SOneTo indexing and support \ for svd(M; full=true)

timholy:
#613 didn't have adequate test coverage for solving problems like `svd(A) \ v` when `A` is non-square. It worked for the "thin" decomposition, but `svd(A; full=Val(true)) \ v` failed.

While implementing this I noticed what appeared to be missing support for indexing StaticArrays with `SOneTo`. This led to a change in one test result (see bea22b412c1f1330823f060cebfdf7515ac48d2d), so this is worth thinking about carefully and is a breaking change.

### DIFF
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -80,7 +80,7 @@ using StaticArrays, Test, LinearAlgebra
 
         @test m[:, 1:2] isa Matrix
         @test m[:, [true, false, false]] isa Matrix
-        @test m[:, SOneTo(2)] isa MMatrix{2, 2, Int}
+        @test m[:, SOneTo(2)] isa SMatrix{2, 2, Int}
         @test m[:, :] isa SMatrix{2, 3, Int}
         @test m[:, 1] isa SVector{2, Int}
         @test m[2, :] isa SVector{3, Int}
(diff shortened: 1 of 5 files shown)


c42f:
Oddly, the MacOS build for julia-1.0 failed with a timeout. There wasn't a lot to see in the build log so I've restarted it.

test/abstractarray.jl
```diff
@@ -80,7 +80,7 @@ using StaticArrays, Test, LinearAlgebra
 
         @test m[:, 1:2] isa Matrix
         @test m[:, [true, false, false]] isa Matrix
-        @test m[:, SOneTo(2)] isa MMatrix{2, 2, Int}
+        @test m[:, SOneTo(2)] isa SMatrix{2, 2, Int}
```
c42f: I think this new behavior is preferable because it's more consistent. It's technically breaking, but I think the chances of this breaking the average user of `StaticArrays` are low.

test/indexing.jl
```diff
@@ -101,6 +101,10 @@ using StaticArrays, Test
         @test (@inferred getindex(sm, :, SVector(2,1))) === @SMatrix [3 1; 4 2]
         @test (@inferred getindex(sm, 1, :)) === @SVector [1,3]
         @test (@inferred getindex(sm, :, 1)) === @SVector [1,2]
+
+        # SOneTo
+        @testinf sm[SOneTo(1),:] === @SMatrix [1 3]
+        @testinf sm[:,SOneTo(1)] === @SMatrix [1;2]
```
c42f: Could you add an equivalent test for `MMatrix` just to assert the `MMatrix` type is preserved?

[edit - I went ahead and added this]

Issue #623, opened 2019-06-30
2x2 complex exponential of zeros matrix is undefined


sdaxen:
Exponential of sized complex 2x2 matrix of zeros produces `NaN`s. This is not the case for a regular `Matrix`.

```julia
julia> using LinearAlgebra, StaticArrays

julia> exp(@SMatrix zeros(ComplexF64, 2, 2))
2×2 SArray{Tuple{2,2},Complex{Float64},2,4}:
 NaN+NaN*im  NaN+NaN*im
 NaN+NaN*im  NaN+NaN*im

julia> exp(@SMatrix zeros(ComplexF64, 1, 1))
1×1 SArray{Tuple{1,1},Complex{Float64},2,1}:
 1.0 + 0.0im

julia> exp(@SMatrix zeros(ComplexF64, 3, 3))
3×3 SArray{Tuple{3,3},Complex{Float64},2,9}:
 1.0+0.0im  0.0+0.0im  0.0+0.0im
 0.0+0.0im  1.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im  1.0+0.0im

julia> exp(zeros(ComplexF64, 2, 2))
2×2 Array{Complex{Float64},2}:
 1.0+0.0im  0.0+0.0im
 0.0-0.0im  1.0+0.0im

julia> exp(@SMatrix zeros(Float64, 2, 2))
2×2 SArray{Tuple{2,2},Float64,2,4}:
 1.0  0.0
 0.0  1.0
```

mschauer:
`(a-d)^2 + 4*b*c == 0` should be special cased like in the real case, to avoid division by zero in

https://github.com/JuliaArrays/StaticArrays.jl/blob/86c16e323c3b9ed6c7812f0e900368a5c918ccde/src/expm.jl#L55

sdaxen:
Just submitted a pull request with special-casing.

mschauer:
Closed by #624

PR #624, opened 2019-06-30
Avoid division by 0 in 2x2 complex exp

sdaxen:
See https://github.com/JuliaArrays/StaticArrays.jl/issues/623

### DIFF
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -50,15 +50,15 @@ end
     @inbounds d = A[4]
 
     z = sqrt((a - d)*(a - d) + 4*b*c )
-    e = exp((a + d - z)/2)
-    f = exp((a + d + z)/2)
+    e = expm1((a + d - z) / 2)
+    f = expm1((a + d + z) / 2)
     ϵ = eps()
     g = abs2(z) < ϵ^2 ? exp((a + d) / 2) * (1 + z^2 / 24) : (f - e) / z
 
-    m11 = (g * (a - d) + f + e) / 2
+    m11 = (g * (a - d) + f + e) / 2 + 1
     m12 = g * b
     m21 = g * c
-    m22 = (-g * (a - d) + f + e) / 2
+    m22 = (-g * (a - d) + f + e) / 2 + 1
 
     (newtype)((m11, m21, m12, m22))
 end
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -52,12 +52,13 @@ end
     z = sqrt((a - d)*(a - d) + 4*b*c )
     e = exp((a + d - z)/2)
     f = exp((a + d + z)/2)
-    zr = inv(z)
+    ϵ = eps()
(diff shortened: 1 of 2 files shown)


test/expm.jl
```diff
@@ -6,6 +6,7 @@ using StaticArrays, Test, LinearAlgebra
     @test exp(@SMatrix [4 2; -2 1])::SMatrix ≈ exp([4 2; -2 1])
     @test exp(@SMatrix [4 2; 2 1])::SMatrix ≈ exp([4 2; 2 1])
     @test exp(@SMatrix [ -3+1im  -1+2im;-4-5im   5-2im])::SMatrix ≈ exp(Complex{Float64}[ -3+1im  -1+2im;-4-5im   5-2im])
+    @test exp(@SMatrix [2+2im  1-1im; -2-2im  6+2im])::SMatrix ≈ exp(Complex{Float64}[2+2im  1-1im; -2-2im  6+2im])
```
mschauer:
```suggestion
    # test for case `(a[1,1] - a[2,2])^2 + 4*a[2,1]*a[1,2] == 0`
    @test exp(@SMatrix [2+2im  1-1im; -2-2im  6+2im])::SMatrix ≈ exp(Complex{Float64}[2+2im  1-1im; -2-2im  6+2im])
```

src/expm.jl
```diff
@@ -52,12 +52,13 @@ end
     z = sqrt((a - d)*(a - d) + 4*b*c )
     e = exp((a + d - z)/2)
```
mschauer: we could compute expm1 here
```suggestion
    e = expm1((a + d - z)/2)
```

src/expm.jl
```diff
@@ -52,12 +52,13 @@ end
...
     f = exp((a + d + z)/2)
-    zr = inv(z)
+    ϵ = eps()
+    g = abs2(z) < ϵ^2 ? exp((a + d) / 2) * (1 + z^2 / 24) : (f - e) / z
 
-    m11 = (-e*(a - d - z) + f*(a - d + z)) * zr/2  
-    m12 = (f-e) * b * zr
-    m21 = (f-e) * c * zr
-    m22 = (-e*(-a + d - z) + f*(-a + d + z)) * zr/2
+    m11 = (g * (a - d) + f + e) / 2
```
mschauer:
```suggestion
    m11 = (g * (a - d) + f + e + 2) / 2
```
etc...

sdaxen:
I don't know how much `expm1` gets us here, but since the most common case where `z==0` is when we exponentiate the zeros matrix, I agree it makes sense to use it.

src/expm.jl
sdaxen: I don't know how much `expm1` gets us here, but since the most common case where `z==0` is when we exponentiate the zeros matrix, I agree it makes sense to use it.

c42f:
Looks great to me, thanks!

Issue #518, opened 2018-10-17
Inconsistency with `Scalar` and conversion vs wrap


timholy:
Given that `Scalar <: SArray`, the behavior of
https://github.com/JuliaArrays/StaticArrays.jl/blob/552bc7596c1639d1f66bf781084d3961b36dbd53/src/Scalar.jl#L10
is inconsistent with
https://github.com/JuliaArrays/StaticArrays.jl/blob/552bc7596c1639d1f66bf781084d3961b36dbd53/src/convert.jl#L5

Fortunately, in a few cases there's a very welcome ambiguity error:
```julia
julia> sa = Scalar((1,))
Scalar{Int64}((1,))

julia> Scalar(sa)
ERROR: MethodError: Scalar{T} where T(::Scalar{T}) is ambiguous. Candidates:
  (::Type{Scalar{T} where T})(a::AbstractArray{T,0} where T) in StaticArrays at /home/tim/.julia/dev/StaticArrays/src/Scalar.jl:11
  (::Type{Scalar{T} where T})(a::AbstractArray) in StaticArrays at /home/tim/.julia/dev/StaticArrays/src/Scalar.jl:10
  (::Type{SA})(a::StaticArray) where SA<:StaticArray in StaticArrays at /home/tim/.julia/dev/StaticArrays/src/convert.jl:4
Possible fix, define
  Scalar{T} where T(::StaticArray{S,T,0} where T where S<:Tuple)
Stacktrace:
 [1] top-level scope at none:0
```
but mostly this just silently does things differently for 0 dimensions than for any other dimension.

I don't think you can really have both. Can you choose which behavior you want to keep?

andyferris:
Right. From a user perspective, we tend to make available the "varargs list of elements" constructor, like `SVector(x, y, z)`.

Do you know what the current best practice is for array constructors vs `convert`? What do other packages do? I guess this is a bit like the current `Number` debate... except we've been moving *toward* having an `AbstractArray` constructor interface...

andyferris:
I don't think this is a `Scalar` issue - it holds for all length-1 static arrays.

We have the "varargs list of elements" constructor on purpose. I particularly don't like special cases where certain lengths (typically 0 or 1) don't follow the overall pattern. Currently, this seems odd to me:

```julia
julia> SVector(1:3)
ERROR: UndefVarError: S not defined
Stacktrace:
[...]

PR #626, opened 2019-07-08
Add FieldMatrix: a rank-two tensor generalisation of FieldVector

cfbaptista:
`FieldVector` already facilitates the creation of vector types which can be used inside `Vector{<:FieldVector}` arrays to represent vector fields. I also have the need to represent rank-two tensor fields (gradients of vector fields for example). Therefore, I have implemented the `FieldMatrix` type to facilitate this. Others might find this useful too.

Inheriting from this type makes it easy to create your own rank-two N x M tensor types in a similar fashion that `FieldVector` does for vectors types. Here is an example:

```
struct Stress <: FieldMatrix{3, 3, Float64}
    xx::Float64
    xy::Float64
    xz::Float64
    yx::Float64
    yy::Float64
    yz::Float64
    zx::Float64
    zy::Float64
    zz::Float64
end
```

Tests can be found in `test/FieldMatrix.jl`. These tests mimick the tests from `test/FieldVector.jl`. All tests from `test/runtests.jl`  (with `test/FieldMatrix.jl` included) passed without any warning/error.

Please let me know whether this is useful and/or improvements are needed before this can be accepted. For example, I'm not too sure about the constructors I provided. They work, but might not yet be optimal performance-wise.


tkoolen:
Should we generalize things more and have a `FieldArray` instead?

cfbaptista:
@tkoolen that would indeed support all use cases. People from solid mechanics, for example, sometimes use rank-four stiffness tensors. However, performance for large static arrays is no good, right?

fredrikekre:
> struct Stress

> People from solid mechanics, for example, sometimes use rank-four stiffness tensors.


Completely off-topic, but have you seen https://github.com/KristofferC/Tensors.jl ?

c42f:
This looks pretty neat. Agreed that it would be good to generalize to arbitrary dimensions if you're willing to take a go at that. Then I imagine `FieldMatrix` would become a type alias of `FieldArray`.

Somewhat tangential, but have you seen https://github.com/JuliaDiffEq/LabelledArrays.jl ? It's an interesting alternative approach where the array type is identified by the names of the fields. (LabelledArrays wasn't possible when we originally created `StaticArrays.FieldVector` but became workable with the introduction of `getproperty`.)

cfbaptista:
@fredrikekre I have now :-)

I had a quick glance and it seems that https://github.com/KristofferC/Tensors.jl supports at least all of my needs. Please correct me if I'm wrong, but I think types which inherit from `FieldMatrix` are more generic than tensors from `Tensor.jl` because:

1. Tensor.jl only supports tensors with the same length in each dimension (i.e. N, N x N and N x N x N x N)
2. The length in each dimension can only be 1, 2 or 3.
3. Tensor.jl only supports real-valued tensors (T <: Real)

These limitations do not bother me because I don't need more, but I don't think this applies to everyone. Also I noticed strange behaviour when adding a literal value to a `Tensors.Tensor`:

```
using Tensors
T = Tensor{2, 3, Float64}(rand(3, 3))
U = T .+ 1.0

println(typeof(T))
Tensor{2,3,Float64,9}

println(typeof(U))
Array{Float64,2}
```

I do not like that the result is heap allocated.

cfbaptista:
@c42f Thanks and I also agree that it would be good to generalise to `FieldArray`. Both `FieldVector` and `FieldMatrix` can be typealises.

https://github.com/JuliaDiffEq/LabelledArrays.jl looks very promising indeed. I would be very interested in some benchmarks with functions from `LinearAlgebra`. It appears to be built on top on StaticArrays, so I'm sure it's performant.

cfbaptista:
As of commit 89f40d8 `FieldArray` has been added and `FieldVector` and `FieldMatrix` are defined as abstract types which are subtypes of `StaticArray`. Inheriting from `FieldArray` makes it possible to define your own tensor types of arbitrary rank.

The constructors might need some additional work and some actual testing of higher-rank tensors should be added.

c42f:
Some relevant thoughts about our constructor woes over at https://github.com/JuliaArrays/StaticArrays.jl/issues/518#issuecomment-509881044.

With that proposal, the way to efficiently and unambiguously construct `Stress` (without users needing to define their own constructor) would be

```julia
Stress(@SA[1 2 3;
           4 5 6;
           7 8 9])
```

src/FieldArray.jl
```diff
@@ -0,0 +1,85 @@
...
+
+    struct Point3D <: FieldVector{3, Float64}
+        x::Float64
+        y::Float64
+        z::Float64
+    end
+"""
+abstract type FieldVector{N, T} <: FieldArray{Tuple{N}, T, 1} end
+
+@inline (::Type{FA})(x::Tuple{Vararg{Any, N}}) where {N, FA <: FieldArray} = if length(FA) == length(x)
```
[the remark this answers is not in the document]
cfbaptista: I agree.

src/FieldArray.jl
```diff
@@ -0,0 +1,85 @@
...
+                                                                                FA(x...)
+                                                                             else
+                                                                                throw(DimensionMismatch("No precise constructor for $FA found. Length of input was $(length(x))."))
+                                                                             end
+
+@propagate_inbounds getindex(a::FieldArray, i::Int) = getfield(a, i)
```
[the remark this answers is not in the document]
cfbaptista: I am going to change the order of the fields in the documentation example.

cfbaptista:
Thanks for pointing this out @c42f, the order of the fields in the example in OP was an oversight of mine. It should have been:

```
struct Stress <: FieldMatrix{3, 3, Float64}
    xx::Float64
    yx::Float64
    zx::Float64
    xy::Float64
    yy::Float64
    zy::Float64
    xz::Float64
    yz::Float64
    zz::Float64
end
```

The same modification should also be applied to `Stiffness` in the example for `FieldArray`.

I would not go the column-major way because row-major is the standard / more common in Julia. We could explicitly mention it in the documentation, however, I think column-major is expected behaviour.

I fail to see the need for the macro in constructing an object of type `Stress`. We can already do this:

`Stress(rand(3, 3))`

c42f:
Excellent, thanks for sorting out the field ordering.

> I would not go the column-major way because row-major is the standard / more common in Julia. We could explicitly mention it in the documentation, however, I think column-major is expected behaviour.

Column major is expected by the subset of Julia users who know to think about storage orders. However, the average user of multidimensional arrays would generally not need to know about this so it's worth mentioning in the doc strings.

> Stress(rand(3, 3))

This makes the compiler generate rather slow code compared to `Stress(@SArray(rand(3,3)))` because it forces the allocation of a temporary `Matrix` to hold the result from `rand(3,3)`, followed by copying that into `Stress`:

```
[ Info: Recompiling stale cache file /home/tcfoster/.julia/compiled/v1.1/StaticArrays/yY9vm.ji for StaticArrays [90137ffa-7385-5640-81b9-e52037218182]

julia> struct Stress <: FieldMatrix{3, 3, Float64}
           xx::Float64
           yx::Float64
           zx::Float64
           xy::Float64
           yy::Float64
           zy::Float64
           xz::Float64
           yz::Float64
           zz::Float64
       end

julia> @btime Stress(rand(3,3))
  46.995 ns (1 allocation: 160 bytes)
3×3 Stress:
 0.592865  0.26077   0.870697
 0.900356  0.129086  0.6465  
 0.157575  0.422919  0.219663

julia> @btime Stress(@SArray(rand(3,3)))
  17.234 ns (0 allocations: 0 bytes)
3×3 Stress:
 0.20502   0.166018  0.699433
 0.790852  0.96874   0.207043
 0.292533  0.03907   0.831171
```

For cases where there's less work being compared to generating the random numbers this will only be more pronounced.

cfbaptista:
Yes, I meant column-major instead of row-major. I edited my post.

Ah I didn't know it was slow. Then I agree with the macro as well.

c42f:
By the way, I hope you can you see what I mean about the perils of the multi-argument constructor? With the latest version:

```julia
julia> using StaticArrays

julia> struct Stress <: FieldMatrix{3, 3, Float64}
           xx::Float64
           yx::Float64
           zx::Float64
           xy::Float64
           yy::Float64
           zy::Float64
           xz::Float64
           yz::Float64
           zz::Float64
       end

julia> Stress(1,2,3,  # Column major ordering with misleading formatting
              4,5,6,
              7,8,9)
3×3 Stress:  # Oops, interpreted as the transpose of the above
 1.0  4.0  7.0
 2.0  5.0  8.0
 3.0  6.0  9.0
```

This isn't your fault! The problem of constructors is a difficult problem we've been struggling with for some time (see #518).

However I think it's worth mentioning this potential pitfall somewhere. It would be possible to disallow the multi-argument constructor by default in an effort to prevent these issues though that might be heavy handed.  It would be less bad with a shorter `@SA` macro as proposed in #518.

cfbaptista:
Yes, I do see the issue with the multi-argument constructor. I will make a mention of this in the documentation.

I'm in for `@SA`. It's short and it saves memory.

However, `Stress` is a struct with 9 fields, so I don't know how you want to disallow a constructor with 9 arguments. This might be a Julia feature I am not aware of.

test/FieldMatrix.jl
```diff
@@ -0,0 +1,104 @@
...
+    @testset "Mutable Tensor2x2" begin
+        eval(quote
+            mutable struct Tensor2x2{T} <: FieldMatrix{2, 2, T}
+                xx::T
+                xy::T
+                yx::T
+                yy::T
+            end
+
+            StaticArrays.similar_type(::Type{P2D}, ::Type{T}, s::Size{(2,2)}) where {P2D<:Tensor2x2,T} = Tensor2x2{T}
```
c42f: No real need to change it, but note this can be written 

```julia
StaticArrays.similar_type(::Type{<:Tensor2x2}, ::Type{T}, s::Size{(2,2)}) where {T} = Tensor2x2{T}
```

test/FieldMatrix.jl
```diff
@@ -0,0 +1,104 @@
...
+        @test_throws Exception p[2] = 4.0
+        @test_throws Exception p[2, 3] = 6.0
+
+        @test size(p) === (3,3)
+        @test length(p) === 9
+        @test eltype(p) === Float64
+
+        @testinf Tuple(p) === (1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0)
+
+        @test (p + p) === Tensor3x3(2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0)
```
c42f: This could do with an `@inferred` on the left hand side

src/FieldArray.jl
```diff
@@ -0,0 +1,103 @@
...
+        yxxy::Float64
+        xyxy::Float64
+        yyxy::Float64
+        xxyy::Float64
+        yxyy::Float64
+        xyyy::Float64
+        yyyy::Float64
+    end
+
+Do note that `FieldArray` and its derivatives are column major.
```
c42f:
```suggestion
Note that you must define the fields of any `FieldArray` subtype in column major order. If you
want to use an alternative ordering you will need to pay special attention in providing your
own definitions of `getindex`, `setindex!` and tuple conversion.
```

src/FieldArray.jl
```diff
@@ -0,0 +1,103 @@
...
+        zx::Float64
+        xy::Float64
+        yy::Float64
+        zy::Float64
+        xz::Float64
+        yz::Float64
+        zz::Float64
+    end
+
+Do note that `FieldMatrix` and its derivatives are column major. Constructing Stress like this
```
c42f:
```suggestion
Note that the fields of any subtype of `FieldMatrix` must be defined in column major order.
This means that formatting of constructors for literal `FieldMatrix` can be confusing. For example
```

src/FieldArray.jl
```diff
@@ -0,0 +1,103 @@
...
+                   4.0, 5.0, 6.0,
+                   7.0, 8.0, 9.0)
+
+    3×3 Stress:
+     1.0  4.0  7.0
+     2.0  5.0  8.0
+     3.0  6.0  9.0
+
+
+will give you the transpose of what the multi-argument formatting suggests.
```
c42f:
```suggestion
will give you the transpose of what the multi-argument formatting suggests. For clarity,
you may consider using the alternative

    sigma = Stress(@SArray[1 2 3;
                           4 5 6;
                           7 8 9])
```

tkoolen:
I'm with @c42f, thanks very much for the effort!

c42f:
> However, Stress is a struct with 9 fields, so I don't know how you want to disallow a constructor with 9 arguments. This might be a Julia feature I am not aware of.

Well anything is possible with an inner constructor but that doesn't help us here because any general constructors we define for `FieldArray` must be an outer constructor. I think we can revisit this later if we feel the need, perhaps as part of deciding exactly what to do with #518.

test/FieldMatrix.jl
cfbaptista: That's something I copied from the pre-existing `test/FieldVector.jl`. I am applying your suggestion to both `test/FieldVector.jl` and `test/FieldMatrix.jl`.

c42f: Sure no problem. Historically (julia ~ 0.5 when this package was started?) this syntax wasn't possible.

test/FieldMatrix.jl
cfbaptista: Sure thing!

src/FieldArray.jl
cfbaptista: Done

src/FieldArray.jl
cfbaptista: Done

Issue #616, opened 2019-06-06
New Release 0.11.0


c42f:
Time for a new release (https://github.com/JuliaArrays/StaticArrays.jl/pull/613#issuecomment-499117218)

Let's see how registrator + tagbot works out...

@JuliaRegistrator register

Release notes:

## Breaking changes
* Remove deprecated functionality (#600, #601)

## New features
* Add `SHermitianCompact` (#358)
* Support `\` for SVD factorization (#613)
* Support for `svd(::StaticMatrix, full=true)` (#596)
* Allow using static arrays as index for views (#586)

## Performance
* Performance improvements for `vecdot(x,y)` (#603)
* Improved `A\B` performance for size(A)=(2or3, 2or3) and size(B) = (2or3, n) (#584)
* Added faster inverse for 4x4 matrices (#597)

## Bug fixes and minor changes
* Fix crash in `eig()` for diagonal 2x2 complex matrices (#615)
* Fix README examples for current Julia (#611)
* Test on julia version 1.2 (#610)
* Fix constructor for `SVector{0}` from `Array` (#608)
* Fix A\B where A is nonsquare (#607)
* Allow empty indexing of zero-dimensional MArrays (#581)
* Fix for zero dimensional `broadcast!` (#590)

c42f:
@JuliaRegistrator register

c42f:
Ok, that seems to have worked.

It looks like the manually written release notes were't picked up due to retriggering registrator but that's easily fixed.

PR #632, opened 2019-07-16
Deprecate ::Scalar +/- :StaticArray, fixes #627.

### LINKED ISSUE #627: Deprecate ::Number +/- ::StaticArray ?
This is inconsistent with `::Number +/- ::Array`:
```julia
julia> using StaticArrays

julia> SA = @SMatrix rand(2,2); A = Matrix(SA);

julia> 1 + SA
 1.75609  1.63302
 1.98185  1.06612

julia> 1 + A
ERROR: MethodError: no method matching +(::Int64, ::Array{Float64,2})
```

fredrikekre:


andyferris:
I agree with @c42f that we should make a couple of releases (with and without the deprecation).

What's the process with that these days? Edit the version in the Project.toml file and then use JuliaRegistrator?

c42f:
Yep it's that simple: change the version number in Project.toml and trigger JuliaRegistrator with

@ JuliaRegistrator register()

on an issue or commit comment (obviously fix the formatting to actually trigger the bot)

Tagbot allows you to write release notes in the issue too, though you need to include them with every JuliaRegistrator invocation; see my clumsy attempt at https://github.com/JuliaArrays/StaticArrays.jl/issues/616. I'm somewhat inclined to let Tagbot create the release notes in the future because it saves a lot of work and Tagbot does a pretty good job.

PR #636, opened 2019-07-22
WIP new `@SA` macro and comprehension support

andyferris:
New `@SA` macro designed to be a shorter, more focussed and hopefully more correct version of the `@SArray` macro. This macro can be used with array literal sytax like `@SA[1,2,3]` (even without a space!).

Typed comprehension using one-based indexing and up to two dimensions is currently supported. Some work remains to generalize this but we can rely on constant propagation for things like `@SA T[f(x) for x in 1:3]` (the old macro would `eval` the `1:3` part).

@c42f this is the other reasonable alternative to #633. Infact the PR includes that work as a subset - I suppose we should play a bit with both and see what we prefer. I suspect untyped comprehensions mightn't be too different to implement (there is type promotion logic in the `SArray` constructor after all). The other required generalizations are to support more than two ranges and supporting multidimensional inputs in the comprehension (note that `Array`s seem to take the Cartesian product of the input `keys`). We might also consider going beyond one-based indexing in the input arrays (but I'm not convinced that is necessary).


c42f:
Nice, will try to take a closer look tomorrow.

src/SArray.jl
```diff
@@ -237,3 +237,112 @@ end
...
+    elseif isa(ex, Expr) && ex.head == :typed_comprehension
+        if length(ex.args) != 2 || !isa(ex.args[2], Expr) || ex.args[2].head != :generator
+            error("Expected generator in typed comprehension, e.g. Float64[f(i,j) for i = 1:3, j = 1:3]")
+        end
+        T = ex.args[1]
+        ex = ex.args[2]
+        n_ranges = length(ex.args) - 1
+        range_vars = [ex.args[i+1].args[1] for i = 1:n_ranges]
+        ranges = [ex.args[i+1].args[2] for i = 1:n_ranges]
+        func = esc(:(($(Expr(:tuple, range_vars...)) -> $(ex.args[1]))))
```
c42f: I wonder if we might be able to remove some of this logic and just pass the generator along to `typed_sarray_comprehension`.  Ie, make this lowering less about the syntax, and more about just lowering to something other than `typed_comprehension` which is mangled by Base lowering...

andyferris:
cc @EvertSchippers @MeganDawson

c42f:
Replaced by #633

PR #639, opened 2019-07-26
Add docstrings for push, pop, pushfirst, popfirst, insert, deleteat

garrison:
Also tested the new docstrings locally

The main "critique" I could make is that putting `-> vec` may imply the original `vec` is returned rather than a modified one.  Suggestions welcome for making this more clear.

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -1,3 +1,18 @@
+"""
+    push(vec::StaticVector, item)
+
+Return a new `StaticVector` with `item` inserted on the end of `vec`.
+
+# Examples
+```jldoctest
+julia> push(@SVector[1, 2, 3], 4)
+4-element SArray{Tuple{4},Int64,1,4} with indices SOneTo(4):
+ 1
+ 2
+ 3
+ 4
+```
+"""
 @inline push(vec::StaticVector, x) = _push(Size(vec), vec, x)
 @generated function _push(::Size{s}, vec::StaticVector, x) where {s}
     newlen = s[1] + 1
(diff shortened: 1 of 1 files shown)


garrison:
Also, I purposely left out `setindex` because it should really be documented in Base.  However, I just realized that there is actually a slight difference between how `setindex` behaves here and in Base:  in Base (for a `Tuple`) it will be a no-op if the last index is out of bounds, while here (for a `StaticVector`) it will throw a `BoundsError`.  So maybe they actually each need a docstring after all.

c42f:
Excellent, thanks! It seems that `Base.setindex` is not documented nor exported, so I think it would be worth documenting the `setindex` that we have here.

Regarding `-> vec` and making it clear that the input `vec` isn't modified, I suppose you could reword the text slightly. Eg for `push`:

> Return a new vector with `item` inserted on the end of `vec`

and similarly for the others.

src/deque.jl
```diff
@@ -1,3 +1,18 @@
+"""
+    push(vec::StaticVector, item) -> vec
+
+Insert an `item` at the end of `vec` and return the resulting vector.
```
fredrikekre: "Insert" makes it sound a bit like it is mutating. Maybe something like "Create a new vector with `item` in the end" or similar? (Same on the other docstrings.)

garrison:
Changes made as requested.  Please let me know if any other comments.

fredrikekre:
> putting -> vec may imply the original vec is returned rather than a modified one.

I suggest just dropping that.

garrison:
Done

c42f:
Looks great to me, thanks!

PR #640, opened 2019-07-30
Fix type piracy in test/qr.jl, fixes #437

### LINKED ISSUE #437: Don't commit type piracy in test/qr.jl
https://github.com/JuliaArrays/StaticArrays.jl/blob/77d051fe474e582398e62971742d4562fb57f232/test/qr.jl#L3-L8

denius:


c42f:
Thanks for this!

The build error was something spurious-looking related to building the docs (surprising and confusing but it disappeared on a restart of the docs job)

PR #643, opened 2019-08-01
Simplify broadcast code generation

### LINKED ISSUE #642: Broadcast operations resulting in Scalars fail
julia> Scalar(1) .+ 1
    ERROR: BoundsError: attempt to access 0-element Array{Int64,1} at index [1]
    Stacktrace:
     [1] getindex(::Array{Int64,1}, ::Int64) at ./array.jl:729
     [2] #s162#227 at /home/eivind/.julia/dev/StaticArrays/src/broadcast.jl:118 [inlined]

TheBB:
This simplifies code according to a tip left in a comment, by using `CartesianIndices` instead of generating indices manually. This also fixes #642 by not trying to increment an empty index.

I broke a long line there according to personal taste, if that's okay.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -105,30 +105,15 @@ scalar_getindex(x::Tuple{<: Any}) = x[1]
         end
     end
 
-    exprs = Array{Expr}(undef, newsize)
-    more = prod(newsize) > 0
-    current_ind = ones(Int, length(newsize))
     sizes = [sz.parameters[1] for sz ∈ s.parameters]
-
-    while more
-        exprs_vals = [(!(a[i] <: AbstractArray) ? :(scalar_getindex(a[$i])) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))])) for i = 1:length(sizes)]
-        exprs[current_ind...] = :(f($(exprs_vals...)))
-
-        # increment current_ind (maybe use CartesianIndices?)
-        current_ind[1] += 1
-        for i ∈ 1:length(newsize)
-            if current_ind[i] > newsize[i]
-                if i == length(newsize)
-                    more = false
-                    break
-                else
-                    current_ind[i] = 1
-                    current_ind[i+1] += 1
-                end
-            else
-                break
-            end
-        end
+    indices = CartesianIndices(newsize)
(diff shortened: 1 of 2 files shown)


src/broadcast.jl
```diff
@@ -105,30 +105,13 @@ scalar_getindex(x::Tuple{<: Any}) = x[1]
...
-                    break
-                else
-                    current_ind[i] = 1
-                    current_ind[i+1] += 1
-                end
-            else
-                break
-            end
-        end
+    exprs = map(CartesianIndices(newsize)) do current_ind
```
tkoolen: Hmm, I don't think you can safely use closures in function generators.

TheBB: I've never been able to trigger an error that way, and the tests seem to agree. But I'm happy to rewrite anyway.

c42f: Oddly enough, it seems that you can't use closures in the generated code, not the generator itself:

```julia
julia> @generated function foo()
           :(()->1)
       end

julia> foo()
ERROR: generated function body is not pure. this likely means it contains a closure or comprehension.
Stacktrace:
 [1] top-level scope at none:0
```

andyferris: That's right. The generated code goes through a restricted, "pure" version of lowering, that can't modify the type tree or method tables. A closure adds both a new type and the closure method itself. 

On the other hand, while the generator itself is considered `@pure` it can contain closures/comprehensions (since they can be desugared deterministically during initial code lowering).

AFAICT these restrictions allowed Jameson to build a sound implementation of dependency-based recompilation (OTOH it's unclear to me whether they are *necessary* to create a sound implementation of dependency-based recompilation).

c42f: I don't it's necessary; more viewed as a limitation of the current implementation — see https://discourse.julialang.org/t/a-method-to-remove-the-use-of-runtime-eval-and-invokelatest-as-well-as-support-closures-in-generated-functions-come-with-an-implemented-proptype/27450/2

c42f:
I haven't looked through this in detail, but it's a lot neater, Andy has approved and the tests pass, so I merged it. Thank you @TheBB!

PR #644, opened 2019-08-01
Enable linear indexing with multidimensional index

TheBB:
This PR attempts to allow indexing operations of the form `indexee[indexer]` where both *indexer* and *indexee* may be multidimensional.

This was previously possible if *indexer* was one-dimensional (seems to have gone down the linear indexing branch of the code) or if *indexee* was one-dimensional, but *indexer* may be multidimensional (this goes to the multidimensional indexing path, and there's a test for it so this must have been intentional).

This sends all `getindex(::StaticArray, ::StaticArray)` down the linear indexing path (where they belong, IMO) by updating the methods of form `getindex(::StaticArray, ::StaticVector)` to accept arrays for the second argument. The other changes are natural consequences. Of course the same goes for setindex.

Note, I had to change the second argument from `StaticVector{<:Any, Int}` to `StaticArray{<:Tuple, Int}`, otherwise Julia resolved the ambiguity with the multidimensional branch the wrong way. That branch only works if the number of indices are equal to the number of dimensions in the indexee. That can wait for another day.

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -112,15 +112,15 @@ end
     end
 end
 
-@propagate_inbounds function getindex(a::StaticArray, inds::StaticVector{<:Any, Int})
-    _getindex(a, Length(inds), inds)
+@propagate_inbounds function getindex(a::StaticArray, inds::StaticArray{<:Tuple, Int})
+    _getindex(a, Size(inds), inds)
 end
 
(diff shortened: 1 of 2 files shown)


c42f:
For some reason it surprised me that the `StaticArray{<:Tuple, Int}` trick works. But yes, this looks good to me.

PR #646, opened 2019-08-09
add `in` method

stev47:
This change makes `in` unroll the loop (compared to `Base.in`) and is
significantly faster.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -203,6 +203,8 @@ end
 @inline any(a::StaticArray{<:Tuple,Bool}; dims=:) = reduce(|, a; dims=dims, init=false) # (benchmarking needed)
 @inline any(f::Function, a::StaticArray; dims=:) = mapreduce(x->f(x)::Bool, |, a; dims=dims, init=false) # (benchmarking needed)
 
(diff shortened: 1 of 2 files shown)


c42f:
Looks good to me, the generated code is much better for numeric types. It's much faster for short vectors and still a lot better for static arrays of length 16:

```julia
julia> f(x) = x in (1,3,2,5,6,7,8,10,11,12,13,70,80,81,82,16)

julia> g(x) = x in SVector(1,3,2,5,6,7,8,10,11,12,13,70,80,81,82,16)
g (generic function with 1 method)

# Wrap in mapreduce to prevent optimizer from removing `f` entirely (??)
julia> @btime mapreduce(f, |, $(rand(Int, 1000)))
  6.086 μs (0 allocations: 0 bytes)
false

julia> @btime mapreduce(g, |, $(rand(Int, 1000)))
  1.276 μs (0 allocations: 0 bytes)
false
```

Side note: here's an entertaining optimization opportunity exposed by this PR:
```julia
julia> g(x) = x in SVector((1,2,3,4))
g (generic function with 1 method)

julia> @code_native debuginfo=:none g(1)
	.text
	addq	$-1, %rdi
	cmpq	$4, %rdi
	setb	%al
	retq
	nopl	(%rax)
```

PR #652, opened 2019-08-23
Fix eltype in empty static array construction

### LINKED ISSUE #651: Type Stability of Constructor
There are some type stability issues (at least for StaticArrays v0.11.0) for an empty `SVector`:
```julia
julia> using StaticArrays

julia> SVector{0, Float64}([])
0-element SArray{Tuple{0},Any,1,0} with indices SOneTo(0)

julia> SVector{0, Float64}(Float64[])
0-element SArray{Tuple{0},Float64,1,0} with indices SOneTo(0)

julia> SVector{0, Float64}(Float32[])
0-element SArray{Tuple{0},Float32,1,0} with indices SOneTo(0)
```
I expected to get a `0-element SArray{Tuple{0},Float64,1,0} with indices SOneTo(0)` in all three cases, similar to the non-empty case:
```julia
julia> SVector{1, Float64}([2])
1-element SArray{Tuple{1},Float64,1,1} with indices SOneTo(1):
 2.0

julia> SVector{1, Float64}(Float64[2])
1-element SArray{Tuple{1},Float64,1,1} with indices SOneTo(1):                      
 2.0                                                                                
                                                                                    
julia> SVector{1, Float64}(Float32[2])
1-element SArray{Tuple{1},Float64,1,1} with indices SOneTo(1):                      
 2.0                                                                                
```

This must have worked at some point in the past, since I got errors caused by this recently on Travis for one of my packages (see https://travis-ci.org/ranocha/SummationByPartsOperators.jl/builds/575385284). The last build before (where everything was fine) has been on 2019-03-07.

c42f:
Fixes #651

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -26,7 +26,8 @@ end
     return _convert(SA, a, Length(SA))
 end
 
-@inline _convert(SA, a, ::Length{L}) where L = SA(unroll_tuple(a, Length(SA)))
+@inline _convert(SA, a, l::Length) = SA(unroll_tuple(a, l))
+@inline _convert(SA::Type{<:StaticArray{<:Tuple,T}}, a, ::Length{0}) where T = similar_type(SA, T)(())
 @inline _convert(SA, a, ::Length{0}) = similar_type(SA, eltype(a))(())
 
 length_val(a::T) where {T <: StaticArrayLike} = length_val(Size(T))
--- a/test/convert.jl
+++ b/test/convert.jl
@@ -10,4 +10,8 @@ end # testset
     # Issue #520
     @testinf SVector{0}(Int8[]) === SVector{0,Int8}()
     @testinf SMatrix{0,0}(zeros(0,0)) === SMatrix{0,0,Float64}(())
+
+    # Issue #651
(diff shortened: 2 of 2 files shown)


src/convert.jl
```diff
@@ -26,7 +26,8 @@ end
     return _convert(SA, a, Length(SA))
 end
 
-@inline _convert(SA, a, ::Length{L}) where L = SA(unroll_tuple(a, Length(SA)))
+@inline _convert(SA, a, l::Length) = SA(unroll_tuple(a, l))
+@inline _convert(SA::Type{<:StaticArray{<:Tuple,T}}, a, ::Length{0}) where T = similar_type(SA, T)(())
```
andyferris: I suspect a type assertion would be good here (that the output is indeeed an `SA`).

c42f: We could do that. Which error condition are you trying to cover, is it protection for the user to avoid stack overflows when they have a broken `similar_type` (constructor calling `_convert` calling constructor etc) or some such?

andyferris: Just guarding against the case that `similar_type(SA)` isn’t an `SA` (like rotations) which means constructors and conversion might return a rather surprising result.

PR #656, opened 2019-09-15
SA type aliases for constructors with precise eltype

### LINKED ISSUE #24: Naming for small vectors

c42f:
The discussion at #655 has stimulated what I feel is finally a reasonable solution to constructing small static arrays with precise eltype without too much typing:

```julia
SA_F64 = SA{Float64}
SA_F32 = SA{Float32}

```

I don't love the underscore there, but `SAF64` looks very confusing.  Other than that, I think the naming here is consistent with the precedent set by `Base.ComplexF64`.

Better naming options welcome!

I believe this (along with #633) finally fixes #24 to the extent that we can have a solution to that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ Pkg.add("StaticArrays")  # or Pkg.clone("https://github.com/JuliaArrays/StaticAr
 using StaticArrays
 using LinearAlgebra
 
+# Use the convenience constructor type `SA` to create vectors and matrices
+SA[1, 2, 3]     isa SVector{3,Int}
+SA_F64[1, 2, 3] isa SVector{3,Float64}
+SA_F32[1, 2, 3] isa SVector{3,Float32}
+SA[1 2; 3 4]     isa SMatrix{2,2,Int}
+SA_F64[1 2; 3 4] isa SMatrix{2,2,Float64}
+
 # Create an SVector using various forms, using constructors, functions or macros
 v1 = SVector(1, 2, 3)
 v1.data === (1, 2, 3) # SVector uses a tuple for internal storage
--- a/docs/src/pages/quickstart.md
+++ b/docs/src/pages/quickstart.md
@@ -5,6 +5,13 @@ Pkg.add("StaticArrays")  # or Pkg.clone("https://github.com/JuliaArrays/StaticAr
 using StaticArrays
 using LinearAlgebra
 
+# Use the convenience constructor type `SA` to create vectors and matrices
+SA[1, 2, 3]     isa SVector{3,Int}
+SA_F64[1, 2, 3] isa SVector{3,Float64}
+SA_F32[1, 2, 3] isa SVector{3,Float32}
+SA[1 2; 3 4]     isa SMatrix{2,2,Int}
+SA_F64[1 2; 3 4] isa SMatrix{2,2,Float64}
+
 # Create an SVector using various forms, using constructors, functions or macros
 v1 = SVector(1, 2, 3)
 v1.data === (1, 2, 3) # SVector uses a tuple for internal storage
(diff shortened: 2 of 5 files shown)


andyferris:
I do have mixed feelings about the way it "looks". Also contrast with `N0f8` (which might suggest `SAf64`?).

Do any other community members have some opinions on this?

andyferris:
The other interesting case is `ComplexF16`, `ComplexF32`, `ComplexF64`.

c42f:
The `SAf64` alternative would be ok.

But I think the existing alternative is better because it really emphasizes the `SA` part separately. This is kind of important because we've chosen to overload the typed array construction syntax for this special tag type `SA`.

```julia
julia> const SAf64 = SA_F64
SA{Float64}

julia> SA_F64[1,2,3]
3-element SArray{Tuple{3},Float64,1,3} with indices SOneTo(3):
 1.0
 2.0
 3.0

julia> SAf64[1,2,3]
3-element SArray{Tuple{3},Float64,1,3} with indices SOneTo(3):
 1.0
 2.0
 3.0
```

c42f:
Does anyone have better ideas for naming or should we just merge this? I'd like to have something equivalent to it to go with SA in a new release.

Any further thoughts @andyferris; do you buy my argument or remain unconvinced? :-)

c42f:
Well, unless anyone's inclined to paint this bikeshed a different color, I think I'll merge this tomorrow or so.

c42f:
In the name of forward progress I went ahead and merged this; I think/hope these names are a good tradeoff!

cdsousa:
Hi all,

Is a decrease in performance from using `@StaticArray[]` to using `SA[]` expected?

```julia

julia> f(x) = @SArray[1,2,x]

julia> g(x) = SA[1,2,x]

julia> @code_llvm f(3)
# short stuff

julia> @code_llvm g(3)
# long stuff

julia> @btime f(3)
  0.023 ns (0 allocations: 0 bytes)
3-element SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 1
 2
 3

julia> @btime g(3)
  249.480 ns (2 allocations: 64 bytes)
3-element SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 1
 2
 3

julia> versioninfo()
Julia Version 1.2.0
Commit c6da87ff4b (2019-08-20 00:03 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = "/usr/share/code/code"

julia> import Pkg; Pkg.status("StaticArrays")
    Status `~/.julia/environments/v1.2/Project.toml`
  [90137ffa] StaticArrays v0.11.0 #master (https://github.com/JuliaArrays/StaticArrays.jl.git)
```

c42f:
Yikes no! Looks like I made a mistake of leaving in some spurious `where T` when refactoring which the tests did't pick up. Does #665 fix it for you?

c42f:
By the way @cdsousa, thank you very much for reporting this problem.

These kind of performance problems are hard to detect without proper benchmarking, and proper benchmarking is hard without dedicated hardware... instead we rely heavily on `@inferred`, but sometimes (as in this case) `@inferred` can't be used to detect problems with inlining.

cdsousa:
@c42f, it is fixed indeed! And thank you very much for all the awesome work.

c42f:
Phew :-) You had me worried for a minute there (I did check this originally, but if I was mistaken it would nix the whole idea of having `SA`).

PR #657, opened 2019-09-17
[WIP] Nested views with static indexing

mateuszbaran:
This allows `StaticIndexing` to appear at any position when making a view.

I'm not sure if it counts as type piracy or not but this shouldn't cause any issues.

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -369,8 +369,7 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
 
 # unsafe_view
 
-# I was considering catching arbitrary `Union{StaticIndexing,Base.ViewIndex}`
-# indices here but they shouldn't ever be mixed here.
+# unsafe_view need only deal with vargs of `StaticIndexing`, as wrapped by to_indices.
 # i1 is explicitly specified to avoid ambiguities with Base
 Base.unsafe_view(A::AbstractArray, i1::StaticIndexing, indices::StaticIndexing...) = Base.unsafe_view(A, unwrap(i1), map(unwrap, indices)...)
 
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -372,4 +372,4 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
 unwrap_if_needed(i::StaticIndexing) = unwrap(i)
 unwrap_if_needed(i::Base.ViewIndex) = i
 
-Base.unsafe_view(A::AbstractArray, is::Union{StaticIndexing{T},Base.ViewIndex}...) where T = Base.unsafe_view(A, map(unwrap_if_needed, is)...)
+Base.unsafe_view(A::AbstractArray, is::Union{StaticIndexing,Base.ViewIndex}...) = Base.unsafe_view(A, map(unwrap_if_needed, is)...)
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -369,11 +369,13 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
 
 # unsafe_view
 
-unwrap_if_needed(i::StaticIndexing) = unwrap(i)
-unwrap_if_needed(i::Base.ViewIndex) = i
+# I was considering catching arbitrary `Union{StaticIndexing,Base.ViewIndex}`
+# indices here but they shouldn't ever be mixed here.
+Base.unsafe_view(A::AbstractArray, indices::StaticIndexing...) = Base.unsafe_view(A, map(unwrap, indices)...)
 
-Base.unsafe_view(A::AbstractArray, is::Union{StaticIndexing,Base.ViewIndex}...) = Base.unsafe_view(A, map(unwrap_if_needed, is)...)
-
-# views of views need a new method for Base._indices_sub
-
-Base._indices_sub(i::StaticIndexing, I...) = Base._indices_sub(unwrap(i), I...)
+# Views of views need a new method for Base.SubArray because storing indices
+# wrapped in StaticIndexing in field indices of SubArray causes all sorts of problems.
+# Additionally, in some cases the SubArray constructor may be called directly
+# instead of unsafe_view so we need this method too (Base._maybe_reindex
+# is a good example)
+Base.SubArray(A::AbstractArray, indices::NTuple{<:Any,StaticIndexing}) = Base.SubArray(A, map(unwrap, indices))
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -371,7 +371,8 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
 
 # I was considering catching arbitrary `Union{StaticIndexing,Base.ViewIndex}`
 # indices here but they shouldn't ever be mixed here.
-Base.unsafe_view(A::AbstractArray, indices::StaticIndexing...) = Base.unsafe_view(A, map(unwrap, indices)...)
+# i1 is explicitly specified to avoid ambiguities with Base
+Base.unsafe_view(A::AbstractArray, i1::StaticIndexing, indices::StaticIndexing...) = Base.unsafe_view(A, unwrap(i1), map(unwrap, indices)...)
 
 # Views of views need a new method for Base.SubArray because storing indices
 # wrapped in StaticIndexing in field indices of SubArray causes all sorts of problems.
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -369,4 +369,7 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
 
 # unsafe_view
 
-Base.unsafe_view(A::AbstractArray, i::StaticIndexing{T}) where T = Base.unsafe_view(A, unwrap(i))
+unwrap_if_needed(i::StaticIndexing) = unwrap(i)
+unwrap_if_needed(i::Base.ViewIndex) = i
+
+Base.unsafe_view(A::AbstractArray, is::Union{StaticIndexing{T},Base.ViewIndex}...) where T = Base.unsafe_view(A, map(unwrap_if_needed, is)...)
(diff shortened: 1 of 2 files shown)


mateuszbaran:
I've added another minor change that makes views of views work. I'm working on statically sized views right now so it's possible I'll uncover some other bugs.

src/indexing.jl
```diff
@@ -369,4 +369,11 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
 
 # unsafe_view
 
-Base.unsafe_view(A::AbstractArray, i::StaticIndexing{T}) where T = Base.unsafe_view(A, unwrap(i))
+unwrap_if_needed(i::StaticIndexing) = unwrap(i)
+unwrap_if_needed(i::Base.ViewIndex) = i
```
c42f: It looks like you could just add a method to `unwrap` itself rather than having `unwrap_if_needed`.

src/indexing.jl
```diff
@@ -369,4 +369,11 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
...
+
+Base.unsafe_view(A::AbstractArray, is::Union{StaticIndexing,Base.ViewIndex}...) = Base.unsafe_view(A, map(unwrap_if_needed, is)...)
```
c42f: This method signature does make me uneasy because it can also match cases where none of the argument types are defined in this package.

However it seems the method specificity rules make it mostly work (or at least I couldn't find a case where there's an ambiguity).

I did find the following bug though in the process:

```julia
julia> a = rand(3,4,5,6);

julia> b = @view a[:,:,1,1]
3×4 view(::Array{Float64,4}, :, :, 1, 1) with eltype Float64:
 0.450313  0.311055  0.747741  0.966306
 0.984626  0.313968  0.798511  0.71977 
 0.486325  0.42368   0.743077  0.285142

julia> @view b[SA[1,2],1]
2-element view(::Array{Float64,4}, StaticArrays.StaticIndexing{SArray{Tuple{2},Int64,1,2}}([1, 2]), StaticArrays.StaticIndexing{Int64}(1), StaticArrays.StaticIndexing{Int64}(1), StaticArrays.StaticIndexing{Int64}(1)) with eltype Float64 with indices SOneTo(2):
#undef

```

c42f:
Ah the build has failed. Looks like we're relying a lot on internals here which might be different in julia 1.0 and 1.1. Can you check whether those cases can be made to work as well?

src/indexing.jl
mateuszbaran: Thanks for looking at it :slightly_smiling_face: . I've already discovered this problem but I don't know yet how to fix it. I'm working on it.

mateuszbaran:
I'll check Julia 1.0 and 1.1 (I'm using 1.2 right now). Consider it a work in progress.

src/indexing.jl
mateuszbaran: `unwrap` has only one method that works on `StaticIndexing`, should I add this?
```julia
unwrap(i::Base.ViewIndex) = i
```

mateuszbaran:
The new version seems to work correctly. I've also added some comments.

src/indexing.jl
```diff
@@ -369,4 +369,14 @@ Base.checkindex(B::Type{Bool}, inds::AbstractUnitRange, i::StaticIndexing{T}) wh
 
 # unsafe_view
 
-Base.unsafe_view(A::AbstractArray, i::StaticIndexing{T}) where T = Base.unsafe_view(A, unwrap(i))
+# I was considering catching arbitrary `Union{StaticIndexing,Base.ViewIndex}`
+# indices here but they shouldn't ever be mixed here.
+# i1 is explicitly specified to avoid ambiguities with Base
```
c42f: I like these comments a lot: the indexing code is quite tricky and it's good to have some context for why things are like they are.

One minor thing - you could remove the reference to `I` here. (The fact that it's you who's thought carefully about this is definitely relevant to the discussion on this PR, but will be somewhat irrelevant to people looking at the code later.) Perhaps could be written:

```julia
# unsafe_view need only deal with vargs of `StaticIndexing`, as wrapped by to_indices.
# i1 is explicitly specified to avoid ambiguities with Base
```

mateuszbaran:
Thanks! I've changed the comment so it should be OK to merge it now. Do you want to have the commits squashed?

> I'm looking forward to the day that multidimensional mixtures of static and dynamic axes "just works" rather than requiring unwrapping, etc.

This should happen soon, depending on your definition of "just works" :slightly_smiling_face: . `HybridArrays` already generates decent code for many common access patterns. There are so many possible combinations though that I don't know which should be prioritized (apart from the ones that I need).

c42f:
Squashing is definitely a good idea small changes like this, but it's fairly easy to do that with the github UI without loosing the CI results or waiting for CI to rerun. So I just did that. The main downside is the maintainer needs to clean up the commit message (so I hope you're happy with what I wrote ;-) )

`HybridArrays` looks really great. If you need small changes to StaticArrays (to the traits and such) to make things smoother please don't hesitate to make a PR.

Also I wonder whether some things like `SizedArray` might eventually be subsumed by `HybridArray` given that the underlying storage mechanism will be basically the same. There's probably a big story for adding extra traits to dispatch which may need to involve `Base`.

Anyway, for now I think it's ideal for you to add the functionality you need and see where it leads.

mateuszbaran:
> Squashing is definitely a good idea small changes like this, but it's fairly easy to do that with the github UI without loosing the CI results or waiting for CI to rerun. So I just did that. The main downside is the maintainer needs to clean up the commit message (so I hope you're happy with what I wrote / copy & pasted ;-) )

Great, I'm happy with the message :slightly_smiling_face: .



> Also I wonder whether some things like `SizedArray` might eventually be subsumed by `HybridArray` given that the underlying storage mechanism will be basically the same. There's probably a big story for adding extra traits to dispatch which may need to involve `Base` and `LinearAlgebra`.

I don't think `HybridArray` can subsume `SizedArray`. Notice that `HybridArray` derives from `AbstractArray`, not `StaticArray`. A lot of code of `StaticArrays` assumes that all dimensions have statically known sizes and that's not the case for `HybridArray`. The general idea of `HybridArray` is to forward all operations to the wrapped array unless an operation involves only a part of the array with a statically known size (in which case I use the `SizedArrays` approach of unrolling everything).

c42f:
> HybridArray derives from AbstractArray, not StaticArray

Very true, but I view the fact that this matters at all as a design flaw of the ecosystem. For example, `Hermitian{<:StaticArray}` is also not a static array subtype, but we have to handle it in our static linear algebra code or people rightfully get confused.

This has led to a temporary non-extensible solution `StaticMatrixLike`:

https://github.com/JuliaArrays/StaticArrays.jl/blob/935dc85f0fe290672cc2a60bafbe5259795ddbca/src/StaticArrays.jl#L83-L94

but the real solution to these woes is to have better trait-based dispatch. Eg, have those traits in `LinearAlgebra` itself which would let us control dispatch in a more flexible way.

> unless an operation involves only a part of the array with a statically known size

But `SizedArray` is just a degenerate example of this case.

> in which case I use the SizedArrays approach of unrolling everything

I wonder, did you try just extracting into an `SArray` in that case and letting the compiler do the work? This can be a zero cost abstraction in many cases because the the compiler can figure out that the temporary SArray never needs to be constructed at all.  It probably wouldn't work in all cases but it might get you surprisingly far.

mateuszbaran:
> Very true, but I view the fact that this matters at all as a design flaw of the ecosystem. For example, `Hermitian{<:StaticArray}` is also not a static array subtype, but we have to handle it in our static linear algebra code or people rightfully get confused.

> But `SizedArray` is just a degenerate example of this case.

Yes, right, but making `HybridArray` fast enough to fully replace `SizedArray` is much more work than I'm planning to do in the short term.

> but the real solution to these woes is to have better trait-based dispatch. Eg, have those traits in `LinearAlgebra` itself which would let us control dispatch in a more flexible way.

Changing dispatch in `LinearAlgebra` to trait-based sounds like a Julia 2.0 plan. An easier thing to do would be converting `StaticArrayLike` and friends to THTT-based traits. That would be a small improvement.

> I wonder, did you try just extracting into an `SArray` in that case and letting the compiler do the work? This can be a zero cost abstraction in many cases because the the compiler can figure out that the temporary SArray never needs to be constructed at all. It probably wouldn't work in all cases but it might get you surprisingly far.

That's exactly what `getindex` does for `HybridArray`. I'm planning to just do `getindex`, `setindex!` and broadcasting for `HybridArray` and a few minor things like `hcat`. A lot of other operations can be done by composing these three with `StaticArrays` or by just accessing the underlying array. For example, I'm not planning to implement matrix multiplication of `HybridArray`s (but I could help if someone else wanted to do the job).

c42f:
> An easier thing to do would be converting StaticArrayLike and friends to THTT-based traits

Yeah, that's mainly what I had in mind.

If we got some kind of native trait/interface/protocol support in the language itself that would be great, but I don't think there's a timeline for that or a concrete design.

In any case it sounds like we're on the same page. It will be interesting to see where things go :-)

andyferris:
> Yeah, that's mainly what I had in mind.

Me too.

I've been hoping the trait can simply be whatever is returned by `axes` or `keys` or `eachindex`. (This would unify dynamic, hybrid and static arrays quite easily).

c42f:
> I've been hoping the trait can simply be whatever is returned by axes or keys or eachindex.

That should work quite well for shape. I'm less sure about container type `similar_type`?

andyferris:
> I'm less sure about container type `similar_type`?

It's tricky... this is the other huge problem to solve, no?

It's really nice to do the hard (computationally intensive) work with mutable containers (I mean, clearly operations that create an `SArray` aren't "scalable", since the amount of code necessarily scales with the number of elements of the container). I also think it's great to support a "functional" workflow and I find I basically never mutate the output of "functional operations" (like mutating the output of `map` or the output of matrix-matrix multiplication - these usually get passed as-is to the next operation).

I'm thinking a patern like:

```julia
function f(array)
    out = similar(array)
    f!(out, array)
    return freeze(out)
end

function f!(out, array)
    return _f!(eachindex(out, array), out, array)
end

function _f!(indices, out, array)
    # Here the code might be specialized on the indices... e.g. small static indices might involve unrolled loops, hybrid indices might unroll one of two nested loops, etc
end
```

Sometimes you may `thaw` the output of `f(array)`, which is fine (semantically - `similar` and `copyto!`), and clever compiler people may even let us do that one day without an extra memory copy whenever we can prove that it's safe.

andyferris:
> That should work quite well for shape

I'm thinking of slightly beyound the shape... the indices contain their "staticness" and the indices of a sparse matrix could allow a way to find the non-zero elements, and the indices of a hash map might know not just the keys but the layout in the hash map (tokens).

In a lot of cases I've thought of, this information is sufficient to write an efficient `_f!` method (above).

PR #658, opened 2019-09-19
Base.strides methods for statically sized arrays

mateuszbaran:
I need these methods for `HybridArrays` (views use them) and I think they belong here.

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -16,6 +16,9 @@ end
 Base.axes(rv::Adjoint{<:Any,<:StaticVector})   = (SOneTo(1), axes(rv.parent)...)
 Base.axes(rv::Transpose{<:Any,<:StaticVector}) = (SOneTo(1), axes(rv.parent)...)
 
+Base.strides(a::Union{SArray,MArray}) = Base.size_to_strides(1, size(a)...)
+Base.strides(a::SizedArray) = strides(a.data)
+
 function Base.summary(io::IO, a, inds::Tuple{SOneTo, Vararg{SOneTo}})
     print(io, Base.dims2string(length.(inds)), " ")
     Base.showarg(io, a, true)
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -16,7 +16,8 @@ end
 Base.axes(rv::Adjoint{<:Any,<:StaticVector})   = (SOneTo(1), axes(rv.parent)...)
 Base.axes(rv::Transpose{<:Any,<:StaticVector}) = (SOneTo(1), axes(rv.parent)...)
 
-Base.strides(a::Union{SArray,MArray}) = Base.size_to_strides(1, size(a)...)
+# Base.strides is intentionally not defined for SArray, see PR #658 for discussion
+Base.strides(a::MArray) = Base.size_to_strides(1, size(a)...)
 Base.strides(a::SizedArray) = strides(a.data)
 
 function Base.summary(io::IO, a, inds::Tuple{SOneTo, Vararg{SOneTo}})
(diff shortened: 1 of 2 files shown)


mateuszbaran:
These two functions need strides:
https://github.com/JuliaLang/julia/blob/30940a7aea98ca33c49549e5cb86a0c53bd523ca/base/subarray.jl#L388
https://github.com/JuliaLang/julia/blob/30940a7aea98ca33c49549e5cb86a0c53bd523ca/base/subarray.jl#L411

I decided to make views work with arbitrary static/hybrid arrays in `HybridArrays` (I copied tests for views from Base) and then make the cases I need fast. It doesn't look that complicated.

timholy:
For an `SArray`, is there even a pointer you can safely pass? I don't think they are guaranteed to be heap-allocated...

mateuszbaran:
Sorry, I should've written about it in the previous post. Yes, pointers (and even the whole concept of views) doesn't make much sense for `SArray` but it certainly does for `MArray` and `SizedArray`. Is this good enough reason to not define `strides` for `SArray`?

c42f:
Thanks Tim for cutting right to the heart of the matter! It seemed confusing that this would be needed or even useful for `SArray`.

> Is this good enough reason to not define strides for SArray?

Yes I think it would be best to not support this for `SArray` as it doesn't make a lot of sense. Other than that I think this looks fine to merge.

andyferris:
> For an `SArray`, is there even a pointer you can safely pass? 

If this is regarding `ccall` from `SArray{..,T}` to `Ptr{T}` (with `isbitstype(T)`), I wonder if we could override `cconvert`/`unsafe_convert` to make an intermediate `RefValue` (that would be optimized away to the stack)?

c42f:
Oh, you mean `cconvert(Ptr{T}, ::SArray)` returning a `RefValue` containing the static array? That's an interesting thought, it seems plausible and efficient!

Where does that leave us with `strides` I wonder...  It doesn't apply to the `SArray` per se but could still be useful when combined with ccall and seems harmless enough to have it defined.

andyferris:
Isn’t `strides` all about supporting BLAS and LAPACK?

c42f:
> Isn’t `strides` all about supporting BLAS and LAPACK?

Ok I checked. It looks like `stride` is also used as an optimization in `write(::IO, ::SubArray)` but that's about the only use in `Base`. The other 100 uses or so are as you say.

mateuszbaran:
I really appreciate this discussion, it's a good opportunity to learn something new :+1: 

I've removed the method for `SArray` so it should be good now.

c42f:
Thanks, I'm not sure the mentioned `cconvert` for `SArray` would make sense (given that you may want to pass those by either value or pointer to `ccall`?). So this seems good for now.

mateuszbaran:
I didn't even know about `cconvert` before this thread so I don't know if that would be useful. `strides` method for `SArray` can always be added when someone actually needs this.

c42f:
Yes exactly :+1:

I suspect we don't actually want this automatic conversion anyway because it's valid enough to pass `SVector`s by value to C functions (I think? I didn't try it.) and users might want to do that.

lhupe:
> `strides` method for `SArray` can always be added when someone actually needs this.

Hi, I actually need `strides` for StaticArrays. I want to save StaticArrays to plain HDF5 using `HDF5.jl`, which requires `strides` to be implemented (afaik only to check whether the stride of the first dimension is one, which seems to be a hard requirement of the format). As you can imagine, I'd like to avoid having to convert all StaticArrays to regular Arrays just for saving. 

I have checked that adding a `strides` method for SArrays solves the problem, so this seems to be the cleanest solution.

c42f:
@lhupe not all arrays define `strides` so I think this may just be a bug in HDF5.jl. Can you report it over there?

For a non-StaticArrays example, I'd think that writing a `Range` should work, and yet we have

```
julia> h5write("test.h5", "x", 1:3)
ERROR: MethodError: no method matching strides(::UnitRange{Int64})
```

Issue #641, opened 2019-07-31
reduce(op, a::StaticArray) is ambiguous


colinxs:
```
julia> svec = SVector([1,2], [1,2])
julia> reduce(vcat, svec)
ERROR: MethodError: reduce(::typeof(hcat), ::SArray{Tuple{2},Array{Int64,1},1,2}) is ambiguous. Candidates:
  reduce(op, a::StaticArray) in StaticArrays at /home/colinxs/.julia/packages/StaticArrays/VyRz3/src/mapreduce.jl:168
  reduce(::typeof(hcat), A::AbstractArray{#s72,1} where #s72<:(Union{AbstractArray{T,1}, AbstractArray{T,2}} where T)) in Base at abstractarray.jl:1324
Possible fix, define
  reduce(::typeof(hcat), ::StaticArray{S,#s72,1} where S<:Tuple where #s72<:(Union{AbstractArray{T,1}, AbstractArray{T,2}} where T))
```
This is on StaticArrays 0.11 and Julia 1.1.1. As far as I've seen this only happens for vcat/hcat.

I believe it's related to https://github.com/JuliaLang/julia/pull/27188 when specializations were added for vcat and hcat.

c42f:
Yep, this is one of the few remaining ambiguities. I guess we just need to work around it by adding a few specializations.

colinxs:
Thanks for the prompt response! I figured you were already aware but flagged this just in case :).

c42f:
Well I'm not sure we're aware of which ambiguities cause real problems so thanks for reporting it.

c42f:
Fixed in #662, thanks to @mateuszbaran

PR #662, opened 2019-09-26
fix for issue #641

mateuszbaran:
Just two methods for disambiguation.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -89,9 +89,9 @@ end
         @inbounds return $expr
     end
 end
-    
+
 @generated function _mapreduce(f, op, dims::Colon, nt::NamedTuple{(:init,)},
-                               ::Size{S}, a::StaticArray...) where {S}    
+                               ::Size{S}, a::StaticArray...) where {S}
     expr = :(nt.init)
     for i ∈ 1:prod(S)
         tmp = [:(a[$j][$i]) for j ∈ 1:length(a)]
@@ -124,7 +124,7 @@ end
                                ::Size{S}, a::StaticArray) where {S,D}
     N = length(S)
     Snew = ([n==D ? 1 : S[n] for n = 1:N]...,)
-    
+
     exprs = Array{Expr}(undef, Snew)
     itr = [1:n for n ∈ Snew]
     for i ∈ Base.product(itr...)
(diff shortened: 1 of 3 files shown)


c42f:
Awesome, thanks!

I guess this reduces the number of ambiguities detected at the top of runtests.jl - if you could also reduce the threshold correspondingly that would be great. (May be julia version dependent, so might be worth checking on version 1.0 as well. Also, we haven't updated that for a while, so it might not be exactly equal to the current threshold.)

mateuszbaran:
Yes, I'll see if that reduces the number of ambiguities.

mateuszbaran:
Julia 1.0 still reports 7 ambiguities, including two for `reduce` but I think they are quite hard to trigger.

src/mapreduce.jl
```diff
@@ -176,6 +176,13 @@ end
 
 @inline reduce(op, a::StaticArray; dims=:, kw...) = _reduce(op, a, dims, kw.data)
 
+# disambiguation
+reduce(::typeof(vcat), A::StaticVector{<:Any,<:AbstractVecOrMat}) =
+    Base._typed_vcat(mapreduce(eltype, promote_type, A), A)
```
c42f: It's worthwhile to get the ambiguities down to zero and keep them there, so I had a closer look. Method specificity is a complex heuristic so it can be hard to figure out which methods to write but in this case it's because of the `Any` in `StaticVector`. Amusingly the following broader-looking signature seems to fix it:

```julia
reduce(::typeof(vcat), A::StaticArray{<:Tuple,<:AbstractVecOrMat})
```

Also, perhaps we should dispatch to our own internal `_reduce` here rather than `Base._typed_vcat`?

mateuszbaran: That's quite interesting indeed. I'll change the signature then.

> Also, perhaps we should dispatch to our own internal `_reduce` here rather than `Base._typed_vcat`?

That's a good question. `Base._typed_vcat` preallocates output for the whole array that will be returned using `similar` with calculated new size. That should be faster than our `_reduce` for sized arrays of normal `Array`s but it doesn't return a static array when it gets a static array of static arrays. So I think we should have two methods here: one for static arrays of static arrays that calls our `_reduce` and one for everything else. Does it sound reasonable?

c42f: It's a bit of a weird special case, given that we don't have any other special cases of `reduce(op, a::StaticArray; dims=:, kw...)`; it makes me think that maybe this signature should have been tighter in some way (though I'm not sure quite how).

Anyway, yes, let's do as you suggest for now.

mateuszbaran: I've made these changes and now `StaticArrays` has two ambiguities less :tada: .

test/mapreduce.jl
```diff
@@ -157,4 +157,13 @@ using Statistics: mean
...
         @test @inferred(v + v) == SVector(SVector(6, 4, 2), SVector(10, 14, 18))
     end
+    @testset "hcat and vcat" begin
+        # issue #641
+        v = SVector([1,2], [3,4])
+        @test reduce(vcat, v) == [1,2,3,4]
+        @test reduce(hcat, v) == [1 3; 2 4]
+        v2 = SVector(SVector(1,2), SVector(3,4))
+        @test @inferred(reduce(vcat, v2)) === @SVector [1,2,3,4]
+        @test @inferred(reduce(hcat, v2)) === @SMatrix [1 3; 2 4]
```
c42f: Tests look spot on :+1:

PR #663, opened 2019-09-26
Broadcasting with tuples (fixes #485)

### LINKED ISSUE #485: broadcasting of Tuples and SVectors is broken

mateuszbaran:
I think this looks more or less right.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -95,7 +95,6 @@ end
 
 scalar_getindex(x) = x
 scalar_getindex(x::Ref) = x[]
-scalar_getindex(x::Tuple{<: Any}) = x[1]
 
 @generated function _broadcast(f, ::Size{newsize}, s::Tuple{Vararg{Size}}, a...) where newsize
     first_staticarray = 0
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -55,6 +55,7 @@ broadcast_indices(A::StaticArray) = indices(A)
 @inline broadcast_sizes() = ()
 @inline broadcast_size(a) = Size()
 @inline broadcast_size(a::AbstractArray) = Size(a)
+@inline broadcast_size(a::NTuple{N}) where N = Size(N)
 
 function broadcasted_index(oldsize, newindex)
     index = ones(Int, length(oldsize))
@@ -110,7 +111,7 @@ scalar_getindex(x::Tuple{<: Any}) = x[1]
     exprs = similar(indices, Expr)
     for (j, current_ind) ∈ enumerate(indices)
         exprs_vals = [
-            (!(a[i] <: AbstractArray) ? :(scalar_getindex(a[$i])) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))]))
+            (!(a[i] <: AbstractArray || a[i] <: Tuple) ? :(scalar_getindex(a[$i])) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))]))
             for i = 1:length(sizes)
         ]
         exprs[j] = :(f($(exprs_vals...)))
@@ -139,7 +140,7 @@ end
     exprs = similar(indices, Expr)
     for (j, current_ind) ∈ enumerate(indices)
         exprs_vals = [
-            (!(as[i] <: AbstractArray) ? :(as[$i][]) : :(as[$i][$(broadcasted_index(sizes[i], current_ind))]))
+            (!(as[i] <: AbstractArray || as[i] <: Tuple) ? :(as[$i][]) : :(as[$i][$(broadcasted_index(sizes[i], current_ind))]))
             for i = 1:length(sizes)
         ]
         exprs[j] = :(dest[$j] = f($(exprs_vals...)))
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -216,15 +216,18 @@ end
 
     @testset "broadcasting with tuples" begin
         # issue 485
-        v = SVector(1,2,3)
-        @test v .+ (10, 20, 30) == SVector(11, 22, 33)
-        vm = MVector(1,2,3)
-        vm .+= (10, 20, 30)
-        @test vm == SVector(11, 22, 33)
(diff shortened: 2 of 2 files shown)


src/broadcast.jl
```diff
@@ -110,7 +111,7 @@ scalar_getindex(x::Tuple{<: Any}) = x[1]
     exprs = similar(indices, Expr)
     for (j, current_ind) ∈ enumerate(indices)
         exprs_vals = [
-            (!(a[i] <: AbstractArray) ? :(scalar_getindex(a[$i])) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))]))
+            (!(a[i] <: AbstractArray || a[i] <: Tuple) ? :(scalar_getindex(a[$i])) : :(a[$i][$(broadcasted_index(sizes[i], current_ind))]))
```
c42f: I guess this means we should kill off the `scalar_getindex(x::Tuple)` further up? TBH I'm not sure why that was ever there.

mateuszbaran: We could, it looks like it was there to allow broadcasting with one-element tuples.

c42f: Ok, I think we should remove it for consistency, and add some tests to check that the functionality still works. AFAICT cases like

```julia
SA[1,2,3] .* (SA[1,0,0],)
```

used to work because of the `scalar_getindex` hack and your changes preserves that functionality while also allowing more things to work.

TBH the fact that we check `a[i] <: AbstractArray` makes me nervous and it seems like these problems will recur with other user-defined types (because we may not respect the `broadcastable` API? I'm not an expert on how this is set up though).

Regardless of that what you've got here is a definite improvement so we should go with it.

c42f:
I went ahead and removed that version of `scalar_getindex` for consistency.  Also I added a test to check that the previous functionality enabled by that still works.

In the process I thought I could tighten the tests up a bit more. They now check:
* That single element tuple broadcast with SA still works
* That expected things are inferred
* That the return types of these broadcasts are precisely as expected

Also I tried out the new `SA` convenience constructor.

I pushed all that to your branch — hope that's ok!  Please have a look over it and see whether it makes sense. (And we'll let full CI run again to keep me honest :-) )

test/broadcast.jl
```diff
@@ -213,4 +213,21 @@ end
...
+        @test @inferred((SA[10 20; 30 40]) .+ (1,2))     === SA{Int}[11 21; 32 42]
+
+        add_bc!(m, v) = m .+= v  # Helper function; @inferred gets confused by .+= syntax
+        @test @inferred(add_bc!(MVector((1,2,3)), (10, 20, 30)))   ::MVector{3,Int}   == SA[11, 22, 33]
+        @test @inferred(add_bc!(MMatrix(SA[10 20; 30 40]), (1,2))) ::MMatrix{2,2,Int} == SA[11 21; 32 42]
+
+        # Tuples of SA
+        @test SA[1,2,3] .* (SA[1,0],) === SVector{3,SVector{2,Int}}(((1,0), (2,0), (3,0)))
+        # Unfortunately this case of nested broadcasting is not inferred
+        @test_broken @inferred(SA[1,2,3] .* (SA[1,0],))
```
c42f: BTW, this case wasn't inferred before either, so we haven't made it worse in this PR.

src/broadcast.jl
mateuszbaran: > TBH the fact that we check `a[i] <: AbstractArray` makes me nervous and it seems like these problems will recur with other user-defined types (because we may not respect the `broadcastable` API? I'm not an expert on how this is set up though).

Yes, it's a bad hack that will bite us in the future but as far as I understand the broadcast machinery (I'm not an expert though) this unrolled broadcasting can't be expected to work with arbitrary types participating in broadcasting anyway. A potentially better solution would be checking `Base.BroadcastStyle` of `a[i]` for `Style{Tuple}` and `AbstractArrayStyle`, so that could be done as well. But constructing `Base.BroadcastStyle` in a generated function doesn't sound like a great idea either. Or we could construct `Base.BroadcastStyle` in `Base.copy(B::Broadcasted{StaticArrayStyle{M}})`.

This can be improved gradually so we can make this a little bit better in this PR first.

mateuszbaran:
> I pushed all that to your branch — hope that's ok! Please have a look over it and see whether it makes sense. (And we'll let full CI run again to keep me honest :-) )

That's great, thanks :slightly_smiling_face: . I'll look over your changes.

test/broadcast.jl
mateuszbaran: This inference failure is weird. `Base.copy(B::Broadcasted{StaticArrayStyle{M}})` fails to inline and then its return type is not inferred. But `Cthulhu.jl` says everything's OK when I descend to that method of `copy`. There seems to be a problem outside `StaticArrays`.

mateuszbaran:
> Please have a look over it and see whether it makes sense.

I think the code makes sense :+1: .

c42f:
Great, merged!

PR #664, opened 2019-09-26
Avoid eltype degrading to `Union{}` for empty map/broadcast

### LINKED ISSUE #528: Adding n x 0 matrices results in Union{} eltype
On 0.9.1:

```julia
julia> using StaticArrays

julia> x = zeros(SMatrix{3, 0, Float64})
3×0 SArray{Tuple{3,0},Float64,2,0}

julia> x + x
3×0 SArray{Tuple{3,0},Union{},2,0}
```

while on 0.8.3:

```julia
julia> using StaticArrays

julia> x = zeros(SMatrix{3, 0, Float64})
3×0 SArray{Tuple{3,0},Float64,2,0}

julia> x + x
3×0 SArray{Tuple{3,0},Float64,2,0}
```

c42f:
This reverts to using `Core.Compiler.return_type` for map/broadcast, but
only in the very restricted case that the output container is completely
empty.

This is consistent with the way that return_type is used in Base for
collect and broadcast for empty collections only.

Fixes #528 using the strategy from https://github.com/JuliaArrays/StaticArrays.jl/issues/528#issuecomment-517110844

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -97,11 +97,15 @@ scalar_getindex(x::Ref) = x[]
 scalar_getindex(x::Tuple{<: Any}) = x[1]
 
 @generated function _broadcast(f, ::Size{newsize}, s::Tuple{Vararg{Size}}, a...) where newsize
-    first_staticarray = 0
-    for i = 1:length(a)
-        if a[i] <: StaticArray
-            first_staticarray = a[i]
-            break
+    first_staticarray = a[findfirst(ai -> ai <: StaticArray, a)]
+
+    if prod(newsize) == 0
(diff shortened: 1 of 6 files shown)


c42f:
@tkoolen I think this finally fixes #528 in a way which is safe. Or at least, it's no worse than what Base already does for mapping an empty collection. If it's good enough for Base, I think it's good enough for us :-)

tkoolen:
Thank you! Makes sense and looks good to me once tests pass.

c42f:
Eh, that's a lot of test failures! That's what I get for pushing right after a last minute rebase. Lucky we have CI to keep us honest :-D  Let's see what it says now.

PR #666, opened 2019-09-30
Conversion of SizedArray to Array reshapes

mateuszbaran:
I'm trying to port some improvements I've made in `HybridArrays` to `SizedArray` here. Full support for wrapping arbitrary `AbstractArray`s is a fairly large project and noticeable increase in complexity of `SizedArray` so I'm not sure if it's actually a good idea.

Anyway, here is a small change that:
1) makes sure that converting back to `Array` retains given static shape,
2) changes `Array(::SizedArray)` so that it makes a copy (since `Array(::Array)` also makes one).

### DIFF
--- a/src/SizedArray.jl
+++ b/src/SizedArray.jl
@@ -16,6 +16,14 @@ struct SizedArray{S <: Tuple, T, N, M} <: StaticArray{S, T, N}
         if length(a) != tuple_prod(S)
             error("Dimensions $(size(a)) don't match static size $S")
         end
+        if size(a) != size_to_tuple(S)
+            Base.depwarn("Construction of `SizedArray` with an `Array` of a different
+                size is deprecated. If you need this functionality report it at
+                https://github.com/JuliaArrays/StaticArrays.jl/pull/666 .
+                Calling `sa = reshape(a::Array, s::Size)` will actually reshape
+                array `a` in the future and converting `sa` back to `Array` will
+                return an `Array` of shape `s`.", :SizedArray)
+        end
         new{S,T,N,M}(a)
     end
 
@@ -53,13 +61,13 @@ end
 @inline convert(::Type{SA}, sa::SA) where {SA<:SizedArray} = sa
 
 # Back to Array (unfortunately need both convert and construct to overide other methods)
-@inline Array(sa::SizedArray{S}) where {S} = Array(reshape(sa.data, size_to_tuple(S)))
-@inline Array{T}(sa::SizedArray{S,T}) where {T,S} = Array(reshape(sa.data, size_to_tuple(S)))
-@inline Array{T,N}(sa::SizedArray{S,T,N}) where {T,S,N} = Array(reshape(sa.data, size_to_tuple(S)))
+@inline Array(sa::SizedArray{S}) where {S} = sa.data
+@inline Array{T}(sa::SizedArray{S,T}) where {T,S} = sa.data
+@inline Array{T,N}(sa::SizedArray{S,T,N}) where {T,S,N} = sa.data
 
-@inline convert(::Type{Array}, sa::SizedArray{S}) where {S} = convert(Array, reshape(sa.data, size_to_tuple(S)))
-@inline convert(::Type{Array{T}}, sa::SizedArray{S,T}) where {T,S} = convert(Array{T}, reshape(sa.data, size_to_tuple(S)))
-@inline convert(::Type{Array{T,N}}, sa::SizedArray{S,T,N}) where {T,S,N} = convert(Array{T,N}, reshape(sa.data, size_to_tuple(S)))
+@inline convert(::Type{Array}, sa::SizedArray{S}) where {S} = sa.data
+@inline convert(::Type{Array{T}}, sa::SizedArray{S,T}) where {T,S} = sa.data
+@inline convert(::Type{Array{T,N}}, sa::SizedArray{S,T,N}) where {T,S,N} = sa.data
 
 @propagate_inbounds getindex(a::SizedArray, i::Int) = getindex(a.data, i)
 @propagate_inbounds setindex!(a::SizedArray, v, i::Int) = setindex!(a.data, v, i)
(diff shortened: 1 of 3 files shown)


src/SizedArray.jl
```diff
@@ -53,13 +53,13 @@ end
...
+@inline Array(sa::SizedArray{S}) where {S} = Array(reshape(sa.data, size_to_tuple(S)))
+@inline Array{T}(sa::SizedArray{S,T}) where {T,S} = Array(reshape(sa.data, size_to_tuple(S)))
+@inline Array{T,N}(sa::SizedArray{S,T,N}) where {T,S,N} = Array(reshape(sa.data, size_to_tuple(S)))
 
-@inline convert(::Type{Array}, sa::SizedArray) = sa.data
-@inline convert(::Type{Array{T}}, sa::SizedArray{S,T}) where {T,S} = sa.data
-@inline convert(::Type{Array{T,N}}, sa::SizedArray{S,T,N}) where {T,S,N} = sa.data
+@inline convert(::Type{Array}, sa::SizedArray{S}) where {S} = Array(reshape(sa.data, size_to_tuple(S)))
+@inline convert(::Type{Array{T}}, sa::SizedArray{S,T}) where {T,S} = Array(reshape(sa.data, size_to_tuple(S)))
+@inline convert(::Type{Array{T,N}}, sa::SizedArray{S,T,N}) where {T,S,N} = Array(reshape(sa.data, size_to_tuple(S)))
```
c42f: It seems that `convert` doesn't copy in `Base`:

```julia
julia> a = [1 2]
1×2 Array{Int64,2}:
 1  2

julia> b = convert(Array, a)
1×2 Array{Int64,2}:
 1  2

julia> b[1] = 0
0

julia> a
1×2 Array{Int64,2}:
 0  2
```

mateuszbaran:
> I'm less sure about `convert`; it looks like `Base` doesn't necessarily make a copy here, so we would be justified in only unwrapping. But the fact that `SizedArray` is a weird "wrap plus reshape" container (rather than just a wrapper which preserves shape) is kind of unfortunate and forces us to return a `reshape` for consistency.

`reshape` doesn't copy when there is no need to actually reshape so doing `reshape` + `convert` in `convert` should be reasonable. I'll make a change.

> But the fact that `SizedArray` is a weird "wrap plus reshape" container (rather than just a wrapper which preserves shape) is kind of unfortunate and forces us to return a `reshape` for consistency.

I'm actually sort of using this feature in `Manifolds.jl` (through `SizedAbstractArray`) for reshaped views. It will be replaced with `SSubArray` from `HybridArrays.jl` when I get to it though. When I was playing with this it didn't really work as generally as `StaticArrays.jl` seems to allow. Essentially undef initialization appears impossible (or unreasonable) unless `M=N` or `M=1` (and IIRC at the moment only the `M=N` case works in `StaticArrays`). I really think all other cases should be disallowed.

mateuszbaran:
OK, that sounds reasonable. I doubt anyone actually uses this "functionality" intentionally.

src/SizedArray.jl
```diff
@@ -16,6 +16,14 @@ struct SizedArray{S <: Tuple, T, N, M} <: StaticArray{S, T, N}
         if length(a) != tuple_prod(S)
             error("Dimensions $(size(a)) don't match static size $S")
         end
+        if size(a) != size_to_tuple(S)
+            Base.depwarn("Construction of `SizedArray` with an `Array` of a different
+                size is deprecated. If you need this functionality report it at
+                https://github.com/JuliaArrays/StaticArrays.jl/pull/666 .
+                Calling `sa = reshape(a::Array, s::Size)` will actually reshape
+                array `a` in the future and converting `sa` back to `Array` will
+                return an `Array` of shape `s`.", :SizedArray)
```
c42f: Oh sorry! This isn't quite what I meant. I meant that we should make this depwarn in the conversions and constructors (just a tweak to what you had already).

Having said that... this is an clearer alternative to resolve these issues by clarifying the design rather than hacking around the `SizedArray`-might-also-reshape design that we have already.

I do think it might be better.  Any thoughts @andyferris?

src/SizedArray.jl
```diff
@@ -53,11 +61,11 @@ end
 @inline convert(::Type{SA}, sa::SA) where {SA<:SizedArray} = sa
 
 # Back to Array (unfortunately need both convert and construct to overide other methods)
-@inline Array(sa::SizedArray) = sa.data
+@inline Array(sa::SizedArray{S}) where {S} = sa.data
```
c42f: Should copy though?  Also don't need the `where {S}` now.

mateuszbaran: OK, I'll keep this change then. It should copy and `Array{T}` should also change the element type, right? But not reshape for now.

src/SizedArray.jl
mateuszbaran: Yes, I think keeping the old behavior + depwarn for the next version and then changing it in the following one to something reasonable would be least surprising.

src/SizedArray.jl
mateuszbaran: I forgot that these constructors don't change element type, nevermind.

I've changed them so that they make copies. Is this the right set of changes for the next version of `StaticArrays.jl`?

c42f: > should also change the element type, right

Hmm, good question actually. I think we'll fail to do the right thing for `convert`, because for non-matching eltype we will dispatch to the fallback `convert(::Array, ::StaticArray)`? I guess this has been a bug for a while.

c42f:
This is a subtle and tricky deprecation. I'd appreciate any thoughts from @andyferris who originally implemented `SizedArray`.

Andy — to summarize, the way that `SizedArray` unwraps when converting back to `Array` is kind of inconsistent. It doesn't preserve the shape for example (the constructors also don't create a copy which is also inconsistent).

So... what to do. One way forward (currently implemented here) is to remove the arbitrary `reshape` support from `SizedArray` so that unwrapping is more straightforward. I think this would clarify the design in a nice way but it would remove some functionality.

(Side note: I just noticed that our `reshape` for `MArray` is arguably broken because it doesn't share the underlying data. Ugh.)

andyferris:
I'm totally OK with ditching the ability to reshape an array. That was actually an over-ambitious attempt to step towards hybrid arrays at a time where I was also worried about reshaping and/or nesting things. I probably should have deleted it a while ago...

PR #669, opened 2019-10-02
RFC: Deprecate calling of Size to get SizedArray

c42f:
This feature makes it hard to find code which actually uses SizedArray (eg see https://github.com/JuliaArrays/StaticArrays.jl/pull/666#pullrequestreview-295372179), and it blesses `SizedArray` in an odd fashion by providing a special syntax for constructing it using the trait, which is completely different from the way that other `StaticArray` subtypes are constructed.

So I feel like it's an oddity which we should deprecate, perhaps.

If we do this, it's definitely breaking! But I also kind of suspect that a relatively small number of users actually use `SizedArray`. That's quite hard to assess though with a code search for the reasons mentioned...

One thing I'm not so sure about here is "what if you have `s::Size`, and you want to make a `SizedArray` which is that size"?  The natural thing to do would be to use `similar_type(a, s)`, but that currently produces an `SArray`.  For example, consider `reshape`:

```
# The following is the new version in this PR
reshape(a::Array, ::Size{S}) where {S} = SizedArray{Tuple{S...}}(a)
# vs the arguably-more-natural
reshape(a::Array, s::Size) = similar_type(SizedArray, s)(a)
```

Unfortunately I don't think we can really change `similar_type` to return a `SizedArray` for efficiency reasons. So maybe the current solution is ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ similar_type(m3) == SArray{Tuple{3,3},Int64,2,9}
 Size(m3) === Size(3,3)
 
 # A standard Array can be wrapped into a SizedArray
-m4 = Size(3,3)(rand(3,3))
+m4 = SizedMatrix{3,3}(rand(3,3))
 inv(m4) # Take advantage of specialized fast methods
 
 # reshape() uses Size() or types to specify size:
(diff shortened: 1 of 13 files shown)


c42f:
Thanks for having a look. Let's do this then.

c42f:
The `freeze` idea is good I think. It provides a way forward for temporary mutable arrays without needing to resolve the general and difficult design problem of "mutating immutables" and all the difficulties that implies (the problem of bypassing inner constructors etc).

PR #671, opened 2019-10-16
Avoid calling promote_tuple_eltype in generated function to fix issue

andreasnoack:
with promotion of user defined types. This is a minimal version of #670. I still think it would be beneficial to avoid generated functions when they aren't needed but it it's not urgent. In contrast this current issue is causing issues when using `StaticArrays` with `ForwardDiff` so it would be useful to get fixed immediately. I'd appreciate if you could make a bugfix release after merging this.


c42f:
Sounds fine to me. We are overdue for a new release but it will contain breaking changes so we'd need to backport this to release immediately. Easy enough, just some work.

PR #672, opened 2019-10-16
Release 0.11.1 backports (do not merge)

c42f:
This is a PR to run CI for a quick backported bugfix release as requested by @andreasnoack in #671. I've added that patch plus a few other extremely simple seemingly low risk fixes. We're overdue for a release 0.12.0 containing several features and a few deprecations but that will have to wait a little longer.

(Perhaps there's a better way to run CI across the various platforms?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ notifications:
 branches:
   only:
     - master
+    - /release-.*/
     - /^v[0-9]+\.[0-9]+\.[0-9]+$/ # version tags for Documenter.jl, see # 298
 # matrix:
 #   allow_failures:
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "0.11.0"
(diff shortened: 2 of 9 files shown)


c42f:
@andreasnoack if this passes CI, feel free to create a new release 0.11.1 from this branch using JuliaRegistrator (I may not have time to do so tomorrow).

c42f:
@JuliaRegistrator register branch=release-0.11

fredrikekre:
> Perhaps there's a better way to run CI across the various platforms?

Add `/^release-.*$/` to this list: https://github.com/JuliaArrays/StaticArrays.jl/blob/d1c6288290e8dad29a2376ef3eb4dad89c1bf02d/.travis.yml#L17-L19

c42f:
> Add `/^release-.*$/` to this list

Thanks! I already did this in the last commit on this branch but we should add that patch to master as well (actually your version is more precise so we should use that).

I was more looking for a place to aggregate the CI results from appveyor and travis here at github and it seemed like a PR might be the way to do that.

c42f:
Seems to have gone through ok.

PR #677, opened 2019-10-22
Enable CI with julia-1.3

c42f:
I just noticed this was missing. Hopefully there will be no problems given we pass tests on nightly...


c42f:
Wow that Appveyor run was slow. Let's lighten it up a bit and restrict 1.1 and 1.2 testing to travis.

PR #679, opened 2019-10-24
Remove allocations with `norm(x, p)`

### LINKED ISSUE #678: norm(x::SVector, Inf) allocates
```julia
julia> using StaticArrays, LinearAlgebra, BenchmarkTools

julia> x = @SVector rand(10);

julia> # allocates:
       @benchmark norm($x, Inf)
BenchmarkTools.Trial: 
  memory estimate:  128 bytes
  allocs estimate:  3
  --------------
  minimum time:     43.023 ns (0.00% GC)
  median time:      44.243 ns (0.00% GC)
  mean time:        49.218 ns (6.71% GC)
  maximum time:     1.751 μs (93.92% GC)
  --------------
  samples:          10000
  evals/sample:     990

julia> # does not allocate
       @benchmark maximum(abs.($x))
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     0.015 ns (0.00% GC)
  median time:      0.017 ns (0.00% GC)
  mean time:        0.018 ns (0.00% GC)
  maximum time:     0.151 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

```

c42f:
The allcations were probably due to lack of proper inlining of the
`init=...` keyword, but we're not reducing empty vectors here (there's a
special case for that) so we shouldn't need `init` at all.

Should fix #678

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -267,13 +267,13 @@ _norm_p0(x) = x == 0 ? zero(x) : one(x)
     return quote
         $(Expr(:meta, :inline))
         if p == Inf
-            return mapreduce(abs, max, a; init=$(zero(real(eltype(a)))))
+            return mapreduce(abs, max, a)
         elseif p == 1
             @inbounds return $expr_p1
         elseif p == 2
             return norm(a)
         elseif p == 0
-            return mapreduce(_norm_p0, +, a; init=$(zero(real(eltype(a)))))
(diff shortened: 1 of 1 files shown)


lassepe:
I just tested this and it fixes #678 for me. LGTM

c42f:
Thanks for checking! Let's merge it then.

Issue #661, opened 2019-09-23
eigen in non-diagonal matrices


ronisbr:
Hi!

I am getting this error when trying to use `eigen` on non-diagonal matrices:

```julia
julia> Imc = @SMatrix([446. 100. 0.; 0. 1149. 100.; 0. 0. 886])
3×3 SArray{Tuple{3,3},Float64,2,9} with indices SOneTo(3)×SOneTo(3):
 446.0   100.0    0.0
   0.0  1149.0  100.0
   0.0     0.0  886.0

julia> eigen(Imc)
ERROR: Only hermitian matrices are diagonalizable by *StaticArrays*. Non-Hermitian matrices should be converted to `Array` first.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] _eig at /Users/ronan.arraes/.julia/packages/StaticArrays/3KEjZ/src/eigen.jl:120 [inlined]
 [3] #eigen#492 at /Users/ronan.arraes/.julia/packages/StaticArrays/3KEjZ/src/eigen.jl:386 [inlined]
 [4] eigen(::SArray{Tuple{3,3},Float64,2,9}) at /Users/ronan.arraes/.julia/packages/StaticArrays/3KEjZ/src/eigen.jl:386
 [5] top-level scope at REPL[43]:1
```

It works with diagonal matrices:

```julia
julia> Imc = @SMatrix([446. 0. 0.; 0. 1149. 0.; 0. 0. 886])
3×3 SArray{Tuple{3,3},Float64,2,9} with indices SOneTo(3)×SOneTo(3):
 446.0     0.0    0.0
   0.0  1149.0    0.0
   0.0     0.0  886.0

julia> eigen(Imc)
Eigen{Float64,Float64,SArray{Tuple{3,3},Float64,2,9},SArray{Tuple{3},Float64,1,3}}
eigenvalues:
3-element SArray{Tuple{3},Float64,1,3} with indices SOneTo(3):
  446.0
  886.0
 1149.0
eigenvectors:
3×3 SArray{Tuple{3,3},Float64,2,9} with indices SOneTo(3)×SOneTo(3):
 1.0  0.0  0.0
 0.0  0.0  1.0
 0.0  1.0  0.0
```

c42f:
[...]

PR #683, opened 2019-10-30
Handle non-hermitian matrices in eigen

CarpeNecopinum:
Planned to help with #661 

For the (previously unhandled) non-hermitian branch of eigen, this PR falls back to `eigen(Array(A),...)`. To add in type stability, the type of the expected eigenvalues is added as a parameter that defaults to `real(eltype(A))` (which is consistent with the old implementation). Passing a complex type here allows the user to use `eigen(StaticMatrix)` even if they are not completely positive that the matrix will always be symmetric/hermitian.

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -123,26 +123,33 @@ end
     return SVector{s[1], T}(vals)
 end
 
+# Utility to rewrap `Eigen` of normal `Array` into an Eigen containing `SArray`.
+@inline function _make_static(s::Size, E::Eigen{T,V}) where {T,V}
+    Eigen(similar_type(SVector, V, Size(s[1]))(E.values),
+          similar_type(SMatrix, T, s)(E.vectors))
+end
 
 @inline function _eig(s::Size, A::T, permute, scale) where {T <: StaticMatrix}
-    # For the non-hermitian branch, fall back to LinearAlgebra
     if ishermitian(A)
-        eivals, eivecs = _eig(s, Hermitian(A), permute, scale)
-        return eivals, eivecs
+        return _eig(s, Hermitian(A), permute, scale)
     else
-        eivals, eivecs = eigen(Array(A); permute = permute, scale = scale)
-        return SVector{s[1]}(eivals), SMatrix{s[1],s[2]}(eivecs)
+        # For the non-hermitian branch fall back to LinearAlgebra eigen().
+        # Eigenvalues could be real or complex so a Union of concrete types is
+        # inferred. Having _make_static a separate function allows inference to
+        # preserve the union of concrete types:
+        #   Union{E{A,B},E{C,D}} -> Union{E{SA,SB},E{SC,SD}}
+        _make_static(s, eigen(Array(A); permute = permute, scale = scale))
     end
 end
 
 @inline function _eig(s::Size, A::LinearAlgebra.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
     E = eigen(Hermitian(Array(parent(A))))
-    return (SVector{s[1], T}(E.values), SMatrix{s[1], s[2], eltype(A)}(E.vectors))
+    return Eigen(SVector{s[1], T}(E.values), SMatrix{s[1], s[2], eltype(A)}(E.vectors))
 end
 
 
 @inline function _eig(::Size{(1,1)}, A::LinearAlgebra.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
-    @inbounds return (SVector{1,T}((real(A[1]),)), SMatrix{1,1,eltype(A)}(I))
+    @inbounds return Eigen(SVector{1,T}((real(A[1]),)), SMatrix{1,1,eltype(A)}(I))
 end
 
 @inline function _eig(::Size{(2,2)}, A::LinearAlgebra.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
@@ -171,7 +178,7 @@ end
             vecs = @SMatrix [ v11  v21 ;
                               v12  v22 ]
 
-            return (vals, vecs)
+            return Eigen(vals, vecs)
         end
     else # A.uplo == 'L'
         if !iszero(a[2]) # A is not diagonal
@@ -195,7 +202,7 @@ end
             vecs = @SMatrix [ v11  v21 ;
                               v12  v22 ]
 
-            return (vals,vecs)
+            return Eigen(vals,vecs)
         end
     end
 
@@ -211,7 +218,7 @@ end
         vecs = @SMatrix [convert(TA, 0) convert(TA, 1);
                          convert(TA, 1) convert(TA, 0)]
     end
-    return (vals,vecs)
+    return Eigen(vals,vecs)
 end
 
 # A small part of the code in the following method was inspired by works of David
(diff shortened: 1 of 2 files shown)


CarpeNecopinum:
Gonna add in some tests (hopefully soon). Though I'm running into a weird issue where 
```
eigen(@SMatrix randn(4,4))
```
throws `InexactError` (as it should), but
```
@test_throws Exception eigen(@SMatrix randn(4,4))
```
fails because `No exception thrown`. Which I'm kinda scratching head about...

c42f:
Thanks for the patch! That error *is* weird. `@test_throws Exception eigen(@SMatrix randn(4,4))` works for me on this branch.

Maybe a better way to fix this would be to follow the Base API regardless of the type instability. The type instability would presumably only be of `Union{A,B}` form where `A` and `B` are concrete types. That's often not so bad due to the union splitting optimization so I think it would be best to pursue that and check with a benchmark before considering breaking from the `Base` API. Generally this package tries hard to keep with `Base` if possible.

To recover the current behavior we should check that users are able to use `eigen(Hermitian(A))` - that should give full type stability.

CarpeNecopinum:
Ah yeah, you're right!
```
Base.return_types(eigen, Tuple{Array{Float64,2}})
1-element Array{Any,1}:
 Union{Eigen{Complex{Float64},Complex{Float64},Array{Complex{Float64},2},Array{Complex{Float64},1}}, Eigen{Float64,Float64,Array{Float64,2},Array{Float64,1}}}
```

Funnily enough, I had some `Eigen{_A,_B,_C,_D} where {_A,_B,_C,_D}` as inferred return type at some point when I was working on it, which is why there is quite the amount of type-juggling in my implementation.

CarpeNecopinum:
Gonna add some tests regarding type stability and handling of non-hermitian matrices soon™

src/eigen.jl
```diff
@@ -382,7 +383,7 @@ end
     return (SVector(eig1, eig2, eig3), hcat(eigvec1, eigvec2, eigvec3))
 end
 
-@inline function eigen(A::StaticMatrix; permute::Bool=true, scale::Bool=true)
+@inline function eigen(A::StaticMatrix; permute::Bool=true, scale::Bool=true) where {T}
```
c42f: This `where {T}` looks unnecessary now.

test/eigen.jl
```diff
@@ -213,8 +213,8 @@ using StaticArrays, Test, LinearAlgebra
         @test eigvals(m) ≈ sort(m_d)
         @test eigvals(Hermitian(m)) ≈ sort(m_d)
 
-        # not Hermitian
-        @test_throws Exception eigen(@SMatrix randn(4,4))
+        # not Hermitian - we can handle these now
+        # @test_throws Exception eigen(@SMatrix randn(4,4))
```
c42f: Feel free to delete if it's obsolete! Though I guess we should have a test for `eigen(Hermitan(SMatrix(...))))` doing "something sensible"

src/eigen.jl
CarpeNecopinum: Oh right, good catch

test/eigen.jl
CarpeNecopinum: Done, now testing the handling of rotation matrices (which are the first type of non-hermitian matrices with testable properties I came up with) and type stability on hermitian matrices.

CarpeNecopinum:
Should be all done and ready to merge from my side now...

c42f:
Thanks this looks pretty neat but it seems inference is more confused by the StaticArrays code than by the Base implementation:

```
julia> Core.Compiler.return_type(eigen, (SArray{Tuple{3,3},Float64,2,9},))
Eigen

julia> Core.Compiler.return_type(eigen, (Matrix{Float64},))
Union{Eigen{Complex{Float64},Complex{Float64},Array{Complex{Float64},2},Array{Complex{Float64},1}}, Eigen{Float64,Float64,Array{Float64,2},Array{Float64,1}}}
```

Is it possible to convince type inference to return a `Union` like the `Base` version? Maybe some well-placed type asserts would do the trick?

c42f:
It looks like the type instability here could be due to the keyword args, or could be due to the Tuple of unions (rather than Union of Tuples) here:

```julia
julia> Core.Compiler.return_type(StaticArrays._eig, (Size{(3,3)}, SArray{Tuple{3,3},Float64,2,9}, Bool, Bool))
Tuple{Union{SArray{Tuple{3},Complex{Float64},1,3}, SArray{Tuple{3},Float64,1,3}},
      Union{SArray{Tuple{3,3},Complex{Float64},2,9}, SArray{Tuple{3,3},Float64,2,9}}}
```

c42f:
Ok, the problem is that when transforming the `Eigen` containing a normal `Matrix` and `Vector` into an `Eigen` containing `StaticArray`s, the compiler can't infer the output type unless we encourage it with a specific trick.  I've got what I think is a functioning version but it needs a little cleanup... will push later.

c42f:
Ok, I've pushed some changes to this branch. Have a look and let me know what you think.

Unfortunately, while these changes *do* help type inference, it seems that the compiler isn't able to make full use of this information right now (see https://github.com/JuliaLang/julia/issues/33895)

CarpeNecopinum:
That's a pretty good find! Code looks good, too. I just find it slightly weird that we de- and reconstruct an `Eigen` multiple times, but since it's just an immutable with 2 fields, it shouldn't be a problem.

Could've sworn though that I checked the inferred type of `eigen(::StaticMatrix)` in my branch and it turned up as a Union as it did for the LinearAlgebra variant. Hope I'll get around to testing that again soon...

c42f:
>  I just find it slightly weird that we de- and reconstruct an `Eigen` multiple times

Yes, well we can't really avoid this if we are going to unwrap the result of `eigen(::Matrix)` and rewrap it using static arrays. (Granted we could use the internals of `LinearAlgebra.LAPACK` to avoid the rewrapping, but I wouldn't go that far unless there's some evidence that it really helps.)

CarpeNecopinum:
> Could've sworn though that I checked the inferred type of eigen(::StaticMatrix) in my branch and it turned up as a Union as it did for the LinearAlgebra variant. Hope I'll get around to testing that again soon...

Okay, looks like I remembered that wrong... but I have found an (imho) easier fix:

```
@inline function _eig(s::Size, A::T, permute, scale) where {T <: StaticMatrix}
    # For the non-hermitian branch, fall back to LinearAlgebra
    svec(x::Array{T}) = SVector{s[1],T}(x)
    smat(x::Array{T}) = SMatrix{s[1],s[2]}(x)

    if ishermitian(A)
        _eig(s, Hermitian(A), permute, scale)
        return eivals, eivecs
    else
        eivals, eivecs = eigen(Array(A); permute = permute, scale = scale)
        return svec(eivals), smat(eivecs)
    end
end
```
Just having the `smat` and `svec` functions here does the trick already.

CarpeNecopinum:
> Could've sworn though that I checked the inferred type of eigen(::StaticMatrix) in my branch and it turned up as a Union as it did for the LinearAlgebra variant. Hope I'll get around to testing that again soon...

Appears I was mistaken... can't seem to come up with an easier fix for now, either.

c42f:
Over at #693 we've found a tool for testing small unions of concrete types so I added one extra test. Other than that I plan to merge this once CI has passed again.

From the discussion at https://github.com/JuliaLang/julia/issues/33895, it seems possible to make use of the type information in the small union, but it's only really effective if the result of `eigen` is passed to another function.

PR #686, opened 2019-11-07
A tool for testing whether codegen fully inlines a call tree.

c42f:
Based on #685 (CC @tkf). I'll rebase once some solution to that issue is merged.

The implementation here is certainly hacky as it does a very simple minded inspection of the LLVM code as a string. But it should be much better to have some hacky tests for code quality rather than inspecting the code by eye!

### DIFF
--- a/test/initializers.jl
+++ b/test/initializers.jl
@@ -18,6 +18,7 @@ SA_test_hcat(x,T) = SA{T}[1 x x]
 @test @inferred(SA_test_hcat(2.0)) === SMatrix{1,3,Float64}((1,2,2))
 @test @inferred(SA_test_hcat(2,Float32))   === SMatrix{1,3,Float32}((1,2,2))
 
+# hvcat needs to be in a function for the row argument to constant propagate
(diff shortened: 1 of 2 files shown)


tkf:
FYI, I fond this kind of test sometimes interfere with `--code-coverage=none` so I wrote https://github.com/tkf/IRTest.jl/blob/master/src/IRTest.jl and use it like this https://github.com/tkf/BangBang.jl/blob/68b59c03471713c1bdfb4f8e6aff4ce3b63b5b5a/test/test_ir.jl#L4

c42f:
Thanks, that's good to know. Are you doing similar tests on the output of `@code_llvm`? I don't know much about LLVM IR, so this test macro is pretty hacky.

tkf:
Mine is pretty hacky, tool. Most of the time I look for substrings like `fmul <4 x double>` or the actual return value if I want to make sure const-prop works.

c42f:
This seemed to pass tests so I guess I didn't need IRTest.jl for this particular use case. I think I'll merge this right away because it improves the testing for #685 somewhat.

PR #689, opened 2019-11-15
Move concatenation into abstractarray.jl

c42f:
This is where it belongs; it's array functionality rather than linear algebra.

Also return a throw expression for size mismatches rather than throwing
from function generators.


c42f:
This one is (almost) entirely boring code movement. CI says yes, so I think I'll just merge it.

PR #694, opened 2019-11-27
Improve 2x2 eigen

tfgast:
Using the fact that the eigenvectors must be orthonormal, v21 and v22, can be computed from v21 and v11. In addition to being more performant, this is actually more robust in some cases, for instance with [1.0 -9.465257061381386e-20; -9.465257061381386e-20 1.0], the previous code would return duplicated eigenvectors.


c42f:
Looks great to me, thanks!

Could you add a test or tests which cover the cases which the previous code got wrong? Is this characterized by the case where the off diagonal elements are less than `eps(eltype(A))`?

Would this fix also allow us to remove the special case for where `A` is detected to be diagonal? Just looking at the code that seems unlikely to be numerically reliable... and ultimate the root cause of this bug.

tfgast:
I've improved the handling in the case of underflow as well. I wasn't able to remove the special case for diagonal matrices, as we still need to avoid dividing by zero.  I've also added some more tests.

test/eigen.jl
```diff
@@ -75,24 +75,36 @@ using StaticArrays, Test, LinearAlgebra
...
+    @testset "2×2 degenerate cases" for (i, j) in ((1 , 1), (1, 2), (2, 1)), uplo in (:U, :L)
+        fmin = floatmin(Float64)
+        pfmin = prevfloat(fmin)
+        nfmin = nextfloat(fmin)
+        A = SMatrix{2,2,Float64}((i, 0, 0, j))
+        E = eigen(Symmetric(A, uplo))
+        @test eigvecs(E) * SDiagonal(eigvals(E)) * eigvecs(E)' ≈ A
+        A = SMatrix{2,2,Float64}((i, pfmin, pfmin, j))
+        E = eigen(Symmetric(A, uplo))
+        @test eigvecs(E) * SDiagonal(eigvals(E)) * eigvecs(E)' ≈ A
```
c42f: Cool, so can you add some comment about exactly which cases we are straddling here by examining the boundary between `pfmin` vs `nfmin`?  We seem to have have `x^2 == 0` for both of those and `i ± x == 1.0`, `i*j  ± x == i*j` etc...

c42f: I'd love to merge this, but I'm unsure whether these tests are testing the right thing. Do you have some thoughts on the above query?

tfgast: I looked into it again.  You were right, they were testing the wrong thing.  When I tested the right thing it wasn't working properly.  I put some time in to make it work and simplify things, mostly by using the hypot function which already has a lot of work put into making it accurate.

test/eigen.jl
```diff
@@ -75,24 +75,31 @@ using StaticArrays, Test, LinearAlgebra
...
-        m2 = SMatrix{2,2}(m2_a)
-        @test (@inferred_maybe_allow SVector{2,ComplexF64} eigvals(m1, m2)) ≈ eigvals(m1_a, m2_a)
-        @test (@inferred_maybe_allow SVector{2,ComplexF64} eigvals(Symmetric(m1), Symmetric(m2))) ≈ eigvals(Symmetric(m1_a), Symmetric(m2_a))
+    # issue #523, #694
+    zero = 0.0
+    smallest_non_zero = nextfloat(zero)
+    smallest_normal = floatmin(zero)
+    largest_subnormal = prevfloat(smallest_normal)
+    epsilon = eps(1.0)
+    one_p_epsilon = 1.0 + epsilon
```
c42f, line 86: Quibble: Better written `nextfloat(1.0)` — it's the same in this case, but if you'd written `1.0 - epsilon` that's not equal to `prevfloat(1.0)` which is a bit of a gotcha given that the boundary in floating point discretization density lies on the powers of two.

test/eigen.jl
```diff
@@ -75,24 +75,31 @@ using StaticArrays, Test, LinearAlgebra
...
+    degenerate = (zero, -1, 1, smallest_non_zero, smallest_normal, largest_subnormal, epsilon, one_p_epsilon, -one_p_epsilon)
+    @testset "2×2 degenerate cases" for (i, j, k) in zip(degenerate,degenerate,degenerate), uplo in (:U, :L)
```
c42f, line 88: I don't think this `zip` does what you want because `i,j,k` are always the same? Though I'm not sure what you want!

Perhaps you meant to use `for (i, j, k) in Iterators.product(degenerate,degenerate,degenerate)`?

tfgast: You're totally right, that's what I meant

test/eigen.jl
tfgast: :thumbsup:

andyferris:
Where is this one at?

> Also, how does this change as a whole impact performance?

Do we have an answer to this one?

thchr:
I tested the performance impact of this just now: as far as I can see, there is a speed-up for `Hermitian{ComplexF64, ...}` matrices but a slow-down for `Hermitian{Float64, ...}` matrices (not completely clear why to me):

#### Hermitian{ComplexF64, ...} matrices
*Master:*
```jl
julia> @benchmark StaticArrays._eig(Size{(2,2)}(), A, true, true) setup=begin
           a = @SMatrix rand(ComplexF64,2,2)
           A = Hermitian(a + a')
       end
BenchmarkTools.Trial: 10000 samples with 940 evaluations.
 Range (min … max):  101.915 ns … 324.681 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     102.766 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   105.722 ns ±   9.921 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅█▁          ▁▃▁  ▁▁                                          ▁
  ███▆▅▄▆▆▆▅▅▅▆███▇█████▇▇▇▆▅▅▅▅▅▅▄▅▅▅▄▅▄▄▅▄▅▄▄▄▅▅▄▄▄▅▅▅▄▄▄▄▃▄▃ █
  102 ns        Histogram: log(frequency) by time        149 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

*PR:*
```jl
julia> @benchmark StaticArrays._eig(Size{(2,2)}(), A, true, true) setup=begin
           a = @SMatrix rand(ComplexF64,2,2)
           A = Hermitian(a + a')
       end
BenchmarkTools.Trial: 10000 samples with 979 evaluations.
 Range (min … max):  64.147 ns … 324.821 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     66.394 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   67.999 ns ±   7.887 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄▆▆█▂            ▁▂▁▁  ▁                                     ▁
  █████▆▄▅▃▆▅▆▆▆▄▆██████████▇▇▇▇▆▆▅▅▅▅▅▅▄▅▅▅▄▆▄▅▅▄▅▁▄▅▄▅▃▅▅▄▄▄ █
  64.1 ns       Histogram: log(frequency) by time       104 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

#### Hermitian{Float64, ...} matrices

*Master:*
```jl
julia> @benchmark StaticArrays._eig(Size{(2,2)}(), A, true, true) setup=begin
           a = @SMatrix rand(Float64,2,2)
           A = Hermitian(a + a')
       end
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  8.909 ns … 69.670 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     9.009 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   9.230 ns ±  1.770 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▄                                                         ▁
  ██▆▄▄▄▄▁▃▄▅▄▃▃▃▄▁▁▁▄▅▇▇▆▅▄▅▄▄▃▄▄▅▄▄▄▁▃▃▁▃▃▁▁▁▃▁▁▁▁▁▃▁▄▃▃▄▅ █
  8.91 ns      Histogram: log(frequency) by time     18.9 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

*PR:*
```jl
julia> @benchmark StaticArrays._eig(Size{(2,2)}(), A, true, true) setup=begin
           a = @SMatrix rand(Float64,2,2)
           A = Hermitian(a + a')
       end
BenchmarkTools.Trial: 10000 samples with 994 evaluations.
 Range (min … max):  29.879 ns … 204.930 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     31.489 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   32.438 ns ±   5.469 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁▆▃ █▃                                                       ▁
  ███████▆▆▅▄▄▄▅▁▃▄▄▅▆▅▆▅▅▄▁▄▅▇█▇█▇▇▆▇▇▆▆▇▅▇▅▇▇▅▅▅▆▆▆█▇▇▆▅▆▆▅▆ █
  29.9 ns       Histogram: log(frequency) by time      51.6 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

PR #704, opened 2019-12-23
fix broadcasts with only adjoint/transpose SVectors

marius311:
A minor thing I ran into / fixed.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -97,7 +97,7 @@ scalar_getindex(x) = x
 scalar_getindex(x::Ref) = x[]
 
 @generated function _broadcast(f, ::Size{newsize}, s::Tuple{Vararg{Size}}, a...) where newsize
-    first_staticarray = a[findfirst(ai -> ai <: StaticArray, a)]
+    first_staticarray = a[findfirst(ai -> ai <: Union{StaticArray, Transpose{<:Any, <:StaticArray}, Adjoint{<:Any, <:StaticArray}}, a)]
 
     if prod(newsize) == 0
         # Use inference to get eltype in empty case (see also comments in _map)
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -115,6 +115,8 @@ end
         # Issue 382: infinite recursion in Base.Broadcast.broadcast_indices with Adjoint
         @test @inferred(SVector(1,1)' .+ [1, 1]) == [2 2; 2 2]
         @test @inferred(transpose(SVector(1,1)) .+ [1, 1]) == [2 2; 2 2]
+        
+        @test @inferred(v1' + v2') === SVector(2,6)'
     end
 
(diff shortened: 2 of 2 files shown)


mateuszbaran:
Thanks! I have a few comments:
  1. This change sometimes modifies the result type of broadcasts that mix different static arrays (for example `SArray` and `MArray`), for example in expressions like `S' .+ M` vs `S .+ M` where `S` is a `SArray` and `M` is an `MArray`. I don't know which result type is better, but the old behavior can be kept if necessary by only looking for `StaticArray`-like things only when no non-wrapped `StaticArray`s were found.
  2. You can use the `StaticArrayLike` union here instead of your custom one.

src/broadcast.jl
```diff
@@ -97,7 +97,7 @@ scalar_getindex(x) = x
 scalar_getindex(x::Ref) = x[]
 
 @generated function _broadcast(f, ::Size{newsize}, s::Tuple{Vararg{Size}}, a...) where newsize
-    first_staticarray = a[findfirst(ai -> ai <: StaticArray, a)]
+    first_staticarray = a[findfirst(ai -> ai <: Union{StaticArray, Transpose{<:Any, <:StaticArray}, Adjoint{<:Any, <:StaticArray}}, a)]
```
tkoolen: Should we be using [`StaticArrayLike`](https://github.com/JuliaArrays/StaticArrays.jl/blob/103e9d4dabfdb64ccb378c8539d15136dde964de/src/StaticArrays.jl#L79-L98) here?

c42f: Yes ideally we'd add `StaticArrayLike` support to this method. But we'd need to make sure that `similar_type(first_staticarray, ...)` makes sense in all the cases that covers, and I think we'd need to add to `BroadcastStyle()` as well to intercept those wrapped arrays correctly.

marius311: Yea, the Union I chose was specifically to have it be consistent with the definitions for [BroadcastStyle](https://github.com/JuliaArrays/StaticArrays.jl/blob/b4f7e5e9bef5204f30275a86f61baba804fb1b64/src/broadcast.jl#L13-L15), so currently I think its at least consistent, if not exhaustive. 

Up to you guys if you want to merge this as-is or not.

c42f:
I added a minor change to be clearer about what the test is actually testing.

I do think we should use `StaticArrayLike` more consistently but for now I'm happy to merge this as a bugfix rather than an enhancement (pending CI passing again of course).

c42f:
>     1. This change sometimes modifies the result type of broadcasts

That's true... though I think it's easiest just to go with this change as I it follows the rule of "always use the leftmost StaticArray (Like)" with `similar_type`.

In general I feel that this existing leftmost rule isn't satisfying and we should probably do something better. But for now it has the advantage of being simple.

PR #706, opened 2019-12-28
Views of zero-dimensional arrays

### LINKED ISSUE #705: StackOverflowError when viewing zero-dimensional arrays
Not that you'd ever want to view a zero-dimensional array, but:
```julia
julia> A=zeros();

julia> view(A)
0-dimensional view(::Array{Float64,0}) with eltype Float64:
0.0

julia> using StaticArrays

julia> view(A)
ERROR: StackOverflowError:
Stacktrace:
 [1] SubArray(::Array{Float64,0}, ::Tuple{}) at /home/colinxs/.julia/packages/StaticArrays/1g9bq/src/indexing.jl:381 (repeats 80000 times)
```
This happens with 0.12.1 and #master. 

Looks like https://github.com/JuliaArrays/StaticArrays.jl/blob/103e9d4dabfdb64ccb378c8539d15136dde964de/src/indexing.jl#L381 is the culprit.

```julia
Julia Version 1.3.0
Commit 46ce4d7933 (2019-11-26 06:09 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

mateuszbaran:
Fixes #705 .

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -378,9 +378,6 @@ Base.unsafe_view(A::AbstractArray, i1::StaticIndexing, indices::StaticIndexing..
 # Additionally, in some cases the SubArray constructor may be called directly
 # instead of unsafe_view so we need this method too (Base._maybe_reindex
 # is a good example)
-Base.SubArray(A::AbstractArray, indices::NTuple{<:Any,StaticIndexing}) = Base.SubArray(A, map(unwrap, indices))
-# this method is defined to prevent infinite recursion when viewing a
-# zero-dimensional array (see issue #705)
-function Base.SubArray(A::AbstractArray, indices::Tuple{})
-    return invoke(Base.SubArray, Tuple{AbstractArray, Tuple}, A, indices)
-end
+# the tuple indices has to have at least one element to prevent infinite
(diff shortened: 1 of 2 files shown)


src/indexing.jl
```diff
@@ -379,3 +379,8 @@ Base.unsafe_view(A::AbstractArray, i1::StaticIndexing, indices::StaticIndexing..
 # instead of unsafe_view so we need this method too (Base._maybe_reindex
 # is a good example)
 Base.SubArray(A::AbstractArray, indices::NTuple{<:Any,StaticIndexing}) = Base.SubArray(A, map(unwrap, indices))
```
martinholters: Isn't it sufficient to restrict the second argument to non-empty tuples here? I.e. `Tuple{StaticIndexing,Vararg{StaticIndexing}}`?

mateuszbaran: I didn't know I can put `Vararg` there, good idea.

mateuszbaran: I've updated the PR, it should be good now.

PR #711, opened 2020-01-03
Fix MArray constructor docstring

simonbyrne:
`MArray{S,T,L}()` throws an error.


c42f:
Thanks Simon.

Hmm, it looks like we document the `L` parameter incorrectly (#712)

Issue #696, opened 2019-12-04
Eigendecomposition of 3x3 symmetric Float64 matrices is not very accurate


mateuszbaran:
While working on the SPD manifold for Manifolds.jl we discovered some numerical accuracy problems that appear when using `MMatrix` but not `Matrix`. I've been able to identify one source of this problem, that is eigendecomposition of symmetric 3x3 matrices with Float64 elements. It's static implementation is much less accurate then what `LinearAlgebra` has. Example:

```julia
julia> using LinearAlgebra, StaticArrays

julia> x = @MArray [1.0 0.0 0.0; 0.0 1.7500000000000002 -0.4330127018922193; 0.0 -0.4330127018922193 1.25]
3×3 MArray{Tuple{3,3},Float64,2,9} with indices SOneTo(3)×SOneTo(3):
 1.0   0.0        0.0     
 0.0   1.75      -0.433013
 0.0  -0.433013   1.25    

julia> meig = eigen(Symmetric(x))
Eigen{Float64,Float64,SArray{Tuple{3,3},Float64,2,9},SArray{Tuple{3},Float64,1,3}}
eigenvalues:
3-element SArray{Tuple{3},Float64,1,3} with indices SOneTo(3):
 0.9999999929755253
 1.0000000070244748
 2.0               
eigenvectors:
3×3 SArray{Tuple{3,3},Float64,2,9} with indices SOneTo(3)×SOneTo(3):
 1.0  -0.0        0.0     
 0.0   0.5        0.866025
 0.0   0.866025  -0.5     

julia> eig = eigen(Symmetric(Array(x)))
Eigen{Float64,Float64,Array{Float64,2},Array{Float64,1}}
eigenvalues:
3-element Array{Float64,1}:
 1.0               
 1.0000000000000009
 2.0               
eigenvectors:
3×3 Array{Float64,2}:
 1.0  0.0        0.0     
 0.0  0.5        0.866025
 0.0  0.866025  -0.5     

julia> norm(x - meig.vectors * Diagonal(meig.values) * meig.vectors')
9.934107421073401e-9

julia> norm(x - eig.vectors * Diagonal(eig.values) * eig.vectors')
7.406111315403549e-16
```
For now we just disabled problematic tests but I think this issue should at least be known to other users of this functionality.

c42f:
[...]

PR #714, opened 2020-01-13
WIP: Improve precision of 3x3 eigen analysis

CarpeNecopinum:
Addresses #696 by implementing the iterative method presented in https://www.geometrictools.com/Documentation/RobustEigenSymmetric3x3.pdf for 3x3 symmetric eigen analysis.

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -226,7 +226,7 @@ end
 # under the Boost Software License, Version 1.0 (included at the end of this file)
 # TODO ensure right-handedness of the eigenvalue matrix
 # TODO extend the method to complex hermitian
-@inline function StaticArrays._eig(::Size{(3,3)}, A::LinearAlgebra.HermOrSym{T}, permute, scale) where {T <: Real}
+@inline function _eig(::Size{(3,3)}, A::LinearAlgebra.HermOrSym{T}, permute, scale) where {T <: Real}
     function converged(aggressive, bdiag0, bdiag1, bsuper)
         if aggressive
             bsuper == 0
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -221,187 +221,159 @@ end
     return Eigen(vals,vecs)
 end
 
-# A small part of the code in the following method was inspired by works of David
-# Eberly, Geometric Tools LLC, in code released under the Boost Software
-# License (included at the end of this file).
+# Port of https://www.geometrictools.com/GTEngine/Include/Mathematics/GteSymmetricEigensolver3x3.h
+# released by David Eberly, Geometric Tools, Redmond WA 98052
+# under the Boost Software License, Version 1.0 (included at the end of this file)
+# TODO ensure right-handedness of the eigenvalue matrix
 # TODO extend the method to complex hermitian
-@inline function _eig(::Size{(3,3)}, A::LinearAlgebra.HermOrSym{T}, permute, scale) where {T <: Real}
-    S = arithmetic_closure(T)
-    Sreal = real(S)
-
-    @inbounds a11 = convert(Sreal, A.data[1])
-    @inbounds a22 = convert(Sreal, A.data[5])
-    @inbounds a33 = convert(Sreal, A.data[9])
-    if A.uplo == 'U'
-        @inbounds a12 = convert(S, A.data[4])
-        @inbounds a13 = convert(S, A.data[7])
-        @inbounds a23 = convert(S, A.data[8])
-    else
-        @inbounds a12 = conj(convert(S, A.data[2]))
-        @inbounds a13 = conj(convert(S, A.data[3]))
-        @inbounds a23 = conj(convert(S, A.data[6]))
-    end
-
-    p1 = abs2(a12) + abs2(a13) + abs2(a23)
-    if (p1 == 0)
-        # Matrix is diagonal
-        v1 = SVector(one(S),  zero(S), zero(S))
-        v2 = SVector(zero(S), one(S),  zero(S))
-        v3 = SVector(zero(S), zero(S), one(S) )
-
-        if a11 < a22
-            if a22 < a33
-                return Eigen(SVector((a11, a22, a33)), hcat(v1,v2,v3))
-            elseif a33 < a11
-                return Eigen(SVector((a33, a11, a22)), hcat(v3,v1,v2))
-            else
-                return Eigen(SVector((a11, a33, a22)), hcat(v1,v3,v2))
-            end
-        else #a22 < a11
-            if a11 < a33
-                return Eigen(SVector((a22, a11, a33)), hcat(v2,v1,v3))
-            elseif a33 < a22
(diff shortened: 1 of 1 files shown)


CarpeNecopinum:
One weird issue with this when I ran the tests was that the type stability tests on 3x3 Hermitians fail again. Trying in the REPL, however, they appear to be type stable:

```jl
julia> Base.return_types(eigen, (Hermitian{Float32,SArray{Tuple{3,3},Float32,2,9}},))
1-element Array{Any,1}:
 Eigen{Float32,Float32,SArray{Tuple{3,3},Float32,2,9},SArray{Tuple{3},Float32,1,3}}

julia> @inferred eigen(Hermitian(x))
... no error ...
```

Performance is also really good (~33% speedup of my normal estimation over the current, less precise method), so I feel like the `@inferred` macro is over-sensitive here...

src/eigen.jl
```diff
@@ -221,188 +221,160 @@ end
...
-            end
-        else #a22 < a11
-            if a11 < a33
-                return Eigen(SVector((a22, a11, a33)), hcat(v2,v1,v3))
-            elseif a33 < a22
-                return Eigen(SVector((a33, a22, a11)), hcat(v3,v2,v1))
-            else
-                return Eigen(SVector((a22, a33, a11)), hcat(v2,v3,v1))
-            end
+@inline function StaticArrays._eig(::Size{(3,3)}, A::LinearAlgebra.HermOrSym{T}, permute, scale) where {T <: Real}
```
c42f:
```suggestion
@inline function _eig(::Size{(3,3)}, A::LinearAlgebra.HermOrSym{T}, permute, scale) where {T <: Real}
```

src/eigen.jl
```diff
@@ -221,188 +221,160 @@ end
...
+    function converged(aggressive, bdiag0, bdiag1, bsuper)
+        if aggressive
+            bsuper == 0
+        else
+            summe = abs(bdiag0) + abs(bdiag1)
```
c42f: `summe`?

c42f:
> One weird issue with this when I ran the tests was that the type stability tests on 3x3 Hermitians fail again

This is a (confusing!) consequence of Julia's scoping rules. The issue here is that closures like `_sortperm3` share bindings with their parent function, so the variable `perm` is the *same variable* in the scope of `_eig` and in the closure `_sortperm3`.

To fix this you can move `_sortperm3` outside the `_eig` so that it's not considered as a closure. Alternatively, annotate `perm` as `local perm` within `_sortperm3`, or use a different variable name.

src/eigen.jl
```diff
@@ -221,188 +221,160 @@ end
...
+    term1 = c * a01 + s * a11
+    b00 = c * term0 + s * term1
+    b01 = s * term0 - c * term1
+    term0 = s * a00 - c * a01
+    term1 = s * a01 - c * a11
+    b11 = s * term0 - c * term1
+    b12 = s * a02 - c * a12
+    b22 = a22
+
+    max_iteration = 2 * (1 + 53 + 1021) # 2 * (1 + mantissa_bits - minimum exponent) for Float64
```
mateuszbaran: You can use something like `precision(eltype(A)) - exponent(floatmin(eltype(A)))` to make it more generic.

src/eigen.jl
CarpeNecopinum: Fixed with a7f709d

src/eigen.jl
CarpeNecopinum: Was `sum` in the original and I'm always kinda anxious of shadowing Base verbs.
Fixed by a474f02.

src/eigen.jl
CarpeNecopinum: Thanks, implemented as 48fff92.

src/eigen.jl
```diff
@@ -221,188 +221,197 @@ end
...
-    # rows of A - eig2*I. However, we need to "solve" the remaining 2x2 subspace
-    # problem in case the cross products are identically or nearly zero
-
-    # The remaing 2x2 subspace is:
-    @inbounds if abs(eigvec1[1]) < abs(eigvec1[2]) # safe to set one component to zero, depending on this
-        orthogonal1 = SVector(-eigvec1[3], zero(S), eigvec1[1]) / sqrt(abs2(eigvec1[1]) + abs2(eigvec1[3]))
-    else
-        orthogonal1 = SVector(zero(S), eigvec1[3], -eigvec1[2]) / sqrt(abs2(eigvec1[2]) + abs2(eigvec1[3]))
+    function _sortperm3(v)
+        local perm = SVector(1,2,3)
```
c42f: This seems ok as a workaround... though do note that having inner functions inside a complex implementation like `_eig` without the intent to have them capture any state always runs the risk that you will accidentally share bindings. For example, if `max_abs` in `get_cos_sin` happened to be a variable name in `_eig` you would again hit the same problem.

If you don't intend to use these as closures (ie, you don't want to capture any state from the outer function) it's a fragile solution to add `local` on a piecemeal basis, overall I think it would be better to just move these helper functions outside of `_eig`. Put another way: the placement of `_sortperm3` is *quite* significant in ways other than its visibility outside of `_eig`: I believe the closure-or-not distinction should be considered as a priority vs the other side effect of placement — whether `_sortperm3` is visible at the module level.

c42f: (I do realize I suggested adding `local` as one of the options to fix this :grimacing:. But on further consideration I think I was wrong about that. Unless I've missed something and you really do capture some state here.)

CarpeNecopinum: Yeah, it's a thing I wasn't too happy about either... was more of a "getting this fixed and testable as fast as possible" type of deal.

Would be cool to have some kind of way to annotate an inner function as "non-capturing" / "only here because it isn't used elsewhere".

andyferris: I agree with Chris's general advice here (it's a shame scoping can become a footgun...) - but I wouldn't block this PR on that. Are there any other outstanding issues?

src/eigen.jl
```diff
@@ -221,188 +221,197 @@ end
     return Eigen(vals,vecs)
 end
 
-# A small part of the code in the following method was inspired by works of David
-# Eberly, Geometric Tools LLC, in code released under the Boost Software
-# License (included at the end of this file).
+# Port of https://www.geometrictools.com/GTEngine/Include/Mathematics/GteSymmetricEigensolver3x3.h
+# released by David Eberly, Geometric Tools, Redmond WA 98052
+# under the Boost Software License, Version 1.0 (included at the end of this file)
```
c42f:
```suggestion
# under the Boost Software License, Version 1.0 (included at the end of this file)
# The original documentation states
# (see https://www.geometrictools.com/Documentation/RobustEigenSymmetric3x3.pdf )
```

src/eigen.jl
c42f: Yeah, it's not blocked on that; more just that I'm loosing track of things.

PR #720, opened 2020-01-29
add undef tags to docs for MMatrix and MVector

simonbyrne:
I fixed MArray in #711, but missed these.


c42f:
looks good to me, thanks :+1:

PR #722, opened 2020-01-31
Update SDiagonal.jl

githubtomtom:
as per Chris Foster, the original syntax was historical for Julia 0.5


c42f:
Thanks, I think this is good. I'm pretty sure this is completely equivalent, though I do wish the julia documentation about constructors was clearer.

PR #725, opened 2020-01-31
Update SizedArray.jl

githubtomtom:
historical syntax for Julia 0.5


c42f:
Hi thanks for these. If you do more cleanup could I ask that all related work be submitted as a single pull request?

For example, the four pull requests I just merged were all fixing the same thing in different files. It makes for easier review and less redundant history if you do it in a single PR. Thanks!

githubtomtom:
sure. thanks.

PR #728, opened 2020-02-10
replace plain error with DimensionMismatch

matbesancon:


matbesancon:
Some boundcheck propagation and tagging were also missing

src/MArray.jl
```diff
@@ -134,7 +134,7 @@ macro MArray(ex)
             s2s = map(i -> ((isa(ex.args[i], Expr) && ex.args[i].head == :row) ? length(ex.args[i].args) : 1), 1:s1)
             s2 = minimum(s2s)
             if maximum(s2s) != s2
-                error("Rows must be of matching lengths")
+                throw(DimensionMismatch("Rows must be of matching lengths"))
```
c42f: I feel like the malformed AST which would be passed in this case is more an `ArgumentError` to the macro than a `DimensionMismatch`.  The same could be said of the other macros.

matbesancon:
just updated the errors in macros

c42f:
Thanks, I think there's a few you missed.

Eh, we have too many of these macros :-/

c42f:
There, I went and made those consistent as well. Let's check CI again.

matbesancon:
Thanks! CI is passing, but I doubt the errors were covered

PR #731, opened 2020-02-14
Default implementation of similar_type for FieldArray

### LINKED ISSUE #729: Summing two `FieldVectors` produces an `SArray`
Example:
```
julia> struct Point3D <: StaticArrays.FieldVector{3, Float64}
           x::Float64
           y::Float64
           z::Float64
       end

julia> a = Point3D(rand(3)); b = Point3D(rand(3)); a+b
3-element StaticArrays.SArray{Tuple{3},Float64,1,3} with indices SOneTo(3):
 0.31581254718043894
 0.7451396786025122
 0.4810013436998537
```

Obviously, it's not hard to change it back into a `Point3D`.  

I'm wondering if it's reasonable and feasible for this to be automatic?  

For example, is it as simple as defining `+` for FieldVector`s?  Are there good reasons not to do this?

(Presumably this definition would propagate to `sum`, `mean`, etc.)

c42f:
This should free users from defining similar_type in many cases, though
they'll still need to do so when their type is parametric on the eltype.
(There's no general way for us to know how to reparameterize such
user-defined types.)

Fixes #729

### DIFF
--- a/src/FieldArray.jl
+++ b/src/FieldArray.jl
@@ -6,7 +6,15 @@ will automatically define `getindex` and `setindex!` appropriately. An immutable
 `FieldArray` will be as performant as an `SArray` of similar length and element type,
 while a mutable `FieldArray` will behave similarly to an `MArray`.
 
-For example:
+Note that you must define the fields of any `FieldArray` subtype in column major order. If you
+want to use an alternative ordering you will need to pay special attention in providing your
+own definitions of `getindex`, `setindex!` and tuple conversion.
+
+If you define a `FieldArray` which is parametric on the element type you should
+consider defining `similar_type` as in the `FieldVector` example.
+
+
+# Example
 
     struct Stiffness <: FieldArray{Tuple{2,2,2,2}, Float64, 4}
         xxxx::Float64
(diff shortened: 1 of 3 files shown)


kmsquire:
Thanks!  I'm wondering if I'm not understanding you with regard to parametric types, because this seems to work on your branch:

```julia
julia> struct Point{T} <: FieldVector{3, T}
           x::T
           y::T
           z::T
       end

julia> p = Point([1,2,3]); p+p
3-element Point{Int64} with indices SOneTo(3):
 2
 4
 6
```

c42f:
You're right that does work because the eltype doesn't change.

The following doesn't work though because the eltype changes:

```julia
julia> struct V{T} <: FieldVector{3, T}
           x::T
           y::T
           z::T
       end

julia> V(1,2,3) ./ V(1,2,3)
3-element SArray{Tuple{3},Float64,1,3} with indices SOneTo(3):
 1.0
 1.0
 1.0
```

c42f:
I mean, this isn't exactly harmful, but turning into `SVector` might not be what you want either.

kmsquire:
Got it, thanks!

PR #733, opened 2020-02-17
Update SArray.jl

githubtomtom:
constructors for undef

### DIFF
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -50,6 +50,28 @@ end
     end
 end
 
+"""
+    SArray{S, T, N, L}(undef)
+    SArray{S, T, N}(undef)
+    SArray{S, T}(undef)
+"""
+@inline SArray{S, T, N, L}(::UndefInitializer) where {S <: Tuple, T, N, L} = 
+    SArray{S, T, N, L}(Array{T}(undef, size_to_tuple(S) ) )
+
+@generated function SArray{S, T, N}(::UndefInitializer) where {S <: Tuple, T, N}
+    return quote
+        @_inline_meta
+        SArray{S, T, N, $(tuple_prod(S)) }(undef)
+    end
+end
+
+@generated function SArray{S, T}(::UndefInitializer) where {S <: Tuple, T}
+    return quote
+        @_inline_meta
+        SArray{S, T, N, $(tuple_length(S)), $(tuple_prod(S)) }(undef)
+    end
+end
+
 @inline SArray(a::StaticArray) = SArray{size_tuple(Size(a))}(Tuple(a))
 
 ####################
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -54,6 +54,11 @@ end
     SArray{S, T, N, L}(undef)
     SArray{S, T, N}(undef)
     SArray{S, T}(undef)
+    SArray{S}(undef)
+    SVector{S, T}(undef)
+    SVector{S}(undef)
+    SMatrix{S1, S2, T}(undef)
+    SMatrix{S1, S2}(undef)
 """
 @inline SArray{S, T, N, L}(::UndefInitializer) where {S <: Tuple, T, N, L} = 
     SArray{S, T, N, L}(Array{T}(undef, size_to_tuple(S) ) )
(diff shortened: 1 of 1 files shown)


src/SArray.jl
```diff
@@ -50,6 +50,28 @@ end
     end
 end
 
+"""
+    SArray{S, T, N, L}(undef)
+    SArray{S, T, N}(undef)
+    SArray{S, T}(undef)
+"""
+@inline SArray{S, T, N, L}(::UndefInitializer) where {S <: Tuple, T, N, L} = 
+    SArray{S, T, N, L}(Array{T}(undef, size_to_tuple(S) ) )
```
c42f: This allocates a normal `Array`, and then discards it which would be a nasty performance trap. If we must supply this constructor, you should use an `MArray` here.

githubtomtom:
it’s useful. For example I’m creating a mutable struct that has a field of SVector:

mutable struct A
    x::SVector{3, Int}
end

A() = A(SVector{3, Int}(undef)

a = A()
now I can have a type A object that has undefined content.

the content could be modified later like:
a.x = SVector{3, Int}(v)

githubtomtom:
if SVector{3, Int}(undef)  is not defined, instead I have to define like:
A() = A(SVector{3, Int}(Vector{Int}(undef, 3) )

so I define the constructor for undef for convenience cases like this.

src/SArray.jl
githubtomtom: I would like to use SVector as one of the fields in a **mutable** struct. So, the whole mutable object has to be constructed in the first place (by creating a `undef SVector`), and then we could mutate the field of it.

c42f:
Well the use of `undef` should generally be avoided unless you really need it for performance. In your example I think it would be preferable to just use `zeros`:

```
mutable struct A
    x::SVector{3, Int}
end

A() = A(zeros(SVector{3, Int}))
```

This will certainly be far more efficient than using `A(SVector{3, Int}(Vector{Int}(undef, 3) )` and more predictable as well.

For mutable structs you can also express undefined initialization using an inner constructor, for example

```julia
mutable struct A
    x::SVector{3, Int}
    A() = new()
end
```

though again, just using `zeros` would likely be preferable here.

githubtomtom:
ok thanks.

githubtomtom:
just curious as `SVector{3, Int}()` is not defined, what actually happens when we call `new()` ?

c42f:
> just curious as `SVector{3, Int}()` is not defined, what actually happens when we call `new()` ?

The result will be whatever happened to be in memory at the time (no constructor is called).  Immutables like `SVector{3,Int}` are stored inline in the parent structure, so their value will just be some uninitialized memory.

If you were storing a mutable struct like `MVector` the wrapper `A` would have a pointer to the `MVector` which will be set to zeros (displayed as `#undef`).

As noted above, I'd suggest avoiding `new()` unless you really need it. Initializing `x` to something well defined will be both easier to work with and quite cheap at runtime.

c42f:
As of right now, I feel we should not support undef initialization of `SArray` because there's usually a better alternative. However I'm always willing to be persuaded otherwise if you can show a clear use case where the alternatives don't work.

Having said that, we could treat this as a documentation issue in two ways: (1) define `SArray(undef)` to throw a nice self-explanatory error (2) Add a section to the documentation which discusses this. Possibly in the `SArray` constructor docs, possibly a section in the docs which discusses patterns for working with immutable arrays.

githubtomtom:
ok thanks.

c42f:
I'll close this for the reasons discussed.

I hope one of the alternatives worked for you?

andyferris:
Agree - use zeros or leave it undefined in your inner constructor to _your_ struct.

(Also word to the wise: leaving references undef will have performance impact in some cases in the future - for ease of implementation, the upcoming inlining of structs with immutable references will be disabled if it’s fields aren’t all defined in all inner constructors. Not sure if this should impact the design of MArray or anything, but it’s interesting)

c42f:
> Also word to the wise: leaving references undef will have performance impact in some cases in the future - for ease of implementation, the upcoming inlining of structs with immutable references will be disabled if it’s fields aren’t all defined in all inner constructors. Not sure if this should impact the design of MArray or anything, but it’s interesting

Interesting, thanks.

BTW, using `new()` without initializing all fields *already* has a performance impact for structs containing non-inline fields: in codegen, the compiler needs to insert an extra test and branch for whether the field is null. If it can prove this never happens I believe this test is elided.

PR #736, opened 2020-02-18
Minimal benchmark setup

tkf:
See #737. Please comment on how we organize `benchmark/*` here.

### DIFF
--- a/benchmark/bench_qr.jl
+++ b/benchmark/bench_qr.jl
@@ -1,0 +1,18 @@
+module BenchQR
+
+using BenchmarkTools
+using LinearAlgebra
+using StaticArrays
+
+suite = BenchmarkGroup()
+
+for K = 1:22
+    a = rand(SMatrix{K,K,Float64,K*K})
+    m = Matrix(a)
+    s = suite["S=$K"] = BenchmarkGroup()
+    s["SMatrix"] = @benchmarkable qr($a)
+    s["Matrix"] = @benchmarkable qr($m)
+end
+
+end  # module
+BenchQR.suite
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,0 +1,8 @@
+using BenchmarkTools
+SUITE = BenchmarkGroup()
+for file in sort(readdir(@__DIR__))
+    if startswith(file, "bench_") && endswith(file, ".jl")
+        SUITE[file[length("bench_") + 1:end - length(".jl")]] =
+            include(file)
+    end
+end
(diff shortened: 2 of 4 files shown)


benchmark/benchmarks.jl
```diff
@@ -0,0 +1,12 @@
+# Each file of the form "bench_$(name).jl" in this directory is `include`d and
+# its last statement is assumed to be a `BenchmarkGroup`.  This group is added
+# to the top-level group `SUITE` with the `$name` extracted from the file name.
```
tkf: What do you think about this strategy? I use this for all my packages and it works well for me.  This is useful for running each file separately like this: `run(include("benchmark/bench_qr.jl"))`.

c42f: This sounds completely fine to me!

---

PS: My fantasy interface might be more akin to `@testset` where the groups are assembled "magically" using dynamic scope

```julia
@benchmark_group "all" begin
    @benchmark_group "QR $K" for K=1:22
        a = rand(SMatrix{K,K,Float64,K*K})
        m = Matrix(a)
        @benchmarkable qr($a)
        @benchmarkable qr($m)
     end
     # other groups  or includes()...
end
```

but unlike `Test`, to allow running these selectively. But no doubt that's an issue for BenchmarkTools ;-)

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,12 @@
...
+
+using BenchmarkTools
+SUITE = BenchmarkGroup()
```
KristofferC: could be `const` I guess?

benchmark/bench_qr.jl
```diff
@@ -0,0 +1,18 @@
+module BenchQR
+
+using BenchmarkTools
+using LinearAlgebra
+using StaticArrays
+
+suite = BenchmarkGroup()
```
KristofferC: could be `const` I guess?

benchmark/benchmarks.jl
KristofferC: > PS: My fantasy interface might be more akin to @testset where the groups are assembled "magically" using dynamic scope

Ref https://github.com/JuliaCI/PkgBenchmark.jl/issues/9#issuecomment-257952067.

benchmark/benchmarks.jl
tkf: OK, initially I didn't want `const` to avoid seeing the redefinition warning but it's not a good motivation anymore since I use modules.

benchmark/benchmarks.jl
c42f: Thanks for the crossref :+1:

I do strongly disagree with Jarret on that one. I think dynamic scoping (not global scope mind you!) can be fairly clear and non-magical for this kind of use as long as it's clearly documented.

PR #737, opened 2020-02-18
Setup BenchmarkCI

tkf:
This PR sets up [BenchmarkCI.jl](https://github.com/tkf/BenchmarkCI.jl).  This is what happens if/after this PR is merged:

* Every push to the PR triggers a CI job that runs benchmarks defined in `benchmark/benchmarks.jl` for the head of PR branch and the master.
* The benchmark result is posted as comment like this https://github.com/JuliaArrays/StaticArrays.jl/pull/737#issuecomment-587964525 for every push, when the PR is created from the branch _from this repository_.
* If the PR is created from a fork, CI job is run but the comment will not be posted.
    * As a simple solution, the contributors can create a PR in their GitHub repository from the same branch. This will run the benchmark CI on their PR.

Since the benchmarks are run against `master` by the default setting, I tweaked the setting to use the branch `tkf/pkgbenchmark` (#736).  This way, we can check that it is working. We should revert this commit once #736 is merged.

Questions:

* Do we want to run benchmarks for every push on every PR?
* Do we want to get a comment for every push on every PR?
* Is this how we want to organize `benchmark/*`?  Let's discuss this in #736.

For example, if we want to just run benchmarks (without spammed by the comments), we can just remove these lines:

https://github.com/JuliaArrays/StaticArrays.jl/blob/84ca6a42f1c86ea49ff35998df21dca976d926f1/.github/workflows/benchmark.yml#L18-L21

TODOs before merging this PR:

- [x] merge #736
- [x] remove `baseline = "origin/tkf/pkgbenchmark"` c157922939fdc013baedb2c3c06f33219d3e8ba1


tkf:
@c42f I think this is good to go, if you want to use the default setup.

andyferris:
Cool! I for one think this is a great idea.

If I had the energy I'd move the existing CI to github workflows too!

c42f:
* Do we want to run benchmarks for every push on every PR?
* Do we want to get a comment for every push on every PR?

I feel the comments might get a bit excessive, but maybe we try it and see?

tkf:
OK. I'll merge this as-is and open a PR for an alternative setup to suppress comments. You can merge that setup if you want to avoid spamming from BenchmarkCI.

Issue #734, opened 2020-02-18
Matrix Multiplication for large SizedArray


bjack205:
I get the following error when multiplying two SizedArrays
`conversion to pointer not defined for SizedArray{Tuple{20,20},Float64,2,2}`

This occurs when the the `_mul` method passes the arrays to BLAS, and the error happens on the `ccal` in `mul_blas!`.

c42f:
I haven't been able to reproduce this yet. The following works for me:

```julia
julia> a = rand(SizedMatrix{20,20});

julia> a*a
20×20 SArray{Tuple{20,20},Float64,2,400} with indices SOneTo(20)×SOneTo(20):
...
```

(You'll notice it produces an `SArray`, however that is changed in the latest master branch.)

Could you provide a complete self contained example of the failure?

bjack205:
Sorry, my bad, I forgot to mention I'm using the in-place matrix multiply
```julia
julia> a = rand(SizedMatrix{20,20});
julia> b = rand(SizedMatrix{20,20});
julia> c = zeros(SizedMatrix{20,20});
julia> mul!(c,a,b)
ERROR: conversion to pointer not defined for SizedArray{Tuple{20,20},Float64,2,2}
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] unsafe_convert(::Type{Ptr{Float64}}, ::SizedArray{Tuple{20,20},Float64,2,2}) at ./pointer.jl:67
 [3] macro expansion at /home/brian/.julia/packages/StaticArrays/1g9bq/src/matrix_multiply.jl:307 [inlined]
 [4] mul_blas!(::Size{(20, 20)}, ::SizedArray{Tuple{20,20},Float64,2,2}, ::Size{(20, 20)}, ::Size{(20, 20)}, ::SizedArray{Tuple{20,20},Float64,2,2}, ::SizedArray{Tuple{20,20},Float64,2,2}) at /home/brian/.julia/packages/StaticArrays/1g9bq/src/matrix_multiply.jl:289
 [5] macro expansion at /home/brian/.julia/packages/StaticArrays/1g9bq/src/matrix_multiply.jl:266 [inlined]
 [6] _mul! at /home/brian/.julia/packages/StaticArrays/1g9bq/src/matrix_multiply.jl:248 [inlined]
[...]

PR #738, opened 2020-02-19
Add in-place multiply-add

bjack205:
This adds support for the new 5-argument in-place matrix multiplication in Julia v1.3 (#357 ).
We can probably consolidate the multiply-add and with the matrix multiply methods, but I've left them separate for now.

I've also added a quick "hack" to avoid an invalid pointer error (#734) when doing in-place multiplication with SizedArrays large enough that they get passed to BLAS.


andyferris:
Certainly seems like a pretty good idea - did you do any benchmarking comparing performance to naive multiply-then-add?

bjack205:
Yep, here's a few benchmarks

Test script:
```julia
julia> α,β = 2.0, 1.0
julia> N = 3 
julia> a = @MMatrix rand(N,N)
julia> b = @MMatrix rand(N,N)
julia> c = @MMatrix rand(N,N)
julia> A,B,C = Array(a), Array(b), Array(c)
julia> @btime mul!($C, $A, $B, 1.0, 1.0);
julia> @btime mul!($c, $a, $b, 1.0, 1.0);
```
And the results:
### N = 3
| Version | @btime result |
| ---------- | ------------------ |
| Base Array | 16.470 ns (0 allocations: 0 bytes) |
|  Master branch | 11.839 ns (0 allocations: 0 bytes) |
| This version | 8.847 ns (0 allocations: 0 bytes) |

### N = 5
| Version | @btime result |
| ---------- | ------------------ |
| Base Array | 121.191 ns (0 allocations: 0 bytes) |
| Master branch | 182.495 ns (4 allocations: 224 bytes) |
| This version | 22.558 ns (0 allocations: 0 bytes) |

### N = 20
| Version | @btime result |
| ---------- | ------------------ |
| Base Array | 581.967 ns (0 allocations: 0 bytes) |
| Master branch | 4.618 μs (4 allocations: 224 bytes) |
| This version | 575.928 ns (0 allocations: 0 bytes) |

src/matrix_multiply.jl
```diff
@@ -284,8 +289,28 @@ end
...
+        Sa::Size{sa}, Sb::Size{sb},
+        a::Transpose{<:Any, <:SizedArray}, b::SizedArray{<:Any,T},
+        α::Real=one(T), β::Real=zero(T)) where {s,sa,sb, T <: BlasFloat} =
+    BLAS.gemm!('T','N', α, a.parent.data, b.data, β , c.data)
+
+@inline mul_blas!(Sc::Size{s}, c::StaticMatrix{<:Any, <:Any, T},
+        Sa::Size{sa}, Sb::Size{sb},
+        a::Transpose{<:Any, <:StaticMatrix}, b::StaticMatrix{<:Any, <:Any, T},
+        α::Real=one(T), β::Real=zero(T)) where {s,sa,sb, T <: BlasFloat} =
+    BLAS.gemm!('T','N', α, a.parent, b, β , c)
```
c42f: This rather a lot of duplication which will be hard to maintain.

I think all you need is a utility function to "get the data pointer and transposed-ness" out of `a`, `b`, `c`? Then you can collapse all these down again into one function.

src/matrix_multiply_add.jl
```diff
@@ -0,0 +1,263 @@
...
+    _mul!(Size(dest), dest, Size(A), Size(B), A, B, α, β)
+@inline mul!(dest::StaticVecOrMat, A::StaticVector, B::Transpose{<:Any, <:StaticVector}, α::Real, β::Real) =
+    _mul!(Size(dest), dest, Size(A), Size(B), A, B, α, β)
+@inline mul!(dest::StaticVecOrMat, A::StaticVector, B::Adjoint{<:Any, <:StaticVector}, α::Real, β::Real) =
+    _mul!(Size(dest), dest, Size(A), Size(B), A, B, α, β)
+
+@inline mul!(dest::StaticVector, A::Transpose{<:Any, <:StaticMatrix}, B::StaticVector, α::Real, β::Real) =
+    _tmul!(Size(dest), dest, Size(A.parent), Size(B), A.parent, B, α, β)
+@inline mul!(dest::StaticMatrix, A::Transpose{<:Any, <:StaticMatrix}, B::StaticMatrix, α::Real, β::Real) =
+    _mul!(Size(dest), dest, Size(A), Size(B), A, B, α, β)
```
c42f: Is it possible to arrange this dispatch a bit more cleanly rather than having quite so much code duplication here?

For example, you can define a type alias similar to `StaticMatrixLike` which captures both `StaticMatrix`, `Transpose` (and possibly `Adjoint`).

Then a single overload of `mul!` may be sufficient.

bjack205:
Thanks for the feedback! This is my first PR so I certainly have a lot to learn!

I've rehashed my solution and made it shorter and hopefully more along the lines of what you're looking for. I've merged the 5-argument `mul!` with the 3-argument `mul!` and unified the way transposes are handled. 

The transposes are particularly tricky since they tend to allocate at the slightest provocation. My solution avoids this allocation with a function barrier and instead caches a transpose flag in with the type (see `TSize` in `matrix_multiply_add.jl`). This ended up with a fairly straight-forward solution, IMHO.

The benchmarks are pretty much the same as they were before. 
The three-argument multiplication doesn't incur any performance hit for small matrices, but for the medium-sized "chunked" version, my implementation runs a little slower.

bjack205:
I missed a test case that happens to be breaking pretty much everything. All my tests passed in magic numbers for alpha and beta, but when they're variables the `MulAddMul` allocates (due to type instability). I've tried a few things but can't figure out how to recover from the type instability. I'm not familiar enough with Julia to really understand why my implementation doesn't work while the one in `LinearAlgebra/matmul.jl` works, since they seem fairly similar to me. I'm open to any ideas on how to fix this, because otherwise this attempt isn't worth much.

bjack205:
Okay, I think it should be pretty close with the most recent commit. I hard-coded the `MulAddMul` for both the 5-argument and 3-argument versions to avoid type instability and resultant allocations. 

NOTE THAT USING `MulAddMul` IS A BREAKING CHANGE, since this is a feature in Julia v1.3. I don't see this as a problem, since the 5-argument `mul!` was just made available in v1.3, anyways.

I also fixed the performance decline for the chunked unrolling, so the new version has improved or equivalent performance on all sizes. 

I guess I should also give some motivation for why I think these methods are important. In my package `TrajectoryOptimization.jl` I've been battling getting the core algorithm to not allocate any memory. Normal `SArrays` work great, but we found that compilation times went through the roof when we tried larger problems. To my knowledge (which, admittedly, isn't all that much when it comes to Julia), `SizedArray` with `mul!` is the only solution that works nearly as fast as `SArrays` at small sizes (thanks to loop unrolling!), scales well to large arrays (thanks to BLAS), and doesn't incur any memory allocations.

Newest benchmarks:
test script:
```julia
function benchmark_matmul(N1,N2,ArrayType=MArray)
    if ArrayType <: MArray
        Mat = MMatrix
        Vec = MVector
    elseif ArrayType <: SizedArray
        Mat = SizedMatrix
        Vec = SizedVector
    end
    α,β = 1.0, 1.0
    A = rand(Mat{N1,N2})
    B = rand(Mat{N2,N2})
    C = rand(Mat{N1,N2})
    println("C = A*B")
    @btime mul!($C,$A,$B)
    println("C += A*B")
    @btime mul!($C,$A,$B,$α,$β)
    println("B = A'C")
    @btime mul!($B,Transpose($A),$C)
    println("B += A'C")
    @btime mul!($B,Transpose($A),$C,$α,$β)
end

```
### `MArray`
| Size | branch | `C = A*B` | `C += A*B` | `B = A'B` | `B += A'B` |
| ---- | ---- | ---- | ---- | --- | ---- | 
| 3x3 | master | 6.6 ns (0 allocs) | 193.4 ns (3 allocs) | 9.2 ns (0 allocs) | 193.8 ns (3 allocs) |
| 3x3 | new      | 6.7 ns (0 allocs) | 8.1 ns (0 allocs)    | 6.6 ns (0 allocs) | 8.1 ns (0 allocs) |
| 10x10 | master | 123.3 ns (0 allocs) | 889.5 ns (7 allocs) | 489.6 ns (0 allocs) | 683.9 ns (3 allocs) |
| 10x10 | new      | 124.1 ns (0 allocs) | 134.1 ns (0 allocs) | 196.9 ns (0 allocs) | 207.2 ns (3 allocs) |
| 20x20 | master | 613.7 ns (0 allocs) | 4.83 μs (7 allocs) | 3.76 μs (0 allocs) | 4.21 μs (3 allocs) |
| 20x20 | new      | 613.7 ns (0 allocs) | 580.9 ns (0 allocs) | 613.4 ns (0 allocs) | 591 ns (3 allocs) |

### `SizedArray`
| Size | branch | `C = A*B` | `C += A*B` | `B = A'B` | `B += A'B` |
| ---- | ---- | ---- | ---- | --- | ---- | 
| 3x3 | master | 10 ns (0 allocs) | 195 ns (3 allocs) | 10 ns (0 allocs) | 194 ns (0 allocs) |
| 3x3 | new | 10 ns (0 allocs) | 11 ns (0 allocs) | 11 ns (0 allocs) | 11 ns (0 allocs) |
| 10x10 | master | 119 (0 allocs) | 905 ns (7 allocs) | 406 ns (0 allocs) | 680 ns (3 allocs) |
| 10x10 | new | 125 ns (0 allocs) | 129 ns (0 allocs) | 195 ns (0 allocs) | 199 ns (0 allocs) |
| 20x20 | master | error | error | error | error |
| 20x20 | new | 617 ns (0 allocs) | 581 ns (0 allocs) | 612 ns (0 allocs) | 576 ns (0 allocs) |

src/matrix_multiply_add.jl
```diff
@@ -0,0 +1,326 @@
+import LinearAlgebra.MulAddMul
```
tkf: `MulAddMul` was meant to be an implementation detail of 5-arg `mul!`.  It's simple piece of code so the safest thing to do is to vendor it here, if you want to use it.  But `MulAddMul` has been causing some troubles in LinearAlgebra so I'm not entirely sure if that's the best solution.  Is just using `alpha = true` and `beta = false` as default not an option?

bjack205: Good idea about creating a copy in StaticArrays, so that we don't break compatibility with previous Julia versions. It's useful since the boolean type parameters are available at compile time so we can conditionally include or exclude terms in the generated code. Although looking at the the benchmarks there isn't much performance benefit to excluding the extra addition and multiplications, so maybe we shouldn't even bother special-casing the extra terms.

bjack205: I removed the dependency on `LinearAlgebra.MulAddMul` so it no longer breaks backwards  compatibility. I replaced it with something similar but less prone to type-instability.

bjack205:
I fixed the backwards-compatibility problem. Are there further changes that need to be made?

src/matrix_multiply_add.jl
```diff
@@ -0,0 +1,352 @@
...
+    C = _get_raw_data(c)
+    BLAS.gemm!(mat_char(tA), mat_char(tB), T(alpha(_add)), A, B, T(beta(_add)), C)
+end
+
+# if C is transposed, transpose the entire expression
+@inline mul_blas!(Sc::TSize{<:Any,true}, c::StaticMatrix, Sa::TSize, Sb::TSize,
+        a::StaticMatrix, b::StaticMatrix, _add::MulAddMul) =
+    mul_blas!(transpose(Sc), c, transpose(Sb), transpose(Sa), b, a, _add)
+
+# TODO: Get this version of mul_blas! working so there's backward-compatibility with older Julia versions
```
c42f, line 285: Is this TODO still needed?

test/matrix_multiply_add.jl
```diff
@@ -0,0 +1,185 @@
...
+    # matrix-transpose × vector
+    mul!(b,At,c)
+    @test b ≈ A'c
+    mul!(b,At,c,2.0,0.0)
+    @test b ≈ 2A'c
+    mul!(b,At,c,1.0,2.0)
+    @test b ≈ 5A'c
+
+    bmark = @benchmark mul!($c,$A,$b) samples=10 evals=10
+    @test minimum(bmark).allocs == 0
```
c42f, line 96: Is it possible to use `@allocated` or does that not work properly? (I know we've had some problems with it lately related to changes in Base.)

c42f:
Oops it looks like I caused a conflict by merging some unrelated test tools. I merged into your branch which will hopefully fix that up.

Generally this is looking quite good now (a lot less scary in terms of amount of code added!)

Great job with the table of benchmarks there's some really impressive improvements there. Could you add the benchmarks to `benchmark` directory so that we retain a copy of them and run them in the future? You can see the results of comparing benchmarks across versions in the github actions "Benchmark" job. There's other examples there of how to hook them up.

test/matrix_multiply_add.jl
```diff
@@ -0,0 +1,185 @@
...
+#     println("C = A*B")
+#     @btime mul!($C,$A,$B)
+#     println("C = A*B + C")
+#     @btime mul!($C,$A,$B,$α,$β)
+#     println("B = A'C")
+#     @btime mul!($B,Transpose($A),$C)
+#     println("B = A'C + B")
+#     @btime mul!($B,Transpose($A),$C,$α,$β)
+# end
+# benchmark_matmul(20,20,SizedArray)
```
c42f, line 26: Let's move this into the `benchmarks` directory so it'll be kept up to date.

src/matrix_multiply_add.jl
bjack205: Nope. Fixed and deleted out the old code.

test/matrix_multiply_add.jl
bjack205: I couldn't get it to not record allocations with `@allocated`. I'm guessing `@benchmark` will be more accurate? I've added commented out lines using `@test_noalloc` below each test, as well as the allocations recorded.

test/matrix_multiply_add.jl
bjack205: Done.

test/matrix_multiply_add.jl
```diff
@@ -0,0 +1,178 @@
...
+using LinearAlgebra
+using BenchmarkTools
+using Test
+
+macro test_noalloc(ex)
+    esc(quote
+        $ex
+        @test(@allocated($ex) == 0)
+    end)
+end
```
c42f, line 11: If you've found `@benchmark` reliable for this, let's just make `test_noalloc` use that instead? Something like:

```
macro test_noalloc(ex)
    quote
        bmark = @benchmark $(esc(ex)) samples=10 evals=10
        @test minimum(bmark).allocs == 0
    end
end
```

You could put this in `test_util` as well and we can use it everywhere instead of `@allocated`? Couple of questions:
* We don't actually want to do a benchmark here so it seems a bit heavy weight. We can put up with that though, unless it slows down the tests a lot?
* Can you reduce the samples and evals?
* Is the `minimum` there required, or will anything other than the maximum do? (Are we measuring the right thing?)

c42f: Ugh, there's something really broken about the way Julia or BenchmarkTools processes the correct-seeming use of `esc` in the nested macro above.

Also I didn't realize quite what `minimum` does; how about this instead:

```julia
macro test_noalloc(ex)
    ex = :(@test minimum(@benchmark($ex, samples=10, evals=10)).allocs == 0)
    ex.args[2] = __source__  # trick: make `@test` aware of the source location of `@test_noalloc`
    esc(ex)
end
```

tkf: `@test_noalloc` macro sounds like a good idea. We can swap `@benchmark` with something else if these tests slow down the whole test suite a lot.  My guess is that it'd be OK as long as you have something like `samples=10 evals=10`.

> Is the `minimum` there required

`bmark` is the `Trial` whose `.allocs` field is already an `Int`:

https://github.com/JuliaCI/BenchmarkTools.jl/blob/cf7cada0b3f41a00c70ef03891ffae29c31f8573/src/trials.jl#L5-L11

It seems that `Base.push!(t::Trial, time, gctime, memory, allocs)` (which is used to accumulate the benchmark trial results) already takes `maximum` over all the trials:

https://github.com/JuliaCI/BenchmarkTools.jl/blob/cf7cada0b3f41a00c70ef03891ffae29c31f8573/src/trials.jl#L29

(Also, `minimum` is the minimum over the wall times. So I don't think we need it anyway.)

bjack205: Your suggested `test_noalloc(ex)` macro didn't work for me? When I use it within a function (not in global scope) it can't find the local arguments.

andyferris: Try put a `$` in front of the local arguments (it's a feature of `@benchmark` to make sure that it captures them properly).

c42f: > Your suggested test_noalloc(ex) macro didn't work for me?

Andy's suggestion seems to work. But oh wow, I had a look at `BenchmarkTools.generate_benchmark_definition`, it's kind of a horror show :grimacing: it calls `Core.eval` in the macro and really seems to go out of its way to subvert the usual macro expansion rules in a weird way.

Maybe it has to be that way for some reason, but it sure is doing some weird stuff.

c42f: If it's too hard to wrap `@benchmark` reliably I'm completely ok with just having it written out by hand in a few places.

I tried looking more into why `@allocated` is failing but it quickly led to a very deep rabbit hole which I'm still trying to climb out of.

c42f:
I went ahead and merged this, we can sort the `@allocated` mysteries out another time.

Thanks for a very nice contribution!

PR #740, opened 2020-02-19
Update convert.jl

githubtomtom:
supports promote() to SArray

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -40,3 +40,22 @@ length_val(a::Type{T}) where {T<:StaticArrayLike} = length_val(Size(T))
         @inbounds return $(Expr(:tuple, exprs...))
     end
 end
+
+# promote() to SArray
+@inline Base.promote_rule(::Type{<: SVector{S, T} },
+                        y::Type{<: AbstractArray{T} } ) where {S, T} = (@boundscheck length(y) == S || error();
+                                                                        SVector{S, T} )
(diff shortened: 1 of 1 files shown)


andyferris:
OK, this is pretty interesting - what is the desired behavior here? That operations like `SVector(1,2,3) + [1,2,3]` returns an `SVector`? Or manual usage of `promote`?

I note that `promote` was designed for `Number` and operations such as `*` can invoke automatically `promote` - while the definitions here might not make great sense for that case. Also, should the `length` be checked to be compatible, or the `size`, or the `axes` (remembering we are coming from `AbstractArray`)?

githubtomtom:
actually I was try to do  SVector(1,2,3) .+ [1,2,3] (broadcasting) returning a SVector. Unfortunately overriding promote_rule() seems not able to achieve that automatically. I’m not familiar with broadcasting interface at the moment, would try to extend that piece later.

By the way, now we can manually use promote() which is nice I think.

c42f:
> actually I was try to do  SVector(1,2,3) .+ [1,2,3] (broadcasting) returning a SVector

Oh right! great to know thanks.

In that case I suggest we fix broadcasting rather than messing with `promote` which is generally for use with scalars.  Unfortunately it may not be simple; the broadcasting machinery is already quite complex.

c42f:
I've created #744 to track fixing broadcast as I think that's the correct way to fix your underlying issue.

PR #741, opened 2020-02-19
Suppress comments from BenchmarkCI

tkf:
See: https://github.com/JuliaArrays/StaticArrays.jl/pull/737#issuecomment-588475251

It's good to go (if this setup is desired) but I submit it as a draft to avoid accidentally merging this.


c42f:
Thanks! Let's see how the existing config goes for a bit and we can merge this if it seems necessary.

c42f:
I'm going to merge this to cut down on the number of notifications.

The HTML version is kinda nice, but the text version in the job logs is just as good I think.

PR #742, opened 2020-02-20
Port perf/benchmark_matrix_ops.jl to benchmark/*.jl

tkf:
This PR adds `benchmark/bench_matrix_ops.jl` which is a straight-forward port of `perf/benchmark_matrix_ops.jl`.

### DIFF
--- a/benchmark/bench_matrix_ops.jl
+++ b/benchmark/bench_matrix_ops.jl
@@ -21,7 +21,7 @@ for f in [det, inv, exp]
     for N in matrix_sizes
         SA = @SMatrix rand(N, N)
         A = Array(SA)
-        s2 = s1["$N"] = BenchmarkGroup()
+        s2 = s1[string(N, pad=2)] = BenchmarkGroup()
         s2["SMatrix"] = @benchmarkable $f($SA)
         s2["Matrix"] = @benchmarkable $f($A)
     end
@@ -35,7 +35,7 @@ for f in [*, \]
         SB = @SMatrix rand(N, N)
         A = Array(SA)
         B = Array(SB)
-        s2 = s1["$N"] = BenchmarkGroup()
+        s2 = s1[string(N, pad=2)] = BenchmarkGroup()
         s2["SMatrix"] = @benchmarkable $f($SA, $SB)
         s2["Matrix"] = @benchmarkable $f($A, $B)
     end
--- a/benchmark/bench_matrix_ops.jl
+++ b/benchmark/bench_matrix_ops.jl
@@ -1,21 +1,21 @@
 module BenchMatrixOps
 
+import Random
 using BenchmarkTools
 using LinearAlgebra
-using Random
 using StaticArrays
 
 const suite = BenchmarkGroup()
 const Nmax = 20
 
 # Use same arrays across processes (at least with the same Julia version):
-const RNG = MersenneTwister(1234)
+Random.seed!(1234)
 
 # Unary operators
 for f in [det, inv, exp]
     s1 = suite["$f"] = BenchmarkGroup()
     for N in 1:Nmax
-        SA = @SMatrix rand(RNG, N, N)
+        SA = @SMatrix rand(N, N)
         A = Array(SA)
         s2 = s1["$N"] = BenchmarkGroup()
         s2["SMatrix"] = @benchmarkable $f($SA)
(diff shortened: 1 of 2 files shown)


benchmark/bench_matrix_ops.jl
```diff
@@ -0,0 +1,41 @@
...
+        SA = @SMatrix rand(RNG, N, N)
+        A = Array(SA)
+        s2 = s1["$N"] = BenchmarkGroup()
+        s2["SMatrix"] = @benchmarkable $f($SA)
+        s2["Matrix"] = @benchmarkable $f($A)
+    end
+end
+
+# Binary operators
+for f in [*, \]
```
tkf: `perf/benchmark_matrix_ops.jl` only benchmarks `\`:

https://github.com/JuliaArrays/StaticArrays.jl/blob/5ead6c8693104daeccbed91dcc26dc4eb679d750/perf/benchmark_matrix_ops.jl#L5

I guess it was because `*` was handled in the extensive benchmarks in `perf/benchmark*.jl`.  But I guess it's reasonable to do this here, at least until they are ported to BenchmarkTools-based benchmarks?

benchmark/bench_matrix_ops.jl
```diff
@@ -0,0 +1,45 @@
+module BenchMatrixOps
+
+import Random
+using BenchmarkTools
+using LinearAlgebra
+using StaticArrays
+
+const suite = BenchmarkGroup()
+const matrix_sizes = if haskey(ENV, "GITHUB_EVENT_PATH")
+    (1, 2, 3, 4, 10, 20)
```
tkf: * `matrix_sizes = 1:20` => total CI time = 37 min https://github.com/JuliaArrays/StaticArrays.jl/runs/456805444
* `matrix_sizes = (1, 2, 3, 4, 10, 20)` => total CI time = 15 min https://github.com/JuliaArrays/StaticArrays.jl/runs/456815324

c42f: Great :+1:

If we're already at 15 min, I worry a bit about will happen for a comprehensive benchmark suite :grimacing:

benchmark/bench_matrix_ops.jl
c42f: Sounds good to me.

benchmark/bench_matrix_ops.jl
c42f: (I guess using the same treatment for QR would probably help with that 15 mins though)

benchmark/bench_matrix_ops.jl
```diff
@@ -0,0 +1,45 @@
...
+# Use same arrays across processes (at least with the same Julia version):
+Random.seed!(1234)
+
+# Unary operators
+for f in [det, inv, exp]
+    s1 = suite["$f"] = BenchmarkGroup()
+    for N in matrix_sizes
+        SA = @SMatrix rand(N, N)
+        A = Array(SA)
+        s2 = s1["$N"] = BenchmarkGroup()
```
c42f: To make these sort correctly in the benchmark results the following might work?

```suggestion
        s2 = s1[string(N, pad=2)] = BenchmarkGroup()
```

benchmark/bench_matrix_ops.jl
```diff
@@ -0,0 +1,45 @@
...
+
+# Binary operators
+for f in [*, \]
+    s1 = suite["$f"] = BenchmarkGroup()
+    for N in matrix_sizes
+        SA = @SMatrix rand(N, N)
+        SB = @SMatrix rand(N, N)
+        A = Array(SA)
+        B = Array(SB)
+        s2 = s1["$N"] = BenchmarkGroup()
```
c42f:
```suggestion
        s2 = s1[string(N, pad=2)] = BenchmarkGroup()
```

benchmark/bench_matrix_ops.jl
tkf: Yeah, I have the same worry. I guess we can't do the "full sweep" in CI and be a bit less greedy about the parameters we try.

An option to reduce CI time is to tune benchmarks separately and store the JSON file somewhere. But I think it's easy to get it stale and cumbersome to add benchmarks...

BTW, a less ad-hoc way to do this kind of filtering is to use `@tagged` from BenchmarkToos.jl: https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/manual.md#indexing-into-a-benchmarkgroup-using-tagged I don't think PkgBenchmark.jl support this though.

c42f: I also wonder how much of this is compile time; some of the algorithms are probably quite terrible for N=20.

tkf: Yeah, one thing I can't copy easily from `perf/*.jl` is the compile-time benchmarks. It'd be great if BenchmarkTools reports compile-time while doing the tuning. I don't think it is difficult but just tedious...

PR #743, opened 2020-02-20
Update SArray.jl

githubtomtom:
avoid calling `@inline (::Type{SA})(a::StaticArray) where {SA<:StaticArray} = SA(Tuple(a))` in `convert.jl`.
it allows us to define function that returns a SVector but expects a AbstractVector argument e.g.
```
fun(x::AbstractVector) = 2 * (SVector{length(x)}(x) )
```
with the proposed method, we can skip adding the following:
```
fun(x::SVector) = 2x
```

### DIFF
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -55,6 +55,7 @@ end
 ####################
 ## SArray methods ##
 ####################
+@inline SArray{S, T, N, L}(x::SArray{S, T, N, L} ) where {S, T, N, L} = x
 
 @propagate_inbounds function getindex(v::SArray, i::Int)
     v.data[i]


c42f:
I don't really understand what you're trying to do here, sorry! Could you provide code which you want to work but which doesn't work currently?

To me it seems we can already "write a function which accepts an `AbstractVector` and returns an `SVector`:

```julia
julia> fun(x::AbstractVector) = 2 * (SVector{length(x)}(x) )
fun (generic function with 1 method)
```

It works with both SVector and normal Vector:

```julia
julia> fun(SA[1,2,3])
3-element SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 2
 4
 6

julia> fun([1,2,3])
3-element SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 2
 4
 6
```

githubtomtom:
yes it works now. But now calling `SVector{N}(x)` would **inevitably** call `@inline (::Type{SA})(a::StaticArray) where {SA<:StaticArray} = SA(Tuple(a))` which is **wasteful** when `a`itself has  exactly the same `SArray` type.

with**out** my suggestion, if I want to **avoid** the wasteful call, I need to define `fun(x::SVector) = 2x`.

my suggestion is that we could simply (inline) return itself, so that a call like:
```
x = SVector{3, Int}(1, 2, 3)
fun(x)
```
would translate into `2 * SVector{3, Int}(x)` then inlined translated into `2 * x`.

c42f:
Right, thanks for explaining. We could merge this I guess; it seems harmless enough for `SArray`.

However: did you benchmark? The compiler is **very** good at removing this kind of thing completely for immutable types. I'd be surprised if you could measure a difference.

In fact, for your `fun` above, I have *identical* native code generated.

Before this PR:

```
julia> @code_native fun(SA[1,2,3])
	.text
; ┌ @ REPL[10]:1 within `fun'
	movq	%rdi, %rax
; │┌ @ linalg.jl:20 within `*'
; ││┌ @ broadcast.jl:758 within `broadcast'
; │││┌ @ broadcast.jl:820 within `materialize'
; ││││┌ @ broadcast.jl:26 within `copy'
; │││││┌ @ broadcast.jl:100 within `_broadcast'
; ││││││┌ @ broadcast.jl:125 within `macro expansion'
; │││││││┌ @ int.jl:54 within `*'
	vmovdqu	(%rsi), %xmm0
	vpaddq	%xmm0, %xmm0, %xmm0
	movq	16(%rsi), %rcx
	addq	%rcx, %rcx
; │└└└└└└└
	vmovdqu	%xmm0, (%rdi)
	movq	%rcx, 16(%rdi)
	retq
	nopl	(%rax,%rax)
; └
```

After the change in this PR:

```
julia> @code_native fun(SA[1,2,3])
	.text
; ┌ @ REPL[2]:1 within `fun'
	movq	%rdi, %rax
; │┌ @ linalg.jl:20 within `*'
; ││┌ @ broadcast.jl:758 within `broadcast'
; │││┌ @ broadcast.jl:820 within `materialize'
; ││││┌ @ broadcast.jl:26 within `copy'
; │││││┌ @ broadcast.jl:100 within `_broadcast'
; ││││││┌ @ broadcast.jl:125 within `macro expansion'
; │││││││┌ @ int.jl:54 within `*'
	vmovdqu	(%rsi), %xmm0
	vpaddq	%xmm0, %xmm0, %xmm0
	movq	16(%rsi), %rcx
	addq	%rcx, %rcx
; │└└└└└└└
	vmovdqu	%xmm0, (%rdi)
	movq	%rcx, 16(%rdi)
	retq
	nopl	(%rax,%rax)
; └
```

Notice how the line number information is all inside broadcast.jl?  That means the constructor itself generates zero lines of code, regardless of the `Tuple` or not.

c42f:
As discussed, I don't think we need this so I'll close the issue.

However, if you do have a benchmark which shows a difference I'm happy to reconsider.

Issue #746, opened 2020-02-24
Conversion to AbstractArray not defined


daanhb:
Conversion of static vectors and matrices to an AbstractVector or AbstractArray with a wider element type seems to fail with an error:
```julia
julia> using StaticArrays

julia> convert(AbstractVector{BigFloat}, SVector(1.0, 2.0))
ERROR: setindex!() with non-isbitstype eltype is not supported by StaticArrays. Consider using SizedArray.
...
```
Yet, a type-stable conversion is possible because:
```julia
julia> convert(SVector{2,BigFloat}, SVector(1.0, 2.0)) isa AbstractVector{BigFloat}
true
```
Conversion to an abstract vector or abstract array with the same element type succeeds (via a fallback in Base).

I use conversions like this elsewhere in generic code that updates the element type of a vector. Currently, I have to add special cases for static vectors.

Would this just be a matter of adding conversion routines that invoke the right static array constructor? E.g., though it should be more general than this:
```julia
julia> import Base.convert

julia> convert(::Type{AbstractArray{T}}, v::SVector{N,S}) where {N,S,T} = convert(SVector{N,T}, v)
convert (generic function with 205 methods)

julia> convert(::Type{AbstractVector{T}}, v::SVector{N,S}) where {N,S,T} = convert(SVector{N,T}, v)
convert (generic function with 206 methods)

julia> convert(AbstractVector{BigFloat}, SVector(1.0, 2.0))
2-element SArray{Tuple{2},BigFloat,1,2} with indices SOneTo(2):
 1.0
 2.0

julia> ans isa AbstractVector{BigFloat}
true
```
This is not as problematic as going in the other direction (converting from abstract vectors to static vectors).

c42f:
This seems fairly clear cut — `Base` defines it so we should too.

A little use of `similar_type` and matching the `AbstractArray{N}` dimension `N` will probably work.

daanhb:
Thanks! Judging by the definitions in base [here](https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L15), the conversions to `AbstractArray{T}` and `AbstractArray{T,N}` end up calling the `AbstractArray{T}` and `AbstractArray{T,N}` constructor. So, it seems nicer to define those.

Would adding the following lines to `convert.jl` in StaticArrays be sufficient for all cases? Is this the right way to do it?
```julia
AbstractArray{T}(sa::StaticArray{S,T}) where {S,T} = sa
[...]

PR #747, opened 2020-02-25
support conversion to AbstractArray

daanhb:
Julia Base defines conversion to the abstract types `AbstractArray{T}` and `AbstractArray{T,N}` by invoking their (abstract) constructors. These fail with an error for static vectors (see #746).

This pull request defines the `AbstractArray{T}` and `AbstractArray{T,N}` constructors for static arrays in terms of `similar_type`. This enables the following:
```julia
julia> convert(AbstractVector{Float64}, SVector(1,2))
2-element SArray{Tuple{2},Float64,1,2} with indices SOneTo(2):
 1.0
 2.0
```
One use case is to generically change the element type of a vector, in the example above from `Int` to `Float64`.

I've added tests for SVector, SMatrix, MVector and MMatrix. Perhaps there are others?
Also, should the conversions perhaps be `@inline`?

### DIFF
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -134,23 +134,49 @@ using StaticArrays, Test, LinearAlgebra
         # Issue #746
         # conversion to AbstractArray changes the eltype from Int to Float64
         sv = SVector(1,2)
-        @test convert(AbstractArray{Float64}, sv) isa SVector{2,Float64}
-        @test convert(AbstractVector{Float64}, sv) isa SVector{2,Float64}
+        @test @inferred(convert(AbstractArray{Float64}, sv)) isa SVector{2,Float64}
+        @test @inferred(convert(AbstractVector{Float64}, sv)) isa SVector{2,Float64}
         @test convert(AbstractArray{Float64}, sv) == sv
         @test convert(AbstractArray{Int}, sv) === sv
         sm = SMatrix{2,2}(1,2,3,4)
-        @test convert(AbstractArray{Float64,2}, sm) isa SMatrix{2,2,Float64}
+        @test @inferred(convert(AbstractArray{Float64,2}, sm)) isa SMatrix{2,2,Float64}
         @test convert(AbstractArray{Float64,2}, sm) == sm
         @test convert(AbstractArray{Int,2}, sm) === sm
         mv = MVector(1, 2, 3)
-        @test convert(AbstractArray{Float64}, mv) isa MVector{3,Float64}
-        @test convert(AbstractVector{Float64}, mv) isa MVector{3,Float64}
+        @test @inferred(convert(AbstractArray{Float64}, mv)) isa MVector{3,Float64}
+        @test @inferred(convert(AbstractVector{Float64}, mv)) isa MVector{3,Float64}
         @test convert(AbstractArray{Float64}, mv) == mv
         @test convert(AbstractArray{Int}, mv) === mv
         mm = MMatrix{2, 2}(1, 2, 3, 4)
-        @test convert(AbstractArray{Float64,2}, mm) isa MMatrix{2,2,Float64}
+        @test @inferred(convert(AbstractArray{Float64,2}, mm)) isa MMatrix{2,2,Float64}
         @test convert(AbstractArray{Float64,2}, mm) == mm
         @test convert(AbstractArray{Int,2}, mm) === mm
+
+        # Test some of the types in StaticMatrixLike
+        sym = Symmetric(SA[1 2; 2 3])
+        @test @inferred(convert(AbstractArray{Float64}, sym)) isa Symmetric{Float64,SMatrix{2,2,Float64,4}}
+        @test @inferred(convert(AbstractArray{Float64,2}, sym)) isa Symmetric{Float64,SMatrix{2,2,Float64,4}}
+        @test convert(AbstractArray{Float64}, sym) == sym
+        her = Hermitian(SA[1 2+im; 2-im 3])
+        @test @inferred(convert(AbstractArray{ComplexF64}, her)) isa Hermitian{ComplexF64,SMatrix{2,2,ComplexF64,4}}
+        @test convert(AbstractArray{ComplexF64}, her) == her
+        diag = Diagonal(SVector(1,2))
+        @test @inferred(convert(AbstractArray{Float64}, diag)) isa Diagonal{Float64,SVector{2,Float64}}
+        @test convert(AbstractArray{Float64}, diag) == diag
+        # The following cases currently convert the SMatrix into an MMatrix, because
+        # the constructor in Base invokes `similar`, rather than `convert`, on the static array
+        # trans = Transpose(SVector(1,2))
+        # @test @inferred(convert(AbstractArray{Float64}, trans)) isa Transpose{Float64,SVector{2,Float64}}
+        # adj = Adjoint(SVector(1,2))
+        # @test @inferred(convert(AbstractArray{Float64}, adj)) isa Adjoint{Float64,SVector{2,Float64}}
+        # uptri = UpperTriangular(SA[1 2; 0 3])
+        # @test @inferred(convert(AbstractArray{Float64}, uptri)) isa UpperTriangular{Float64,SMatrix{2,2,Float64,4}}
+        # lotri = LowerTriangular(SA[1 0; 2 3])
+        # @test @inferred(convert(AbstractArray{Float64}, lotri)) isa LowerTriangular{Float64,SMatrix{2,2,Float64,4}}
+        # unituptri = UnitUpperTriangular(SA[1 2; 0 1])
+        # @test @inferred(convert(AbstractArray{Float64}, unituptri)) isa UnitUpperTriangular{Float64,SMatrix{2,2,Float64,4}}
+        # unitlotri = UnitLowerTriangular(SA[1 0; 2 1])
+        # @test @inferred(convert(AbstractArray{Float64}, unitlotri)) isa UnitLowerTriangular{Float64,SMatrix{2,2,Float64,4}}
     end
 end
 
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -129,6 +129,25 @@ using StaticArrays, Test, LinearAlgebra
         m = MVector(1, 2, 3)
         @test @inferred(reverse(m))::typeof(m) == MVector(3, 2, 1)
     end
+
+    @testset "Conversion to AbstractArray" begin
+        # Issue #746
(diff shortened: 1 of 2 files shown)


daanhb:
The plot thickens: I ended up testing all of the matrices in `StaticMatrixLike` (thanks for the suggestion) because some of them actually do not work. Chasing this, it works for `Symmetric`, `Hermitian` and `Diagonal` because they explicitly define the conversion to `AbstractMatrix`. See, e.g., [here](https://github.com/JuliaLang/julia/blob/434c6bbead0edc59f1d8c64c55a979b5898d994e/stdlib/LinearAlgebra/src/symmetric.jl#L276).

However, the other types (`Transpose`, `Adjoint` and the `Triangular` family) fall back to a default constructor in Base defined in [array.jl](https://github.com/JuliaLang/julia/blob/434c6bbead0edc59f1d8c64c55a979b5898d994e/base/array.jl#L541), which invokes `similar`. Hence, an `SArray` becomes an `MArray` even with the current changes:
```julia
julia> tri = UpperTriangular(SA[1 2; 0 3])
2×2 UpperTriangular{Int64,SArray{Tuple{2,2},Int64,2,4}}:
 1  2
 ⋅  3

julia> convert(AbstractArray{Float64}, tri)
2×2 UpperTriangular{Float64,MArray{Tuple{2,2},Float64,2,4}}:
 1.0  2.0
  ⋅   3.0
```
At least it does not throw an error and the result is correct, just not optimal.

I've added tests for the types that do not completely work, but commented them out for now. This can not be fixed in `StaticArrays` and should probably be pursued in Base.

daanhb:
Hmm, I thought about filing an issue with Julia, but it suddenly is not such a clear-cut case anymore. Construction of arrays has seen a great deal of discussion and people may have different expectations, but even regardless of that currently `AbstractArray{T,N}(myarray)` in Base typically returns something mutable. So, changing that would be breaking.

Worse, in fact it is breaking for StaticArrays too, because I did not realize before that this line actually works:
```julia
julia> using StaticArrays

julia> convert(AbstractVector{Float64}, SVector(1, 2))
2-element MArray{Tuple{2},Float64,1,2} with indices SOneTo(2):
 1.0
 2.0
```
The original example in #746 failed not because the constructor is not defined, but because the fallback in Base did not work for the non-isbits type `BigFloat`... In hindsight, that is exactly what the error message was telling me. D-oh! That also means the issue with my original error message was never the absence of the constructor, but the unrelated issue that assigment of `BigFloat` to mutable static arrays is not supported. (Which I guess is intentional.) I'll let myself out :-)

In my mind `convert(AbstractArray{T,N}, myarray)` would complement `similar(myarray, T)`. `similar` deliberately throws away data and mutability, but the conversion would retain data and mutability. However, as things stand, by default and by design the conversion does retain data but also discards mutability. It can be, and actually mostly is, implemented in terms of `similar` and does not bring anything new to the table. That means I'm left looking for an idiomatic way to preserve both data and mutability :-)

A simple example unrelated to static arrays is the following:
```julia
julia> convert(AbstractVector{Float64}, 1:3)
3-element Array{Float64,1}:
 1.0
 2.0
 3.0
```
Which method would give me `1.0:1.0:3.0` instead?

daanhb:
I guess one alternative for my original use case may be to just use broadcast:
```julia
julia> BigFloat.(SVector(1,2))
2-element SArray{Tuple{2},BigFloat,1,2} with indices SOneTo(2):
 1.0
 2.0
```
This typically doesn't preserve structure though:
```julia
julia> Float64.(1:3)
3-element Array{Float64,1}:
 1.0
 2.0
 3.0
```

daanhb:
On ~~second~~ third thought, last comment for the day, things are still not entirely consistent as they are now without this pull request because conversion to a vector with the same element type returns itself even if it is immutable:
```julia
julia> convert(AbstractVector{Int}, SVector(1,2))
2-element SArray{Tuple{2},Int64,1,2} with indices SOneTo(2):
 1
 2

julia> convert(AbstractVector{Float64}, SVector(1,2))
2-element MArray{Tuple{2},Float64,1,2} with indices SOneTo(2):
 1.0
 2.0
```
The [fallback conversion](https://github.com/JuliaLang/julia/blob/434c6bbead0edc59f1d8c64c55a979b5898d994e/base/array.jl#L541) to a mutable container may just be because, well, one needs a mutable container in order to be able to copy data back into it.

c42f:
> That also means the issue with my original error message was never the absence of the constructor, but the unrelated issue that assigment of `BigFloat` to mutable static arrays is not supported. (Which I guess is intentional.)

This is arguably a bug but not really possible to fix efficiently and generically due to Julia's semantics of immutable and mutable objects. You may want a `SizedArray` or mutable `FieldArray` for this case instead. See, however, https://github.com/JuliaArrays/StaticArrays.jl/pull/749 where I've fixed this.

> In my mind `convert(AbstractArray{T,N}, myarray)` would complement `similar(myarray, T)`. similar deliberately throws away data and mutability, but the conversion would retain data and mutability.

This is how I expect it to behave: preserving "as much as possible" about the type of the source array, but changing its `eltype`. I reckon we just implement it this way. It may break some things, but only by removing mutability (so the breakage will be clearly visible). That's just as well, because depwarning this would be difficult!

daanhb:
Just to be sure, I've checked what happens in Julia itself with conversions to `AbstractArray`. It is used very sparingly in Base, very liberally in the LinearAlgebra package, and just two times in SparseArrays. (This is based simply on grep, I could have missed many things.)

To test I've added a few lines (only three in total) to base/range.jl and stdlib/LinearAlgebra/src/triangular.jl, starting from Julia-1.4.0-rc2, to get the following behaviour:
```julia
julia> convert(AbstractVector{Float64}, 1:5)
1.0:1.0:5.0

julia> using LinearAlgebra

julia> tri = UpperTriangular(Diagonal(1:3))
3×3 UpperTriangular{Int64,Diagonal{Int64,UnitRange{Int64}}}:
 1  0  0
 ⋅  2  0
 ⋅  ⋅  3

julia> convert(AbstractMatrix{Float64}, tri)
3×3 UpperTriangular{Float64,Diagonal{Float64,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}}:
 1.0  0.0  0.0
  ⋅   2.0  0.0
  ⋅    ⋅   3.0

julia> convert(AbstractMatrix{BigInt}, tri)
3×3 UpperTriangular{BigInt,Diagonal{BigInt,UnitRange{BigInt}}}:
 1  0  0
 ⋅  2  0
 ⋅  ⋅  3
```
In other words, the conversion preserves the immutable range type, and both UpperTriangular and Diagonal recursively invoke `AbstractArray` conversion on their data.

The results with static vectors would be similar, but I wanted to check Julia's test suite so I went for the immutable ranges. 
The good news is that the entire test suite simply passes!

Two small things:
- I did have to restrict the element type in `AbstractArray{T,1}(r::AbstractRange)` to be `<:Real` because Ranges do not like complex numbers. This actually comes up in a test that does `complex(1:5)`, and the conversion is triggered by [this line](https://github.com/JuliaLang/julia/blob/012b270df6c524bf3058e092d68aefc369647d67/base/complex.jl#L1029) in complex.jl.
- One conversion to `AbstractArray` in SparseArrays looks worrisome [here](https://github.com/JuliaLang/julia/blob/012b270df6c524bf3058e092d68aefc369647d67/stdlib/SparseArrays/src/sparsevector.jl#L1791). This code seems to assume that the result after conversion is mutable, even though the method itself is out-of-place. This would have led to an error, if a test had been included with an immutable type.

daanhb:
> > In my mind `convert(AbstractArray{T,N}, myarray)` would complement `similar(myarray, T)`. similar deliberately throws away data and mutability, but the conversion would retain data and mutability.
> 
> This is how I expect it to behave: preserving "as much as possible" about the type of the source array, but changing its `eltype`. I reckon we just implement it this way. It may break some things, but only by removing mutability (so the breakage will be clearly visible). That's just as well, because depwarning this would be difficult!

I agree with this, the breakage will be clearly visible - and hopefully limited. It would help if there was a generic function for the following two behaviours:
- give me the most efficient array that is like the one I have but with a different element type
- give me a mutable array that is like the one I have but with a different element type

The latter can be implemented purely in terms of `similar`, but the former can not and for optimal results it needs help from the specific concrete Array. (If it turns out to be helpful elsewhere, it could be part of the `AbstractArray` interface to optionally implement the `AbstractArray{T,N}` constructor this way..)

test/abstractarray.jl
```diff
@@ -129,6 +129,55 @@ using StaticArrays, Test, LinearAlgebra
...
+        # adj = Adjoint(SVector(1,2))
+        # @test @inferred(convert(AbstractArray{Float64}, adj)) isa Adjoint{Float64,SVector{2,Float64}}
+        # uptri = UpperTriangular(SA[1 2; 0 3])
+        # @test @inferred(convert(AbstractArray{Float64}, uptri)) isa UpperTriangular{Float64,SMatrix{2,2,Float64,4}}
+        # lotri = LowerTriangular(SA[1 0; 2 3])
+        # @test @inferred(convert(AbstractArray{Float64}, lotri)) isa LowerTriangular{Float64,SMatrix{2,2,Float64,4}}
+        # unituptri = UnitUpperTriangular(SA[1 2; 0 1])
+        # @test @inferred(convert(AbstractArray{Float64}, unituptri)) isa UnitUpperTriangular{Float64,SMatrix{2,2,Float64,4}}
+        # unitlotri = UnitLowerTriangular(SA[1 0; 2 1])
+        # @test @inferred(convert(AbstractArray{Float64}, unitlotri)) isa UnitLowerTriangular{Float64,SMatrix{2,2,Float64,4}}
```
c42f: I guess we can we work around this by adding new constructors, eg, `Transpose(m::StaticMatrix) = ...`?

Either that or uncomment these and leave them as `@test_broken` for now?

daanhb: The construction of the matrices is fine, fortunately, with all these types, it is just the conversion. I think that, in order to fix it here, we would have to intercept functions like `AbstractMatrix{T}(A::UpperTriangular{T,<:SArray})`. That seems a bit intrusive, so I've added `@test_broken` for now like you suggest here. In the meantime I did file an issue with Julia (https://github.com/JuliaLang/julia/issues/34995) to hopefully fix it upstream.

c42f: Defining `AbstractMatrix{T}(A::UpperTriangular{T,<:SArray})` seems somewhat ok to me — we already do these kinds of things elsewhere in the package.

Having said that, it's definitely not ideal — thanks for submitting the upstream issue.

PR #749, opened 2020-02-27
RFC: setindex! for non-isbits MArrays

c42f:
Here I've tried implementing this for the sake of people using MArray in generic code, even though it's unlikely to be fast. (Furthermore I can't think of a way to make it fast. It's like we need a weird kind of mutable pseudo-container which is guaranteed to be inlined as a part of the parent struct, such that you can't access it separately. Ironically a mutable `FieldArray` should work like this.)

Overall I'm somewhat inclined to do this but not decided. If we leave this as-is, it doesn't leave a good option for people wanting to write generic code using `MArray`. It seems we can have:
* "just works" but potentially quite slow. Could be nice for generic code, especially if people want to put in `BigFloat` or suchlike on occasion.
* Fails hard but clearly directs people to use `SizedArray`. Good for aspiring performance hackers but kind of bad for everyone else.

The generated code for this isn't great. Basically `m[i] = v` turns into a big pile of `(ifelse(i == 1, v, m[1]), ifelse(i == 2, v, m[2]), ...)` to construct a new tuple which `m.data` will be set to.

Thoughts?


tkf:
For the "fail hard" option, wouldn't it be better to disallow even creating `MArray`s with non-isbits elements? If so, I think defining `setindex!` here makes sense if it is allowed to construct in the first place.

> Good for aspiring performance hackers

Shouldn't people who wants to optimize the hell out of their code use `SArray` in the first place?  Gearing towards user-friendliness in `MArray` API makes sense to me.

(@KristofferC If the benchmark comment from the bot is too noisy, we can turn off the comment. See #741)

c42f:
> Can we use `@GC.preserve` on both `v` and `val` and use something like the isbitstype path? We basically want to copy a pointer... so long as the garbage collection is conservative enough this should be safe, no?

Well I kind of think not? The storage `Tuple` is meant to be immutable so if we go around overwriting elements doesn't that violate Julia semantics? Consequences could be the compiler proving something const in inference and making decisions based on that which are violated at the runtime.

Related - I'm not exactly sure why we get away with apparently the exact same thing for bits types! (I guess the inlining of "immutable" storage into the mutable parent structure has some weird semantic consequences.)

> Or is there something about the generational GC that makes this more complex than I would guess?

Yes! The GC is generational. Objects are split into old and new generations. In order to collect the new generation independently, it needs to specially track any old objects which point to new ones. This means you need to inform the GC whenever you set a pointer field (so it can update the internal data structures). This is the function of the obscurely named runtime function `jl_gc_wb()`. So in principle we could write the pointer and ccall `jl_gc_wb` or equivalent. It might be fun to try but I'm not sure this can make sense for tuples which are meant to be immutable :-)

c42f:
> For the "fail hard" option, wouldn't it be better to disallow even creating `MArray`s with non-isbits elements? If so, I think defining `setindex!` here makes sense if it is allowed to construct in the first place.

You might be right about that. We currently have another bug where initializing with `undef` basically makes a non-isbits `MArray` unusable.

So perhaps this PR would have better been an issue: "What to do about MArray with non-isbits elements?"

> > Good for aspiring performance hackers
> 
> Shouldn't people who wants to optimize the hell out of their code use `SArray` in the first place? Gearing towards user-friendliness in `MArray` API makes sense to me.

Temporary `MArray`s can be quite good for in-place algorithms; very efficient and simple to use, and the compiler seems quite willing to reason about them effectively. Especially for some of the linear algebra algorithms, people have to go to extreme lengths to be able to use the immutable versions.

When I say "extreme lengths" I guess it's a matter of perspective: I find imperative in-place numerical linear algebra more natural to reason about compared to the very functional style that SArray demands.

> (@KristofferC If the benchmark comment from the bot is too noisy, we can turn off the comment. See #741)

Yes I think it's a bit noisy (especially when I do stupid things like breaking the tests, as I did here :grimacing:) I've merged that now to reduce notification volume.

tkf:
> When I say "extreme lengths" I guess it's a matter of perspective

I guess this is what I meant to say by "optimize the hell out."  I wanted to mention that `MArray` lies somewhere between `SArray` and `Array` in terms of ease-of-use spectrum and (maybe "user-friendliness" wasn't the best word for it).  Since `SArray` covers the extreme portion of this spectrum, I thought accepting more generic use with `MArray` seems to be a good balance.  That is to say, if you have a super performance-sensitive code and you want to avoid bitten by non-isbits `MArray` at all cost, you still can use `SArray`.

By the way, can the `unsafe_store!` path support union of isbits types like `Union{Float64,Missing}`?

c42f:
> By the way, can the `unsafe_store!` path support union of isbits types like `Union{Float64,Missing}`?

Excellent question, I checked the data layout and I think the answer is a conclusive "no for now" because tuples-of-unions are not inlined. Indeed I think the covariance of tuples makes this impossible right now because a data storage `NTuple{2,Union{Missing,Float64}}` is not a concrete type?

Possibly it would help to have systematic treatment of small union selector bits along the lines of https://github.com/JuliaLang/julia/issues/33120 but I guess we would still need a new builtin `InvariantTuple` to make this work.

c42f:
> I checked the data layout and I think the answer is a conclusive "no for now" because tuples-of-unions are not inlined

Actually very recently we have https://github.com/JuliaLang/julia/pull/34126 — if that were to merge we might actually be in a position to fix this without a huge performance hit. Though we'd still need to figure out how to maintain correct GC invariants.

tkf:
> I guess we would still need a new builtin `InvariantTuple` to make this work

Can we just insert a "value wrapper" to get an invariant tuple type?

```julia
struct InvariantValue{T}  # "value wrapper"
    value::T
end

struct InvariantTuple{T <: Tuple{Vararg{InvariantValue}}}
    values::T
end

Base.getindex(t::InvariantTuple, i::Integer) = t.values[i].value

# Some helper functions
(::Base.Colon)(x::S, ::Type{T}) where {T, S <: T} = InvariantValue{T}(x)
invariant(args::InvariantValue...) = InvariantTuple(args)

t1 = invariant(1:Union{Missing,Int}, 2.0:Union{Missing,Float64})
t2 = invariant(1:Union{Missing,Int}, missing:Union{Missing,Float64})
typeof(t1) === typeof(t2)
```

(Alternatively, we can also create a "typed linked list." Though I guess this is much more compiler-friendly.)

I guess what we need here is actually

```julia
struct InvariantNTuple{N,T}
    values::NTuple{N,InvariantValue{T}}
end
```


which may be more compiler friendly (naively thinking, fewer type parameters sounds good)?

goulart-paul:
Can someone comment on the status of this PR?   I find myself in the "people wanting to write generic code using MArray" group described above, particularly to keep open the option of using BigFloats.

ChrisRackauckas:
> Revisit if https://github.com/JuliaLang/julia/pull/34126 is merged

It merged, so it should be fine now?

Issue #748, opened 2020-02-27
Discussion: should we use divide-and-conquer reduce for instruction-level parallelism?


tkf:
As I discussed briefly with @c42f in https://github.com/JuliaArrays/StaticArrays.jl/pull/702#discussion_r379269102, using the divide-and-conquer approach in `reduce` on static arrays may be useful for leveraging instruction-level parallelism (ILP).  I cooked up two (somewhat contrived?) examples and it turned out that using divide-and-conquer approach can be 3x to 4x faster (in my laptop) than sequential (`foldl`) approach.  This is the case when "parallelizing" both compute- and memory- bound instructions.

I also made two versions of benchmarks with and without `@simd`.  I was surprised that putting `@simd` didn't help improving the `foldl` version.

Anyway, I think it seems to be a pretty good motivation for using this approach in StaticArrays.jl, although it'd be even nicer if there are more "real-world" benchmarks like this. What do you think?  Does it make sense to use this approach in StaticArrays.jl?

### Code

I'm using `NTuple` as a model of `SVector` here. `mapmapreduce!` is the main function that I benchmarked.

```julia
module ILPBenchmarks

@inline tmapreduce(f::F, op::O, init, ::Tuple{}) where {F,O} = init
@inline tmapreduce(f::F, op::O, init, x::Tuple{Any}) where {F,O} = f(x)
@inline tmapreduce(f::F, op::O, init, (a, b)::Tuple{Any,Any}) where {F,O} = op(f(a), f(b))
@inline function tmapreduce(f::F, op::O, init, v::Tuple) where {F,O}
    left, right = _halve(v)
    return op(tmapreduce(f, op, init, left), tmapreduce(f, op, init, right))
end

@generated function _halve(v::NTuple{N,Any}) where {N}
    m = N ÷ 2
    quote
        (($([:(v[$i]) for i in 1:m]...),), ($([:(v[$i]) for i in m+1:N]...),))
    end
end

@inline foldlargs(op::F, acc) where {F} = acc
@inline foldlargs(op::F, acc, x, xs...) where {F} = foldlargs(op, op(acc, x), xs...)

@inline tmapfoldl(f::F, op::O, init, xs::Tuple) where {F,O} =
    foldlargs((acc, x) -> op(acc, f(x)), init, xs...)

function mapmapreduce!(mapreducer::R, g::G, f::F, op::O, ys, xs, init) where {R,G,F,O}
    for i in eachindex(xs, ys)
[...]

PR #750, opened 2020-02-29
Define mapfoldl/foldl for static arrays

tkf:
This PR includes some improvements discussed in https://github.com/JuliaArrays/StaticArrays.jl/pull/702#discussion_r379200418:

* Document how `_InitialValue` works and explain some idea behind it.
* Use `_InitialValue` instead of passing around `kw.data :: NamedTuple`.
* Use `_InitialValue` to collapse the foldl implementations with and without `init` into one.
* Add `mapfoldl` and `foldl` on static arrays:
  1. Rename the reduce "kernels" `_mapreduce` to `_mapfoldl`.
  2. Add `_mapreduce` back as an alias to `_mapfoldl`.
     * This way, functions (e.g., `sum`) wrapping reduce can just call `_mapreduce`.
     * I added this indirection to avoid another big diff in the future #751.  It may make sense to use the divide-and-conquer version for `_mapreduce` #748.
  3. Define `Base.mapfoldl` and `Base.foldl` as a simple wrapper of `_mapfoldl`.

It's not a tiny diff but mostly all the changes are simple refactoring.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -1,3 +1,23 @@
+"""
+    _InitialValue
+
+A singleton type for representing "universal" initial value (identity element).
+
+The idea is that, given `op` for `mapfoldl`, virtually, we define an "extended"
+version of it by
+
+    op′(::_InitialValue, x) = x
+    op′(acc, x) = op(acc, x)
+
+This is just a conceptually useful model to have in mind and we don't actually
+define `op′` here  (yet?).  But see `Base.BottomRF` for how it might work in
+action.
+
+(It is related to that you can always turn a semigroup without an identity into
+a monoid by "adjoining" an element that acts as the identity.)
+"""
+struct _InitialValue end
+
 @inline _first(a1, as...) = a1
 
 ################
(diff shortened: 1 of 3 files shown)


src/mapreduce.jl
```diff
@@ -86,28 +106,21 @@ end
...
     expr = :(f($(tmp...)))
-    for i ∈ 2:prod(S)
-        tmp = [:(a[$j][$i]) for j ∈ 1:length(a)]
-        expr = :(op($expr, f($(tmp...))))
-    end
-    return quote
-        @_inline_meta
-        @inbounds return $expr
+    if init === _InitialValue
+        expr = :(Base.reduce_first(op, $expr))
```
tkf: `Base.reduce_first` wasn't called before so `sum(SA[true])` returned `true`; now it returns `1 :: Int`.  This is consistent with `sum([true])`.

src/mapreduce.jl
```diff
@@ -188,20 +181,33 @@ end
...
+################
+## (map)foldl ##
+################
+
+@inline mapfoldl(f, op::R, a::StaticArray; init = _InitialValue()) where {R} =
+    _mapfoldl(f, op, :, init, Size(a), a)
+@inline foldl(op::R, a::StaticArray; init = _InitialValue()) where {R} =
+    _foldl(op, a, :, init)
+@inline _foldl(op::R, a, dims, init = _InitialValue()) where {R} =
+    _mapfoldl(identity, op, dims, init, Size(a), a)
```
c42f: Do we need the `R` here? It doesn't seem to be used.

tkf: This is yet another "ritual" that I do just by habit to avoid a bad interaction with the specialization heuristics: https://github.com/JuliaLang/julia/issues/32761#issuecomment-518773597

> If I remember correctly, the `f::F where F` specialization was only required if you didn't call `f` directly inside the function body but instead just passed it to another function — probably a non-inlined function. Or at least that's the way it worked a long time ago (year+).

I thought this can happen when `_foldl` is happened to be at the boundary of inlining decision but I don't have a proof.  If you don't want include this kind of "cargo cult" code, I'm happy to remove the no-op type parameters.

c42f: Eek, I didn't know about this corner of the specialization heuristics. Would be nice if `@specialize` did this rather than having to introduce `where` syntax.

Anyway, it seems this has been documented, so maybe we should just add a small comment and link to https://docs.julialang.org/en/v1.5-dev/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing-1

I'm still a bit unclear on how that relates to inlining. Presumably this has no effect if all callees are inlined, but if `_mapfoldl` was not inlined we'll get hit with dynamic dispatch without this?

tkf: > if `_mapfoldl` was not inlined we'll get hit with dynamic dispatch without this?

That's my understanding. I thought it's something like the `g_func` example in the documentation.

PR #752, opened 2020-03-12
Migrate to github actions for CI

c42f:
Our CI on Travis is really slow now, given that we try to test quite comprehensively.

Github actions seems to allow for more parallel jobs (and it integrates really nicely with the PR workflow), so let's try it out.

I've tried to slim down the build matrix from the full cartesian product, but the docs aren't quite clear on how this works, so perhaps I've got it wrong.  Could probably do some more exclusions too. See
https://github.community/t5/GitHub-Actions/Improve-matrix-exclusion-feature/td-p/32592
https://github.community/t5/GitHub-Actions/How-to-conditionally-include-exclude-items-in-matrix-eg-based-on/td-p/37871


c42f:
Ok, well at this point I'm confused. I'm getting

> Your workflow file was invalid: The pipeline is not valid. .github/workflows/ci.yml (Line: 28, Col: 13): Matrix include mapping does not contain any extra values to include,.github/workflows/ci.yml (Line: 29, Col: 13): Matrix include mapping does not contain any extra values to include

So something to do with the `include` isn't parsing correctly, but I can't see what's different from https://github.com/julia-actions/setup-julia#matrix-testing

tkf:
I was confused how `include` works in GitHub actions, too. But it seems quite different from Travis:

> You can add additional configuration options to a build matrix job that **already exists**. For example, [...]
>
> ```yaml
> runs-on: ${{ matrix.os }}
> strategy:
>   matrix:
>     os: [macos-latest, windows-latest, ubuntu-18.04]
>     node: [4, 6, 8, 10]
>     include:
>       # includes a new variable of npm with a value of 2 for the matrix leg matching the os and version
>       - os: windows-latest
>         node: 4
>         npm: 2
> ```
>
> --- https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-configurations-in-a-matrix-build

So it's about adding new option key-value pairs, not about adding new job?  Confusingly, `exclude` seems to work like Travis. I don't know what is happening with this asymmetry...

Anyway, it seems emulating `include` on Travis requires creating a big cartesian product and removing the ones we don't want to run.  Alternatively, maybe creating multiple YAML files is easier?

martinholters:
One thing I've noticed is that of TagBot creates a release, that not result in a tag push event bring triggered. So no CI run and hence no stable docs deployment. Might be worth also triggering on release because of that.

tkf:
@martinholters This configuration takes care of it:

https://github.com/JuliaArrays/StaticArrays.jl/blob/5f0d53fdd4be37f7de12304840096e9834e613c7/.github/workflows/TagBot.yml#L13

See https://github.com/JuliaRegistries/TagBot#ssh-deploy-keys

c42f:
Wow I missed that `include` and `exclude` are weirdly asymmetric. Yikes.

Ok, squashed those together and merged this. Thanks for your help!

Issue #758, opened 2020-04-01
No method to convert from UniformScaling


danielmatz:
If I have a struct such as

```julia
struct Foo
    x::SMatrix{3, 3, Float64}
end
```

and I want to initialize it using `I` (`UniformScaling{Bool}`), the following does not work:

```julia
Foo(I)
```

The constructor that Julia creates calls convert, but there isn't a matching method. Instead, I have to do

```julia
Foo(SMatrix{3, 3, Float64}(I))
```

Could we add a convert method? Perhaps one that just forwards to the constructor.

c42f:
Yes, we should make this work by adding an appropriate `convert` method.

By the way, `SMatrix{3, 3, Float64}` is not a concrete type; you want `SMatrix{3,3,Float64,9}` instead.

c42f:
#759 should fix this for you

danielmatz:
> By the way, SMatrix{3, 3, Float64} is not a concrete type; you want SMatrix{3,3,Float64,9} instead.

Argh, I keep forgetting about that. Thank you.

And thank you for the fix!

c42f:
> Argh, I keep forgetting about that.

Yeah it's really unfortunate that it's required :-(

I'm fairly sure there's nothing we can do about it though. Not until we have compiler support for doing something better.

c42f:
Thanks for the bug report anyway :-)

andyferris:
This is indeed an annoying case, @danielmatz. I am sympathetic. But I should refer to my post on #759:

> Umm, sorry, but I really think we shouldn't do this! If you are transforming an object by adding or removing information (e.g. `size`), you should use a constructor not `convert`.
> 
> `convert` relies on a whole bunch of assumptions. Or rather, methods in `Base` rely on a bunch of assumptions about `convert`. The `hash` should be the same before and after the convert. They must be `isequal`, and `isless` has to define the same ordering. It should conserve the transitivity of `isequal`, for example if you define this than `isequal(convert(SMatrix{2,2}, I), convert(SMatrix{3,3}, I))` shouldn't differ from `isequal(I, I)`, which must be `true`. It's also typically bi-directional/information conserving. (There are some accommodations made for converting `Float64` to `Float32` but even these are annyoing - and they are a different "flavour" of information change compared to adding/removing a `size`).
> 
[...]

PR #759, opened 2020-04-01
Fix convert(::Type{<:StaticArray}, ::UniformScaling)

c42f:
Fix for #758

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -102,6 +102,7 @@ end
 # because we do not have `SArray{Tuple{2,3}} <: StaticMatrix`.
 (::Type{SM})(I::UniformScaling) where {SM<:(StaticArray{Tuple{N,M}} where {N,M})} =
     _scalar_matrix(Size(SM), SM, I.λ)
+convert(T::Type{<:StaticArrayLike}, I::UniformScaling) = T(I)
 
 # Construct a matrix with the scalar λ on the diagonal and zeros off the
 # diagonal. The matrix can be non-square.
--- a/test/linalg.jl
+++ b/test/linalg.jl
@@ -75,6 +75,13 @@ using LinearAlgebra: checksquare
         @test @inferred(SDiagonal{3}(2.0I))::SDiagonal{3,Float64} == 2I3x3
         @test @inferred(SDiagonal{3,Float64}(I))::SDiagonal{3,Float64} == I3x3
         @test @inferred(SDiagonal{3,Float32}(2.0I))::SDiagonal{3,Float32} == 2I3x3
+
+        # convert from UniformScaling falls back to constructors
+        @test @inferred(convert(SMatrix{2,2,Float64}, I)) === SA_F64[1 0; 0 1]
+        @test @inferred(convert(MMatrix{2,2,Float64}, I))::MMatrix{2,2,Float64} == SA[1 0; 0 1]
+        @test @inferred(convert(SMatrix{2,2}, I)) === SA{Bool}[1 0; 0 1]
+        @test @inferred(convert(SArray{Tuple{2,2}}, I)) === SA{Bool}[1 0; 0 1]
+        @test @inferred(convert(SDiagonal{2}, I)) === Diagonal{Bool,SArray{Tuple{2},Bool,1,2}}((1,1))
     end
 
     @testset "diagm()" begin


andyferris:
Umm, sorry, but I really think we shouldn't do this! If you are transforming an object by adding or removing information (e.g. `size`), you should use a constructor not `convert`.

`convert` relies on a whole bunch of assumptions. Or rather, methods in `Base` rely on a bunch of assumptions about `convert`. The `hash` should be the same before and after the convert. They must be `isequal`, and `isless` has to define the same ordering. It should conserve the transitivity of `isequal`, for example if you define this than `isequal(convert(SMatrix{2,2}, I), convert(SMatrix{3,3}, I))` shouldn't differ from `isequal(I, I)`, which must be `true`. It's also typically bi-directional/information conserving. (There are some accommodations made for converting `Float64` to `Float32` but even these are annyoing - and they are a different "flavour" of information change compared to adding/removing a `size`).

We can't do any of these with `I` and an `AbstractArray`, because the `size`/`axes` features in `isequal`, `hash`, `isless`, etc. Users currently can't use `convert(Matrix{Float64}, I)` so I don't see it as terribly surprising/going to change anyones generic code.

c42f:
Hm, well it's easy to revert.

But I think this means that `Base` is actually wrong to use `convert` in the lowering of constructors, and it should instead be lowering constructors... to call constructors?

andyferris:
> But I think this means that Base is actually wrong to use convert in the lowering of constructors, and it should instead be lowering constructors... to call constructors?

Yeah I was wondering about that, too. `convert` shouldn't do anything "wrong" in this case, it's just that using a constructor might cover more useful cases. But on the other hand, a constructor can do just about anything; it might wrap the input and change it's behavior instead of constructing something with similar behaviour - even `AbstractArray` doesn't have requirements on constructors. Thus you don't see a lot of calls to constructors (as type parameters) in generic code (we actually do have a few in `StaticArrays` because we _require_ the constructor from a tuple to have certain behavior).

c42f:
I mean, the situation which triggered this is essentially

```
struct Foo
    x::Bar
end
```
where the default constructor is
```
Foo(x) = Foo(convert(Bar, x))
```

The compiler doesn't give us a choice about this — other than we can override the entire `Foo` constructor — which is far from ideal when `Foo` has many fields.

The problem is that the constructor doesn't need to satisfy all those invariants you mentioned, which suggests that lowering to use `convert` is just the wrong abstraction for default constructor lowering?

c42f:
Well I've reverted this for now, and as always thanks for the useful feedback!

---

For the language there's also the "situation" with default constructors of parametric types which don't behave at all like non-parametric types. I still find the exact behavior surprisingly subtle and difficult to remember so there's probably more work in Julia 2.0 to have all that make sense.

tkf:
Can't we still create a default constructor `Foo(x) = Foo(aggressive_convert(Bar, x))` with the definition `aggressive_convert(T, x) = convert(T, x)`?  I think it's non-breaking so doable in Julia 1.x. (Though we need a better name for `aggressive_convert`.)

Or are you talking about parametric types specifically?

andyferris:
> For the language there's also the "situation" with default constructors of parametric types which don't behave at all like non-parametric types. I still find the exact behavior surprisingly subtle and difficult to remember so there's probably more work in Julia 2.0 to have all that make sense.

> Can't we still create a default constructor Foo(x) = Foo(aggressive_convert(Bar, x)) with the definition aggressive_convert(T, x) = convert(T, x)? I think it's non-breaking so doable in Julia 1.x. (Though we need a better name for aggressive_convert.)

Yes, agreed. I think there is room for something that sits between `convert` (and the implicit assumptions around it) and constructors, and this `aggressive_convert` (modulo naming) seems like a good candidate.

andyferris:
I'm thinking something like these guarantees:

 * `isequal(convert(T, x)::T, x)`
 * `aggressive_convert(T, x)::T == x`
 * `T(x) isa T`

tkf:
> * T(x) isa T

Yeah, I wish we can have this. Currently the julia manual is "allowing" that this does not have to be the case: https://docs.julialang.org/en/latest/manual/conversion-and-promotion/#Constructors-that-don't-return-instances-of-their-own-type-1

But it's definitely one of WAT moments in Julia.

andyferris:
Yeah, definitely a WAT moment; I wouldn't recommend it even in the case stated in the manual (just create a damn `flip` function! Just like `adjoint` vs `Adjoint`).

But if we're talking about Julia 2.0 we may as well hope for shiny features like constraints on generic functions or some kind of interfaces/contracts thing that could be used to enforce this :)

tkf:
> just create a damn `flip` function! Just like `adjoint` vs `Adjoint`

Yeah, totally agree :)

If we get the enforcement of `T(x) :: T` from the compiler, I think we can almost forget about `aggressive_convert`.  Though there is a convention that `MutableContainer(x)` should copy `x`.  I've been wondering if we can have some kind of ownership model so that the constructor can be lowered to something like `Foo(x) = Foo(Bar(move!(x)))` where `Bar(move!(x))` means `Bar` owns `x` so that it doesn't have to copy `x` even if `Bar` is a container.

andyferris:
Well we'd need an ownership model :) And something sensible around mutability/immutability.

I actually think Julia might work really well with ownership and mutation inference, meaning instead of `move!` the compiler just detects if there are any other active (mutable) references. Just like type inference can inline types, ellide dispatch, etc, inferring lifetimes and ownership can (a) make allocation/deallocation static (remove any GC overhead/jitter from carefully written inner loops/hot paths) and (b) if done right could reduce allocations by avoiding unnecessary copies - functional operations like `map` could be automatically done in-place like `map!` where provably safe, etc. Code that follows the rust rules could use static allocation and avoid copies; everything else gets put to the GC and allocations are made as now.

c42f:
Where are the requirements for `convert` documented?  The current docstring for `convert` says

> Convert x to a value of type T.
>
>  If T is an Integer type, an InexactError will be raised if x is not representable by T, for example if x is not integer-valued, or is outside the range supported by T.

Which is woefully inadequate if the ecosystem relies on a bunch of these invariants being true.

c42f:
> Can't we still create a default constructor `Foo(x) = Foo(aggressive_convert(Bar, x))` with the definition `aggressive_convert(T, x) = convert(T, x)`? I think it's non-breaking so doable in Julia 1.x.

Spot on. I was just coming back to write this exact thing. I was considering calling it `construct`, with the idea that in Julia 1.x we would have
```
construct(T,x) = convert(T,x)
```
but that in Julia 2.0 it would become
```
construct(T,x) = T(x)
```

I'm not sure if that's helpful though if in 2.0 we had `T(x) isa T`.

tkf:
Actually I think what I meant was `borrow!` (`share!`?) rather than `move!`.

@andyferris I think my point was that user-defined constructor needs to (effectively) provide two entry points: one that copies the inputs and one that keeps them as shared references. The compiler is free to use the latter even when the code says the former when it can prove it's safe to do it. But I think the default behavior for the constructor has to be the sharing/non-copying one to avoid changing the semantic drastically from 1.x.

> I was considering calling it `construct`

@c42f A bit aside, but I actually prefer `construct(T, properties...)` to mean [`ConstructionBase.constructorof(T)(properties...)`](https://juliaobjects.github.io/ConstructionBase.jl/dev/). I think this is a fundamental building block for dealing with immutable data types.

> in Julia 2.0 it would become
> 
> ```
> construct(T,x) = T(x)
> ```
> 
> I'm not sure if that's helpful though if in 2.0 we had `T(x) isa T`.

I think we still need this for specializations like `construct(T, x::T) where {T} = x` which is useful when, e.g., `T <: Dict`.

PR #762, opened 2020-04-13
Ability to select particular test cases

c42f:
Rearrange tests so that tests from only a single file can be run.  For example, to run only the MVector and SVector tests:

```julia
Pkg.test("StaticArrays", test_args=["MVector", "SVector"])
```


c42f:
... I know I should implement a more general solution to this in `Base.Test`. But `Base.Test` internals are so messy. Maybe one day :-)

Issue #700, opened 2019-12-19
Changes to `one()` implementation seem to break Rotations.jl


c42f:
It seems some recent changes on master have broken the Rotations tests (very likely #690).

This needs detangling before the next 0.12.2 release (Is Rotations depending on an implementation detail of StaticArrays... or did #690 somehow break the public API?)

c42f:
Fixed in #763

PR #763, opened 2020-04-13
Fix accidential API breakage in one()

c42f:
Fix release-blocking bug #700 .

It turns out that Rotations.jl relies on `one(M)` calling the constructor of `M`, so transitioning to use of `simliar_type` in #690 broke `Rotations`. 

It's arguably a bug to call the constructor rather than calling `similar_type`, but let's avoid breaking compatibility for now.

I say it's arguably a bug because some sets of matrices defined by a type cannot represent the multiplicative identity, for example the set of skew-symmetric matrices. StaticArrays should likely assume that generic operations like `one()` must devolve to a matrix type which can represent
anything in GL(n,R).

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -87,30 +87,49 @@ end
 @inline Base.zero(a::SA) where {SA <: StaticArray} = zeros(SA)
 @inline Base.zero(a::Type{SA}) where {SA <: StaticArray} = zeros(SA)
 
+@inline _construct_sametype(a::Type, elements) = a(elements)
+@inline _construct_sametype(a, elements) = typeof(a)(elements)
+
 @inline one(m::StaticMatrixLike) = _one(Size(m), m)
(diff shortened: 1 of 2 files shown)


andyferris:
Good point, Chris. Rotations are also an example where `zero(M)` can’t be an `M`.

Should we patch Rotations.jl if that is where the misunderstanding is?

c42f:
> Rotations are also an example where `zero(M)` can’t be an `M`

Yes indeed! I left a too-long comment in the code with something to that effect :-)

Rough plan was to revert this temporarily, fix `Rotations` to override `zero` and `one`, and eventually put this back again. Maybe in the next major version which might be 1.0.

c42f:
See companion issues:
* Rotations https://github.com/JuliaGeometry/Rotations.jl/issues/114
* StaticArrays https://github.com/JuliaArrays/StaticArrays.jl/issues/765

c42f:
It should be uncontroversial that we need this for the next release, so I've merged it.

Issue #489, opened 2018-09-06
logabsdet(::SMatrix) fails


tpapp:
```julia
julia> VERSION # (this is `master`, but the same happens on 1.0)
v"1.1.0-DEV.199"

julia> using LinearAlgebra, StaticArrays

julia> mx = SMatrix{2,2}(Float64.(1:4)...)
2×2 SArray{Tuple{2,2},Float64,2,4}:
 1.0  3.0
 2.0  4.0

julia> logabsdet(mx)
ERROR: MethodError: no method matching lu!(::SArray{Tuple{2,2},Float64,2,4}, ::Val{true}; check=false)
```

tpapp:
I wonder if I could just add
```julia
function logabsdet(m::SMatrix)
    LU = lu(m)
    d, s = logabsdet(LU.U)
    d, s * _parity(LU.p)
end
```
Can someone confirm that `_parity` counts the row switches? It seems to from the code.

PR #766, opened 2020-04-15
Add singularity checks for lu, method for issuccess.

tpapp:
This makes the implementation of `lu` conform to the API of LinearAlgebra: check by default, optionally skip check.

Tests and methods that use `lu` are modified accordingly.

### DIFF
--- a/src/det.jl
+++ b/src/det.jl
@@ -47,7 +47,7 @@ end
     if prod(S) ≤ 14*14
         quote
             @_inline_meta
-            LUp = lu(A)
+            LUp = lu(A; check = false)
             det(LUp.U)*_parity(LUp.p)
         end
     else
@@ -62,7 +62,7 @@ end
     if prod(S) ≤ 14*14
         quote
             @_inline_meta
-            LUp = lu(A)
+            LUp = lu(A; check = false)
             d, s = logabsdet(LUp.U)
             d + log(s*_parity(LUp.p))
         end
--- a/src/det.jl
+++ b/src/det.jl
@@ -47,8 +47,8 @@ end
     if prod(S) ≤ 14*14
         quote
             @_inline_meta
-            LUp = lu(A; check = false)
-            det(LUp.U)*_parity(LUp.p)
+            _, U, p = _lu(A, Val(true), false)
+            det(UpperTriangular(U))*_parity(p)
         end
     else
         :(@_inline_meta; det(Matrix(A)))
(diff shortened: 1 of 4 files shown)


tpapp:
When this is merged, I would proceed to fix #489, just don't want to introduce conflicts to resolve.

src/det.jl
```diff
@@ -47,7 +47,7 @@ end
     if prod(S) ≤ 14*14
         quote
             @_inline_meta
-            LUp = lu(A)
+            LUp = lu(A; check = false)
```
c42f, line 50: It may be slightly helpful to call `_lu` directly here. In other places in this package we've found that keyword arguments still sometimes cause suboptimal performance. (I'm not sure about this particular case, we never dug to the real root cause in the compiler yet.)

src/lu.jl
```diff
@@ -30,25 +30,54 @@ function Base.show(io::IO, mime::MIME{Symbol("text/plain")}, F::LU)
...
-    L, U, p = _lu(A, pivot)
+function lu(A::StaticMatrix{N,N}, pivot::Union{Val{false},Val{true}}=Val(true);
+            check = true) where {N}
+    L, U, p = _lu(A, pivot, check)
     LU(LowerTriangular(L), UpperTriangular(U), p)
 end
 
-@generated function _lu(A::StaticMatrix{M,N,T}, pivot) where {M,N,T}
+# location of the first zero on the diagonal, 0 when not found
+function __first_zero_on_diagonal(A::StaticMatrix{M,N,T}) where {M,N,T}
```
c42f, line 47: Stylistic nitpick: we've used a single leading underscore elsewhere in this package for internal functions, so I'd somewhat prefer sticking to that.

tpapp: Strictly speaking this is not true (`__lu` is the exception), but of course I am happy to change it.

src/det.jl
tpapp: Of course, I will change it.

tpapp:
I think the reason is that JuliaLang/julia/pull/27336 came after #368, while 0.7/1.0 was in the process of being finalized, and this was missed.

I would argue that since methods for `StaticMatrix` don't claim to do anything else than what is documented for `LinearAlgebra.lu`, not conforming to the latter is a bug, which this PR fixes, so no deprecation is necessary since the previous behavior was not the official API.

martinholters:
Not that I remember in detail, but I'm pretty sure I tried to follow Base API when I did #368. We then just missed to follow along when Base changed... until now.

c42f:
Thanks Tamas, let's go with this as a bugfix then.

PR #768, opened 2020-04-17
Some vcat / hcat overloads for Number + StaticArray

### LINKED ISSUE #602: vcat(::SVector, ::Number) should return an SVector
I assume this behaviour is just an oversight:
```julia
julia> vcat(SVector(1), 2)
2-element Array{Int64,1}:
 1
 2
```
Shall I make a PR overriding `vcat`(and possibly `hcat`)?

c42f:
These extra overloads should be helpful for users though it's not really
possible to fix this in generality until the dispatch mechanism in Base
is improved.

Closes #602


mateuszbaran:
I tried fixing it a bit, is that OK?

c42f:
> I tried fixing it a bit, is that OK?

More than ok, it's very much appreciated!

It appears I'd forgotten all about this.

PR #769, opened 2020-04-22
Remove show(::IO, ::Type{StaticArrays.Blah}) overloads

c42f:
Apparently these can be problematic for compiler-internal reasons, and I
think the printing has been improved such that the remaining cases of
these may not do anything useful. So let's just remove them.


c42f:
By the way, I'd say exactly why removing these is a good idea if I understood it! All I really know is that some obscure and very nasty Pumas.jl issue may have been related to the presence of these methods. And that removing them would do no great harm. CC @andreasnoack

c42f:
See also https://github.com/PainterQubits/Unitful.jl/issues/321

timholy:
Old history here: https://github.com/JuliaLang/julia/issues/13306. Not sure how much of a modern problem it is, but I've been paranoid about specializing `show` for types ever since.

c42f:
Thanks for chiming in here with the links! I only had a very vague memory of some of those past issues.

PR #770, opened 2020-04-22
Define logabsdet(::SMatrix), det & friends for LU

### LINKED ISSUE #489: logabsdet(::SMatrix) fails

tpapp:
1. define a `logabsdet` for `SMatrix`, fixes #489

2. define methods for `det`, `logdet`, `logabsdet` for LU (cf `LinearAlgebra`, which has these methods)

3. make `det`, `logdet`, `logabsdet` fall back to using the LU methods (to avoid code repetition)

4. include `lu.jl` first, since `det.jl` uses code from it

5. incidental consistency fixes

### DIFF
--- a/src/det.jl
+++ b/src/det.jl
@@ -42,13 +42,24 @@ end
     -2*rem(s, 2) + 1
 end
 
+det(F::LU) = det(F.U) * _parity(F.p)
+
+function logdet(F::LU)
+    d, s = logabsdet(F.U)
+    d + log(s * _parity(F.p))
+end
+
+function logabsdet(F::LU)
+    d, s = logabsdet(F.U)
+    d, s * _parity(F.p)
+end
+
 @generated function _det(::Size{S}, A::StaticMatrix) where S
     checksquare(A)
     if prod(S) ≤ 14*14
         quote
             @_inline_meta
-            _, U, p = _lu(A, Val(true), false)
-            det(UpperTriangular(U))*_parity(p)
+            det(lu(A, Val(true); check = false))
         end
     else
         :(@_inline_meta; det(Matrix(A)))
@@ -62,11 +73,22 @@ end
     if prod(S) ≤ 14*14
         quote
             @_inline_meta
-            LUp = lu(A; check = false)
-            d, s = logabsdet(LUp.U)
-            d + log(s*_parity(LUp.p))
+            logdet(lu(A, Val(true); check = false))
         end
     else
         :(@_inline_meta; logdet(drop_sdims(A)))
+    end
+end
+
+@inline logabsdet(A::StaticMatrix) = _logabsdet(Size(A), A)
+@generated function _logabsdet(::Size{S}, A::StaticMatrix) where S
+    checksquare(A)
+    if prod(S) ≤ 14*14
+        quote
+            @_inline_meta
+            logabsdet(lu(A, Val(true); check = false))
+        end
+    else
+        :(@_inline_meta; logabsdet(drop_sdims(A)))
     end
 end
--- a/src/det.jl
+++ b/src/det.jl
@@ -44,6 +44,13 @@ end
 
 det(F::LU) = det(F.U) * _parity(F.p)
 
+function logabsdet(A::Union{LowerTriangular{<:Any,<:StaticMatrix},
+                            UpperTriangular{<:Any,<:StaticMatrix}})
+    checksquare(A)
+    mapreduce(x -> (log(abs(x)), sign(x)), ((l1, s1), (l2, s2)) -> (l1 + l2, s1 * s2),
+              diag(A))
+end
+
 function logdet(F::LU)
     d, s = logabsdet(F.U)
     d + log(s * _parity(F.p))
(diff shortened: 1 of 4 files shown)


src/det.jl
```diff
@@ -42,13 +42,24 @@ end
...
 
+det(F::LU) = det(F.U) * _parity(F.p)
+
+function logdet(F::LU)
+    d, s = logabsdet(F.U)
+    d + log(s * _parity(F.p))
+end
+
+function logabsdet(F::LU)
+    d, s = logabsdet(F.U)
```
andyferris, line 53: What does this dispatch to?

tpapp: Good catch! It does to the default method in `LinearAlgebra` (like it did before). However, this is an opportunity for improvement, so I will do something nicer.

tpapp:
Friendly bump: it would be great if someone could review this. I am happy to make any changes necessary.

src/det.jl
```diff
@@ -44,6 +44,13 @@ end
 
 det(F::LU) = det(F.U) * _parity(F.p)
 
+function logabsdet(A::Union{LowerTriangular{<:Any,<:StaticMatrix},
+                            UpperTriangular{<:Any,<:StaticMatrix}})
+    checksquare(A)
```
andyferris, line 49: My recollection from years ago (yikes!) is that this method is slow. Have you checked if it is ellided? We can do it statically in a variety of ways, the simplest of which would be to add type parameter constraints like `<:StaticMatrix{S,S} where S`

tpapp: According to `@code_warntype`, it is fully elided.

In any case, it is used in multiple places in the package.

tpapp:
Thanks for thinking about these points, I will try to address these in turn.

1. According to `@code_warntype`, `checksquare` is elided.

2. Not sure what benchmarks you would prefer here, since this method errored before. If I compare to doing `logabsdet(A::SMatrix)` vs making it a `Matrix`, the static version is 3-10x faster, depending on matrix size. But note that LU dominates all calculations, so this is really a benchmark of that.

3. `Complex` matrices work with `logabsdet`. `Hermitian` unfortunately doesn't, but it didn't before this PR for `det` either. I would prefer if those things were addressed in another PR. Same applies to better algorithms for `Symmetric`. I can open an issue for these if you prefer.

Unless there are further suggestions, I would really appreciate if this PR could get merged. I am happy to contribute to this package but I like to do it in small pieces, each improving some functionality.

andyferris:
Alright - thanks for that @tpapp!

I agree, small pieces are definitely better :)

andyferris:
Is this or something else causing the latest test failure on master?

https://github.com/JuliaArrays/StaticArrays.jl/actions/runs/89972783

tpapp:
The errors in the log look like
```julia
matrix multiply-add: Test Failed at /home/runner/work/StaticArrays.jl/StaticArrays.jl/test/matrix_multiply_add.jl:9
  Expression: #= /home/runner/work/StaticArrays.jl/StaticArrays.jl/test/matrix_multiply_add.jl:9 =# @allocated(mul!(c, A, b)) == 0
   Evaluated: 12 == 0
Stacktrace:
 [1] test_multiply_add(::Int32, ::Int32, ::Type{T} where T) at /home/runner/work/StaticArrays.jl/StaticArrays.jl/test/matrix_multiply_add.jl:9
 [2] top-level scope at /home/runner/work/StaticArrays.jl/StaticArrays.jl/test/matrix_multiply_add.jl:175
 [3] top-level scope at /buildworker/worker/package_linux32/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1114
 [4] top-level scope at /home/runner/work/StaticArrays.jl/StaticArrays.jl/test/matrix_multiply_add.jl:172
matrix multiply-add: Test Failed at /home/runner/work/StaticArrays.jl/StaticArrays.jl/test/matrix_multiply_add.jl:9
  Expression: #= /home/runner/work/StaticArrays.jl/StaticArrays.jl/test/matrix_multiply_add.jl:9 =# @allocated(mul!(c, A, b)) == 0
   Evaluated: 12 == 0
```
so I do not think it is related. However, if you find that is, just ping me and I am happy to contribute a fix quickly.

c42f:
I suspect those test failures may be due to (1) the unreliability of `@allocated` in 1.4+, and (2) compiler changes in Julia master (though who knows exactly what... may be the inline-allocation changes which just landed in time for 1.4... they'd presumably have a big impact on codegen).

So yes, nothing to do with the code here.

c42f:
Attempted quick fix for CI: https://github.com/JuliaArrays/StaticArrays.jl/pull/776

PR #781, opened 2020-05-04
remove unnecessary definition of `!=`

SimonDanisch:
`!=` should inline to `!(a == b)` anyways, but by defining those, there is a massive amount of method invalidation going on.
Compile times on master:
```
  5.000174 seconds (3.63 M allocations: 227.198 MiB, 0.06% gc time)
```
Compile times on this branch:
```
  4.748623 seconds (3.63 M allocations: 227.200 MiB, 0.05% gc time)
```

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -129,9 +129,6 @@ Length(::Size{S}) where {S} = _Length(S...)
 @pure Base.:(==)(::Size{S}, s::Tuple{Vararg{Int}}) where {S} = S === s
 @pure Base.:(==)(s::Tuple{Vararg{Int}}, ::Size{S}) where {S} = s === S
 
-@pure Base.:(!=)(::Size{S}, s::Tuple{Vararg{Int}}) where {S} = S !== s
-@pure Base.:(!=)(s::Tuple{Vararg{Int}}, ::Size{S}) where {S} = s !== S
-
 @pure Base.prod(::Size{S}) where {S} = prod(S)
 
 Base.LinearIndices(::Size{S}) where {S} = LinearIndices(S)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
With https://github.com/JuliaLang/julia/blob/master/src/gf.c#L1492 set to 1, I found some more.
Biggest offenders are (locs -> lines of code of invalidated methods):
```julia
~300locs
>> StaticArrays.unsafe_convert(...) Tuple{typeof(Base.unsafe_convert), Type{Ptr{T}}, Base.RefValue{FA}} where FA<:StaticArrays.FieldArray{N, T, D} where D where T where N
~500locs
>> StaticArrays.similar(...) Tuple{typeof(Base.similar), Type{A}, StaticArrays.Size{S}} where S where A<:(AbstractArray{T, N} where N where T)
~1000 locs
>> StaticArrays.similar(...) Tuple{typeof(Base.similar), A, Type{T}, StaticArrays.Size{S}} where S where T where A<:(AbstractArray{T, N} where N where T)
~800 locs
>> StaticArrays.convert(...) Tuple{typeof(Base.convert), Type{SA}, Tuple} where SA<:(StaticArrays.StaticArray{S, T, N} where N where T where S<:Tuple)
```

KristofferC:
Many of these seems to be the case where there is something like `Type{T} where T <: ...` which I guess is problematic due to the `Union{}` case.

BeastyBlacksmith:
See also the list in  https://github.com/JuliaPlots/Plots.jl/issues/2676

andyferris:
> Many of these seem to happen where there is something like Type{T} where T <: ... which I guess is problematic due to the Union{} case.

Yeah the `Size` being stored in a form like `Tuple{3}` is annoying; we have `Size{S} where S <: Tuple` when I would prefer something like `Size{S} where S isa Tuple{Vararg{Int}}` :) Similarly for `AbstractArray{T <: Any, N isa Int}` or even the `N` in `Vararg` and `NTuple`. Not only is this more precise but also gets rid of the `Union{}` intersection.

PR #783, opened 2020-05-05
Sized AbstractArray

mateuszbaran:
Allocation-free statically sized views in Julia 1.5 🎉

The code should be more-or-less complete but I need to add more tests.

One thing that is potentially controversial (and subject to change) is allowing one way of reshaping. Sometimes I represent arrays of different sizes as parts of a long 1-dimensional array, I could of course call `reshape` before wrapping in `SizedArray` but that's easy enough to do directly in `SizedArray`, and currently `SizedArray` allows for even more reshaping. Relevant PR: #666.

This PR supersedes #341 .

### DIFF
--- a/src/SizedArray.jl
+++ b/src/SizedArray.jl
@@ -208,12 +208,3 @@ function Base.view(
     new_size = new_out_size(S, indices...)
     return SizedArray{new_size}(view(a.data, indices...))
 end
-
-function Base.view(
-    a::MArray{S},
-    indices::Union{Integer, Colon, StaticVector, Base.Slice, SOneTo}...,
-) where {S}
-    new_size = new_out_size(S, indices...)
-    view_from_invoke = invoke(view, Tuple{AbstractArray, typeof(indices).parameters...}, a, indices...)
-    return SizedArray{new_size}(view_from_invoke)
-end
--- a/src/SizedArray.jl
+++ b/src/SizedArray.jl
@@ -117,7 +117,7 @@ end
 @propagate_inbounds getindex(a::SizedArray, i::Int) = getindex(a.data, i)
 @propagate_inbounds setindex!(a::SizedArray, v, i::Int) = setindex!(a.data, v, i)
 
-const SizedVector{S,T,M} = SizedArray{Tuple{S},T,1,M,Array{T,M}}
+const SizedVector{S,T} = SizedArray{Tuple{S},T,1,1}
 
 @inline function SizedVector{S}(a::TData) where {S,T,TData<:AbstractVector{T}}
     return SizedArray{Tuple{S},T,1,1,TData}(a)
@@ -136,7 +136,7 @@ end
     return SizedVector{S,T}(a.data)
 end
 
-const SizedMatrix{S1,S2,T,M} = SizedArray{Tuple{S1,S2},T,2,M,Array{T,M}}
+const SizedMatrix{S1,S2,T} = SizedArray{Tuple{S1,S2},T,2}
 
 @inline function SizedMatrix{S1,S2}(
     a::TData,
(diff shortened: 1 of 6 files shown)


c42f:
The CI failures here are weird - it's a git error at checkout where supposedly  d2adcd0706644b3fab095bb0b2fe9c1080bea20f doesn't exist. But it exists in my local repo.

mateuszbaran:
I've pushed a new commit and the tests seem to be running OK.

I have one question here: do you think `view(::SizedArray, something...)` should (where possible) return a `SizedArray` wrapping a view of the wrapped `AbstractArray`?

andyferris:
> I have one question here: do you think view(::SizedArray, something...) should (where possible) return a SizedArray wrapping a view of the wrapped AbstractArray?

To me - the ideal behavior has `axes(getindex(a, b)) === axes(view(a, b))`.

If this involves effort I would wait until we support `axes` which are "partially static" (a mixture of `OneTo` and `SOneTo`) and we can clean it all up then.

mateuszbaran:
> To me - the ideal behavior has `axes(getindex(a, b)) === axes(view(a, b))`.

As far as I can tell, this is already the case 🙂 . I'm primarily concerned that without swapping of `SubArray` and `SizedArray` when taking a `view` lots of methods from `StaticArrays.jl` couldn't be used.

mateuszbaran:
I think this is ready for review now. I have finished `view`s of `SizedArray` and `MArray`.

src/SizedArray.jl
```diff
@@ -53,43 +64,156 @@ end
...
+    return Array(reshape(sa.data, size_to_tuple(S)))
+end
+@inline function Base.Array{T,N}(sa::SizedArray{S,T,N}) where {T,S,N}
+    return Array(reshape(sa.data, size_to_tuple(S)))
+end
 
-@inline convert(::Type{Array}, sa::SizedArray) = sa.data
-@inline convert(::Type{Array{T}}, sa::SizedArray{S,T}) where {T,S} = sa.data
-@inline convert(::Type{Array{T,N}}, sa::SizedArray{S,T,N}) where {T,S,N} = sa.data
+@inline function convert(::Type{Array}, sa::SizedArray{S}) where {S}
```
andyferris, line 95: This only valid if `sa.data isa Array` - so we'll need a fallback that converts if not.

andyferris: Oh never mind I see you did it below!

src/SizedArray.jl
```diff
@@ -53,43 +64,156 @@ end
...
+    a::SizedArray{S},
+    indices::Union{Integer, Colon, StaticVector, Base.Slice, SOneTo}...,
+) where {S}
+    new_size = new_out_size(S, indices...)
+    return SizedArray{new_size}(view(a.data, indices...))
+end
 
-function promote_rule(::Type{<:SizedArray{S,T,N,M}}, ::Type{<:SizedArray{S,U,N,M}}) where {S,T,U,N,M}
-    SizedArray{S,promote_type(T,U),N,M}
+function Base.view(
```
andyferris, line 212: We really need to move this to the MArray.jl file.

mateuszbaran: Sure, I'll move it.

mateuszbaran: I think this can be marked as resolved?

mateuszbaran:
I've fixed one remaining problem with views. Failure on nightly is related to a change in printing of types.

chriselrod:
Could you define `parent` now that the `parent` array's properties determine those of the `SizedArray`?

mateuszbaran:
Sure, I've just added `parent` (and `vec`).

mateuszbaran:
Are there any plans to release StaticArrays 0.13 sometime soonish? If not, I think I'll put sized abstract arrays in HybridArrays.jl for now. It would solve many problems with `SSubArray`.

c42f:
@mateuszbaran I approved this, so I was kind of waiting for you to go ahead and merge it. Go on :)

Regarding release, I was imagining we should just release 1.0 next, following the most conservative release plan. IIRC there was one other minor breaking change we should do.

mateuszbaran:
Great! I will merge it.

Releasing 1.0 would be quite a big thing and I'm happy with your plan about it :+1: .

PR #789, opened 2020-05-14
[test] set version specific bounds in detect_ambiguities

sjkelly:
Followup to #787


c42f:
Much appreciated, thanks!

PR #793, opened 2020-05-19
Update CI for Julia 1.4

andyferris:


andyferris:
What are the three (required) checks at the bottom? How do we change them to Julia 1.4?

tkf:
I think that's the status check I requested https://github.com/JuliaArrays/StaticArrays.jl/pull/776#issuecomment-623044559

I don't have access to it, but I think that's in https://github.com/JuliaArrays/StaticArrays.jl/settings/branches

c42f:
Oh that was confusing. I wish the required checks were maintained as part of a yml config and not hidden away in the admin interface :-/

c42f:
Ok, I've updated the config and merged this at the same time to hopefully prevent other PRs from becoming broken.

andyferris:
> I wish the required checks were maintained as part of a yml config and not hidden away in the admin interface :-/

Hmm... I guess you even could self-refrentially maintain a branch protection on the branch itself? Have github check the config prior to a change to check if the pusher has privileges?

PR #801, opened 2020-06-16
Fix `diff` of nested static arrays

c42f:
The eltype of these can just be inferred based on values from the
element tuple (correct algebraic alternative in the type domain would be
to use `zero(eltype(a))` rather than `one(eltype(a))`, but this wouldn't
work for arrays inside static arrays.)

Should fix a bug reported on slack by @pablosanjose

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -302,8 +302,8 @@ end
 
     return quote
         @_inline_meta
-        T = typeof(one(eltype(a)) - one(eltype(a)))
-        @inbounds return similar_type(a, T, Size($Snew))(tuple($(exprs...)))
+        elements = tuple($(exprs...))
+        @inbounds return similar_type(a, eltype(elements), Size($Snew))(elements)
     end
 end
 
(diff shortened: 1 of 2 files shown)


src/mapreduce.jl
```diff
@@ -302,8 +302,8 @@ end
 
     return quote
         @_inline_meta
-        T = typeof(one(eltype(a)) - one(eltype(a)))
-        @inbounds return similar_type(a, T, Size($Snew))(tuple($(exprs...)))
+        elements = tuple($(exprs...))
+        @inbounds return similar_type(a, eltype(elements), Size($Snew))(elements)
```
andyferris, line 306: `eltype(::Tuple)` uses a recursive algorithm with `promote_typejoin`. Not sure how it behaves with long tuples? (Or mixed types?)

Way way way way way way way back we had to use our own `promote_tuple_eltype` implementation to support long tuples gracefully.

c42f: Well today in 1.4.0 we have

```
@inferred(diff(rand(SVector{100,SVector{2,Int}})))
```

being ok so perhaps this is fine?

Would it help if I added that to the tests?

andyferris: I guess? It’s hard to know what people expect us to do with their length-100 static arrays...

tkf: Torture the compiler if it ends up in a fast code and continue doing so until it brings everything down? :trollface:  (That's just my impression. Honestly, I don't know.)

c42f: Yeah, I think that's accurate :-D

PR #808, opened 2020-07-09
Resolve ambiguity for `convert(::Type{<:Scalar}, ::StaticArray)`

Liozou:
This PR fixes a method ambiguity in StaticArrays.jl. I found it while testing my package (which depends on StaticArrays.jl) with [Aqua.jl](https://github.com/JuliaTesting/Aqua.jl) and it popped up because I was overloading `convert` (for something completely unrelated).

Before this PR:
```julia
julia> using StaticArrays

julia> Scalar{Int}[SVector{1,Int}(3)]
ERROR: MethodError: convert(::Type{Scalar{Int64}}, ::SVector{1,Int64}) is ambiguous. Candidates:
  convert(::Type{SA}, a::AbstractArray) where SA<:(Scalar{T} where T) in StaticArrays at /home/liozou/.julia/dev/StaticArrays/src/Scalar.jl:12
  convert(::Type{SA}, sa::StaticArray) where SA<:StaticArray in StaticArrays at /home/liozou/.julia/dev/StaticArrays/src/convert.jl:10
Possible fix, define
  convert(::Type{SA}, ::StaticArray) where SA<:(Scalar{T} where T)
Stacktrace:
 [1] setindex!(A::Vector{Scalar{Int64}}, x::SVector{1,Int64}, i1::Int64)
   @ Base ./array.jl:849
 [2] getindex(#unused#::Type{Scalar{Int64}}, x::SVector{1,Int64})
   @ Base ./array.jl:416
 [3] top-level scope
   @ none:1

julia> Scalar[SVector{1,Float64}(2.0)]
ERROR: MethodError: convert(::Type{Scalar{T} where T}, ::SVector{1,Float64}) is ambiguous. Candidates:
  convert(::Type{SA}, a::AbstractArray) where SA<:(Scalar{T} where T) in StaticArrays at /home/liozou/.julia/dev/StaticArrays/src/Scalar.jl:12
  convert(::Type{SA}, sa::StaticArray) where SA<:StaticArray in StaticArrays at /home/liozou/.julia/dev/StaticArrays/src/convert.jl:10
Possible fix, define
  convert(::Type{SA}, ::StaticArray) where SA<:(Scalar{T} where T)
Stacktrace:
 [1] setindex!(A::Vector{Scalar{T} where T}, x::SVector{1,Float64}, i1::Int64)
   @ Base ./array.jl:849
 [2] getindex(#unused#::Type{Scalar{T} where T}, x::SVector{1,Float64})
   @ Base ./array.jl:416
 [3] top-level scope
   @ none:1
```

With this PR:
```julia
julia> Scalar{Int}[SVector{1,Int}(3)]
1-element Vector{Scalar{Int64}}:
 Scalar{Int64}((3,))

julia> Scalar[SVector{1,Float64}(2.0)]
1-element Vector{Scalar{T} where T}:
 Scalar{Float64}((2.0,))
```

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -8,7 +8,7 @@
 
 # this covers most conversions and "statically-sized reshapes"
 @inline convert(::Type{SA}, sa::StaticArray) where {SA<:StaticArray} = SA(Tuple(sa))
-@inline convert(::Type{SA}, sa::StaticArray) where {SA<:Scalar} = SA((sa[],))
+@inline convert(::Type{SA}, sa::StaticArray) where {SA<:Scalar} = SA((sa[],)) # disambiguation
 @inline convert(::Type{SA}, sa::SA) where {SA<:StaticArray} = sa
 @inline convert(::Type{SA}, x::Tuple) where {SA<:StaticArray} = SA(x) # convert -> constructor. Hopefully no loops...
 
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -8,6 +8,7 @@
 
 # this covers most conversions and "statically-sized reshapes"
 @inline convert(::Type{SA}, sa::StaticArray) where {SA<:StaticArray} = SA(Tuple(sa))
+@inline convert(::Type{SA}, sa::StaticArray) where {SA<:Scalar} = SA((sa[],))
 @inline convert(::Type{SA}, sa::SA) where {SA<:StaticArray} = sa
 @inline convert(::Type{SA}, x::Tuple) where {SA<:StaticArray} = SA(x) # convert -> constructor. Hopefully no loops...
 
--- a/test/ambiguities.jl
+++ b/test/ambiguities.jl
@@ -3,9 +3,9 @@
 
 const allowable_ambiguities =
     if VERSION < v"1.1"
-        4
+        3
     elseif VERSION < v"1.2"
-        2
+        1
     else
         0
     end
--- a/test/ambiguities.jl
+++ b/test/ambiguities.jl
@@ -7,7 +7,7 @@ const allowable_ambiguities =
     elseif VERSION < v"1.2"
         2
     else
-        1
+        0
     end
 
 @test length(detect_ambiguities(Base, LinearAlgebra, StaticArrays)) <= allowable_ambiguities
--- a/test/convert.jl
+++ b/test/convert.jl
@@ -14,4 +14,8 @@ end # testset
     # Issue #651
     @testinf SVector{0,Float64}(Any[]) === SVector{0,Float64}()
     @testinf SVector{0,Float64}(Int8[]) === SVector{0,Float64}()
+
+    # PR #808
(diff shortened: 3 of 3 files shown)


Liozou:
I just noticed how close that PR is to #774 (I don't know how I didn't see that before)... If don't think they are actually equivalent, but feel free to close as a dup if you think both issues should be addressed directly in #774 for instance.
CI failures on nightly look unrelated.

tkf:
Great!  So, can we set this to 0 for modern Julia now?

https://github.com/JuliaArrays/StaticArrays.jl/blob/cee335d3e7f08dcc72075ac5f18bb3a632e8b841/test/ambiguities.jl#L4-L11

src/convert.jl
```diff
@@ -8,6 +8,7 @@
 
 # this covers most conversions and "statically-sized reshapes"
 @inline convert(::Type{SA}, sa::StaticArray) where {SA<:StaticArray} = SA(Tuple(sa))
+@inline convert(::Type{SA}, sa::StaticArray) where {SA<:Scalar} = SA((sa[],))
```
tkf, line 11: Maybe comment that it's not redundant?
```suggestion
@inline convert(::Type{SA}, sa::StaticArray) where {SA<:Scalar} = SA((sa[],))  # disambiguation
```

Liozou:
> Great! So, can we set this to 0 for modern Julia now?
> 
> https://github.com/JuliaArrays/StaticArrays.jl/blob/cee335d3e7f08dcc72075ac5f18bb3a632e8b841/test/ambiguities.jl#L4-L11

It looks like so! I just did that in the last commit.
Nightly is still broken for different reasons but I don't have any ambiguity locally for both v1.5.0-rc1 and master. Since CI smiles upon the other Julia versions, there should be no ambiguity left for julia versions >= v1.2.0! And this also means that any project relying on StaticArrays can safely use Aqua.jl (or any automatic ambiguity detection system) in their test suite without risking a test failure due to StaticArrays, which is great.

Caveat though: I'm talking about ambiguities being detected by `Test.detect_ambiguities(Base, LinearAlgebra, StaticArrays)`, which doesn't seem to see https://github.com/JuliaArrays/StaticArrays.jl/pull/774 for instance, so there might still be some internal ambiguities left.

tkf:
> Caveat though: I'm talking about ambiguities being detected by `Test.detect_ambiguities(Base, LinearAlgebra, StaticArrays)`, which doesn't seem to see #774 for instance, so there might still be some internal ambiguities left.

Oh, interesting.  So it looks like `detect_ambiguities` doesn't detect the ambiguity in the constructor unless we specify `Core` as well?  I guess this is because `parentmodule(Type) === Core`.  The output of `detect_ambiguities(Core, Base, LinearAlgebra, StaticArrays)` is rather sad...

I guess it means I should add `Core` by default in Aqua.jl (though this is a breaking change).

PR #817, opened 2020-08-24
Updated _cholesky function.

chriselrod:
You can benchmark the implementation in this PR versus the current `StaticArray`s by having the current `StaticArray`s installed and running the following script:
```julia
using StaticArrays, LinearAlgebra

@generated function uchol(A::SMatrix{M,M,T}) where {M,T}
    q = Expr(:block)
    for n in 1:M
        for m ∈ n:M
	    r = Expr(:ref, :A, n, m)
	    ln = LineNumberNode(@__LINE__,Symbol(@__FILE__))
	    r = Expr(:macrocall, Symbol("@inbounds"), ln, r)
            push!(q.args, Expr(:(=), Symbol(:L_,m,:_,n), r))
        end
        for k ∈ 1:n-1, m in n:M
            L_m_n = Symbol(:L_,m,:_,n)
            push!(q.args, Expr(:(=), L_m_n, Expr(:call, :muladd, Expr(:call, :(-), Symbol(:L_,m,:_,k)), Symbol(:L_,n,:_,k), L_m_n)))
        end
        L_n_n = Symbol(:L_,n,:_,n)
        push!(q.args, Expr(:(=), L_n_n, :sqrt, L_n_n)))
        Linv_n_n = Symbol(:Linv_,n,:_,n)
        push!(q.args, Expr(:(=), Linv_n_n, Expr(:call, :inv, L_n_n)))
        for m ∈ n+1:M
            push!(q.args, Expr(:(*=), Symbol(:L_,m,:_,n), Linv_n_n))
        end
    end
    ret = Expr(:tuple)
    for n ∈ 1:M
        for m ∈ 1:n
            push!(ret.args, Symbol(:L_,n,:_,m))
        end
        for m ∈ n+1:M
            push!(ret.args, zero(T))
        end
    end
    push!(q.args, Expr(:call, Expr(:curly, :SMatrix, M, M), ret))
    Expr(:block, Expr(:meta, :inline), q)
end

function bench(rg)
    x = Matrix{Float64}(undef, length(rg), 2);
    for (i,r) ∈ enumerate(rg)
        S = @SMatrix(randn(r,3r ÷ 2)) |> x -> x * x';
        x[i,1] = @belapsed cholesky(Symmetric($S)).U
        x[i,2] = @belapsed uchol($S)
    end
    x
end
benches = bench(1:24);
@. benches = (1:24) ^ 3 / 3e9 / benches;

using Plots; plotly();
plot(1:24, benches, labels = ["Current Implementation" "Proposed Implementation"]);
xlabel!("Matrix dimensions.")
ylabel!("GFLOPS.")
```
Results:
newplot(13)

If people are willing, we could replace the `sqrt` implementation with
```julia
fast_sqrt(x) = sqrt(x)
sqrt(x::Union{Float32,Float64}) = Base.sqrt_llvm(x)
```
This would not do domain checks on each `sqrt`, meaning it'll return `NaN`s instead of throwing an error if the matrix isn't positive definite.

The primary benefit of this PR is that it generates code for arbitrary square sizes, instead of special cased for a single size. It is by no means optimal, but it seems to be an improvement.

Currently, it switches to the old LAPACK definition above 24x24 instead of above 3x3. But by the looks of it, I should increase this threshold if anything.

### DIFF
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -14,7 +14,7 @@ end
 
 @generated function _cholesky(::Size{S}, A::StaticMatrix{M,M}) where {S,M}
     @assert (M,M) == S
-    M > 24 && return :(_cholesky_large(Size{$S}(), :A))
+    M > 24 && return :(_cholesky_large(Size{$S}(), A))
     q = Expr(:block)
     for n ∈ 1:M
         for m ∈ n:M
@@ -52,6 +52,6 @@ end
 
 # Otherwise default algorithm returning wrapped SizedArray
 @inline _cholesky_large(::Size{S}, A::StaticArray) where {S} =
-    SizedArray{Tuple{S...}}(Matrix(cholesky(Hermitian(Matrix(A))).U))
+    similar_type(A)(cholesky(Hermitian(Matrix(A))).U)
 
 LinearAlgebra.hermitian_type(::Type{SA}) where {T, S, SA<:SArray{S,T}} = Hermitian{T,SA}
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -1,5 +1,5 @@
 # Generic Cholesky decomposition for fixed-size matrices, mostly unrolled
-non_hermitian_error() = throw(L∈earAlgebra.PosDefException(-1))
+non_hermitian_error() = throw(LinearAlgebra.PosDefException(-1))
 @inline function LinearAlgebra.cholesky(A::StaticMatrix)
     ishermitian(A) || non_hermitian_error()
     C = _cholesky(Size(A), A)
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -18,21 +18,22 @@ end
     q = Expr(:block)
     for n in 1:M
         for m ∈ n:M
-	    r = Expr(:ref, :A, n, m)
-	    ln = LineNumberNode(@__LINE__,Symbol(@__FILE__))
-	    r = Expr(:macrocall, Symbol("@inbounds"), ln, r)
-            push!(q.args, Expr(:(=), Symbol(:L_,m,:_,n), r))
+            L_m_n = Symbol(:L_,m,:_,n)
+            push!(q.args, :($L_m_n = @inbounds A[$n, $m]))
         end
         for k ∈ 1:n-1, m in n:M
             L_m_n = Symbol(:L_,m,:_,n)
-            push!(q.args, Expr(:(=), L_m_n, Expr(:call, :muladd, Expr(:call, :(-), Symbol(:L_,m,:_,k)), Symbol(:L_,n,:_,k), L_m_n)))
+            L_m_k = Symbol(:L_,m,:_,k)
+            L_n_k = Symbol(:L_,n,:_,k)
+            push!(q.args, :($L_m_n = muladd(-$L_m_k, $L_n_k, $L_m_n)))
         end
         L_n_n = Symbol(:L_,n,:_,n)
-        push!(q.args, Expr(:(=), L_n_n, Expr(:call, :sqrt, L_n_n)))
+        push!(q.args, :($L_n_n = sqrt($L_n_n)))
         Linv_n_n = Symbol(:Linv_,n,:_,n)
-        push!(q.args, Expr(:(=), Linv_n_n, Expr(:call, :inv, L_n_n)))
+        push!(q.args, :($Linv_n_n = inv($L_n_n)))
         for m ∈ n+1:M
-            push!(q.args, Expr(:(*=), Symbol(:L_,m,:_,n), Linv_n_n))
+            L_m_n = Symbol(:L_,m,:_,n)
+            push!(q.args, :($L_m_n *= $Linv_n_n))
         end
     end
     push!(q.args, :(T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)))
@@ -42,13 +43,12 @@ end
             push!(ret.args, Symbol(:L_,n,:_,m))
         end
         for m ∈ n+1:M
-            push!(ret.args, Expr(:call, :zero, :T))
+            push!(ret.args, :(zero(T)))
         end
     end
-    push!(q.args, Expr(:call, Expr(:call, :similar_type, :A, :T), ret))
-    Expr(:block, Expr(:meta, :inline), q)
+    push!(q.args, :(similar_type(A, T)($ret)))
+    return Expr(:block, Expr(:meta, :inline), q)
 end
-
 
 # Otherwise default algorithm returning wrapped SizedArray
 @inline _cholesky_large(::Size{S}, A::StaticArray) where {S} =
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -28,7 +28,7 @@ end
             push!(q.args, Expr(:(=), L_m_n, Expr(:call, :muladd, Expr(:call, :(-), Symbol(:L_,m,:_,k)), Symbol(:L_,n,:_,k), L_m_n)))
         end
         L_n_n = Symbol(:L_,n,:_,n)
-        push!(q.args, Expr(:(=), L_n_n, Expr(:call, Expr(:(.), :Base, QuoteNode(:sqrt_llvm)), L_n_n)))
+        push!(q.args, Expr(:(=), L_n_n, Expr(:call, :sqrt, L_n_n)))
         Linv_n_n = Symbol(:Linv_,n,:_,n)
         push!(q.args, Expr(:(=), Linv_n_n, Expr(:call, :inv, L_n_n)))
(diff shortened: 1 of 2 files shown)


mateuszbaran:
Thanks, this is a very nice improvement. I'm just not sure about not using quotes. They are used in StaticArrays quite often and personally I find them easier to read than building expressions without them. I could rewrite it if you want.

chriselrod:
> I'm just not sure about not using quotes. They are used in StaticArrays quite often and personally I find them easier to read than building expressions without them. I could rewrite it if you want.

Go ahead.
But I'd like to point out that building an `Expr` object by pushing into it is generally the easiest way to write unrolled code.
If you start with a loop:
```julia
function loopy_func(A, B)
    M, N = size(A)
    s = zero(eltype(A))
    for n in 1:N, m in 1:M
        s += abs2(A[m,n] - B[n,m])
    end
    s
end
```
And then just replace each line with a `push!`:
```julia
function loopy_func_gen(::Size{S}, A, B) where {S}
    M, N = S
    q = Expr(:block)
    push!(q.args, :(s = zero(eltype(A))))
    for n in 1:N, m in 1:M
        push!(q.args, :(s += abs2(A[$m,$n] - B[$n,$m])))
    end
    push!(q.args, :s)
    q
end
```

mateuszbaran:
I think something like this is a bit easier to read:
```julia
@generated function _cholesky(::Size{S}, A::StaticMatrix{M,M}) where {S,M}
    @assert (M,M) == S
    M > 24 && return :(_cholesky_large(Size{$S}(), :A))
    q = Expr(:block)
    for n in 1:M
        for m ∈ n:M
            L_m_n = Symbol(:L_,m,:_,n)
            push!(q.args, :($L_m_n = @inbounds A[$n, $m]))
        end
        for k ∈ 1:n-1, m in n:M
            L_m_n = Symbol(:L_,m,:_,n)
            L_m_k = Symbol(:L_,m,:_,k)
            L_n_k = Symbol(:L_,n,:_,k)
            push!(q.args, :($L_m_n = muladd(-$L_m_k, $L_n_k, $L_m_n)))
        end
        L_n_n = Symbol(:L_,n,:_,n)
        push!(q.args, :($L_n_n = sqrt($L_n_n)))
        Linv_n_n = Symbol(:Linv_,n,:_,n)
        push!(q.args, :($Linv_n_n = inv($L_n_n)))
        for m ∈ n+1:M
            L_m_n = Symbol(:L_,m,:_,n)
            push!(q.args, :($L_m_n *= $Linv_n_n))
        end
    end
    push!(q.args, :(T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)))
    ret = Expr(:tuple)
    for n ∈ 1:M
        for m ∈ 1:n
            push!(ret.args, Symbol(:L_,n,:_,m))
        end
        for m ∈ n+1:M
            push!(ret.args, :(zero(T)))
        end
    end
    push!(q.args, :(similar_type(A, T)($ret)))
    return Expr(:block, Expr(:meta, :inline), q)
end
```

src/cholesky.jl
```diff
@@ -12,48 +12,46 @@ end
...
-        @inbounds a23 = (A[8] - a12'*a13) / a22
-        @inbounds a33 = sqrt(A[9] - a13'*a13 - a23'*a23)
-        T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
-        similar_type(A,T)((a11, zero(T), zero(T), a12, a22, zero(T), a13, a23, a33))
-    end
-end
 
 # Otherwise default algorithm returning wrapped SizedArray
-@inline _cholesky(::Size{S}, A::StaticArray) where {S} =
+@inline _cholesky_large(::Size{S}, A::StaticArray) where {S} =
```
c42f, line 54: While we're fixing this, I think we should probably change the result of `_cholesky_large` to return `similar_type`, as I think preserving the type is generally more predictable in composite expressions.

src/cholesky.jl
```diff
@@ -12,48 +12,46 @@ end
...
-        similar_type(A,T)((sqrt(A[1]), ))
+@generated function _cholesky(::Size{S}, A::StaticMatrix{M,M}) where {S,M}
+    @assert (M,M) == S
+    M > 24 && return :(_cholesky_large(Size{$S}(), :A))
+    q = Expr(:block)
+    for n in 1:M
+        for m ∈ n:M
+	    r = Expr(:ref, :A, n, m)
+	    ln = LineNumberNode(@__LINE__,Symbol(@__FILE__))
+	    r = Expr(:macrocall, Symbol("@inbounds"), ln, r)
```
c42f, line 23: Very minor note — it's usually fine/simpler to use `nothing` instead of a `LineNumberNode` if there's not really a meaningful line for the macro invocation.

c42f:
> I think something like this is a bit easier to read

Agreed, I'd be happy to see that version instead. @mateuszbaran I think you have permissions to push it directly to this branch?

Another benefit of using quoted code rather than directly constructing `Expr` is that syntax literals are a very definitive way to construct ASTs. (Will `Expr` have the exact same representation in Julia 2.0? I'm not 100% sure; there's some very powerful things which could be done by making it more expressive. I'll admit this might be a theoretical concern ;-) )

src/cholesky.jl
```diff
@@ -12,48 +12,46 @@ end
...
-    quote
-        $(Expr(:meta, :inline))
-        T = promote_type(typeof(sqrt(one(eltype(A)))), Float32)
-        similar_type(A,T)((sqrt(A[1]), ))
+@generated function _cholesky(::Size{S}, A::StaticMatrix{M,M}) where {S,M}
+    @assert (M,M) == S
+    M > 24 && return :(_cholesky_large(Size{$S}(), :A))
+    q = Expr(:block)
+    for n in 1:M
+        for m ∈ n:M
```
c42f, line 20: Tiny syntax quibble - `in` vs `∈` ? Let's choose one :)

c42f (commented):
It's great to have a proper statically sized Cholesky, thanks! Let's make the use of `similar_type` systematic.

Optionally we could clean some things up a bit, for instance using @mateuszbaran's version but I think this is largely good to go either way.

> This would not do domain checks on each sqrt, meaning it'll return NaNs instead of throwing an error if the matrix isn't positive definite.

How big is the performance benefit?

Can we arrange things so that the `NaN`s will infect the calculation such that we only do a single check for `NaN` at the end? Then throw a `PosDefException` if the check fails?

mateuszbaran:
> > 
> 
> Agreed, I'd be happy to see that version instead. @mateuszbaran I think you have permissions to push it directly to this branch?

I will try pushing these changes then :slightly_smiling_face: .

chriselrod:
Sorry for the delay there. I replaced my version with @mateuszbaran's.

src/cholesky.jl
```diff
@@ -1,5 +1,5 @@
 # Generic Cholesky decomposition for fixed-size matrices, mostly unrolled
-non_hermitian_error() = throw(LinearAlgebra.PosDefException(-1))
+non_hermitian_error() = throw(L∈earAlgebra.PosDefException(-1))
```
mateuszbaran, line 2: Let's not be too excited about replacing `in` with `∈` :smile:

chriselrod: Unfortunately, because `∈` is a unary operator, I couldn't define `const L∈nearAlgebra = LinearAlgebra`. Looks like I have to change it back.

src/cholesky.jl
c42f: Come to think of it, do we have a test for this code path? It seems maybe not - best to add one then.

c42f (approved):
Looks good to me.

I just noticed that we may not have a test for the fallback case. And if you're looking at the tests it would be worth a quick check to see whether we're covering the necessary cases. Generally I think we're mostly ok because this doesn't really branch.  Though it looks like we could do with a `ComplexF64` test case?

src/cholesky.jl
chriselrod: Was removed with the cleaner implementation.

src/cholesky.jl
chriselrod: Done.

c42f (changes requested):
I thought I'd push this PR through by adding a few tests.

But in doing so I found that this code fails for complex eltype.

Anyway, I took the liberty of pushing my test tweaks to this PR branch. The complex case will need fixing before we can merge this. (I think as an easy option it would be ok to restrict the new implementation to only real numbers, though not very satisfying.)

chriselrod:
> But in doing so I found that this code fails for complex eltype.

That was an oversight. Adding a single character fixed this.

c42f:
> Adding a single character (') fixed this.

Perfect, I love a single character fix! At a quick glance I wasn't sure whether this was meant to work for the complex case and whether I could spare the time to fix it :-)

PR #819, opened 2020-08-25
use `SOneTo` for `eachindex()`

stev47:
Currently we already have `eachindex(SA[1, 2, 3, 4]) isa SOneTo` for
vectors but not for matrices. This changes makes `eachindex` consistent.


stev47:
the ci failure seems unrelated

mateuszbaran (approved):

c42f:
I think this is fine :+1:

Side note — a more fundamental fix to this in the future might be to use a `StaticInteger`, thus,
```julia
Base.length(s::StaticArray) = static(prod(Size(s)))
```

PR #820, opened 2020-08-26
Fixes for Julia 1.6 nightly

mateuszbaran:
I'm not sure the printing changes happened at the exact version I've used here but they were quite close.


mateuszbaran:
I didn't see that #810 does the same thing...

PR #825, opened 2020-09-24
Reducing load time by restricting dest in 5-argument mul! method

mateuszbaran:
This PR restricts the type of `dest` in a 5-argument method of `mul!`. This type was too wide anyway in 0.12.4 and didn't work with anything other than what I left in `StaticVecOrMatLikeForFiveArgMulDest`. Here are the load times:
```
Julia 1.5, before:

  1.466016 seconds (3.77 M allocations: 271.731 MiB)

Julia 1.5, after:

  1.129601 seconds (2.93 M allocations: 212.950 MiB)

Julia 1.6.0-DEV.1025 (2020-09-23), before:

  1.143195 seconds (2.57 M allocations: 201.527 MiB)

Julia 1.6.0-DEV.1025 (2020-09-23), after:

  0.784221 seconds (1.73 M allocations: 142.976 MiB)
```
So it drops by about 0.35 s. Do you prefer fixing this functionality for other types of `dest` or reducing load time?

### DIFF
--- a/src/matrix_multiply_add.jl
+++ b/src/matrix_multiply_add.jl
@@ -60,10 +60,18 @@ Base.transpose(::TSize{S,:any}) where {S,T} = TSize{reverse(S),:transpose}()
 @inline mul_parent(A::StaticMatrix) = A
 @inline mul_parent(A::StaticVector) = A
 
+# Using the full StaticVecOrMatLike in dest of that one method of mul! takes a lot of load time
+const StaticVecOrMatLikeForFiveArgMulDest{T} = Union{
+    StaticVector{<:Any, T},
+    StaticMatrix{<:Any, <:Any, T},
+    Transpose{T, <:StaticVecOrMat{T}},
+    Adjoint{T, <:StaticVecOrMat{T}}
+}
+
 # 5-argument matrix multiplication
 #    To avoid allocations, strip away Transpose type and store tranpose info in Size
-@inline LinearAlgebra.mul!(dest::StaticVecOrMatLike, A::StaticVecOrMatLike, B::StaticVecOrMatLike,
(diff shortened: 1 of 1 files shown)


mateuszbaran:
Crazy idea: let's split some of the less commonly used structured multiplication patterns into a separate library. Depending on how much exactly is cut here, we could achieve even about 0.4s loading time on Julia 1.6.

KristofferC:
How about if you just loop over the union and definite one method each for each concrete type of `dest` instead of defining it for the `Union`? Just interesting to see if it does something.

mateuszbaran:
I tried looping over all `UnionAll` types grouped in `StaticVecOrMatLike` and separately defining methods. It didn't reduce load time at all.

mateuszbaran:
@c42f any thoughts about this?

mateuszbaran:
Bump?

c42f (approved):
Thanks for bumping this, I think it's fine to merge as-is.

If somebody wants to fix it later, that's also fine, but in the meantime we may as well take the benefits.

I must say I'm very surprised by the improvement in time for changing this one method. What led you to focus on it?

mateuszbaran:
OK, I'll just solve the conflict then and merge this.

> What led you to focus on it?

I was wondering which methods cause the most slowdown and did sort of a binary search: commenting out half of the methods, then checking load time, if the time dropped significantly then focusing on this half and if not on the other one.

andyferris:
> I was wondering which methods cause the most slowdown and did sort of a binary search: commenting out half of the methods, then checking load time, if the time dropped significantly then focusing on this half and if not on the other one.

Ah - sometimes the simplest methods are the best!

PR #826, opened 2020-09-24
Add `rot180` etc. for SMatrix

mcabbott:
Before this, it returned an MMatrix:
```
julia> rot180(SA[1 2; 3 4])
 4  3
 2  1

julia> @btime rot180(m)  setup=(m=@SMatrix rand(5,5));
  12.852 ns (1 allocation: 208 bytes)
```
Afterwards, `4.373 ns (0 allocations: 0 bytes)`. (But no quicker for `rotl90`.)

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -199,6 +199,17 @@ reshape(a::Array, ::Size{S}) where {S} = SizedArray{Tuple{S...}}(a)
     return Expr(:tuple, (:(v[$i]) for i = N:(-1):1)...)
 end
 
+@eval @generated function Base.rot180(A::SMatrix{M,N}) where {M,N}
+    exs = rot180([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
+    return :(SMatrix{M,N}($(exs...)))
+end
+for rot in [:rotl90, :rotr90]
+    @eval @generated function Base.$rot(A::SMatrix{M,N}) where {M,N}
+        exs = $rot([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
+        return :(SMatrix{N,M}($(exs...)))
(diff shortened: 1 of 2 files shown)


src/abstractarray.jl
```diff
@@ -199,6 +199,17 @@ reshape(a::Array, ::Size{S}) where {S} = SizedArray{Tuple{S...}}(a)
     return Expr(:tuple, (:(v[$i]) for i = N:(-1):1)...)
 end
 
+@eval @generated function Base.rot180(A::SMatrix{M,N}) where {M,N}
```
mateuszbaran, line 202: Why is this `@eval @generated` instead of just `@generated`?

mcabbott: Because I copy-pasted it out of the loop, sorry...

mateuszbaran (approved):

src/abstractarray.jl
```diff
@@ -199,6 +199,17 @@ reshape(a::Array, ::Size{S}) where {S} = SizedArray{Tuple{S...}}(a)
     return Expr(:tuple, (:(v[$i]) for i = N:(-1):1)...)
 end
 
+@generated function Base.rot180(A::SMatrix{M,N}) where {M,N}
+    exs = rot180([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
+    return :(SMatrix{M,N}($(exs...)))
+end
+for rot in [:rotl90, :rotr90]
+    @eval @generated function Base.$rot(A::SMatrix{M,N}) where {M,N}
+        exs = $rot([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
```
c42f, line 208: Using `$rot` to implement this symbolically is somehow very pleasing :-)

Issue #782, opened 2020-05-05
zero-length MVector to SVector conversion fails (type info is lost)


tpapp:
```julia
julia> using StaticArrays

julia> m = MVector{0,Int}()
0-element MArray{Tuple{0},Int64,1,0} with indices SOneTo(0)

julia> SVector(m)
ERROR: UndefVarError: T not defined
Stacktrace:
 [1] SArray{Tuple{0},T,1,0} where T(::Tuple{}) at /home/tamas/.julia/packages/StaticArrays/mlIi1/src/SVector.jl:19
 [2] SArray at /home/tamas/.julia/packages/StaticArrays/mlIi1/src/SVector.jl:18 [inlined]
 [3] SArray{Tuple{S},T,1,S} where T where S(::MArray{Tuple{0},Int64,1,0}) at /home/tamas/.julia/packages/StaticArrays/mlIi1/src/convert.jl:5
 [4] top-level scope at REPL[32]:1
 [5] eval(::Module, ::Any) at ./boot.jl:331
 [6] eval_user_input(::Any, ::REPL.REPLBackend) at /home/tamas/src/julia-1.4/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [7] run_backend(::REPL.REPLBackend) at /home/tamas/.julia/packages/Revise/MgvIv/src/Revise.jl:1023
 [8] top-level scope at none:0
```

I am happy to make a PR, just let me know what the preferred solution is. Maybe special-case 0-length? Or carry the type in the conversion?

c42f:
This is one of those somewhat embarrassing issues which has been around for ages and which nobody got around to fixing. I think this is the same as https://github.com/JuliaArrays/StaticArrays.jl/issues/557, but maybe having two issues open for this will encourage someone to finally fix it!

I'd be very happy to have a PR fixing this. It shouldn't need to specialize the size-0 case, but make sure that `T` is preserved. Tricky things to watch out for when modifying the constructors are that the resulting internal code works for both types and values, empty and nonempty and for all staticarray subtypes.

PR #831, opened 2020-09-26
Fixing constructors from 0-element arrays

### LINKED ISSUE #557: error creating empty MVector
This seems like it should work:
```jl
julia> MVector(SVector{0,Float64}())
ERROR: UndefVarError: T not defined
Stacktrace:
 [1] MArray{Tuple{0},T,1,0} where T(::Tuple{}) at /Users/stevenj/.julia/packages/StaticArrays/3ENSR/src/MVector.jl:20
 [2] Type at /Users/stevenj/.julia/packages/StaticArrays/3ENSR/src/MVector.jl:19 [inlined]
 [3] MArray{Tuple{S},T,1,S} where T where S(::SArray{Tuple{0},Float64,1,0}) at /Users/stevenj/.julia/packages/StaticArrays/3ENSR/src/convert.jl:5
 [4] top-level scope at none:0
```

mateuszbaran:
Maybe not the prettiest solution but still fixes #557 and #782.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -71,7 +71,7 @@ end
     end
 end
 
-@inline MArray(a::StaticArray{S,T}) where {S<:Tuple,T} = MArray{size_tuple(Size(a)),T}(Tuple(a))
+@inline MArray(a::StaticArray{S,T}) where {S<:Tuple,T} = MArray{S,T}(Tuple(a))
 
 ####################
 ## MArray methods ##
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -71,7 +71,7 @@ end
     end
 end
 
-@inline MArray(a::StaticArray) = MArray{size_tuple(Size(a))}(Tuple(a))
+@inline MArray(a::StaticArray{S,T}) where {S<:Tuple,T} = MArray{size_tuple(Size(a)),T}(Tuple(a))
 
 ####################
 ## MArray methods ##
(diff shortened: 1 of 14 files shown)


mateuszbaran:
That failure looks like a Julia 1.0 bug and I don't see an obvious way to work around it 😕 .

mateuszbaran:
Apart from not solving the problem of converting empty `StaticArray`s to `SizedArray` or `SizedMatrix` on Julia 1.0, this seems to work now :slightly_smiling_face: . Anyway, I think these two cases can be fixed in another PR if someone considers it worth fixing.

c42f (approved):
Ack, sorry for the delay in replying to this. I'm a little unsure about this solution in the sense that it requires a new constructor for each StaticArray subtype. I feel like that shouldn't be necessary...

However, I also know that the generic dispatch for StaticArray constructors is quite hairy and hard to modify and that this does fix the bugs. So I think on balance we should just go with it.

(Regarding 1.0 - at this point it's basically EOL and I'm looking forward to removing support for it once 1.6 (or 1.7?) is declared the new LTS. So I'm not too worried about that.)

mateuszbaran:
> I'm a little unsure about this solution in the sense that it requires a new constructor for each StaticArray subtype. I feel like that shouldn't be necessary...

After working with this a bit I feel that a part of hairiness of static array constructors is trying to define things for all subtypes of `StaticArray`. But I may be wrong. At the very least not all subtypes of `StaticArray` are constructible just from elements.

src/MArray.jl
```diff
@@ -71,7 +71,7 @@ end
     end
 end
 
-@inline MArray(a::StaticArray) = MArray{size_tuple(Size(a))}(Tuple(a))
+@inline MArray(a::StaticArray{S,T}) where {S<:Tuple,T} = MArray{size_tuple(Size(a)),T}(Tuple(a))
```
mateuszbaran, line 74:
```suggestion
@inline MArray(a::StaticArray{S,T}) where {S<:Tuple,T} = MArray{S,T}(Tuple(a))
```

PR #833, opened 2020-09-28
Change order of function definition in matrix multiplication code

mateuszbaran:
This is a bugfix for a bug introduced in #814 . In some circumstances generated functions from `matrix_multiply_add.jl` don't see methods of `gen_by_access` and `uplo_access` defined in `matrix_multiply.jl` (generated functions somehow get instantiated in an earlier world age). This moves all required methods so that they are defined earlier. Unfortunately I don't know how to make a test for this.

Thanks @krzysztofrzecki for discovering this problem!


KristofferC:
> (generated functions somehow get instantiated in an earlier world age)

The generator in a generated function only has access to methods in the world age of when the method is defined.

mateuszbaran:
Yes, the part that I don't understand is why did it work before :smile: . One of the many mysteries of generated functions.

c42f:
> Yes, the part that I don't understand is why did it work before

As far as I know precompiled modules are loaded into the main session as a single world age. That is, functions within the module behave as if they were all defined at the same time.

So the order shouldn't matter unless the generated function is actually called during precompilation itself.

In what circumstances did the bug manifest?

c42f (approved):
Looks fine to me btw. Would be nice to know why it's necessary, but :man_shrugging:

KristofferC:
Maybe by running with `--compiled-modules=no`?

c42f:
Ah of course, that would do it!

I knew I'd encountered this kind of thing before in some package. I remember now it was Uniful.jl and there was a need to run with `--compiled-modules=no` due to some pycall problem on windows, or some such thing.

mateuszbaran:
> In what circumstances did the bug manifest?

Matrix multiplication was invoked from pyjulia with `compiled_modules=False` so it looks like that's it :+1: .

PR #837, opened 2020-10-09
More general division by triangular matrices

mateuszbaran:
I've added right division by triangular matrices, division of structured matrices and removed some related dead code.

### DIFF
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1,11 +1,19 @@
-@inline transpose(A::LinearAlgebra.LowerTriangular{<:Any,<:StaticMatrix}) =
-    LinearAlgebra.UpperTriangular(transpose(A.data))
-@inline adjoint(A::LinearAlgebra.LowerTriangular{<:Any,<:StaticMatrix}) =
-    LinearAlgebra.UpperTriangular(adjoint(A.data))
-@inline transpose(A::LinearAlgebra.UpperTriangular{<:Any,<:StaticMatrix}) =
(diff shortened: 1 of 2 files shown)


src/triangular.jl
```diff
@@ -1,11 +1,19 @@
...
+@inline adjoint(A::UnitLowerTriangular{<:Any,<:StaticMatrix}) =
+    UnitUpperTriangular(adjoint(A.data))
+@inline transpose(A::UpperTriangular{<:Any,<:StaticMatrix}) =
+    LowerTriangular(transpose(A.data))
+@inline adjoint(A::UpperTriangular{<:Any,<:StaticMatrix}) =
+    LowerTriangular(adjoint(A.data))
+@inline transpose(A::UnitUpperTriangular{<:Any,<:StaticMatrix}) =
+    UnitLowerTriangular(transpose(A.data))
+@inline adjoint(A::UnitUpperTriangular{<:Any,<:StaticMatrix}) =
+    UnitLowerTriangular(adjoint(A.data))
```
c42f, line 16: It's surprising that Base doesn't seem to include these rules.

c42f (approved):
As always, I love a good PR which removes code... but this adds functionality as well, very nice :)

src/triangular.jl
mateuszbaran: Is it worth opening an issue in Julia repository?

c42f: Sure. I can't think of a reason these shouldn't be structure-preserving.

PR #843, opened 2020-10-27
Adjust to upcoming Base Vararg change

Keno:
There's two changes here:
1. Drop unnecessary `<:` in Vararg. Tuples are already covariant and this form will be deprecated.
2. Use Base.unwrapva, rather than open coding the equivalent, since the implementation will change.

With this change, everything should work with or without https://github.com/JuliaLang/julia/pull/38136.

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -153,9 +153,9 @@ number of dimensions, and their dimensions match as determined by [`dimmatch`](@
 """
 @pure sizematch(::Size{S1}, ::Size{S2}) where {S1, S2} = sizematch(S1, S2)
 @inline sizematch(::Tuple{}, ::Tuple{}) = true
-@inline sizematch(S1::Tuple{Vararg{<:StaticDimension, N}}, S2::Tuple{Vararg{<:StaticDimension, N}}) where {N} =
+@inline sizematch(S1::Tuple{Vararg{StaticDimension, N}}, S2::Tuple{Vararg{StaticDimension, N}}) where {N} =
     dimmatch(S1[1], S2[1]) && sizematch(Base.tail(S1), Base.tail(S2))
-@inline sizematch(::Tuple{Vararg{<:StaticDimension}}, ::Tuple{Vararg{<:StaticDimension}}) = false # mismatch in number of dimensions
+@inline sizematch(::Tuple{Vararg{StaticDimension}}, ::Tuple{Vararg{StaticDimension}}) = false # mismatch in number of dimensions
 
 """
     sizematch(::Size, A::AbstractArray)
--- a/src/util.jl
+++ b/src/util.jl
@@ -16,10 +16,7 @@ end
 @generated function promote_tuple_eltype(::Union{T,Type{T}}) where T <: Tuple
     t = Union{}
     for i = 1:length(T.parameters)
-        tmp = T.parameters[i]
-        if tmp <: Vararg
-            tmp = tmp.parameters[1]
-        end
+        tmp = Base.unwrapva(T.parameters[i])
         t = :(promote_type($t, $tmp))
     end
     return quote


Keno:
@mateuszbaran @c42f @andyferris If this looks good, could I get it merged/tagged? I'd like to run PkgEval on my Base change, but of course a lot of packages depend on StaticArrays, so won't get tested without this.

mateuszbaran:
I think it looks fine and can be merged. Tagging is a bit of an issue though because master currently has a few slightly breaking changes and we may need to tag a breaking release.

fredrikekre (approved):

fredrikekre:
> Tagging is a bit of an issue though because master currently has a few slightly breaking changes and we may need to tag a breaking release.

That means a PkgEval won't be possible soon anyway since it will take time for all the ecosystem to upgrade... What are the changes? Are they truly BREAKING or are they "technically breaking"?

KristofferC:
Perhaps try running a PkgEval with the StaticArrays master vs the latest StaticArray release and see what the results are.

Keno:
Hmm? 0.12.5 was tagged from master two days ago and is the latest commit. Am I missing something?

Keno:
Oh, I see. It wasn't merged.

Keno:
I'll leave it to the maintainers, but if the next release would be a while, it'd be great if we could tag an 0.12.5 that's just 0.12.4 + this commit, such that the Base PR can move forward.

mateuszbaran:
> That means a PkgEval won't be possible soon anyway since it will take time for all the ecosystem to upgrade... What are the changes? Are they truly BREAKING or are they "technically breaking"? This will be extremely disruptive.

Nothing truly major. I think the biggest change is #783 that slightly changed how `SizedArray` works (including changing what `view` returns for `MArray` and `SizedArray` and what `parent` returns for `SizedArray`). #819 may be a small surprise for people who use `eachindex` on static arrays. #814 may cause some performance regressions on Julia <1.5 because it was tuned for the new memory layout.

Given how few packages actually use `SizedArray` according to JuliaHub it would most likely be less disruptive to tag master as a non-breaking release. From a quick look Altro.jl. RobotDynamics.jl and TrajectoryOptimization.jl may need some patching, as well as one package maintained by me (HybridArrays.jl but I already have a patch).

andyferris (approved):

andyferris:
@c42f what if we release 0.13.0 under the premise of it being a “release candidate” for 1.0.0?

Keno:
If the next release won't be a patch release, then I'd still like to request a 0.12.5 as 0.12.4 + this patch, so existing tagged packages with compat upper bounds on 0.12 will pick up this change.

andyferris:
I see. Sounds like we'll have to do something, then. It's not to hard to backport a few bugfixes.

> I'll leave it to the maintainers

Aww, Keno, I kinda thought you were a maintainer ;)

Keno:
I'm pretty sure I used to have access to this repo, but that got dropped in one of the many rounds of access reshuffle ;). That's ok though. My evil plan as always been to make this package unnecessary one day, so I'll just wait until then.

andyferris:
OK I just re-invited you. I kinda think of you as our `Core` maintainer :)

I also created a release-0.12 branch, with just a couple of very conservative cherry-picks.

> My evil plan as always been to make this package unnecessary one day, so I'll just wait until then.

My hope and dream, too.

andyferris:
This seems to be pointing to the new branch now. I suspect this will be automatic. https://github.com/JuliaRegistries/General/pull/23652

Keno:
Thanks @andyferris

c42f:
> I also created a release-0.12 branch, with just a couple of very conservative cherry-picks.

Perfect, you beat me to it.

> > My evil plan as always been to make this package unnecessary one day, so I'll just wait until then.
> 
> My hope and dream, too.

Haha, yes indeed.

PR #844, opened 2020-10-27
overload Base.rest

simeonschaub:
This makes the new lhs slurping syntax return an `S/MVector` if the rhs is an `S/MArray`. See JuliaLang/julia#37410

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -288,6 +288,10 @@ end
 end
 
 if VERSION >= v"1.6.0-DEV.1334"
-    Base.rest(a::SArray, st=(nothing, 0)) = SVector(Base.rest(Tuple(a), st[2] + 1))
-    Base.rest(a::MArray, st=(nothing, 0)) = MVector(Base.rest(Tuple(a), st[2] + 1))
+    # FIXME: This always assumes one-based linear indexing and that subtypes of StaticArray
+    # don't overload iterate
+    @inline function Base.rest(a::StaticArray, (_, i) = (nothing, 0))
+        newlen = StaticArrays.tuple_prod(Size(a)) - i
+        return similar_type(typeof(a), Size(newlen))(Base.rest(Tuple(a), i + 1))
+    end
 end
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -290,8 +290,8 @@ end
 if VERSION >= v"1.6.0-DEV.1334"
     # FIXME: This always assumes one-based linear indexing and that subtypes of StaticArray
     # don't overload iterate
-    @inline function Base.rest(a::StaticArray, (_, i) = (nothing, 0))
-        newlen = StaticArrays.tuple_prod(Size(a)) - i
+    @inline function Base.rest(a::StaticArray{S}, (_, i) = (nothing, 0)) where {S}
+        newlen = tuple_prod(S) - i
         return similar_type(typeof(a), Size(newlen))(Base.rest(Tuple(a), i + 1))
     end
 end
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -286,3 +286,8 @@ end
         @inbounds return similar_type(a, promote_type(eltype(a), eltype(b)), Size($Snew))(tuple($(exprs...)))
     end
 end
+
+if VERSION >= v"1.6.0-DEV.1334"
+    Base.rest(a::SArray, st=(nothing, 0)) = SVector(Base.rest(Tuple(a), st[2] + 1))
+    Base.rest(a::MArray, st=(nothing, 0)) = MVector(Base.rest(Tuple(a), st[2] + 1))
+end
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -274,7 +274,7 @@ end
     @test @inferred(hcat(SA[1 2 3])) === SA[1 2 3]
 end
 
-if VERSION >= v"1.6.0-DEV.1334"
+@static if VERSION >= v"1.6.0-DEV.1334"
     @testset "Base.rest" begin
         x = @SVector[1, 2, 3]
         @test Base.rest(x) == x
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -276,26 +276,36 @@ end
 
 @static if VERSION >= v"1.6.0-DEV.1334"
     @testset "Base.rest" begin
-        x = @SVector[1, 2, 3]
+        x = SA[1, 2, 3]
         @test Base.rest(x) == x
         a, b... = x
-        @test b == @SVector[2, 3]
+        @test b == SA[2, 3]
     
-        x = @SMatrix[1 2; 3 4]
+        x = SA[1 2; 3 4]
         @test Base.rest(x) == vec(x)
         a, b... = x
-        @test b == @SVector[3, 2, 4]
+        @test b == SA[3, 2, 4]
+
+        a, b... = SA[1]
+        @test b == []
+        @test b isa SVector{0}
     
-        x = @MVector[1, 2, 3]
-        @test Base.rest(x) == x
-        @test pointer(Base.rest(x)) != pointer(x)
-        a, b... = x
-        @test b == @MVector[2, 3]
-    
-        x = @MMatrix[1 2; 3 4]
-        @test Base.rest(x) == vec(x)
-        @test pointer(Base.rest(x)) != pointer(x)
-        a, b... = x
-        @test b == @MVector[3, 2, 4]
+        for (Vec, Mat) in [(MVector, MMatrix), (SizedVector, SizedMatrix)]
+            x = Vec(1, 2, 3)
+            @test Base.rest(x) == x
+            @test pointer(Base.rest(x)) != pointer(x)
+            a, b... = x
+            @test b == Vec(2, 3)
+        
+            x = Mat{2,2}(1, 2, 3, 4)
+            @test Base.rest(x) == vec(x)
+            @test pointer(Base.rest(x)) != pointer(x)
+            a, b... = x
+            @test b == Vec(2, 3, 4)
+
+            a, b... = Vec(1)
+            @test b == []
+            @test b isa Vec{0}
+        end
     end
 end
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -273,3 +273,29 @@ end
     @test @inferred(hcat(SA[1,2,3])) === SMatrix{3,1}(1,2,3)
     @test @inferred(hcat(SA[1 2 3])) === SA[1 2 3]
 end
+
+if VERSION >= v"1.6.0-DEV.1334"
+    @testset "Base.rest" begin
+        x = @SVector[1, 2, 3]
+        @test Base.rest(x) == x
+        a, b... = x
+        @test b == @SVector[2, 3]
+    
+        x = @SMatrix[1 2; 3 4]
+        @test Base.rest(x) == vec(x)
+        a, b... = x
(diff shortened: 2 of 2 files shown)


test/abstractarray.jl
```diff
@@ -273,3 +273,29 @@ end
     @test @inferred(hcat(SA[1,2,3])) === SMatrix{3,1}(1,2,3)
     @test @inferred(hcat(SA[1 2 3])) === SA[1 2 3]
 end
+
+if VERSION >= v"1.6.0-DEV.1334"
```
mateuszbaran, line 277:
```suggestion
@static if VERSION >= v"1.6.0-DEV.1334"
```
This should prevent failures on older versions.

mateuszbaran:
Thanks, that's interesting. Could you also add a method for `SizedArray`?

simeonschaub:
I don't think that we can implement this for `SizedArray` efficiently, since `Base.rest` relies on the iteration state of the wrapped `AbstractArray`, so we can't say anything about the resulting length in general in a constant-prop-friendly way. All we could do is implementing it as something like `let r=Base.rest(a, st...); SizedArray{length(r)}(r) end`, but I am not sure if this is a good idea.

andyferris (changes requested):
Ah, awesome, thanks. :)

Could you make this instead for generic
`StaticVector` using `similar_type` and decrementing the length appropriately? 

I’m not sure about matrices, etc, any ideas?

mateuszbaran:
> I don't think that we can implement this for `SizedArray` efficiently, since `Base.rest` relies on the iteration state of the wrapped `AbstractArray`, so we can't say anything about the resulting length in general in a constant-prop-friendly way. All we could do is implementing it as something like `let r=Base.rest(a, st...); SizedArray{length(r)}(r) end`, but I am not sure if this is a good idea.

I've changed your implementation to
```julia
@inline Base.rest(a::SizedArray{S}, st=(nothing, 0)) where S = SizedVector{StaticArrays.tuple_prod(S)-st[2]}(Base.rest(Tuple(a), st[2] + 1))
@inline Base.rest(a::SArray, st=(nothing, 0)) = SVector(Base.rest(Tuple(a), st[2] + 1))
@inline Base.rest(a::MArray, st=(nothing, 0)) = MVector(Base.rest(Tuple(a), st[2] + 1))
```
and it constant-propagates fine?

```julia
julia> using StaticArrays

julia> @inline Base.rest(a::SizedArray{S}, st=(nothing, 0)) where S = SizedVector{StaticArrays.tuple_prod(S)-st[2]}(Base.rest(Tuple(a), st[2] + 1))

julia> @inline Base.rest(a::SArray, st=(nothing, 0)) = SVector(Base.rest(Tuple(a), st[2] + 1))

julia> @inline Base.rest(a::MArray, st=(nothing, 0)) = MVector(Base.rest(Tuple(a), st[2] + 1))

julia> as = SizedMatrix{2,2}([1 2; 3 4])
2×2 SizedMatrix{2, 2, Int64, 2, Matrix{Int64}} with indices SOneTo(2)×SOneTo(2):
 1  2
 3  4

julia> function ff(sss); a, b... = sss; return sum(b); end
ff (generic function with 1 method)

julia> ff(as)
9

julia> @code_typed ff(as)
CodeInfo(
1 ─ %1  = Base.getfield(sss, :data)::Matrix{Int64}
│         Base.arrayref(true, %1, 1)::Int64
│   %3  = Base.getfield(sss, :data)::Matrix{Int64}
│         Base.arrayref(false, %3, 1)::Int64
│   %5  = Base.getfield(sss, :data)::Matrix{Int64}
│   %6  = Base.arrayref(false, %5, 2)::Int64
│   %7  = Base.getfield(sss, :data)::Matrix{Int64}
│   %8  = Base.arrayref(false, %7, 3)::Int64
│   %9  = Base.getfield(sss, :data)::Matrix{Int64}
│   %10 = Base.arrayref(false, %9, 4)::Int64
│   %11 = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Vector{Int64}, svec(Any, Int64), 0, :(:ccall), Vector{Int64}, 3, 3))::Vector{Int64}
│         Base.arrayset(false, %11, %6, 1)::Vector{Int64}
│         Base.arrayset(false, %11, %8, 2)::Vector{Int64}
│         Base.arrayset(false, %11, %10, 3)::Vector{Int64}
│   %15 = Base.arrayref(false, %11, 1)::Int64
│   %16 = Base.arrayref(false, %11, 2)::Int64
│   %17 = Base.add_int(%15, %16)::Int64
│   %18 = Base.arrayref(false, %11, 3)::Int64
│   %19 = Base.add_int(%17, %18)::Int64
└──       return %19
) => Int64
```

simeonschaub:
Ah, I didn't look into this closely enough. I thought `SizedArray` just forwarded calls to `iterate` to the underlying `AbstractArray`, but it looks like it always uses the standard Base implementation of `iterate` for `AbstractArray`s with `IndexLinear()`, so specializing `Base.rest` on `SizedArray` as well seems sensible to me.

> ```julia
> julia> @inline Base.rest(a::SizedArray{S}, st=(nothing, 0)) where S = SizedVector{StaticArrays.tuple_prod(S)-st[2]}(Base.rest(Tuple(a), st[2] + 1))
> ```

Is converting `a` to a `Tuple` here a good idea? I thought the point of using `SizedArray` over `S/MArray` was to avoid excessive specializations due to large tuples, but I might be wrong.

simeonschaub:
> Could you make this instead for generic
> `StaticVector` using `similar_type` and decrementing the length appropriately?

I am not sure, what assumtion `StaticArray` wants to make about how its subtypes implement the iteration protocoll. I don't think it's safe to assume in general that the iteration state always consists of linear indices, which makes `Base.rest` quite hard to implement generically for all `StaticArray`s.

> I’m not sure about matrices, etc, any ideas?

The way it's implemented in Base, `Base.rest` always returns `AbstractVector` for an `AbstractArray` of any dimensions, which made the most sense to me.

mateuszbaran:
> Ah, I didn't look into this closely enough. I thought `SizedArray` just forwarded calls to `iterate` to the underlying `AbstractArray`, but it looks like it always uses the standard Base implementation of `iterate` for `AbstractArray`s with `IndexLinear()`, so specializing `Base.rest` on `SizedArray` as well seems sensible to me.

Until recently it just wrapped a normal `Array` and I've just extended it to arbitrary `AbstractArray`s to the point that I needed it. Maybe it would make sense for `SizedArray` to forward iteration to the wrapped `AbstractArray`, I don't know enough about iteration to tell. At the very least it's commonly assumed in `SizedArray` that elements are linearly indexable starting with 1. Reviewing and modifying the entire codebase to support other iteration methods would be a big project.

> Is converting `a` to a `Tuple` here a good idea? I thought the point of using `SizedArray` over `S/MArray` was to avoid excessive specializations due to large tuples, but I might be wrong.

As far as I know you get the same specializations unless you operate on the wrapped array. Converting to `Tuple` should be surprising, and people can always use the wrapped array if they choose so.

Currently, at least for me, the main reason to use `SizedArray` is to wrap `SubArray`s.

mateuszbaran:
And one more thing, you need to propagate the element type to the constructor to make it work for static arrays of length 1:
```julia
@inline Base.rest(a::SizedArray{S}, st=(nothing, 0)) where {S,T} = SizedVector{StaticArrays.tuple_prod(S)-st[2],T}(Base.rest(Tuple(a), st[2] + 1))
@inline Base.rest(a::SArray{S,T}, st=(nothing, 0)) where {S,T} = SVector{StaticArrays.tuple_prod(S)-st[2],T}(Base.rest(Tuple(a), st[2] + 1))
@inline Base.rest(a::MArray{S,T}, st=(nothing, 0)) where {S,T} = MVector{StaticArrays.tuple_prod(S)-st[2],T}(Base.rest(Tuple(a), st[2] + 1))
```

```julia
julia> a, b... = SA[1]
1-element SVector{1, Int64} with indices SOneTo(1):
 1

julia> b
0-element SVector{0, Int64} with indices SOneTo(0)
```

I think you should add this as a test case.

simeonschaub:
Hmm, the MacOS runner seems to be hanging during Pkg setup...

andyferris (changes requested):
This is nice; it's also very close to the generic version I requested. Instead of worrying about loops over concrete types it should just be:

```julia
@inline function Base.rest(a::StaticArray, (_, i) = (nothing, 0))
    newlen = StaticArrays.tuple_prod(Size(a)) - i
    return similar_type(typeof(a), Size(newlen))(Base.rest(Tuple(a), i + 1))
end
```

andyferris:
> Is converting a to a Tuple here a good idea? I thought the point of using SizedArray over S/MArray was to avoid excessive specializations due to large tuples, but I might be wrong.

The purpose of `SizedArray` is precisely to get the specializations applied to an already-allocated `Array` (or `AbstractArray` these days) - the point partly being convenience of using an already existing object, partly having a mutable container and not having to reallocate it just to apply `Size` information, and partly because `MArray` only works for `isbits` element types (so - "convenience, speed, correctness").

simeonschaub:
> This is nice; it's also very close to the generic version I requested. Instead of worrying about loops over concrete types it should just be:
> 
> ```julia
> @inline function Base.rest(a::StaticArray, (_, i) = (nothing, 0))
>     newlen = StaticArrays.tuple_prod(Size(a)) - i
>     return similar_type(typeof(a), Size(newlen))(Base.rest(Tuple(a), i + 1))
> end
> ```

See [my comment](https://github.com/JuliaArrays/StaticArrays.jl/pull/844#issuecomment-717398621) above. The problem with that is that it would make the specific implementation of `iterate` part of the implicit API of `StaticArray`. I am not sure, this is a good idea. It would also cause problems, if someone wanted to implement a `StaticArray` with offset indices.
Actually, I believe `SizedArray`s can also have offset indices, so the current implementation would not be valid then.

andyferris:
By default `StaticArray`s have one-based linear indexing - I have written much "generic" code here under this assumption. If you look at SizedArray.jl there is no preservation of the parents `keys` or `axes` or `IndexStyle` (in fact, will it even work at all with an offset array? The index is passed through). Everything, and I mean everything including *Rotations.jl* and other packages, relies on the first few lines of abstractarray.jl.

If someone wanted to specialize a static array that is a bit different they surely could but I'd expect them to overide many of the provided methods, including `rest`.

To be clear - I am very thankful that you are considering such cases, but there is *so* much work to be done to properly support `axes` and Cartesian indexing styles and `Slice` that I feel it would be worth doing the "mixed-static-dynamic-sized" arrays at the same time (e.g. 3xN matrices), which I would do from a clean slate, to be honest (and hey, it wouldn't be the first time that this package (or its predesessors) were rewritten :) ).

andyferris (approved):

src/abstractarray.jl
```diff
@@ -286,3 +286,12 @@ end
         @inbounds return similar_type(a, promote_type(eltype(a), eltype(b)), Size($Snew))(tuple($(exprs...)))
     end
 end
+
+if VERSION >= v"1.6.0-DEV.1334"
+    # FIXME: This always assumes one-based linear indexing and that subtypes of StaticArray
+    # don't overload iterate
+    @inline function Base.rest(a::StaticArray, (_, i) = (nothing, 0))
+        newlen = StaticArrays.tuple_prod(Size(a)) - i
```
mateuszbaran, line 294:
```suggestion
    @inline function Base.rest(a::StaticArray{S}, (_, i) = (nothing, 0)) where {S}
        newlen = tuple_prod(S) - i
```
This should fix the failure on Julia master.

mateuszbaran (approved):

simeonschaub:
Bump. Can this be merged?

PR #845, opened 2020-10-27
require one-based indexing for SizedArray

simeonschaub:
Things like iteration currently fail for SizedArrays with offset axes, so I think it's better to error right array. See https://github.com/JuliaArrays/StaticArrays.jl/pull/844#issuecomment-717598756


mateuszbaran (approved):
I think it's a good idea 👍 .

c42f:
Looks good to me, thanks!

PR #850, opened 2020-11-05
Update to event-driven TagBot.yml

c42f:
As documented at https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249


c42f:
I guess this will work :man_shrugging: we'll find out next we tag a release...

PR #861, opened 2020-12-06
Define elsize to make `pointer` more reliable.

chriselrod:
The added tests don't work on current master. They're in the `SizedArray` tests file because views of `MArray`s create a `SizedArray` wrapper.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.0.1"
+version = "1.0.2"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -279,4 +279,4 @@ function Base.view(
     return SizedArray{new_size}(view_from_invoke)
 end
 
-Base.elsize(::Type{<:MArray{S,T}}) where {S,T} = sizeof(eltype(T))
+Base.elsize(::Type{<:MArray{S,T}}) where {S,T} = sizeof(T)
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -278,3 +278,5 @@ function Base.view(
     view_from_invoke = invoke(view, Tuple{AbstractArray, typeof(indices).parameters...}, a, indices...)
     return SizedArray{new_size}(view_from_invoke)
 end
+
+Base.elsize(::Type{<:MArray{S,T}}) where {S,T} = sizeof(eltype(T))
--- a/src/SizedArray.jl
+++ b/src/SizedArray.jl
@@ -126,7 +126,8 @@ end
 
 Base.parent(sa::SizedArray) = sa.data
 
-Base.pointer(sa::SizedArray) = pointer(sa.data)
+Base.unsafe_convert(::Type{Ptr{T}}, sa::SizedArray) where {T} = Base.unsafe_convert(Ptr{T}, sa.data)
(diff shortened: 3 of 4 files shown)


chriselrod:
All the tests pass locally on Julia 1.5 and master (1.6). The tests that failed on 1.1are tests I added in this PR.

For now, I made the new tests only trigger on Julia >= 1.5.

andyferris (approved):

mateuszbaran:
Why is `elsize` defined for Julia < 1.5 but not tested there?

chriselrod:
Because there was a bug in old Julia versions that got fixed at some point between Julia 1.1 and Julia 1.5.

You can run this code from a released version of `StaticArrays` (you don't need this PR):
```julia
[ Info: Recompiling stale cache file /home/chriselrod/.julia/compiled/v1.1/StaticArrays/yY9vm.ji for StaticArrays [90137ffa-7385-5640-81b9-e52037218182]

julia> Base.unsafe_convert(::Type{Ptr{T}}, sa::SizedArray) where {T} = Base.unsafe_convert(Ptr{T}, sa.data)

julia> Base.elsize(::Type{SizedArray{S,T,M,N,A}}) where {S,T,M,N,A} = Base.elsize(A)

julia> Base.elsize(::Type{<:MArray{S,T}}) where {S,T} = sizeof(T)

julia> A = MMatrix{32,3,Float64}(undef);

julia> av1 = view(A, 1, :);

julia> av2 = view(A, :, 1);

julia> @test pointer(A) == pointer(av1) == pointer(av2)
Test Passed

julia> @test pointer(A, LinearIndices(A)[1,2]) == pointer(av1, 2)
Test Failed at REPL[9]:1
  Expression: pointer(A, (LinearIndices(A))[1, 2]) == pointer(av1, 2)
   Evaluated: Ptr{Float64} @0x00007f72d58e5760 == Ptr{Float64} @0x00007f72d58e5668
ERROR: There was an error during testing

julia> @test pointer(A, LinearIndices(A)[2,1]) == pointer(av2, 2)
Test Passed

julia> @test pointer(A, LinearIndices(A)[4,3]) == pointer(view(A, :, 3), 4) == pointer(view(A, 4, :), 3)
Test Failed at REPL[11]:1
  Expression: pointer(A, (LinearIndices(A))[4, 3]) == pointer(view(A, :, 3), 4) == pointer(view(A, 4, :), 3)
   Evaluated: Ptr{Float64} @0x00007f72d58e5878 == Ptr{Float64} @0x00007f72d58e5878 == Ptr{Float64} @0x00007f72d58e5688
ERROR: There was an error during testing

julia> versioninfo()
Julia Version 1.1.0
```
Basically:
```julia
julia> @test pointer(view(A, 4, :), 3) == pointer(view(A, 4:15, 1), 3) == pointer(A, 6)
Test Passed
```
`pointer(view(A, 4, :), 3)` should give you the pointer to `A[4,3]`, but instead it gives the pointer to `A[6]`, i.e. it starts from `A[4,1]` like it should, but then goes down the row by 2 (`3-1`) instead of across columns by 2.

mateuszbaran (approved):
OK, thanks for explaining it :+1: .

andyferris:
> fixed at some point between Julia 1.1 and Julia 1.5.

Is it worth pointing out that our CI could tell you _exactly_ which point between 1.1 and 1.5 we can enable the test?

(Not that I’m particularly fussed).

chriselrod:
Oh, actually some of the cancelled tests from earlier did finish. Seems it [was still broken on 1.4](https://github.com/JuliaArrays/StaticArrays.jl/runs/1505754290), so 1.5 really is the first where it works.
I could enable the other tests on the older versions, though (those that don't involve pointers on row-views).

andyferris:
Ok - if it was broken on 1.4 then this seems fine.

Issue #373, opened 2018-02-28
Implement isposdef


antoine-levitt:
`isposdef` on a SMatrix is hitting the default fallback calling isposdef!, which fails:
```
julia> a = @SMatrix [2 2;2 2]
2×2 StaticArrays.SArray{Tuple{2,2},Int64,2,4}:
 2  2
 2  2

julia> isposdef(a)
ERROR: MethodError: no method matching isposdef!(::StaticArrays.MArray{Tuple{2,2},Float64,2,4})
Closest candidates are:
```

This happened to me while trying to use DiffEq's DiffEqOperator with a static array.

c42f:
Well, this is still a problem (though the error on julia-1.1 is more explosive).

Looks like we need to either implement `isposdef` or fix our Choelsky implementation to accept the `check` keyword or some such.

VPetukhov:
Maybe a silly question, but why the following wouldn't work?

```julia
isposdef(A::StaticMatrix) = isposdef(cholesky(A))
```

StaticArrays don't call `checkpositivedefinite`, so `check` is actually `false`, and it checks for hermitian inside, so no need to explicitly convert matrix to Hermitian (which is performed in the [LinearAlgebra implementation](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/dense.jl#L91)).

c42f:
Probably not a silly question! There's just limited bandwidth to read up on all the many different linear algebra-related things people would like to do and also tackle improvements to the core static array functionality (of which there are many).

I think your implementation would be correct for `StaticMatrix`, however note that we do not have `Hermitian <: StaticMatrix`, so you'd need an extra case for hermitian or symmetric wrapped static arrays (see also `StaticArrayLike`).

If you think you know how to fix it and it matters to you, do submit a PR :-)

bjack205:
I've submitted a PR #862 that implements `isposdef` for StaticArrays. It adds a simple check in the Cholesky factorization method that allows it to fail gracefully if it's not positive definite. I also added the `check` flag to the `cholesky(::StaticArray)` method. Hopefully this is a good enough solution, because this feature is needed!

PR #862, opened 2020-12-11
Allow cholesky to fail without error

bjack205:
This allows the Cholesky factorization to continue to completion, even for indefinite matrices. This is important since Cholesky is the most efficient way to check for positive definiteness, and it's often extremely useful to attempt a Cholesky factorization and then check to see if it was successful. 

This PR adds a very simple change to the existing `_cholesky` method that checks of the real part of the diagonal element is negative prior to taking the square root, and sets the element to 0 if it is negative. This allows the algorithm to continue without errors (likely accumulating some `Inf` values along the way), and allows for a simple check for positive-definiteness after the factorization is complete. It adds minimal overhead, as shown by the graph below:

chol_comp

This change also allows easy implementation of `isposdef`, addressing issue #373. This is far more computationally efficient than converting to a dynamic matrix, as shown below:
isposdef

It's also cheaper than converting a real matrix to Complex to avoid the domain error for `sqrt` and then checking for imaginary values on the diagonal (plot not shown, but this is pretty intuitive since the conversion will take time, and operating on complex numbers requires at least 2x the number of flops). 

I'm pretty this code works generally for complex values as well, but I'm not well-versed on applying Cholesky to complex-valued matrices. I believe the condition I added into the algorithm is the correct one, namely that the real part of the diagonal elements must be positive prior to taking the square root for the matrix to be positive definite.

### DIFF
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -1,21 +1,50 @@
 # Generic Cholesky decomposition for fixed-size matrices, mostly unrolled
 non_hermitian_error() = throw(LinearAlgebra.PosDefException(-1))
-@inline function LinearAlgebra.cholesky(A::StaticMatrix)
-    ishermitian(A) || non_hermitian_error()
+@inline function LinearAlgebra.cholesky(A::StaticMatrix; check=true)
+    !ishermitian(A) && (check && non_hermitian_error())
     C = _cholesky(Size(A), A)
-    return Cholesky(C, 'U', 0)
+    loc = _checkpsd(C)
+    if check && loc != 0
+        throw(PosDefException(loc))
+    end
+    return Cholesky(C, 'U', loc)
 end
 
-@inline function LinearAlgebra.cholesky(A::LinearAlgebra.RealHermSymComplexHerm{<:Real, <:StaticMatrix})
+"""
+    _checkpsd(C::StaticMatrix{M,M,T})
+
+Checks of an upper-triangular Cholesky factor `C` was successfully factored.
+If the original matrix `A = C'C` was positive definite, return 0. 
+Otherwise, return the location of the first zero element on the diagonal of `C`.
+
+Used to fill in the `info` field of the `LinearAlgebra.Cholesky` type.
+"""
+function _checkpsd(C::StaticMatrix{M,M,T}) where {M,T}
+    ispsd = abs(prod(diag(C))) > zero(real(eltype(C))) 
+    loc = ispsd ? 0 : findfirst(diag(C) .== zero(eltype(C)))
+    return loc
+end
+
+function LinearAlgebra.isposdef(A::StaticMatrix{M,M,T}) where {M,T}
+    C = _cholesky(Size(A), A)
+    return real(prod(diag(C))) > zero(real(T)) 
+end
+
+@inline function LinearAlgebra.cholesky(A::LinearAlgebra.RealHermSymComplexHerm{<:Real, <:StaticMatrix}; check=true)
     C = _cholesky(Size(A), A.data)
-    return Cholesky(C, 'U', 0)
+    loc = _checkpsd(C)
+    if check && loc != 0
+        throw(PosDefException(loc))
+    end
+    return Cholesky(C, 'U', loc)
 end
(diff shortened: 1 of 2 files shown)


src/cholesky.jl
```diff
@@ -1,21 +1,50 @@
...
+"""
+function _checkpsd(C::StaticMatrix{M,M,T}) where {M,T}
+    ispsd = abs(prod(diag(C))) > zero(real(eltype(C))) 
+    loc = ispsd ? 0 : findfirst(diag(C) .== zero(eltype(C)))
+    return loc
 end
 
-@inline function LinearAlgebra.cholesky(A::LinearAlgebra.RealHermSymComplexHerm{<:Real, <:StaticMatrix})
+function LinearAlgebra.isposdef(A::StaticMatrix{M,M,T}) where {M,T}
+    C = _cholesky(Size(A), A)
```
mateuszbaran, line 29: Shouldn't this check first that `A` is Hermitian? Like here: https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/dense.jl#L70

src/cholesky.jl
```diff
@@ -1,21 +1,50 @@
 # Generic Cholesky decomposition for fixed-size matrices, mostly unrolled
 non_hermitian_error() = throw(LinearAlgebra.PosDefException(-1))
-@inline function LinearAlgebra.cholesky(A::StaticMatrix)
-    ishermitian(A) || non_hermitian_error()
+@inline function LinearAlgebra.cholesky(A::StaticMatrix; check=true)
+    !ishermitian(A) && (check && non_hermitian_error())
```
mateuszbaran, line 4:
```suggestion
    check && (!ishermitian(A) && non_hermitian_error())
```
This way if `check` is false then the test whether `A` is Hermitian isn't performed.

src/cholesky.jl
```diff
@@ -1,21 +1,50 @@
...
+    if check && loc != 0
+        throw(PosDefException(loc))
+    end
+    return Cholesky(C, 'U', loc)
+end
+
+"""
+    _checkpsd(C::StaticMatrix{M,M,T})
+
+Checks of an upper-triangular Cholesky factor `C` was successfully factored.
```
mateuszbaran, line 16:
```suggestion
Check if an upper-triangular Cholesky factor `C` was successfully factored.
```

src/cholesky.jl
mateuszbaran: By the way, `isposdef` is defined in `LinearAlgebra` for objects of type `Cholesky`: https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/cholesky.jl#L585 , and it's a different definition.

src/cholesky.jl
bjack205: Improved this so that it first checks if the matrix is Hermitian and then skips computing the factorization if it fails. Also eliminated duplicate code by calling the Hermitian method.

bjack205 (commented):
Thanks for the review. I addressed all your suggestions, eliminating some duplicate code and falling back to the `ispodef` methods in Base. Thanks for the tip!

src/cholesky.jl
mateuszbaran: Thanks, right, it looks better now.

test/chol.jl
```diff
@@ -86,6 +86,28 @@ using LinearAlgebra: PosDefException
             @test (@inferred c \ v) isa SVector{3,elty}
             @test c \ v ≈ c_a \ v_a
         end
+
+        @testset "indefinite matrices" for i = 1:10
```
mateuszbaran, line 90: Is this `i` used inside somewhere?

bjack205: No, just used to iterate over several random matrices, similar to Line 33.

mateuszbaran: Right, it just tests this for 10 random matrices, ok.

mateuszbaran (approved):

bjack205:
Just bumping this. Is this waiting on anything on my end?

mateuszbaran:
I think https://github.com/JuliaArrays/StaticArrays.jl/pull/886 has already solved it in a bit different way.

mateuszbaran:
I'm sorry this couldn't have been merged, things that may have some negative (even small) impact I think need another reviewer and no one else responded.

PR #863, opened 2020-12-26
Reduce latency with a few precompiles

timholy:


andyferris:
Ah, cool. Having some targeted precompiles seems quite reasonable in this package. 

I’m eying off some of the sizes here - things like 2x4 matrix multiplication. Simply out of curiosity - does the same mechanism function correctly from Makie or do the precompiles need to be “package local” for all the machinery to work?

timholy:
> Simply out of curiosity - does the same mechanism function correctly from Makie or do the precompiles need to be “package local” for all the machinery to work?

It depends: if `SomePkg.dostuff()` *inferrably* calls StaticArrays methods, there's no need to have them here because the MethodInstance backedges will link back to methods owned by `SomePkg`. However, if StaticArrays methods are called via runtime dispatch, the only place they can be is here. Basically, they have to attach somewhere and if they are toplevel as far as inference is concerned then the only attachment point is the module that defined them.

Makie has a bit of work to do in improving its inference (https://github.com/JuliaPlots/Makie.jl/issues/792); with a bit of effort there will be no need to have these at all, and especially no need for weird sizes. Happy to delete the non-square sizes if you like.

SimonDanisch (approved):

timholy:
I decided to limit this to square matrices, since Makie can ultimately solve any specific cases with better inferrability. I also generalized the sizes for many of the operations.

PR #867, opened 2021-01-06
Better handling of subtypes of StaticVector in views of SizedArray

mateuszbaran:
While working on https://github.com/mateuszbaran/HybridArrays.jl/issues/31 I noticed that the code for calculating size of views of `SizedArray` doesn't work for some subtypes of `StaticVector`. This PR fixes this problem.

### DIFF
--- a/src/SizedArray.jl
+++ b/src/SizedArray.jl
@@ -200,13 +200,17 @@ end
 
 ### Code that makes views of statically sized arrays also statically sized (where possible)
 
+# _get_static_vector_length is used in a generated function so using a generic function
+# may not be a good idea
+_get_static_vector_length(::Type{<:StaticVector{N}}) where {N} = N
+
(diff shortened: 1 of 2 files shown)


src/SizedArray.jl
```diff
@@ -200,13 +200,17 @@ end
 
 ### Code that makes views of statically sized arrays also statically sized (where possible)
 
+# _get_static_vector_length is used in a generated function so using a generic function
+# may not be a good idea
```
c42f, line 204: Should be ok, I think. How about this?

```suggestion
# Note, _get_static_vector_length is used in a generated function so it's strictly internal and can't be extended
```

c42f (approved):
Looks good to me.

src/SizedArray.jl
mateuszbaran: Thanks, that's better.

PR #872, opened 2021-01-17
Maintenance around Documenter.jl

### LINKED ISSUE #870: Fail deploying documents since switching to GitHub Actions
There are no deployed documents on `gh-pages` branch since the PR #752.
https://github.com/JuliaArrays/StaticArrays.jl/tree/gh-pages

I think, this is because `Documenter.auto_detect_deploy_system()` is failing to detect GitHub Actions, and the function is detecting `Travis` mistakenly.

Here's a screenshot from GitHub Actions.
image

It should be like this:
image

The functions `auto_detect_deploy_system()` is detecting the system from environment variables, so we should check them.
https://github.com/JuliaDocs/Documenter.jl/blob/master/src/deployconfig.jl#L825

hyrodium:
* Fix #870 to deploy the document on github pages.
* Fix jldoctests

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,3 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-
-[compat]
-Documenter = "~0.22"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,3 +1,5 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+
+[compat]
+Documenter = "~0.26"
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -18,4 +18,4 @@ makedocs(;
     sitename = "StaticArrays.jl",
 )
 
-deploydocs(; repo = "github.com/JuliaArrays/StaticArrays.jl")
+deploydocs(; repo = "github.com/juliaarrays/StaticArrays.jl")
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,15 +1,19 @@
-using Documenter, StaticArrays
+using Documenter
+using StaticArrays
 
-makedocs(
-         modules = [StaticArrays],
-         sitename = "StaticArrays.jl",
-         pages = [
-            "Home" => "index.md",
-            "API" => "pages/api.md",
-            "Quick Start" => "pages/quickstart.md",
-            ],
-        )
-
-deploydocs(
-    repo = "github.com/JuliaArrays/StaticArrays.jl.git",
+makedocs(;
+    modules = [StaticArrays],
+    format = Documenter.HTML(
+        canonical = "https://JuliaArrays.github.io/StaticArrays.jl/stable/",
+        assets = ["assets/favicon.ico"],
+    ),
+    pages = [
+        "Home" => "index.md",
+        "API" => "pages/api.md",
+        "Quick Start" => "pages/quickstart.md",
+        ],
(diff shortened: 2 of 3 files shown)


docs/Project.toml
```diff
@@ -1,5 +1,3 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-
-[compat]
-Documenter = "~0.22"
```
fredrikekre, line 5: Might want to keep this, just bump it to the current version.

docs/Project.toml
```diff
@@ -1,5 +1,3 @@
...
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```
fredrikekre, line 3: You can delete this unless you also committ manifest. It is beeing added here: https://github.com/JuliaArrays/StaticArrays.jl/blob/4b09eab5d5fbaba41495a61bf9bac603cc97a28a/.github/workflows/ci.yml#L62

docs/make.jl
```diff
@@ -1,15 +1,19 @@
...
+    format = Documenter.HTML(
+        canonical = "https://JuliaArrays.github.io/StaticArrays.jl/stable/",
+        assets = ["assets/favicon.ico"],
+    ),
+    pages = [
+        "Home" => "index.md",
+        "API" => "pages/api.md",
+        "Quick Start" => "pages/quickstart.md",
+        ],
+    repo = "https://github.com/JuliaArrays/StaticArrays.jl/blob/{commit}{path}#L{line}",
```
fredrikekre, line 15: This should not be necessary to specify for GitHub hosted packages.

docs/make.jl
```diff
@@ -1,15 +1,19 @@
...
-         pages = [
-            "Home" => "index.md",
-            "API" => "pages/api.md",
-            "Quick Start" => "pages/quickstart.md",
-            ],
-        )
-
-deploydocs(
-    repo = "github.com/JuliaArrays/StaticArrays.jl.git",
+makedocs(;
```
fredrikekre, line 4: I suggest this instead of all the individual `using StaticArrays`.
```suggestion
# Setup for doctests in docstrings
DocMeta.setdocmeta!(StaticArrays, :DocTestSetup, :(using StaticArrays))

makedocs(;
```

docs/Project.toml
hyrodium: I think we don't need this `[compat]` table not to repeat the issue.
And, many other repositories also don't have this table.

* https://github.com/JuliaLang/julia/blob/master/doc/Project.toml
* https://github.com/JuliaPlots/PlotDocs.jl/blob/master/docs/Project.toml
* https://github.com/JuliaGraphics/Luxor.jl/blob/master/docs/Project.toml

docs/Project.toml
hyrodium: Thanks, I didn't know the relation.
I'll remove it.

docs/make.jl
hyrodium: Sorry, I just copied from PkgTemplates.jl.
I'll remove it.

docs/make.jl
hyrodium: Thanks, I didn't know the `DocMeta.setdocmeta!`.

docs/Project.toml
fredrikekre: Those packages are out of luck when Documenter 0.26 is released. It is best practice to have it.

src/deque.jl
```diff
@@ -4,7 +4,7 @@
 Return a new `StaticVector` with `item` inserted on the end of `vec`.
 
 # Examples
-```jldoctest
+```jldoctest; setup = :(using StaticArrays)
```
fredrikekre, line 7: These shouldn't be necessary now that you added `setdocmeta!`

docs/Project.toml
hyrodium: I see. thanks for the information!

hyrodium:
Thank you for the detailed review.

c42f:
Very nice, thanks. And thanks Fredrik for the review :)

Issue #878, opened 2021-02-20
static indexing on sparse array doesn't work anymore


Roger-luo:
not sure what is changed, but is causing https://github.com/QuantumBFS/tutorials/issues/17

MWE:

```julia
julia> using StaticArrays, SparseArrays

julia> st = sprand(16, 1, 0.1);

julia> I = @SVector [2, 4];

julia> st[I, 1] = st[I, 1]
ERROR: MethodError: no method matching length(::StaticArrays.StaticIndexing{SVector{2, Int64}})
Closest candidates are:
  length(::Union{Base.KeySet, Base.ValueIterator}) at abstractdict.jl:58
[...]

PR #879, opened 2021-02-27
Improved StaticIndexing (issue #878)

mateuszbaran:
These changes make the example in issue #878 work. I didn't put the exact example in tests because I didn't want to add a new test dependency (SparseArrays) but I can add it if you want.


rafaqz (commented):
Mostly fine, but `StaticIndexing` needs to be `Int` or `AbstractArray` to follow the `to_indices` interface:

> The returned tuple must only contain either Ints or AbstractArrays of scalar indices that are supported by array A

As `A` is untyped here this means for any possible `AbstractArray` so it needs to follow the interface.

mateuszbaran:
Would just subtyping `StaticIndexing{I} <: AbstractVector{Int}` be OK?

rafaqz:
I think so

rafaqz:
It also needs `size` rather than `length`

rafaqz:
I just realised this is still incorrect... now we are wrapping `Int` indices in a `AbstractArray` which will change the indexing behavior.

I don't know why StaticArrays is doing this but I think we need to just limit the `to_indices` here to dispatch on `A::StaticArray` or this will just be wrong in some cases.

Another option is to not wrap `Int` or `CartesianIndex` in `StaticIndices` at all

mbauman:
I suspect the rough idea behind `StaticIndexing` is so that StaticArrays can define implementations like `f(A::AbstractArray, Idx::StaticIndexing...)` without worry of ambiguities.  Making it an `StaticIndexing <: AbstractVector` won't quite work, as far as I understand it, because it needs to wrap integers, matrices and other-dimensional arrays in addition to vectors.

rafaqz:
It really can't dispatch on `AbstractArray` or we will just keep getting confused Discourse posts when it breaks.

Maybe supporting `StaticArray` and `Array` is enough for performance? Or a wider set of Base arrays?

mateuszbaran:
I think it's less wrong to go with the changes that I've just pushed than keep the status quo, so I'm going to merge this (if CI passes and there are no objections). More comprehensive changes can happen in a separate PR (which I'm not planning to work on in the foreseeable future).

PR #891, opened 2021-04-09
Test for unbound args

chriselrod:
I removed all the unbound args and am now testing for `length(detect_unbound_args(StaticArrays)) <= 0`.
Most of them I removed were benign, i.e. would've resulted in an error anyway.

But these two would've resulted in silent allocations and bad performance:
```julia
Base.transpose(::TSize{S,:transpose}) where {S,T} = TSize{reverse(S),:any}()
Base.transpose(::TSize{S,:any}) where {S,T} = TSize{reverse(S),:transpose}()
```
Insidious cases like these are why it's worth checking for unbound arguments, which is easiest if you don't have benign cases you have to ignore.


hyrodium:
Could you resolve the conflicts? It seems great to have the test for unbound args.

src/abstractarray.jl
```diff
@@ -127,7 +127,7 @@ When `a` is an instance or a concrete type the element type `eltype(a)` is
 used. However, when `a` is a `UnionAll` type such as `SMatrix{2,2}`, the
 promoted type of `elements` is used instead.
 """
-@inline function _construct_similar(a, s::Size, elements::NTuple{L,ET}) where {L,ET}
+@inline function _construct_similar(a, s::Size, elements::Tuple{ET,Vararg{ET,L}}) where {L,ET}
```
hyrodium, line 130: Isn't this should be the following?
```suggestion
@inline function _construct_similar(a, s::Size, elements::Tuple{Vararg{ET}}) where ET
```

```julia
julia> (NTuple{L,ET} where {L,ET}) == (Tuple{ET,Vararg{ET,L}} where {L,ET})
false

julia> (NTuple{L,ET} where {L,ET}) == (Tuple{Vararg{ET,L}} where {L,ET})
true

julia> (NTuple{L,ET} where {L,ET}) == (Tuple{Vararg{ET}} where ET)
true
```

chriselrod:
```julia
julia> NTuple{4,Float64} === Tuple{Float64,Vararg{Float64,3}}
true
```
But, if we don't care about `L`, we may as well drop it from the signature.

chriselrod:
Actually, I think I was wrong to accept the suggestion.

`Tuple{Vararg{ET}}` will still be unbound.

hyrodium:
>`Tuple{Vararg{ET}}` will still be unbound.

Ah, I see.

```julia
julia> f(::Tuple{Vararg{ET}}) where ET = ET

julia> f((1,2))
Int64

julia> f((1.0,))
Float64

julia> f(())
ERROR: UndefVarError: ET not defined
Stacktrace:
 [1] f(#unused#::Tuple{})
   @ Main ./REPL[1]:1
 [2] top-level scope
   @ REPL[4]:1
```

hyrodium (approved):
The new method `_construct_similar` does not accept `elements = ()` but since _construct_similar is not used, this is not a problem for now.

https://github.com/JuliaArrays/StaticArrays.jl/blob/76ce2c54bb1ac86923ff84d66e6b82875cea6808/src/linalg.jl#L109-L122

Issue #839, opened 2020-10-17
Support for ConstructorBase


awadell1:
Is there any interest in adding support for ConstructorBase.constructorof to StaticArrays? I think it's as simple as adding the following, likely to convert.jl

```julia
ConstructionBase.constructorof(::Type{SA}) where {SA<:StaticArray} = SA
```

Best,

Alex

mateuszbaran:
What would be the benefits of that? Static array construction is a tricky and convoluted part of `StaticArrays`.

awadell1:
Hey Mateusz!

I'm using a struct to represent the parameters of a UODE, that gets solved by DifferentialEquations and then feed into ForwardDiff.jl to compute the gradiant of a loss function with respect to the model's parameters. Becuase ForwardDiff.jl requiries a vector and not a struct, I've been using Flatten.jl for the conversion.

```julia
struct p{T}
    a::T
    b::T
    c::NTuple{2, T}
end

using Flatten
param = p{Float64}(1, 2, (3, 4))
param_vec = [Flatten.flatten(param)...]
```

The problem arises when I try to use a StaticArray as a field, (Or a LSVector from LabeledArrays.jl). I want to do this because I need to normalize the state vector before feeding it into the NN.

Currently, this breaks down because Flatten.jl uses ConstructorBase.constructorof to determine how to reconstruct an object. 

```julia
using StaticArray, Flatten
a = SVector(1, 2, 3)
Flatten.flatten(a)                      # Get (1, 2, 3) as expected
Flatten.reconstruct(a, (4, 5, 6)        # Expect SVector(4, 5, 6) -> get error

ERROR: DimensionMismatch("No precise constructor for SArray found. Length of input was 3.")
Stacktrace:
 [1] SArray(::Tuple{Tuple{Tuple{Tuple{Int64,Int64,Int64}}}}) at /home/awadell/.julia/packages/StaticArrays/l7lu2/src/convert.jl:1
 [2] StaticArray at /home/awadell/.julia/packages/StaticArrays/l7lu2/src/convert.jl:4 [inlined] (repeats 3 times)
 [3] macro expansion at /home/awadell/.julia/packages/Flatten/oXoLY/src/Flatten.jl:197 [inlined]
[...]

PR #901, opened 2021-04-23
add ConstructionBase interface

rafaqz:
This PR adds an interface for [ConstructionBase.jl](https://github.com/JuliaObjects/ConstructionBase.jl).

This has been requested in #839. I realize there will be resistance to adding a dependency here, with good reason. But ConstrucitonBase.jl is a very small package that seems to make no difference to compile times here, and has some compelling use-cases.

StaticArrays are immutable. Modifying them requires calling their constructor. Packages like Flatten.jl, BangBand.jl, Setfield.jl, Accessors.jl facilitate modifying immutable objects. But objects with types that can't be derived from their fields, like `SArray`, need a `constructorof` method for this to work generically. Setfield.jl can already modify `SArray` by setting array indices, but just replacing e.g. the `SArray` tuple entirely in a generic way (that has no dep on StaticArrays) is not possible.

To illustrate a little more, the main use cases I have for this PR are:
- reconstructing large, complex objects (that often contain `SArrray`) automatically with e.g. all Float64 swapped to Float32 For use on GPU, using Flatten.jl.
- ModelParamaters.jl working with static arrays of parameters - parameter wrappers can be placed anywhere in a model, and used for control or optimisation, and stripped out later. Currently they cant go in a static array because you can't remove them later.
- Swapping `Array` for `CuArray` and similar while re-wrapping `SizedArray`. The goal is that all arrays in an object can be swapped for `CuArray` without needing an explicit Adapt.jl inteface, see e.g. https://github.com/JuliaObjects/ConstructionBase.jl/issues/34

Static arrays are the most common objects that don't work with ConstuctionBase.jl, and the most friction for scaling this kind of functionality. Currently doing these things requires adding  `constructorof` manually in a script.

The other option is for this code to be moved into a glue package (maybe StaticArraysConstructors.jl at JuliaObjects). I haven't done that because a) it would be type pyracy,  and should at least be blessed pyracy, and b) It would be a tiny package (10 lines) probably not worth the maintenance, c) that would require users to load and depend on it to get this functionality, instead of it just working, and d) packages that wanted to ensure that `constructorof` works in most case will need a StaticArrays.jl dependency.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,9 +6,11 @@ version = "1.1.1"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 
 [compat]
 julia = "1"
+ConstructionBase = "1.0, 1.1.0"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -10,6 +10,7 @@ import Base: getindex, setindex!, size, similar, vec, show, length, convert, pro
 
 import Statistics: mean
 
+using ConstructionBase
 using Random
 import Random: rand, randn, randexp, rand!, randn!, randexp!
 using Core.Compiler: return_type
@@ -143,6 +144,7 @@ include("qr.jl")
 include("deque.jl")
 include("flatten.jl")
 include("io.jl")
+include("constructorof.jl")
 
 if Base.VERSION >= v"1.4.2"
     include("precompile.jl")
--- a/src/constructorof.jl
+++ b/src/constructorof.jl
@@ -1,0 +1,15 @@
+struct SArrayConstructor{S,N,L} end
+struct MArrayConstructor{S,N,L} end
+struct SizedArrayConstructor{S,N,M} end
+
+(::SArrayConstructor{S,N,L})(data::NTuple{L,T}) where {S,T,N,L} = SArray{S,T,N,L}(data)
+(::MArrayConstructor{S,N,L})(data::NTuple{L,T}) where {S,T,N,L} = MArray{S,T,N,L}(data)
+(::SizedArrayConstructor{S,N,M})(data::AbstractArray{T,M}) where {S,T,N,M} = 
+    SizedArray{S,T,N,M}(data)
+
+ConstructionBase.constructorof(sa::Type{<:SArray{S,<:Any,N,L}}) where {S,N,L} = 
+    SArrayConstructor{S,N,L}()
+ConstructionBase.constructorof(sa::Type{<:MArray{S,<:Any,N,L}}) where {S,N,L} = 
+    MArrayConstructor{S,N,L}()
+ConstructionBase.constructorof(sa::Type{<:SizedArray{S,<:Any,N,M}}) where {S,N,M} = 
+    SizedArrayConstructor{S,N,M}()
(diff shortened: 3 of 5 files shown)


rafaqz:
The remaining test failures seem to be just ambiguities on 1.6

andyferris:
Thanks for the contribution.

Dealing with immutable data structures is a big part of the core of this package. For example we have the `similar_type` function (instead of `similar`) which returns a constructor and lets you chose the dimensions and element type. The constuctor must accept a `Tuple`. For generic programming with these "constructor factories" you often want to set some other traits of the container: like "I want to use `setindex!` and `push!` or not (`similar` implies mutability, for example, there's a `copy` and `Base.copymutable` (which might or might not be resizable), `empty` and `emptymutable` (which I presume implies resizable)). The `collect` function doesn't let you set any of these and there's no other "value constructor" for `AbstractArray` in `Base`

I'm very glad people are thinking about these problems. If we can have a shared solution that has a chance of migrating towards `Base` that would be fantastic. I keep thinking of how we should do this in `Base` with existing containers like `Array`, `Dict`, `Set`, etc, with at least immutable/mutable/resizable traits, but then there's all the "layout" traits like sparse.

We also use `similar` and `similar_type` on _Rotations.jl_ arrays but they don't return a constructor for the same type. Do you have any idea of how they would support `constructorof`?

In generic code I might want to change the size or element type - so I'm a little curious how you handle this in practice with `constructorof`?

> The other option is for this code to be moved into a glue package (maybe StaticArraysConstructors.jl at JuliaObjects).

Since Julia supports condition code loading - should we consider conditional methods either here or in _ConstructorBase_ as the other option?

rafaqz:
There has been a bunch of thought on this in JuliaObjects and the various packages we've built separately. I guess having something widely supported that works is the way towards getting something similar in base. The idea with ConstructionBase.jl was centralizing the most fundamental methods to work with immutables and share the interface between all the packages that were experimenting with these things. But it's not really user-facing, I don't think most people know it exists until they get an error when `constructorof` doesn't work.

My motivation is modifying immutable objects without knowing anything about their structure or specific methods they may have. e.g. packages like DynamicGrids.jl let users define a model however they want, but I still want to be able to e.g. strip the all the `Floa64` for `Float32`, or get model parameters to generate user interfaces automatically, wherever the parameters happen to be embedded in the object. For float replacement you can just do:

```julia
obj = Flatten.modify(Float32, obj, Float64)
```
As long as `constructorof` works generically on `obj` and its components, or is defined specifically for the types where it doesn't. 

To your specific comments, changing the size would be useful, and probably easy with an `SVector` to use the length of the tuple? I can add that to this PR. Maybe not with `SMatrix` generically? `SizedArray` could also be updated based on the array size instead of staying in the type, although you lose type stability for when you don't need to change the size.

In terms of code loading, do you mean Requires.jl or just custom code?

ConstructionBase.jl is a very small (100 loc)  base package, so we would be wary to add anything to it like that. I've been thinking we could make it even smaller (e.g. ConstructionMethods.jl) which would just have the three empty methods to extend without the generated functions or anything else in ConstructionBase. Would be better as a dependency here?

rafaqz:
I've just been reading over this over again as there was a lot to digest, and I missed some things you mentioned. 

`similar_type` does seem like a very similar concept to `constructorof` - except `constructorof` doesn't specify that it should actually be a type that is returned - it's just something that will construct the object, when given the output of `ConstructionBase.getproperties` as arguments. These are usually just all of the objects fields.

`constructorof` usually leaves the field types free to change, as we can get them from the arguments. It seems `similar_type` allows for a different type `T` in its 2 argument form.

Rotations.jl returning a different object from `constructorof` should be fine too, I don't think there is any restriction on the type of the returned object.

You also mentioned `Dict` and `Set` etc. These probably should have `constructorof` methods defined in ConstructionBase.jl, which would allow setting their fields manually with e.g. `@set` from Setfield.jl.

rafaqz:
Closing this in favor of ConsructionBaseExtras.jl adding these methods with Requires.jl as a stop-gap until we have first class conditional code loading.

PR #909, opened 2021-05-08
Allow `dims::Tuple` in `sum`

### LINKED ISSUE #880: Passing a Static Array to sum() throws an error
Minimal example:
```julia
using StaticArrays
input = @SArray rand(2,2,2);
sum(input, dims=(1,2)) # throws error
```
Error:
```julia
ERROR: LoadError: MethodError: no method matching _mapfoldl(::typeof(identity), ::typeof(+), ::Tuple{Int64,Int64}, ::StaticArrays._InitialValue, ::Size{(2, 2, 2)}, ::SArray{Tuple{2,2,2},Float64,3,8})
```

mcabbott:
Closes #880.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -158,6 +158,12 @@ end
     end
 end
 
+@inline function _mapreduce(f, op, D::Tuple, init, sz::Size{S}, a::StaticArray) where {S}
+    b = _mapreduce(f, op, first(D), init, sz, a)
+    return _mapreduce(f, op, Base.tail(D), init, Size(b), b)
+end
+_mapreduce(f, op, D::Tuple{}, init, sz::Size{S}, a::StaticArray) where {S} = a
+
 @generated function _mapfoldl(f, op, dims::Val{D}, init,
                                ::Size{S}, a::StaticArray) where {S,D}
     N = length(S)
--- a/test/mapreduce.jl
+++ b/test/mapreduce.jl
@@ -102,6 +102,7 @@ using Statistics: mean
         RSArray1 = SArray{Tuple{1,J,K}}  # reduced in dimension 1
         RSArray2 = SArray{Tuple{I,1,K}}  # reduced in dimension 2
         RSArray3 = SArray{Tuple{I,J,1}}  # reduced in dimension 3
+        RSArray13 = SArray{Tuple{1,J,1}}  # reduced in dimension 1 and 3
         a = randn(I,J,K); sa = OSArray(a)
         b = rand(Bool,I,J,K); sb = OSArray(b)
         z = zeros(I,J,K); sz = OSArray(z)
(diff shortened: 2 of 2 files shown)


test/mapreduce.jl
```diff
@@ -111,9 +112,11 @@ using Statistics: mean
         @test sum(sa) === sum(a)
         @test sum(abs2, sa) === sum(abs2, a)
         @test sum(sa, dims=2) === RSArray2(sum(a, dims=2))
+        @test sum(sa, dims=(2,)) === RSArray2(sum(a, dims=2))
         @test sum(sa, dims=Val(2)) === RSArray2(sum(a, dims=2))
         @test sum(abs2, sa; dims=2) === RSArray2(sum(abs2, a, dims=2))
         @test sum(abs2, sa; dims=Val(2)) === RSArray2(sum(abs2, a, dims=2))
+        @test_broken sum(abs2, sa; dims=(1,3)) === RSArray13(sum(abs2, a, dims=(1,3)))
```
mcabbott, line 119: Here's what happens, I'm not very sure why:
```
julia> a == sa
true

julia> sum(a, dims=(1,3))
1×2×1 Array{Float64, 3}:
[:, :, 1] =
 -2.23789  1.60065

julia> sum(sum(a, dims=3), dims=1)
1×2×1 Array{Float64, 3}:
[:, :, 1] =
 -2.23789  1.60065

julia> sum(sa, dims=(1,3))
1×2×1 SArray{Tuple{1, 2, 1}, Float64, 3, 2} with indices SOneTo(1)×SOneTo(2)×SOneTo(1):
[:, :, 1] =
 -2.23789  1.60065

julia> RSArray13(sum(abs2, a, dims = (1, 3)))
1×2×1 SArray{Tuple{1, 2, 1}, Float64, 3, 2} with indices SOneTo(1)×SOneTo(2)×SOneTo(1):
[:, :, 1] =
 3.8265  1.09579
```

src/mapreduce.jl
```diff
@@ -158,6 +158,9 @@ end
     end
 end
 
+@inline _mapreduce(f, op, D::Tuple{<:Any}, init, sz::Size{S}, a::StaticArray) where {S} =
+    _mapreduce(f, op, first(D), init, sz, a)
```
mcabbott, line 162: Things like `sum(abs2, SA[1 2; 3 4], dims = (1,2))` still won't work, that seems like a bigger project. It would be easy to add a method here to make them give a friendly "not yet supported" error, instead of an "internal" error. But opinions vary as to whether that's a good idea.

N5N3: The change needed is not large though, we can use `Base.mapreducedim!` to generated the code we want:
```julia
@generated function _mapfoldl(f, op, dims::Val{D}, init,
                               ::Size{S}, a::StaticArray) where {S,D}
    _init = init === _InitialValue ? :default : :init
    iter = CartesianIndices(S)
    exprs = fill!(similar(iter, Any, Base.reduced_indices(iter, D)), _init)
    rf(ex, nex) = ex === :default ? :(Base.reduce_first(op, $nex)) : :(op($ex,$nex))
    Base.mapreducedim!(I->:(f(a[$(I.I...)])), rf, exprs, iter)
    return quote
        @_inline_meta
        @inbounds elements = tuple($(exprs...))
        @inbounds return similar_type(a, eltype(elements), Size($(size(exprs))))(elements)
    end
end
```
The current test passed locally, and the influence on compile time seems negligible.

Issue #806, opened 2020-07-07
static interface for permutedims


tpapp:
I would like to contribute an implementation for [`permutedims`](https://github.com/JuliaArrays/StaticArrays.jl/blob/68b10c2d67a7f4eadcb0bf7d1e41c986cd584789/src/abstractarray.jl#L194), but first I would like to ask for suggestions about the interface.

In particular, it would be ideal if one could encode the permutation statically. I could imagine a 
```julia
struct StaticPermutation{N}
    permuation::NTuple{N,Int}
    ## optionally: constructor that verifies validity
end
```
type and an interface like
```julia
permutedims(sarray, StaticPermutation(2, 3, 1))
```

but I am interested in suggestions if there is a better solution.

mateuszbaran:
Good idea, although for full type stability you need to encode the entire permutation in the type, such as
```julia
struct StaticPermutation{Permutation<:Tuple}
end
```
and then have something like `permutedims(sarray, StaticPermutation{Tuple{2, 3, 1}}())`. Of course we could have a nicer constructor for permutations.

c42f:
Agreed, we should have `StaticPermutation(2, 3, 1)` encode the permutation purely in the type of `StaticPermutation` so that the output of ``permutedims(sarray, StaticPermutation(2, 3, 1))` would always have a known `Size`. Otherwise, passing `StaticPermutation` would be equivalent to passing a normal `NTuple` (or SVector) as the permutation to `permutedims`.

It may also make logical sense to have the permutation as a kind of `StaticVector`, sorta like

```julia
struct StaticPermutation{N,Permutation<:NTuple{N,Int}} <: StaticVector{N,Int}
end
```

but the extra `N` parameter is annoying (can we avoid it?)

tkf:
I think this is where StaticNumbers.jl is useful.  Something like `permutedims(sarray, (static(2), static(3), static(1)))` is cleaner and more extensible.  You can construct and manipulate the permutation as if everything is done in the value domain rather than in the type domain (e.g., `isperm((static(2), static(3), static(1)))` works and const props without any specific methods).

OTOH, using `Tuple{2, 3, 1}` feels like a hack (although I know `StaticArray` type has been using it since forever).  I understand that it is tempting to use `Tuple` whenever you need "vararag" type parameters.  But I don't think we should be using `Tuple` with non-types as parameters, especially when there are other better ways to do it.

c42f:
[...]

PR #917, opened 2021-05-30
`permutedims` for vectors and matrices

mcabbott:
Before:
```
julia> permutedims(SA[1,2,3])
1×3 reshape(::SVector{3, Int64}, 1, 3) with eltype Int64:
 1  2  3

julia> axes(ans)
(Base.OneTo(1), Base.OneTo(3))

julia> permutedims(SA[1 2; 3 4])
 1  3
 2  4
```
After:
```
julia> permutedims(SA[1,2,3])
1×3 SMatrix{1, 3, Int64, 3} with indices SOneTo(1)×SOneTo(3):
 1  2  3

julia> permutedims(SA[1 2;3 4])
2×2 SMatrix{2, 2, Int64, 4} with indices SOneTo(2)×SOneTo(2):
 1  3
 2  4 
```
Unchanged are all the cases taking `perm`, discussed in #806.
```
julia> permutedims(@SArray(rand(1,2,3)), (3,2,1))
3×2×1 MArray{Tuple{3, 2, 1}, Float64, 3, 6} with indices SOneTo(3)×SOneTo(2)×SOneTo(1):
```

~~I'm not too sure what should happen for MArrays or SizedArrays.~~

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -226,13 +226,19 @@ end
 
 # TODO permutedims? So far just the cases without perm:
 
+Base.permutedims(A::SVector{N}) where {N} = SMatrix{1,N}(A.data...)
+Base.permutedims(A::MVector{N}) where {N} = MMatrix{1,N}(A.data...)
+Base.permutedims(A::SizedVector{N}) where {N} = SizedMatrix{1,N}(permutedims(A.data))
+
 @generated function Base.permutedims(A::SMatrix{M,N}) where {M,N}
     exs = permutedims([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
     return :(SMatrix{N,M}($(exs...)))
 end
-
-Base.permutedims(A::SVector{N}) where {N} = SMatrix{1,N}(A.data...)
-Base.permutedims(A::MVector{N}) where {N} = MMatrix{1,N}(A.data...)
+@generated function Base.permutedims(A::MMatrix{M,N}) where {M,N}
+    exs = permutedims([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
+    return :(MMatrix{N,M}($(exs...)))
+end
+Base.permutedims(A::SizedMatrix{M,N}) where {M,N} = SizedMatrix{N,M}(permutedims(A.data))
 
 #--------------------------------------------------
 # Concatenation
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -228,7 +228,7 @@ end
 
 @generated function Base.permutedims(A::SMatrix{M,N}) where {M,N}
     exs = permutedims([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
-    return :(SMatrix{M,N}($(exs...)))
+    return :(SMatrix{N,M}($(exs...)))
 end
 
 Base.permutedims(A::SVector{N}) where {N} = SMatrix{1,N}(A.data...)
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -224,7 +224,15 @@ for rot in [:rotl90, :rotr90]
     end
 end
 
-# TODO permutedims?
+# TODO permutedims? So far just the cases without perm:
+
+@generated function Base.permutedims(A::SMatrix{M,N}) where {M,N}
+    exs = permutedims([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
+    return :(SMatrix{M,N}($(exs...)))
+end
+
+Base.permutedims(A::SVector{N}) where {N} = SMatrix{1,N}(A.data...)
+Base.permutedims(A::MVector{N}) where {N} = MMatrix{1,N}(A.data...)
 
 #--------------------------------------------------
 # Concatenation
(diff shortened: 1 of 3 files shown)


src/abstractarray.jl
```diff
@@ -227,7 +227,15 @@ for rot in [:rotl90, :rotr90]
...
-# TODO permutedims?
+# TODO permutedims? So far just the cases without perm:
+
+@generated function Base.permutedims(A::SMatrix{M,N}) where {M,N}
+    exs = permutedims([:(getindex(A,$i,$j)) for i in 1:M, j in 1:N])
+    return :(SMatrix{N,M}($(exs...)))
+end
+
+Base.permutedims(A::SVector{N}) where {N} = SMatrix{1,N}(A.data...)
+Base.permutedims(A::MVector{N}) where {N} = MMatrix{1,N}(A.data...)
```
mateuszbaran, line 238: Calling `permutedims` on a normal `Array` gives an `Array` pointing at the same memory, and this does not. I'm not sure though if this data sharing is a part of the interface of `permutedims` or not? Docs are a bit unclear, in particular docstring for `permutedims(v::AbstractVector)` refers to the performed operation as "reshaping", and reshaping indicates data sharing.

mcabbott: Right, I'm not sure what is best here. `permutedims(::Diagonal)` also shares memory, which makes me think that it should return the same type, but whether it's a view or a copy depends on what's convenient. 

`adjoint(::MMatrix)` also makes a copy; I guess `adjoint(::StaticVector)` has other reasons besides this copy/view issue to return a wrapper:
```

julia> d = Diagonal([1,2,3]);

julia> permutedims(d)
3×3 Diagonal{Int64, Vector{Int64}}:
 1  ⋅  ⋅
 ⋅  2  ⋅
 ⋅  ⋅  3

julia> ans[3,3] = 33;

julia> d
3×3 Diagonal{Int64, Vector{Int64}}:
 1  ⋅   ⋅
 ⋅  2   ⋅
 ⋅  ⋅  33

julia> m = @MMatrix [1 2; 3 4];

julia> m'
 1  3
 2  4

julia> ans[2,2] = 22;

julia> m
 1  2
 3  4

julia> SA[1,2,3]'
1×3 adjoint(::SVector{3, Int64}) with eltype Int64 with indices SOneTo(1)×SOneTo(3):
 1  2  3

julia> ans * SA[4,5,6]  # isa Number
32
```

mateuszbaran: OK, so in the case of `MArray` memory sharing would definitely be quite inconvenient, and Base seems to be fine with not sharing memory: https://github.com/JuliaLang/julia/blob/27aaadb9b28e8457d86239e443f8e65d90e878c7/base/permuteddimsarray.jl#L124 . So, this approach to `MVector` and, similarly, `MMatrix` looks OK.

`SizedVector` and `SizedMatrix` should just call appropriate `permutedims` on the wrapped array, and wrap the result in a new `SizedArray`.

mcabbott: For all higher dims, I thought `permutedims(::Array, ...)` never shared, until this `Diagonal` example. So the only concern was whether sharing was part of the contract for `::AbstractVector`, or just an accident of `::Vector`, and I think the conclusion is the latter.

mateuszbaran: Yes, I think that's right.

test/abstractarray.jl
```diff
@@ -201,6 +201,22 @@ using StaticArrays, Test, LinearAlgebra
     end
 end
 
+@testset "permutedims" begin
+    @test @inferred(permutedims(SVector(1,2,3))) === SMatrix{1,3}(1,2,3)
+    @test @inferred(permutedims(MVector(1,2,3))) == MMatrix{1,3}(1,2,3)
+    @test @inferred(permutedims(MVector(1,2,3))) isa MMatrix{1,3}
+    @test @inferred(permutedims(SizedVector{3}(rand(3)))) isa SizedMatrix{1,3}
```
mateuszbaran, line 208: Could you add a test like this here: `@test permutedims(SizedVector{3}([1,2,3])) == [1 2 3]` ?

test/abstractarray.jl
```diff
@@ -201,6 +201,22 @@ using StaticArrays, Test, LinearAlgebra
...
+
+    @test @inferred(permutedims(SMatrix{2,2}(1,2,3,4))) === SMatrix{2,2}(1,3,2,4)
+    @test @inferred(permutedims(MMatrix{2,2}(1,2,3,4))) == MMatrix{2,2}(1,3,2,4)
+    @test @inferred(permutedims(MMatrix{2,2}(1,2,3,4))) isa MMatrix{2,2}
+
+    A = rand(2,3)
+    @test @inferred(permutedims(SMatrix{2,3}(A))) === SMatrix{3,2}(A')
+    @test @inferred(permutedims(MMatrix{2,3}(A))) == MMatrix{3,2}(A')
+    @test @inferred(permutedims(SizedMatrix{2,3}(A))) == SizedMatrix{3,2}(A')
```
mateuszbaran, line 217: Could you add `isa` tests for `MMatrix` and `SizedMatrix`?

mateuszbaran (approved):

mateuszbaran:
Sorry for the delay but could you bump the patch version? I'm going to merge this soon and tag a new release.

PR #919, opened 2021-06-07
Add simple 3- and 4-arg `*` methods

mcabbott:
This PR repairs a small slowdown due to https://github.com/JuliaLang/julia/pull/37898. When all matrices are the same size, that wastes a few ns thinking about their sizes, while this PR restores the left-to-right behaviour: 
```julia
julia> s1,s2,s3,s4,s5 = fill(3,5);

julia> a=@SMatrix rand(s1,s2); b=@SMatrix rand(s2,s3); c=@SMatrix rand(s3,s4); d=@SMatrix rand(s4,s5);

julia> @btime *($(Ref(a))[],$(Ref(b))[],$(Ref(c))[],$(Ref(d))[]);  # after 37898, this thinks about the sizes
  9.291 ns (0 allocations: 0 bytes)

julia> @btime (($(Ref(a))[] * $(Ref(b))[]) * $(Ref(c))[]) * $(Ref(d))[];  # simply left to right
  5.208 ns (0 allocations: 0 bytes)
```
When the matrices are of different sizes, things aren't so clear-cut. Sometimes the re-ordering does speed things up, sometimes it doesn't.

### DIFF
--- a/src/matrix_multiply.jl
+++ b/src/matrix_multiply.jl
@@ -15,11 +15,15 @@ import LinearAlgebra: BlasFloat, matprod, mul!
 @inline *(A::StaticArray{Tuple{N,1},<:Any,2}, B::Adjoint{<:Any,<:StaticVector}) where {N} = vec(A) * B
 @inline *(A::StaticArray{Tuple{N,1},<:Any,2}, B::Transpose{<:Any,<:StaticVector}) where {N} = vec(A) * B
 
+# Avoid LinearAlgebra._quad_matmul's order calculation on equal sizes
+@inline *(A::StaticMatrix{N,N}, B::StaticMatrix{N,N}, C::StaticMatrix{N,N}) where {N} = (A*B)*C
+@inline *(A::StaticMatrix{N,N}, B::StaticMatrix{N,N}, C::StaticMatrix{N,N}, D::StaticMatrix{N,N}) where {N} = ((A*B)*C)*D
+
 """
     mul_result_structure(a::Type, b::Type)
 
 Get a structure wrapper that should be applied to the result of multiplication of matrices
-of given types (a*b). 
+of given types (a*b).
 """
 function mul_result_structure(a, b)
     return identity
@@ -114,7 +118,6 @@ end
         b::Union{Transpose{Tb, <:StaticVector}, Adjoint{Tb, <:StaticVector}}) where {sa, sb, Ta, Tb}
     newsize = (sa[1], sb[2])
     exprs = [:(a[$i]*b[$j]) for i = 1:sa[1], j = 1:sb[2]]
-    
     return quote
         @_inline_meta
         T = promote_op(*, Ta, Tb)
@@ -209,7 +212,7 @@ end
         while m < M
             mu = min(M, m + M_r)
             mrange = m+1:mu
-            
+
             atemps_init = [:($(atemps[k1]) = a[$k1]) for k1 = mrange]
             exprs_init = [:($(tmps[k1,k2])  = $(atemps[k1]) * b[$(1 + (k2-1) * sb[1])]) for k1 = mrange, k2 = nrange]
             atemps_loop_init = [:($(atemps[k1]) = a[$(k1-sa[1]) + $(sa[1])*j]) for k1 = mrange]
--- a/test/matrix_multiply.jl
+++ b/test/matrix_multiply.jl
@@ -173,6 +173,9 @@ mul_wrappers = [
         @test m*transpose(n) === @SMatrix [8 14; 18 32]
         @test transpose(m)*transpose(n) === @SMatrix [11 19; 16 28]
 
+        @test @inferred(m*n*m)  === @SMatrix [49 72; 109 160]
+        @test @inferred(m*n*m*n)  === @SMatrix [386 507; 858 1127]
+
         # check different sizes because there are multiple implementations for matrices of different sizes
         for (mm, nn) in [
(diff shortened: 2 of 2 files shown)


mateuszbaran:
I think ideally 3- and 4-arg methods of `*` would decide during compilation what the best order is, using generated functions and your cost model from that PR. Could you add that or do you need some help?

src/matrix_multiply.jl
```diff
@@ -15,11 +15,15 @@ import LinearAlgebra: BlasFloat, matprod, mul!
 @inline *(A::StaticArray{Tuple{N,1},<:Any,2}, B::Adjoint{<:Any,<:StaticVector}) where {N} = vec(A) * B
 @inline *(A::StaticArray{Tuple{N,1},<:Any,2}, B::Transpose{<:Any,<:StaticVector}) where {N} = vec(A) * B
 
+# Avoid LinearAlgebra._quad_matmul's order calculation on equal sizes
+@inline *(A::StaticMatrix{N,N}, B::StaticMatrix{N,N}, C::StaticMatrix{N,N}) where {N} = (A*B)*C
```
mateuszbaran, line 19: You could also use `StaticMatMulLike` here to be a bit more generic, although currently `StaticMatMulLike` doesn't swap sizes for adjoints and transposes so that would have to be fixed when working on the cost model.

mateuszbaran:
BTW, I think you should reduce the cost of multiplying by diagonal matrices (and probably also things like bidiagonal or tridiagonal) in that PR to base.

mcabbott:
The base PR doesn't do anything for special matrices, since they are all square. But the cost of checking their sizes unnecessarily is usually negligible -- only tiny SMatrices are fast enough that it matters, which is the case this PR wants to fix. I agree that things like Diagonal * Diagonal * Diagonal could be further optimised, there's quite a large set of possibilities and I haven't given any thought to how that might work.

I didn't know about this, but `StaticArrays.StaticMatMulLike{N,N}` looks like a union of static square things which could be used in place of `StaticMatrix{N,N}` here. 

For non-square matrices, I agree you could pull the cost calculation into a generated function. I think I'm going to leave that for another PR, or for someone else if they need it. This one just wants to avoid slowing down a common case.

src/matrix_multiply.jl
mcabbott: I see that `StaticArrays.StaticMatMulLike{3,10}` is a union of three different sizes, is that deliberate or an oversight? Seems like it really ought to be `StaticSquareLike{3}` for the all the square ones, and `StaticMatLike{3,10}` allowing non-square but always `size(A) == (3,10)`.

mateuszbaran: Ultimately I ended up not using the size part of `StaticMatMulLike` so the fact that `s1` and `s2` don't correspond to sizes is essentially an oversight.

mateuszbaran:
Yes, right, optimization of computation of products of two small matrices was complicated enough, and developing a reliable set of heuristics for products of more matrices would be even harder.

Anyway, it's indeed interesting that constant propagation didn't manage to simplify size checks here.

mcabbott:
I was also surprised it didn't get compiled away. Maybe it just needs more things inlined, if I try `@eval LinearAlgebra @inline function _quad_matmul(A,B,C,D) ...` then the time difference in the first message above goes away. Maybe that's a better idea than this PR!

Even with that change, the re-ordering doesn't always speed up SMatrix chains. Here's my test, sometimes one is faster, sometimes the other. Unless the cost calculation is taking as long as the matrix multiplication, it seems that this shows that the optimal order isn't in fact optimal -- so a different cost model would be better for SMatrices?
```
julia> using StaticArrays, Random

julia> s1,s2,s3,s4,s5 = shuffle([2,3,5,7,11])
5-element Vector{Int64}:
  2
  3
  7
  5
 11

julia> a=@SMatrix rand(s1,s2); b=@SMatrix rand(s2,s3); c=@SMatrix rand(s3,s4); d=@SMatrix rand(s4,s5);

julia> @btime *($(Ref(a))[],$(Ref(b))[],$(Ref(c))[],$(Ref(d))[]);  # after 37898, this thinks about the sizes
  29.648 ns (0 allocations: 0 bytes)

julia> @btime (($(Ref(a))[] * $(Ref(b))[]) * $(Ref(c))[]) * $(Ref(d))[];  # simply left to right
  14.640 ns (0 allocations: 0 bytes)
```

mateuszbaran:
So maybe let's no worry too much now about reordering of multiplication for static matrices? Just eliminating run-time size check would definitely be enough.

> if I try `@eval LinearAlgebra @inline function _quad_matmul(A,B,C,D) ...` then the time difference in the first message above goes away. Maybe that's a better idea than this PR!

That indicates that the overhead may be just from the single additional call, hard to tell without looking at `@code_typed`. I think `_quad_matmul` can be safely `@inline` since it's just called in one place in a non-`@inline` function.

mcabbott:
The `@inline` seems safe, but besides solving the above benchmark, I wonder whether it would solve real-world equivalents? Or whether `*` not inlining would reproduce the problem?

> Just eliminating run-time size check

Whether the existing size check is a good idea overall or not will depend on the distribution of non-square cases in "overall". My `shuffle([2,3,5,7,11])` above is sort-of designed to produce marginal cases; if we think the real world has more `10x1` matrices etc. then probably it will pay for itself. (It would of course be easy to remove it entirely, e.g. by overloading `_quad_matmul`)

PR #921, opened 2021-06-10
Fix #920

### LINKED ISSUE #920: `lu` seems to be ambiguous on Julia Nightly
Sometimes I check our Nightly, when they error and I noticed [here](https://github.com/JuliaManifolds/Manifolds.jl/pull/375/checks?check_run_id=2774885756#step:5:947) that the current `lu` decomposition for `5x5` (but not for `4x4`, `3x3` or `2x2`) seems to be ambiguous currently?

```
  MethodError: lu(::SizedMatrix{5, 5, Float64, 2, Matrix{Float64}}, ::Val{true}) is ambiguous. Candidates:
948
    lu(A::AbstractMatrix, ::Val{true}; check) in LinearAlgebra at deprecated.jl:70
949
    lu(A::StaticMatrix{N, N}, pivot::Union{Val{false}, Val{true}}; check) where N in StaticArrays at StaticArrays/uH2MB/src/lu.jl:40
950
    lu(A::StaticMatrix, pivot::Union{Val{false}, Val{true}}; check) in StaticArrays at StaticArrays/uH2MB/src/lu.jl:33
951
  Possible fix, define
952
    lu(::StaticMatrix{N, N}, ::Val{true}) where N
```

This seems to come originally from our matrix exponential call [here](https://github.com/JuliaManifolds/Manifolds.jl/blob/dc1ed9f7316a935b530562fc56df51c4ca93db29/src/groups/special_euclidean.jl#L218).

mateuszbaran:
This fixes the ambiguity on Julia master reported in issue #920 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.2.3"
+version = "1.2.4"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/lu.jl
+++ b/src/lu.jl
@@ -43,6 +43,13 @@ function lu(A::StaticMatrix{N,N}, pivot::Union{Val{false},Val{true}}=Val(true);
     LU(LowerTriangular(L), UpperTriangular(U), p)
 end
 
+if VERSION >= v"1.7-DEV"
+    # disambiguation
+    function lu(A::StaticMatrix{N,N}, pivot::Val{true}) where {N}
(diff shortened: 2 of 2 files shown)


thchr:
This also fixes this related regression:
```jl
using StaticArrays

A  = [1 3 5; 9 7 2; 6 8 4]
SA = SMatrix{3,3}(A)

v  = [1,2,3]
sv = SVector{3}(v)

A\v   # works ≈ [-0.2111, 0.5888, -0.111]
SA\sv # works ...
A\sv  # works ...
SA\v  # throws `MethodError: lu(::SMatrix{3, 3, Float64, 9}, ::Val{true}) is ambiguous.`
```
with a similar error for e.g. `A/SA`.

Would be nice to merge this and tag a new minor release maybe (e.g., to fix CI failures on downstream packages).

PR #929, opened 2021-06-20
Fixes to `norm` and friends (closes #915)

### LINKED ISSUE #915: `norm` of nested `StaticArray`s throws
The implementation of `norm(a::StaticArray)` assumes that the elements of `a` have an `abs2` implementation, i.e. that they are scalars. Unfortunately, that means that if `a`'s elements are also e.g. `SVector`, it will throw:
```jl
julia> a = SVector{2}(SVector{2}(0.0, 0.0), SVector{2}(1.0, 1.0))
2-element SVector{2, SVector{2, Float64}} with indices SOneTo(2):
 [0.0, 0.0]
 [1.0, 1.0]

julia> norm(a)
ERROR: MethodError: no method matching abs2(::SVector{2, Float64})
Closest candidates are:
  abs2(::Bool) at bool.jl:80
  abs2(::Real) at number.jl:151
  abs2(::Complex) at complex.jl:265
  ...
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/StaticArrays/rKBnO/src/linalg.jl:225 [inlined]
 [2] _norm
   @ ~/.julia/packages/StaticArrays/rKBnO/src/linalg.jl:213 [inlined]
 [3] norm(a::SVector{2, SVector{2, Float64}})
   @ StaticArrays ~/.julia/packages/StaticArrays/rKBnO/src/linalg.jl:212
  ...
```
One of the unhappy consequences of this is that things like `isapprox` breaks for nested `StaticArray`s.

The `norm` implementation is here:
https://github.com/JuliaArrays/StaticArrays.jl/blob/d03a5e6aa402dbafab1ed53f44835645aa1746a4/src/linalg.jl#L213-L228

AFAICT LinearAlgebra is avoiding this by calling out to an internal function `norm_sqr` instead of `abs2`, which is just `abs2` for scalar input, but recursively calls `norm` otherwise. There's actually a [`norm_sqr` function in StaticArrays too](https://github.com/JuliaArrays/StaticArrays.jl/blob/d03a5e6aa402dbafab1ed53f44835645aa1746a4/src/linalg.jl#L211), but it seems unused?

If the suggested fix seems acceptable, I can make a PR.

thchr:
This closes the issue noted in #915, i.e. the following now works:
```jl
a  = SA[SA[1, 2], SA[3, 4]]
norm(a)
```
which threw before. It also fixes a similar issue for the "p"-`norm`.

This also fixes a few other issues with `norm` and friends:

- Previously, the implementation of both the `norm` and "p"-`norm` were susceptible to integer overflow because they didn't convert to float early. They do now. This reduces performance a slight bit for `::SVector{Int}` but that seems preferable to quietly overflowing.
- Previously, the "p"-`norm` implementation was type-unstable for `SVector{Int}` inputs (sometimes returning `Float64` sometimes `Int`, depending on `p`). It now always returns a floating point value, consistently with Base.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -208,11 +208,13 @@ end
 
 #--------------------------------------------------
 # Norms
-function _inner_eltype(v::AbstractArray)
-    return eltype(v) <: AbstractArray ? _inner_eltype(first(v)) : eltype(v)
+function _inner_eltype(v::StaticArray)
+    isempty(v) && return float(norm(zero(eltype(v))))
+    return _inner_eltype(first(v))
 end
 _inner_eltype(x::Number) = typeof(x)
-@inline _init_zero(v::AbstractArray) = zero(float(real(_inner_eltype(v))))
+@inline _init_zero(v::StaticArray) = float(norm(zero(_inner_eltype(v))))
+
 @inline function LinearAlgebra.norm_sqr(v::StaticArray)
     return mapreduce(LinearAlgebra.norm_sqr, +, v; init=_init_zero(v))
 end
@@ -235,7 +237,7 @@ end
 end
 
 function _norm_p0(x)
-    T = float(real(_inner_eltype(x)))
+    T = isempty(x) ? eltype(x) : float(norm(zero(first(x))))
     return iszero(x) ? zero(T) :  one(T)
 end
 
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -208,17 +208,24 @@ end
 
 #--------------------------------------------------
 # Norms
-@inline LinearAlgebra.norm_sqr(v::StaticVector) = mapreduce(abs2, +, v; init=zero(real(eltype(v))))
+function _inner_eltype(v::AbstractVector)
+    return eltype(v) <: AbstractVector ? _inner_eltype(first(v)) : eltype(v)
+end
+_inner_eltype(x::Number) = typeof(x)
+@inline _init_zero(v::AbstractVector) = zero(float(real(_inner_eltype(v))))
+@inline function LinearAlgebra.norm_sqr(v::StaticVector)
+    return mapreduce(LinearAlgebra.norm_sqr, +, v; init=_init_zero(v))
+end
 
 @inline norm(a::StaticArray) = _norm(Size(a), a)
 @generated function _norm(::Size{S}, a::StaticArray) where {S}
     if prod(S) == 0
-        return :(zero(real(eltype(a))))
+        return :(_init_zero(a))
     end
 
-    expr = :(abs2(a[1]))
+    expr = :(LinearAlgebra.norm_sqr(a[1]))
     for j = 2:prod(S)
-        expr = :($expr + abs2(a[$j]))
+        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]))
     end
 
     return quote
@@ -227,28 +234,28 @@ end
     end
 end
 
-_norm_p0(x) = x == 0 ? zero(x) : one(x)
+_norm_p0(x) = iszero(x) ? zero(float(_inner_eltype(x))) : one(float(_inner_eltype(x)))
 
 @inline norm(a::StaticArray, p::Real) = _norm(Size(a), a, p)
 @generated function _norm(::Size{S}, a::StaticArray, p::Real) where {S}
     if prod(S) == 0
-        return :(zero(real(eltype(a))))
+        return :(_init_zero(a))
     end
 
-    expr = :(abs(a[1])^p)
+    expr = :(norm(a[1])^p)
     for j = 2:prod(S)
-        expr = :($expr + abs(a[$j])^p)
+        expr = :($expr + norm(a[$j])^p)
     end
 
-    expr_p1 = :(abs(a[1]))
+    expr_p1 = :(norm(a[1]))
     for j = 2:prod(S)
-        expr_p1 = :($expr_p1 + abs(a[$j]))
+        expr_p1 = :($expr_p1 + norm(a[$j]))
     end
 
     return quote
         $(Expr(:meta, :inline))
         if p == Inf
-            return mapreduce(abs, max, a)
+            return mapreduce(norm, max, a)
         elseif p == 1
             @inbounds return $expr_p1
         elseif p == 2
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -208,10 +208,7 @@ end
 
 #--------------------------------------------------
 # Norms
-function _inner_eltype(v::StaticArray)
-    isempty(v) && return float(norm(zero(eltype(v))))
-    return _inner_eltype(first(v))
-end
+_inner_eltype(v::AbstractArray) = isempty(v) ? eltype(v) : _inner_eltype(first(v))
 _inner_eltype(x::Number) = typeof(x)
 @inline _init_zero(v::StaticArray) = float(norm(zero(_inner_eltype(v))))
 
@@ -237,8 +234,8 @@ end
 end
 
 function _norm_p0(x)
-    T = isempty(x) ? eltype(x) : float(norm(zero(first(x))))
-    return iszero(x) ? zero(T) :  one(T)
+    T = _inner_eltype(x)
+    return float(norm(iszero(x) ? zero(T) : one(T)))
 end
 
 @inline norm(a::StaticArray, p::Real) = _norm(Size(a), a, p)
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -234,7 +234,10 @@ end
     end
 end
 
-_norm_p0(x) = iszero(x) ? zero(float(_inner_eltype(x))) : one(float(_inner_eltype(x)))
+function _norm_p0(x)
+    T = float(real(_inner_eltype(x)))
+    return iszero(x) ? zero(T) :  one(T)
+end
 
 @inline norm(a::StaticArray, p::Real) = _norm(Size(a), a, p)
 @generated function _norm(::Size{S}, a::StaticArray, p::Real) where {S}
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -208,12 +208,12 @@ end
 
 #--------------------------------------------------
 # Norms
-function _inner_eltype(v::AbstractVector)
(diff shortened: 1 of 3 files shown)


thchr:
Regarding test failures:
- **Failure in `test/broadcast.jl`**: this one was genuine - I needed to widen the type of a few methods to not just allow `(Static)Vector`s but `StaticArray`s or `AbstractArray`s. Fixed.
- **Failure in `test/ambiguities.jl`**: seems to just be broken generally on higher versions of Julia - I see this in other PRs and in local testing. 
  I see 61 ambiguities on the v1.2.3 of StaticArrays.jl on Julia 1.6.0; it's 49 with this PR, so actually seems better?
- **Failures in `test/arraymath.jl`**: throws on testing that `@allocated(arithmetic_closure(T0)) == 0` for `T0` in `Int64`, `Float32`, `Float64`. Also did that before this PR afaict though (tested locally) - so seems unrelated.
- **Failure in `test/svd.jl`**: an inference failure that also happens on v1.2.3 of StaticArrays.jl on Julia v1.6. Unrelated to this PR. Seems to be due to a failure to constant propagate a keyword from `svd(A; full::Bool)` to a `Val` in `_svd(A, ::Val{<:Bool})`.

I tested that the remaining tests pass locally (on Julia v1.6.0).

thchr:
CI is green on this - anyone up for reviewing this? (could I bother you perhaps @mateuszbaran?)

(I know caring about `norm` on nested arrays might seem low priority - but it would be nice to have this so that approximate comparisons of nested `SVector`s would work).

mateuszbaran:
I'll take a look at this soon. There has recently been some trouble with `reshape` so I don't want to merge more things too quickly.

thchr:
Thanks! 
(For what it's worth: this is rebased on top of your recent reshape fix (and I don't think it should interact with reshape in any way))

src/linalg.jl
```diff
@@ -208,17 +208,24 @@ end
 
 #--------------------------------------------------
 # Norms
-@inline LinearAlgebra.norm_sqr(v::StaticVector) = mapreduce(abs2, +, v; init=zero(real(eltype(v))))
+function _inner_eltype(v::AbstractArray)
+    return eltype(v) <: AbstractArray ? _inner_eltype(first(v)) : eltype(v)
```
mateuszbaran, line 212: Can't you just return `_inner_eltype(first(v))` here? There should be no need to call `eltype`, which might be an abstract type. Note that, for example, this works:
```julia
julia> norm(Vector{Number}[Number[1.0, 2]])
2.23606797749979
```

src/linalg.jl
```diff
@@ -227,28 +234,28 @@ end
     end
 end
 
-_norm_p0(x) = x == 0 ? zero(x) : one(x)
+_norm_p0(x) = iszero(x) ? zero(float(_inner_eltype(x))) : one(float(_inner_eltype(x)))
```
mateuszbaran, line 237: Shouldn't there be a `float(real(...` to handle complex numbers right?

test/linalg.jl
```diff
@@ -202,6 +202,48 @@ StaticArrays.similar_type(::Union{RotMat2,Type{RotMat2}}) = SMatrix{2,2,Float64,
         @test normalize(SVector(1,2,3), 1) ≈ normalize([1,2,3], 1)
         @test normalize!(MVector(1.,2.,3.)) ≈ normalize([1.,2.,3.])
         @test normalize!(MVector(1.,2.,3.), 1) ≈ normalize([1.,2.,3.], 1)
+
+        # nested vectors
+        a  = SA[SA[1, 2], SA[3, 4]]
```
mateuszbaran, line 207: Could you add some tests with complex numbers?

test/linalg.jl
```diff
@@ -202,6 +202,48 @@ StaticArrays.similar_type(::Union{RotMat2,Type{RotMat2}}) = SMatrix{2,2,Float64,
...
+        @test norm(a, 0) ≈ norm(a′, 0)
+
+        # type-stability
+        if VERSION ≥ v"1.2"
+            # only test strict type-stability on v1.2+, since there were Base-related type
+            # instabilities in `norm` prior to https://github.com/JuliaLang/julia/pull/30481
+            @test (@inferred norm(a[1])) == (@inferred norm(a[1], 2))
+            @test (@inferred norm(a)) == (@inferred norm(a, 2))
+            @test (@inferred norm(aa)) == (@inferred norm(aa, 2))
+            @test (@inferred norm(float.(aa))) == (@inferred norm(float.(aa), 2))
```
mateuszbaran, line 239: Could you add an `@inferred` test for norm of `SVector` of length 0?

test/linalg.jl
```diff
@@ -202,6 +202,48 @@ StaticArrays.similar_type(::Union{RotMat2,Type{RotMat2}}) = SMatrix{2,2,Float64,
         @test normalize(SVector(1,2,3), 1) ≈ normalize([1,2,3], 1)
         @test normalize!(MVector(1.,2.,3.)) ≈ normalize([1.,2.,3.])
         @test normalize!(MVector(1.,2.,3.), 1) ≈ normalize([1.,2.,3.], 1)
+
+        # nested vectors
+        a  = SA[SA[1, 2], SA[3, 4]]
+        a′ = convert(Vector{Vector{Int}}, a)
```
mateuszbaran, line 208: That apostrophe looks almost like adjoint, could you change `a′` to `a2`, or `av` or something?

src/linalg.jl
thchr: Nice catch - definitely..!

src/linalg.jl
thchr: Yep, good point - this would have broken at abstractly typed containers. Your suggestion would work fine for non-empty arrays (of course, it would lead to type-unstability during summation for arrays like e.g. `Any[1, 1.0]` but that seems inevitable).

However, it would break for empty arrays. One way out would be then branch on `isempty(v)` and return `eltype(v)` in that case - but then the whole thing is type-unstable (even for non-abstractly typed iterators). Any ideas?

test/linalg.jl
thchr: Sure, done.

test/linalg.jl
thchr: Done.

test/linalg.jl
thchr: Done.

thchr:
Thanks for review and comments @mateuszbaran. There's one issue remaining (in `inner_eltype`, a conflict between desires for abstractly typed arrays and empty arrays) - let me know if you have ideas for how to fix that.

src/linalg.jl
mateuszbaran: Having a branch for `isempty` case is exactly what `LinearAlgebra` stdlib does: https://github.com/JuliaLang/julia/blob/f8d04b27a7514ecd6e9deb4f605f96984075c773/stdlib/LinearAlgebra/src/generic.jl#L625 . This should be type stable in most cases -- `zero(Any)` is undefined so it doesn't cause problems, and if `eltype(v)` is concrete then it's also fine. The problem is when `eltype(v)` is abstract and has `zero` defined, and `zero(eltype(v))` is of different type than in the non-empty `v` path.

src/linalg.jl
```diff
@@ -208,17 +208,24 @@ end
...
 #--------------------------------------------------
 # Norms
-@inline LinearAlgebra.norm_sqr(v::StaticVector) = mapreduce(abs2, +, v; init=zero(real(eltype(v))))
+function _inner_eltype(v::AbstractArray)
+    return eltype(v) <: AbstractArray ? _inner_eltype(first(v)) : eltype(v)
+end
+_inner_eltype(x::Number) = typeof(x)
+@inline _init_zero(v::AbstractArray) = zero(float(real(_inner_eltype(v))))
+@inline function LinearAlgebra.norm_sqr(v::StaticArray)
+    return mapreduce(LinearAlgebra.norm_sqr, +, v; init=_init_zero(v))
```
mateuszbaran, line 217:
```suggestion
    return sum(LinearAlgebra.norm_sqr, v)
```
How about doing it like that? Would there be some drawback?

mateuszbaran: Hm, perhaps the zero-initialization would diverge between this branch and other below... and the zero-initialization of `sum` doesn't fit here.

mateuszbaran:
Having `_init_zero` and `_inner_eltype` defined only for `StaticArray`s and `Number`s would make me more comfortable with this change. Currently it's trying to do things that I don't know how exactly should work. I think all other parts are good so I'd like to see this PR merged in a slightly limited version.

thchr:
Hmm, having `_init_zero` and `_inner_eltype` only defined for `StaticArray` and `Number` has some downsides though: e.g., if it doesn't include `AbstractArray`s, we'd then throw on this:
```jl
norm(SA[[0,0],[1,1]], 0)
```
since we need to infer the element type in `_norm_p0` (which would call `_inner_eltype` on `[0,0]`, hitting a `MethodError`).

To fix that, we can just use `T = isempty(x) ? eltype(x) : float(norm(zero(first(x))))` in the first line of `_norm_p0` instead of `T = _inner_eltype(x)` but then it's basically just code duplication.
I did that for now since that seemed the most consistent with your request to limit the method signatures.

For me though, it seems cleaner to me just to let the signature of `_inner_eltype` be maximally broad? Basically, I'm not sure I see what the downside to defining `_inner_eltype` is? It's a Base function and is only used very locally here. In any case, either approach works, so whichever is good for me.

mateuszbaran:
Well, I think `_norm_p0` should use `_inner_eltype`, and maybe it's actually fine to keep `_inner_eltype` defined for `AbstractArray`s -- I'll take another look at this later.

thchr:
@matbesancon gentle bump: would be good to wrap this one up while we both remember it :)

mateuszbaran:
I've looked at it again, would you be OK with `_norm_p0` calling `_inner_eltype` and `_inner_eltype` being restricted to `AbstractArray`? It should be fine enough then :slightly_smiling_face: .

mateuszbaran (approved):

mateuszbaran:
Thanks! One last thing, could you bump the patch version?

thchr:
Sure: rebased and bumped patch version in latest commit 👍

PR #932, opened 2021-07-01
Fix method ambiguity for `qr` (#931) and lingering ambiguities for `lu`

### LINKED ISSUE #920: `lu` seems to be ambiguous on Julia Nightly

thchr:
This basically does just what #921 did to fix #920 for #931; i.e., removes the ambiguity for `qr`.

While doing that, I noticed that #921 hadn't completely resolved the ambiguity for `lu`; e.g., if called as `lu(A, Val(false))` there would still be an ambiguity error; so I fixed that as well. Similarly, I believe it would previously have been ambiguous if provided with a `check` keyword argument.

Added a few tests as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.2.4"
+version = "1.2.5"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/lu.jl
+++ b/src/lu.jl
@@ -30,28 +30,21 @@ function Base.show(io::IO, mime::MIME{Symbol("text/plain")}, F::LU)
 end
 
 # LU decomposition
-function lu(A::StaticMatrix, pivot::Union{Val{false},Val{true}}=Val(true); check = true)
-    L, U, p = _lu(A, pivot, check)
-    LU(L, U, p)
-end
+for pv in (:true, :false)
+    # ... define each `pivot::Val{true/false}` method individually to avoid ambiguties
+    @eval function lu(A::StaticMatrix, pivot::Val{$pv}; check = true)
+        L, U, p = _lu(A, pivot, check)
+        LU(L, U, p)
+    end
 
-# For the square version, return explicit lower and upper triangular matrices.
-# We would do this for the rectangular case too, but Base doesn't support that.
-function lu(A::StaticMatrix{N,N}, pivot::Union{Val{false},Val{true}}=Val(true);
-            check = true) where {N}
-    L, U, p = _lu(A, pivot, check)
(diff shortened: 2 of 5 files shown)


thchr:
The failure on nightly is unfortunately related to the strategy for resolving these ambiguities: `det` doesn't infer because it calls `lu(::StaticMatrix, Val{...}; check)` which now no longer infers because it `@invoke`s a method with a keyword argument. I.e., this happens:
```jl
f(x::Float64; kw=2) = x+kw
f(x::Int;     kw=2) = x+kw

fInt(x; kw=2) = Base.@invoke f(x::Int; kw)
fFloat(x; kw=2) = Base.@invoke f(x::Float64; kw)
```
where neither `fInt` nor `fFloat` infers, despite both methods of `f` inferring just fine.

So the crux of the problem seems to be that using `@invoke` to define `lu` with a `check` kwarg is bad: but the `check` kwarg is needed occasionally (e.g. in `det` apparently).

What would be a good way to sidestep this? Not inferring on `lu` obviously isn't workable.

thchr:
The issue with `@invoke` and keyword arguments already was already noted previously in https://github.com/JuliaLang/julia/issues/29216.

thchr:
Okay, to fix this, I just dropped the `@invoke` approach entirely and used `@eval` to instead define methods only for `Val{true}` and `Val{false}` pivots, avoiding the need to define the ambiguous `Union{Val{true}, Val{false}` method at all. 
I did this for both `qr` and `lu` for consistency, although it's only really necessary for `lu`.

Locally, all tests pass for me on nightly now.

mateuszbaran:
Thanks for investigating it, this approach looks fine. Could you bump the patch version?

thchr:
Sure: done!

PR #933, opened 2021-07-02
Fix reshaping of a mutable StaticArray

mateuszbaran:
This is an attempt to fix the issue described in https://github.com/JuliaArrays/StaticArrays.jl/pull/907 . I've tried to make changes as limited as possible, so `reshape(::AbstractArray, ::Size)` still makes a copy. I can change that but doing so is not essential to fixing the key problem, that is `reshape(::StaticArray, ::SOneTo...)` making a copy, which is fixed here.

This looks like a change that may have some impact on downstream packages so I hope to get some feedback :slightly_smiling_face: .

cc @KristofferC @dlfivefifty @andyferris .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.2.5"
+version = "1.2.6"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


dlfivefifty:
My understanding is `reshape` should never `copy`...

If this breaks dependent packages this is really a bug in those packages since they shouldn't have assumed it was making a `copy`.

KristofferC:
Yes, should be totally fine to fix this bug.

mateuszbaran:
OK, I've changed it so that no reshape methods (except the one for `SArray`) make a copy. Let's see how it goes.

PR #936, opened 2021-07-15
Add a two-argument show method for SUnitRange.

tpapp:
This allows recovery of containers etc with `SUnitRange` from printed output.

Cf [discussion](https://discourse.julialang.org/t/print-sunitrange-within-tuple-as-is-not-as-a-vector/64678/).


mateuszbaran (approved):
Good idea, thanks!

andyferris (approved):

mateuszbaran:
Could you bump the patch version number for tagging a new release?

tpapp:
Done.

PR #941, opened 2021-07-26
implement `float` and `real` for `StaticArray` (fix #935)

### LINKED ISSUE #935: type-valued float and real functions
Since we often treat StaticArrays almost like scalars, it would be nice to have `float` and `real` functions that act on the *types*.   i.e.
```jl
float(::Type{SArray{S,T,N,L}}) where {S,T,N,L} = SArray{S,float(T),N,L}
real(::Type{SArray{S,T,N,L}}) where {S,T,N,L} = SArray{S,real(T),N,L}
```
analogous to e.g. `float(Int) === Float64` and `real(ComplexF32) == Float32` in `Base`.

Basically `float(T) === typeof(float(zero(T)))`, and similarly for `real`.

thchr:
This just implements the suggestion in #935 for `SArray` and `MArray` and includes some tests.
I put the definitions in `src/convert.jl` - but let me know if they fit better elsewhere.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -13,7 +13,7 @@ import Statistics: mean
 using Random
 import Random: rand, randn, randexp, rand!, randn!, randexp!
 using Core.Compiler: return_type
-import Base: sqrt, exp, log
+import Base: sqrt, exp, log, float, real
 using LinearAlgebra
 import LinearAlgebra: transpose, adjoint, dot, eigvals, eigen, lyap, tr,
                       kron, diag, norm, dot, diagm, lu, svd, svdvals,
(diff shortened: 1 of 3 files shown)


thchr:
Bump: given reactions in #935, I suspect there's not anything controversial about this?

andyferris (commented):
I wonder if it possible to do this for all `StaticArray` with a generic function rather than the macro approach? We have `similar_type` for example.

thchr:
That's a nice idea - thanks: I've added a commit that goes via `similar_type` and `StaticArray` instead.
With this implementation, this should now work also for subtypes of `StaticArray` that extend the two-argument `similar_type`.

thchr:
Bump: this is ready for review and was changed following suggestion in https://github.com/JuliaArrays/StaticArrays.jl/pull/941#pullrequestreview-753320510.

thchr:
Gentle bump.

andyferris (approved):
LGTM

@hyrodium we should fix these up for Rotations.jl, where `similar_type` here won’t preserve the type…

PR #948, opened 2021-08-01
Add `reshape` methods

### LINKED ISSUE #947: Missing methods for `reshape`?
I thought this would work, and produce a SizedArray:
```julia
julia> using StaticArrays

julia> reshape(hcat([1,2,3]), axes(SA[1,2,3]))
ERROR: MethodError: no method matching reshape(::Matrix{Int64}, ::Tuple{SOneTo{3}})

julia> SizedArray{Tuple{3}}([1,2,3])  # what I expected:
3-element SizedVector{3, Int64, Vector{Int64}} with indices SOneTo(3):
 1
 2
 3
```
Context is https://github.com/FluxML/Zygote.jl/pull/1044, trying to make AD more careful about restoring the types of its inputs. Reshaping like this works for OffsetArrays, which in fact (piratically) resolves the above error:
```
julia> using OffsetArrays

julia> reshape(hcat([1,2,3]), axes(SA[1,2,3]))  # this works...
3-element OffsetArray(::Vector{Int64}, 1:3) with eltype Int64 with indices 1:3:
 1
 2
 3

julia> reshape(hcat([1,2,3]), axes(OffsetArray([1,2,3],3)))  # as a size-effect of this ...
3-element OffsetArray(::Vector{Int64}, 4:6) with eltype Int64 with indices 4:6:
 1
 2
 3

julia> reshape(hcat([1,2,3]), (4:6,))  # ... because it's piratically broad.
3-element OffsetArray(::Vector{Int64}, 4:6) with eltype Int64 with indices 4:6:
 1
 2
 3
```

mcabbott:
Fixes #947

But one case doesn't work, and I didn't understand why yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.2.9"
+version = "1.2.10"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -186,6 +186,7 @@ homogenize_shape(shape::Tuple{Vararg{HeterogeneousShape}}) = map(last, shape)
 @inline reshape(a::SArray, s::Size) = similar_type(a, s)(Tuple(a))
 @inline reshape(a::AbstractArray, s::Size) = __reshape(a, ((typeof(s).parameters...)...,), s)
 @inline reshape(a::SArray, s::Tuple{SOneTo,Vararg{SOneTo}}) = reshape(a, homogenize_shape(s))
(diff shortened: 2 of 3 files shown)


mateuszbaran:
I think you have to make a special case for reshaping `AbstractVector`, similarly to `reshape(a::Vector, ::Size{S}) where {S} = SizedArray{Tuple{S...}}(a)`.

mcabbott:
Ok, great. Somehow I assumed the Array inside would be reshaped, but I guess there is no need.

mateuszbaran:
Yes, `SizedArray` doesn't do any reshaping of the wrapped array if there is no need to, and in general it relies on linear indexing to access the wrapped array.

mateuszbaran (approved):

mateuszbaran:
OK, I'm going to merge this tomorrow. Could you bump patch version?

PR #949, opened 2021-08-05
specialize muladd

ranocha:
This avoids hitting the general fallback https://github.com/JuliaLang/julia/blob/5118a1babf9d2963f31e98e427cc99a969d5ca46/base/math.jl#L1167 
Depending on the processor, LLVM can turn this into efficient FMA calls. Depending on the workload, this can speed-up number crunching quite a bit.

I increased the version number since I would really like to be able to use this performance improvement.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.2.10"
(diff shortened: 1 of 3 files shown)


mateuszbaran (approved):

ranocha:
Thanks for approving this so quickly, @mateuszbaran! Looks like all tests pass and we should be good to go

PR #950, opened 2021-08-18
Small tweaks to prevent `==` invalidation

timholy:
These improve inferrence to guard against very minor invalidations
from packages that extend `==`.

### DIFF
--- a/src/matrix_multiply_add.jl
+++ b/src/matrix_multiply_add.jl
@@ -390,11 +390,11 @@ end
 
 function combine_products(expr_list)
     filtered = filter(expr_list) do expr
-        if expr.head != :call || expr.args[1] != :*
+        if expr.head != :call || expr.args[1] !== :*
             error("expected call to *")
         end
         for arg in expr.args[2:end]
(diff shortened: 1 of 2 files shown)


src/matrix_multiply_add.jl
```diff
@@ -390,11 +390,11 @@ end
 
 function combine_products(expr_list)
     filtered = filter(expr_list) do expr
-        if expr.head != :call || expr.args[1] != :*
+        if expr.head != :call || expr.args[1] !== :*
```
mateuszbaran, line 393: Maybe it should also be `expr.head !== :call`?

timholy: As long as all the elements in `expr_list` are known by inference to be `Expr`s, it's not necessary:

```julia
julia> @code_typed (ex->ex.head)(:(sin(3.2)))
CodeInfo(
1 ─ %1 = Base.getfield(ex, :head)::Symbol
└──      return %1
) => Symbol
```

I typically use `===` and `!==` only when it makes a difference (when one of the two sides is uninferred).

mateuszbaran (approved):

PR #951, opened 2021-08-19
Generalize exception type

timholy:
In preparation for https://github.com/JuliaLang/julia/pull/41885

### DIFF
--- a/test/mapreduce.jl
+++ b/test/mapreduce.jl
@@ -90,10 +90,17 @@ using Statistics: mean
 
         # When the mapping and/or reducing functions are unsupported,
         # the error is thrown by `Base.mapreduce_empty`:
-        @test_throws(
-            ArgumentError("reducing over an empty collection is not allowed"),
-            mapreduce(nothing, nothing, SVector{0,Int}())
-        )
+        if Base.VERSION >= v"1.8.0-DEV.363"
+            @test_throws(
+                "reducing over an empty collection is not allowed",
+                mapreduce(nothing, nothing, SVector{0,Int}())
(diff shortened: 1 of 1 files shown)


test/mapreduce.jl
```diff
@@ -90,10 +90,17 @@ using Statistics: mean
...
         # the error is thrown by `Base.mapreduce_empty`:
-        @test_throws(
-            ArgumentError("reducing over an empty collection is not allowed"),
-            mapreduce(nothing, nothing, SVector{0,Int}())
-        )
+        if Base.VERSION >= v"1.8.0-DEV.363"
+            @test_throws(
+                "reducing over an empty collection is not allowed",
+                mapreduce(nothing, nothing, SVector{0,Int}())
+            )
```
c42f, line 97: For anyone (like myself) thinking this use of `@test_throws` looks confusing... see https://github.com/JuliaLang/julia/pull/41888

c42f (approved):
Looks good to me, pending this getting merged upstream.

timholy:
> pending this getting merged upstream.

That's of course fine, but the fact that it returned an ArgumentError was a bit dicey from the start. One can argue that `Union{MethodError,ArgumentError}` is a better type for the error, but the new approach allows you to care only about the message itself (which seems to be what the original author of this test cared about).

timholy:
(clicked the wrong button, sorry)

c42f (approved):
Ah yes, technically this only depends on https://github.com/JuliaLang/julia/pull/41888 which has already been merged. So this seems fine.

PR #970, opened 2021-11-03
Update versions in ci.yml

hyrodium:
[The document](https://juliaarrays.github.io/StaticArrays.jl/dev/) is generated with Julia 1.4.2, but this can be updated.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,10 +54,10 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: julia-actions/setup-julia@latest
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@1
         with:
-          version: '1.4'
+          version: '1'
       - run: julia --project=docs -e '
           using Pkg;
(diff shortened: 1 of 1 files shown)


andyferris (approved):

.github/workflows/ci.yml
```diff
@@ -54,10 +54,10 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: julia-actions/setup-julia@latest
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
         with:
-          version: '1.4'
+          version: '1'
```
c42f, line 60: It may be better to specify a particular version here, for example `1.6`.

The reason is that `jldoctest` blocks are very sensitive to formatting and the output of `Base.show` is not considered part of the stable interface for Base.

So doctests can break between Julia versions.

thchr, line 60:
```suggestion
          version: '1.6'
```

(just so you can press it in the online interface; then I think we could merge this immediately after)

hyrodium:
Close this PR because #985 is better.

Issue #971, opened 2021-11-10
Symmetric(StaticMatrix) causes several problems


vincent-picaud:
Hello,

Thank you for this package,

Starting to use it, I encountered several problems with Symmetric(StaticMatrix). This prevents me to have some generic code (working without modification for StaticMatrices and Julia regular matrix types).

**Example 1: M\v strange behavior**

M\v does not work with this invertible M matrix
```
julia> using LinearAlgebra, StaticArrays

julia> M = Symmetric(@SMatrix Float64[1 1;1 2])
2×2 Symmetric{Float64, SMatrix{2, 2, Float64, 4}}:
 1.0  1.0
 1.0  2.0

julia> v = @SVector Float64[1,2]
2-element SVector{2, Float64} with indices SOneTo(2):
 1.0
 2.0

julia> M\v
ERROR: setindex!(::SMatrix{2, 2, Float64, 4}, value, ::Int) is not defined.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] setindex!(a::SMatrix{2, 2, Float64, 4}, value::Float64, i::Int64)
    @ StaticArrays ~/.julia/packages/StaticArrays/OWJK7/src/indexing.jl:3
  [3] macro expansion
    @ ~/.julia/packages/StaticArrays/OWJK7/src/indexing.jl:66 [inlined]
  [4] _setindex!_scalar
    @ ~/.julia/packages/StaticArrays/OWJK7/src/indexing.jl:46 [inlined]
  [5] setindex!
    @ ~/.julia/packages/StaticArrays/OWJK7/src/indexing.jl:42 [inlined]
  [6] copytri!
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:514 [inlined]
  [7] copytri!
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:510 [inlined]
  [8] lu!(A::Symmetric{Float64, SMatrix{2, 2, Float64, 4}}, pivot::Val{true}; check::Bool)
    @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lu.jl:89
  [9] lu(A::Symmetric{Float64, SMatrix{2, 2, Float64, 4}}, pivot::Val{true}; check::Bool)
[...]

PR #972, opened 2021-11-10
LU decomposition for symmetric and Hermitian static matrices

mateuszbaran:
This is an attempt at solving a part of #971 (specifically, example 1).


src/lu.jl
```diff
@@ -110,6 +112,9 @@ __lu(A::StaticMatrix{M,0,T}, ::Val{Pivot}) where {T,M,Pivot} =
 __lu(A::StaticMatrix{1,1,T}, ::Val{Pivot}) where {T,Pivot} =
     (SMatrix{1,1}(one(T)), A, SVector(1))
 
+__lu(A::LinearAlgebra.HermOrSym{T,<:StaticMatrix{1,1,T}}, ::Val{Pivot}) where {T,Pivot} =
```
thchr, line 115: Out of curiosity, what is the purpose of this additional method? Just a performance optimization?

thchr (approved):
I read through this PR and it seems reasonable and correct to me.

src/lu.jl
mateuszbaran: IIRC yes, that's just an optimization similar to `__lu(A::StaticMatrix{1,1,T}, ::Val{Pivot})`.

mateuszbaran:
I keep getting too many seemingly random CI failures here that need to be resolved before it can be merged.

PR #973, opened 2021-11-29
implement Statistics.median

stev47:
This implements `Statistics.median` based on the existing bitonic
sorting, avoiding unnecessary allocation.
While it is generally suboptimal to sort the whole array, the compiler
manages to skip some branches since only the middle element(s) are used.
Thus `median` is generally faster than `sort`.

Using a dedicated median selection network could yield better
performance and might be considered for future improvement.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.9.9"
+version = "1.9.10"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/ext/StaticArraysStatisticsExt.jl
+++ b/ext/StaticArraysStatisticsExt.jl
@@ -15,6 +15,16 @@ _mean_denom(a, ::Val{D}) where {D} = size(a, D)
 @inline mean(a::StaticArray; dims=:) = _reduce(+, a, dims) / _mean_denom(a, dims)
 @inline mean(f::Function, a::StaticArray; dims=:) = _mapreduce(f, +, dims, _InitialValue(), Size(a), a) / _mean_denom(a, dims)
 
+@inline function median(a::StaticArray; dims = :)
+    if dims == Colon()
+        median(vec(a))
+    else
+        # FIXME: Implement `mapslices` correctly on `StaticArray` to remove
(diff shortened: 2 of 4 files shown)


RoyiAvital:
Any chance to push this forward?

stev47:
> Any chance to push this forward?

updated the MR, waiting for review

mkitti:
This pull request implements median for `StaticVector`. It supports the empty array, missing, and NaN edge cases. For regular functionality, the tests includes a comparison to the `Vector` equivalent of a random `StaticVector` 

While testing random static vectors is useful, defined tests would be welcome. In particular, one could replicate the the tests in Statistics.jl for `median`:
https://github.com/JuliaStats/Statistics.jl/blob/d49c2bf4f81e1efb4980a35fe39c815ef8396297/test/runtests.jl#L31-L92

stev47:
> [...] defined tests would be welcome. In particular, one could replicate the the tests in Statistics.jl for `median`:

ok, I've added the tests from Statistics.jl

RoyiAvital:
@mkitti , Any chance t ping another reviewer to push this?

mateuszbaran:
I tried benchmarking it and there seems to be an issue:
```julia
julia> a = [1, 10, 2, 1.2, -19, 12];

julia> sa = SVector{length(a)}(a);

julia> @btime median($a)
  24.433 ns (2 allocations: 112 bytes)
1.6

julia> @btime median($sa)
  515.203 ns (5 allocations: 176 bytes)
1.6
```
From a quick look it seems like `nanix` being an union might be the culprit. Could you improve the performance here?

stev47:
> I tried benchmarking it and there seems to be an issue:

Can you try in a fresh session? I cannot reproduce your timings (Julia 1.11.2):
```julia
julia> using StaticArrays

julia> using Statistics

julia> a = [1, 10, 2, 1.2, -19, 12];

julia> sa = SVector{length(a)}(a);

julia> @btime median($a)
  31.548 ns (2 allocations: 112 bytes)
1.6

julia> @btime median($sa)
  8.529 ns (0 allocations: 0 bytes)
1.6
```
similar for both Julia 1.10.7 and Julia 1.12.0-DEV.1793 at commit 5ee67b8551.

mateuszbaran (approved):
You're right, it was an issue with my REPL session. Everything looks fine. Let's merge this.

PR #975, opened 2021-12-06
Avoid underflow and overflow in norm()

### LINKED ISSUE #913: norm of StaticArray inaccurate
I have traced a bug in my code to the fact that `norm(::StaticArray)` is considerably less accurate than `norm(::Array)`. In code working with rotations, the norm of a small vector, length 2-4, is often used required to very high precision. 

To verify that the problem is the naive implementation of norm for static arrays, I modified the norm function like so
```julia
@generated function _norm(::Size{S}, a::StaticArray) where {S}
    if prod(S) == 0
        return :(zero(real(eltype(a))))
    end
    
    expr = :(abs2(a[1]))
    for j = 2:prod(S)
        expr = :($expr + abs2(a[$j]))
    end
    
    return quote
        return norm(Array(a))    # <------------- convert to standard array and return norm of this instead 
        $(Expr(:meta, :inline))
        @inbounds return sqrt($expr)
    end
end
```
after which the problem goes away. 

Would it be feasible to have an implementation of norm that is more accurate, at least for small vectors?

MWE:
```julia

julia> a = SA[0, 1e-180]
2-element SVector{2, Float64} with indices SOneTo(2):
 0.0
 1.0e-180

julia> norm(a)
0.0

julia> norm(Array(a))
1.0e-180
```

wsshin:
This PR fixes #913 by avoiding underflow in `norm()`.  In the referenced issue, `norm(SA[0, 1e-180])` generates `0.0` instead of the accurate value `1e-180`, because squaring `1e-180` inside `norm()` produces `1e-360`, whose exponent is less than the [smallest exponent allowed in the IEEE double-precision floating point standard](https://en.wikipedia.org/wiki/Double-precision_floating-point_format#Exponent_encoding): `2^-1022 ≈ 1e-308`.  

The standard practice to avoid this underflow is to pre-normalize the vector by its largest-magnitude entry, calculate the norm, and then undo the normalization outside `norm()`:
```julia
# Calculate norm(v) without underflow.
vmax = maximum(abs, v)
norm_v = vmax * norm(v ./ vmax)
```

This PR implements the above procedure for `StaticArrays`.  This PR also fixes https://github.com/JuliaDiff/ForwardDiff.jl/issues/547.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -252,14 +252,13 @@ end
     return mapreduce(LinearAlgebra.norm_sqr, +, v; init=_init_zero(v))
 end
 
-@inline norm(a::StaticArray) = _norm(Size(a), a)
-@generated function _norm(::Size{S}, a::StaticArray) where {S}
-    if prod(S) == 0
+@generated function norm(a::StaticArray)
+    if prod(Size(a)) == 0
         return :(_init_zero(a))
     end
 
     expr = :(LinearAlgebra.norm_sqr(a[1]/aₘ))
-    for j = 2:prod(S)
+    for j = 2:prod(Size(a))
         expr = :($expr + LinearAlgebra.norm_sqr(a[$j]/aₘ))
     end
 
@@ -280,19 +279,18 @@ function _norm_p0(x)
     return float(norm(iszero(x) ? zero(T) : one(T)))
 end
 
-@inline norm(a::StaticArray, p::Real) = _norm(Size(a), a, p)
-@generated function _norm(::Size{S}, a::StaticArray, p::Real) where {S}
-    if prod(S) == 0
+@generated function norm(a::StaticArray, p::Real)
+    if prod(Size(a)) == 0
         return :(_init_zero(a))
     end
 
     expr = :(norm(a[1]/aₘ)^p)
-    for j = 2:prod(S)
+    for j = 2:prod(Size(a))
         expr = :($expr + norm(a[$j]/aₘ)^p)
     end
 
     expr_p1 = :(norm(a[1]/aₘ))
-    for j = 2:prod(S)
+    for j = 2:prod(Size(a))
         expr_p1 = :($expr_p1 + norm(a[$j]/aₘ))
     end
 
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -229,14 +229,20 @@ end
         return :(_init_zero(a))
     end
 
-    expr = :(LinearAlgebra.norm_sqr(a[1]))
+    expr = :(LinearAlgebra.norm_sqr(a[1]/aₘ))
     for j = 2:prod(S)
-        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]))
+        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]/aₘ))
     end
 
     return quote
         $(Expr(:meta, :inline))
-        @inbounds return sqrt($expr)
+        zero_a = _init_zero(a)
+        aₘ = mapreduce(norm, max, a)::typeof(zero_a)
+        if iszero(aₘ)
+            return zero_a
+        else
+            @inbounds return aₘ * sqrt($expr)
+        end
     end
 end
 
@@ -251,28 +257,32 @@ end
         return :(_init_zero(a))
     end
 
-    expr = :(norm(a[1])^p)
+    expr = :(norm(a[1]/aₘ)^p)
     for j = 2:prod(S)
-        expr = :($expr + norm(a[$j])^p)
+        expr = :($expr + norm(a[$j]/aₘ)^p)
     end
 
-    expr_p1 = :(norm(a[1]))
+    expr_p1 = :(norm(a[1]/aₘ))
     for j = 2:prod(S)
-        expr_p1 = :($expr_p1 + norm(a[$j]))
+        expr_p1 = :($expr_p1 + norm(a[$j]/aₘ))
     end
 
     return quote
         $(Expr(:meta, :inline))
-        if p == Inf
-            return mapreduce(norm, max, a)
+        zero_a = _init_zero(a)
+        aₘ = mapreduce(norm, max, a)::typeof(zero_a)
+        if iszero(aₘ)
+            return zero_a
+        elseif p == Inf
+            return aₘ
         elseif p == 1
-            @inbounds return $expr_p1
+            @inbounds return aₘ * $expr_p1
         elseif p == 2
             return norm(a)
         elseif p == 0
             return mapreduce(_norm_p0, +, a)
         else
-            @inbounds return ($expr)^(inv(p))
+            @inbounds return aₘ * ($expr)^(inv(p))
         end
     end
 end
@@ -466,4 +476,3 @@ end
 # Some shimming for special linear algebra matrix types
 @inline LinearAlgebra.Symmetric(A::StaticMatrix, uplo::Char='U') = (checksquare(A); Symmetric{eltype(A),typeof(A)}(A, uplo))
 @inline LinearAlgebra.Hermitian(A::StaticMatrix, uplo::Char='U') = (checksquare(A); Hermitian{eltype(A),typeof(A)}(A, uplo))
-
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -252,11 +252,7 @@ end
     return mapreduce(LinearAlgebra.norm_sqr, +, v; init=_init_zero(v))
 end
 
-@generated function norm(a::StaticArray)
-    if prod(Size(a)) == 0
-        return :(_init_zero(a))
-    end
-
+@generated function norm_scaled(a::StaticArray)
     expr = :(LinearAlgebra.norm_sqr(a[1]/aₘ))
     for j = 2:prod(Size(a))
         expr = :($expr + LinearAlgebra.norm_sqr(a[$j]/aₘ))
@@ -274,16 +270,33 @@ end
     end
 end
 
+@generated function norm(a::StaticArray)
+    if prod(Size(a)) == 0
+        return :(_init_zero(a))
+    end
+
+    expr = :(LinearAlgebra.norm_sqr(a[1]))
+    for j = 2:prod(Size(a))
+        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]))
+    end
+
+    return quote
+        $(Expr(:meta, :inline))
+        l = @inbounds sqrt($expr)
+        if iszero(l) || isinf(l)
+            return norm_scaled(a)
+        else
+            return l
+        end
+    end
+end
+
 function _norm_p0(x)
     T = _inner_eltype(x)
     return float(norm(iszero(x) ? zero(T) : one(T)))
 end
 
-@generated function norm(a::StaticArray, p::Real)
-    if prod(Size(a)) == 0
-        return :(_init_zero(a))
-    end
-
+@generated function norm_scaled(a::StaticArray, p::Real)
     expr = :(norm(a[1]/aₘ)^p)
     for j = 2:prod(Size(a))
         expr = :($expr + norm(a[$j]/aₘ)^p)
@@ -294,27 +307,57 @@ end
         expr_p1 = :($expr_p1 + norm(a[$j]/aₘ))
     end
 
-    expr_pInf = :(norm(a[1]/aₘ))
-    for j = 2:prod(Size(a))
-        expr_pInf = :(max($expr_pInf, norm(a[$j]/aₘ)))
-    end
-
     return quote
         $(Expr(:meta, :inline))
         zero_a = _init_zero(a)
         aₘ = maxabs_nested(a)
         if iszero(aₘ)
             return zero_a
-        elseif p == Inf
-            return aₘ * $expr_pInf
+        elseif p == 0
+            return mapreduce(_norm_p0, +, a)
         elseif p == 1
             @inbounds return aₘ * $expr_p1
         elseif p == 2
             return norm(a)
-        elseif p == 0
-            return mapreduce(_norm_p0, +, a)
         else
             @inbounds return aₘ * ($expr)^(inv(p))
+        end
+    end
+end
+
+@generated function norm(a::StaticArray, p::Real)
+    if prod(Size(a)) == 0
+        return :(_init_zero(a))
+    end
+
+    expr = :(norm(a[1])^p)
+    for j = 2:prod(Size(a))
+        expr = :($expr + norm(a[$j])^p)
+    end
+
+    expr_p1 = :(norm(a[1]))
+    for j = 2:prod(Size(a))
+        expr_p1 = :($expr_p1 + norm(a[$j]))
+    end
+
+    return quote
+        $(Expr(:meta, :inline))
+        p == Inf && return mapreduce(norm, max, a)  # no need to scale
+        p == 2 && return norm(a)  # norm(a) takes care of scaling
+
+        local l
+        if p == 0
+            l = mapreduce(_norm_p0, +, a)
+        elseif p == 1
+            l = @inbounds $expr_p1
+        else
+            l = @inbounds ($expr)^(inv(p))
+        end
+
+        if iszero(l) || isinf(l)
+            return norm_scaled(a, p)
+        else
+            return l
         end
     end
 end
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -291,7 +291,6 @@ end
 
         iszero(scale) && return _init_zero(a)
         p == 1 && return @inbounds scale * $expr_p1
-        p == 2 && return norm(a)
         return @inbounds scale * ($expr)^(inv(p))
     end
 end
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -224,17 +224,17 @@ _inner_eltype(x::Number) = typeof(x)
 end
 
 @generated function norm_scaled(a::StaticArray)
-    expr = :(LinearAlgebra.norm_sqr(a[1]/aₘ))
+    expr = :(LinearAlgebra.norm_sqr(a[1]/scale))
     for j = 2:prod(Size(a))
-        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]/aₘ))
+        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]/scale))
     end
 
     return quote
         $(Expr(:meta, :inline))
-        aₘ = LinearAlgebra.normInf(a)
+        scale = LinearAlgebra.normInf(a)
 
-        iszero(aₘ) && return aₘ
-        return @inbounds aₘ * sqrt($expr)
+        iszero(scale) && return scale
+        return @inbounds scale * sqrt($expr)
     end
 end
 
@@ -263,24 +263,24 @@ function _norm_p0(x)
 end
 
 @generated function norm_scaled(a::StaticArray, p::Real)
-    expr = :(norm(a[1]/aₘ)^p)
+    expr = :(norm(a[1]/scale)^p)
     for j = 2:prod(Size(a))
-        expr = :($expr + norm(a[$j]/aₘ)^p)
+        expr = :($expr + norm(a[$j]/scale)^p)
     end
 
-    expr_p1 = :(norm(a[1]/aₘ))
+    expr_p1 = :(norm(a[1]/scale))
     for j = 2:prod(Size(a))
-        expr_p1 = :($expr_p1 + norm(a[$j]/aₘ))
+        expr_p1 = :($expr_p1 + norm(a[$j]/scale))
     end
 
     return quote
         $(Expr(:meta, :inline))
-        aₘ = LinearAlgebra.normInf(a)
+        scale = LinearAlgebra.normInf(a)
 
-        iszero(aₘ) && return aₘ
-        p == 1 && return @inbounds aₘ * $expr_p1
+        iszero(scale) && return scale
+        p == 1 && return @inbounds scale * $expr_p1
         p == 2 && return norm(a)
-        return @inbounds aₘ * ($expr)^(inv(p))
+        return @inbounds scale * ($expr)^(inv(p))
     end
 end
 
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -219,6 +219,35 @@ _inner_eltype(v::AbstractArray) = isempty(v) ? eltype(v) : _inner_eltype(first(v
 _inner_eltype(x::Number) = typeof(x)
 @inline _init_zero(v::AbstractArray) = float(norm(zero(_inner_eltype(v))))
 
+@inline maxabs_nested(a::Number) = abs(a)
+
+function maxabs_nested(a::AbstractArray)
+    prod(size(a)) == 0 && (return _init_zero(a))
+
+    m = maxabs_nested(a[1])
+    for j = 2:prod(size(a))
+        m = max(m, maxabs_nested(a[j]))
+    end
+
+    return m
+end
+
+@generated function maxabs_nested(a::StaticArray)
+    if prod(Size(a)) == 0
+        return :(StaticArrays._init_zero(a))
+    end
+
+    expr = :(maxabs_nested(a[1]))
+    for j = 2:prod(Size(a))
+        expr = :(max($expr, maxabs_nested(a[$j])))
+    end
+
+    return quote
+        $(Expr(:meta, :inline))
+        @inbounds return $expr
+    end
+end
+
 @inline function LinearAlgebra.norm_sqr(v::StaticArray)
     return mapreduce(LinearAlgebra.norm_sqr, +, v; init=_init_zero(v))
 end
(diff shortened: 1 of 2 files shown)


src/linalg.jl
```diff
@@ -229,14 +229,20 @@ end
...
     for j = 2:prod(S)
-        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]))
+        expr = :($expr + LinearAlgebra.norm_sqr(a[$j]/aₘ))
     end
 
     return quote
         $(Expr(:meta, :inline))
-        @inbounds return sqrt($expr)
+        zero_a = _init_zero(a)
+        aₘ = mapreduce(norm, max, a)::typeof(zero_a)
```
wsshin, line 240: The return type assertion here is to avoid type instability when `a` is a vector of vectors (https://github.com/JuliaArrays/StaticArrays.jl/blob/master/test/linalg.jl#L278).  `mapreduce` produces a type-unstable output for such `a`.  If there is a better approach without using return type assertion, please make a suggestion.

src/linalg.jl
```diff
@@ -217,26 +217,60 @@ end
...
 
     return quote
         $(Expr(:meta, :inline))
-        @inbounds return sqrt($expr)
+        zero_a = _init_zero(a)
+        aₘ = maxabs_nested(a)
+        if iszero(aₘ)
+            return zero_a
+        else
+            @inbounds return aₘ * sqrt($expr)
```
mcabbott, line 272: This always takes the max, and then for nonzero norm always computes with the division. How does this compare to the speed of the existing code?

The alternative strategy would be just to compute the norm, and only if that is zero or infinite, worry about scaling. More work in the worst case but hopefully most cases in real use will be neither huge nor tiny. 

Xref https://github.com/JuliaLang/julia/pull/43256

src/linalg.jl
```diff
@@ -217,26 +217,60 @@ end
...
+
+    m = maxabs_nested(a[1])
+    for j = 2:prod(size(a))
+        m = max(m, maxabs_nested(a[j]))
+    end
+
+    return m
+end
+
+@generated function maxabs_nested(a::StaticArray)
```
mcabbott, line 235: Do you see speedups from this generated function? Or perhaps it serves another purpose?
```
julia> for N in [3, 10, 30]
       @show N
       x = @SVector randn(N)
       @btime maximum(abs, $x)
       @btime maxabs_nested($x)  # PR, without generated path
       @btime maxabs_nested_g($x)  # with
       end
N = 3
  2.458 ns (0 allocations: 0 bytes)
  2.417 ns (0 allocations: 0 bytes)
  2.416 ns (0 allocations: 0 bytes)
N = 10
  13.652 ns (0 allocations: 0 bytes)
  13.096 ns (0 allocations: 0 bytes)
  13.652 ns (0 allocations: 0 bytes)
N = 30
  77.830 ns (0 allocations: 0 bytes)
  73.826 ns (0 allocations: 0 bytes)
  77.786 ns (0 allocations: 0 bytes)
```

src/linalg.jl
wsshin: Always taking the max and dividing by it is what is done in `LinearAlgebra` to avoid overflow and underflow; see https://github.com/JuliaLang/julia/issues/1861#ref-pullrequest-165754044.

mcabbott: OK I timed things:
```julia
julia> @btime norm($(@SVector rand(10)));  # tagged version
  1.833 ns (0 allocations: 0 bytes)

julia> @btime sqrt(sum(abs2, $(@SVector rand(10))));
  1.833 ns (0 allocations: 0 bytes)

julia> @btime norm_gen($(@SVector rand(10)));  # this PR
  35.750 ns (2 allocations: 112 bytes)

julia> @btime mapreduce(abs, max, $(@SVector rand(10)));
  13.652 ns (0 allocations: 0 bytes)

julia> @btime @fastmath mapreduce(abs, max, $(@SVector rand(10)));
  2.708 ns (0 allocations: 0 bytes)
```
Here `@fastmath` will get `Inf`, `NaN` and `-0.0` wrong, but if it can be safely used, helps a lot.

src/linalg.jl
wsshin: Sorry, I didn't perform performance comparison on this.  I merely thought that `@generated` version would be always better for `StaticArrays`; now that I think about it `@generated` version of `maxabs_nested()` wouldn't be needed because the function is non-allocating anyway.

If the current push passes all CI tests, I will remove the `@generated` version of `maxabs_nested()`.  Thanks!

mcabbott: Might even be simpler just to write `mapreduce(abs, max, x)`. Or normInf I think. Unless you have a compelling reason to write it out for this case.

src/linalg.jl
wsshin: In the current Julia master, the relevant code is [here](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/generic.jl#L510-L519).

mcabbott: Yes. And is quite slow, which 43256 might fix.
```
julia> @btime LinearAlgebra.generic_norm2($(randn(10)));
  20.269 ns (0 allocations: 0 bytes)

julia> @btime sqrt(sum(abs2, $(randn(10))));
  5.208 ns (0 allocations: 0 bytes)

# longer
  
julia> @btime LinearAlgebra.generic_norm2($(randn(10^4)));
  15.750 μs (0 allocations: 0 bytes)

julia> @btime sqrt(sum(abs2, $(randn(10^4))));
  2.083 μs (0 allocations: 0 bytes)

julia> @btime norm($(randn(10^4)));  # BLAS
  12.416 μs (0 allocations: 0 bytes)
```

wsshin: But I understand your concern.  It turns out that `norm(::StaticArray)` is slower than `norm(::AbstractArray)` for the same array size:
```julia-repl
julia> v = rand(10); @btime norm($v);
  20.800 ns (0 allocations: 0 bytes)

julia> sv = @SVector rand(10); @btime norm($sv);
  28.727 ns (0 allocations: 0 bytes)
```
I will experiment with your suggestion.

src/linalg.jl
wsshin: `mapreduce(abs, max, x)` was my initial approach, but it didn't pass the unit tests for nested array `x`.  

`normInf()` sounds great.  I will use it and remove `maxabs_nested()`.

src/linalg.jl
wsshin: Sorry, I didn't know that the PR you Xref'ed was `julia`'s; I thought it was `StaticArrays`'s PR.  Now what you wrote makes much more sense to me.

wsshin: Just pushed a change that performs scaling only when the norm is `0` or `Inf`.  Now the benchmark is much better:
```julia-repl
julia> v = rand(10); @btime norm($v);
  21.335 ns (0 allocations: 0 bytes)

julia> sv = @SVector rand(10); @btime norm($sv);
  5.730 ns (0 allocations: 0 bytes)
```

src/linalg.jl
wsshin: Just pushed a change that uses `normInf()` instead of `maxabs_nested()`.  Please let me know if you have any other suggestions!

wsshin:
The PR is failing tests on Julia nightly (1.8).  `LinearAlgebra.generic_normInf(a)`, which is called by `LinearAlgebra.normInf(a)`, is unstable for an array of arrays `a = SA[SA[1,2], SA[3,4]]`:
```julia-repl
julia> @code_warntype LinearAlgebra.generic_normInf(a)
MethodInstance for LinearAlgebra.generic_normInf(::SVector{2, SVector{2, Int64}})
  from generic_normInf(x) in LinearAlgebra at /Users/wsshin/packages/julia/julia-master/usr/share/julia/stdlib/v1.8/LinearAlgebra/src/generic.jl:454
Arguments
  #self#::Core.Const(LinearAlgebra.generic_normInf)
  x::SVector{2, SVector{2, Int64}}
Body::Any
1 ─ %1 = LinearAlgebra.mapreduce(LinearAlgebra.norm, LinearAlgebra.max, x)::Any
│   %2 = LinearAlgebra.float(%1)::Any
└──      return %2
```
I guess this problem will be eventually fixed in the official Julia 1.8 release, but if this is a concern, maybe we should roll back to using `maxabs_nested()` instead of `normInf()`.  

@mcabbott, please advise.

mcabbott:
Re normInf, I just wondered how much code could be shared. Not sure that `LinearAlgebra.generic_normInf` is the best path. I see that SA's Inf norm is just `mapreduce(norm, max, a)`. When I timed that above, it was the same speed as the hand-written `maxabs_nested`.

However, it might be possible to do better. `norm(x, Inf)` has to allow for NaN (and maybe `missing`) entries. But in this use, after you have checked that the naiive norm is 0 or Inf, there cannot be NaN etc. I think this means `@fastmath max` is safe, which is 4x quicker.

wsshin:
> Re normInf, I just wondered how much code could be shared. Not sure that `LinearAlgebra.generic_normInf` is the best path. I see that SA's Inf norm is just `mapreduce(norm, max, a)`. When I timed that above, it was the same speed as the hand-written `maxabs_nested`.
> 
> However, it might be possible to do better. `norm(x, Inf)` has to allow for NaN (and maybe `missing`) entries. But in this use, after you have checked that the naiive norm is 0 or Inf, there cannot be NaN etc. I think this means `@fastmath max` is safe, which is 4x quicker.

As I wrote earlier, `mapreduce()` was my initial choice of method to calculate the scale factor.  However, `mapreduce(abs, max, a)` fails for `a` of a nested array type (e.g., `a = SA[SA[1,2], SA[3,4]]`) because `abs` is not defined for an element (= array) of such `a`.

To solve this problem, I think you are suggesting to use `mapreduce(norm, max, a)` (with `@fastmath` for performance).  However, this is unstable for some reason:
```julia-repl
julia> VERSION
v"1.8.0-DEV.1202"

julia> a = SA[SA[1,2], SA[3,4]]; @inferred mapreduce(norm, max, a)
ERROR: return type Float64 does not match inferred return type Any
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ REPL[11]:1
```
I tried `@code_warntype mapreduce(norm, max, a)`, but I'm having a hard time understanding the source of instability.

This is why I think defining `maxabs_nested()` is probably a good idea.  That way, we don't have to worry about the unit tests failing in Julia nightly; we have a full control of the stability issue during calculation of the scale factor.

mcabbott:
I don't see this inference failure, BTW:
```
julia> a = SA[SA[1,2], SA[3,4]]; @inferred mapreduce(norm, max, a)
5.0

julia> VERSION
v"1.8.0-DEV.1200"

(jl_WZplTZ) pkg> st StaticArrays
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_WZplTZ/Project.toml`
  [90137ffa] StaticArrays v1.3.0
```

wsshin:
> I don't see this inference failure, BTW:

I think the result is different because the `norm` you used in `mapreduce()` is different from my `norm`: you are using the current master of `StaticArrays`, whereas I am using the present PR modified with `scale = mapreduce(norm, @fastmath(max), a)`.  

I've just pushed the version I am currently using, so please verify if you get the instability.  By the way, the behavior is very strange: the instability depends on the order of executing `norm()` and `mapreduce`.  Restart Julia, and execute `norm` before `mapreduce` to get unstable results:
```julia-repl
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.0-DEV.1202 (2022-01-02)
 _/ |\__'_|_|_|\__'_|  |  Commit fa63a00672 (0 days old master)
|__/                   |

julia> using StaticArrays, LinearAlgebra, Test

julia> a = SA[SA[1,2], SA[3,4]]
2-element SVector{2, SVector{2, Int64}} with indices SOneTo(2):
 [1, 2]
 [3, 4]

julia> @inferred norm(a)  # error
ERROR: return type Float64 does not match inferred return type Any
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ REPL[3]:1

julia> @inferred mapreduce(norm, @fastmath(max), a)  # error
ERROR: return type Float64 does not match inferred return type Any
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ REPL[4]:1
```
On the other hand, restart Julia again, and execute `mapreduce` before `norm` this time to get stable results:
```julia-repl
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.0-DEV.1202 (2022-01-02)
 _/ |\__'_|_|_|\__'_|  |  Commit fa63a00672 (0 days old master)
|__/                   |

julia> using StaticArrays, LinearAlgebra, Test

julia> a = SA[SA[1,2], SA[3,4]]
2-element SVector{2, SVector{2, Int64}} with indices SOneTo(2):
 [1, 2]
 [3, 4]

julia> @inferred mapreduce(norm, @fastmath(max), a)  # no error
5.0

julia> @inferred norm(a)  # no error
5.477225575051661
```
I don't understand this behavior.

wsshin:
@mcabbott, if you don't have any other solution to this instability, I would like to re-introduce `maxabs_nested()` using `@fastmath(max)`, so please let me know.

mcabbott:
Haven't had a chance to look; generated functions are always a pain but load order dep. is pretty surprising. But the type instability involves `norm` again, perhaps it can be tracked down, by commenting out some paths?

thchr:
I just wanted to echo the point that was also made in https://github.com/JuliaArrays/StaticArrays.jl/issues/913#issuecomment-844566315 by @andyferris, namely that:

> This will introduce a calculation of the maximum value, a branch, divisions, etc. It might be a significant slowdown and we’ve previously made different choices of speed vs accuracy to Base.

I.e., personally, I think it is more important that `norm` has near-optimal performance for `StaticArray`s rather than having the same kind of under/overflow guarantees that exist for Base types.

mcabbott:
I think the timing suggested by that comment is like this:
```julia
julia> fastnorm(x) = sqrt(sum(abs2, x));

julia> safenorm(x, y=fastnorm(x)) = iszero(y) ? zero(maximum(abs, x)) : y; 
# not correct, just a branch and a slow function?

julia> xs = [@SVector(rand(10)) for _ in 1:100];

julia> @btime fastnorm.($xs);
  210.360 ns (1 allocation: 896 bytes)

julia> @btime safenorm.($xs);
  236.366 ns (1 allocation: 896 bytes)

julia> xs = [@SVector(rand(3)) for _ in 1:1000]; ys = rand(1000);

julia> @btime $ys .= fastnorm.($xs);
  396.348 ns (0 allocations: 0 bytes)

julia> @btime $ys .= safenorm.($xs);
  472.367 ns (0 allocations: 0 bytes)

julia> @btime $ys .= maximum.(abs, $xs);
  1.333 μs (0 allocations: 0 bytes)
```
So in the best case, all numbers order 1, the cost of adding a branch to check for underflow can be 20%. Actually taking the maximum (which is what LinearAlgebra currently does, always) is much slower.

wsshin:
@mcabbott, thanks for the benchmarks.  Is an extra cost of 20% for branching too much, or it is something we can tolerate?

mcabbott:
That's above my pay grade, I just thought this question ought to have actual numbers attached. Not certain these are the right ones. Would be good to check if they differ on other computers, too.

wsshin:
I think the extra 20% of computation time could be significant to some users, so I think we should probably keep the current `norm()` implementation as @thchr alluded.  

However, there are applications where more accurate norm is desirable in spite of the extra computation time (e.g., automatic differentiation of `norm(::SVector)`).  For such applications, I think we should provide a way to access the present PR's `norm` implementation.  I wonder if we could define some macro for this purpose.  I am not familiar with metaprogramming, but is it easy to define a macro named, for example, `@accu` such that `@accu(norm)(::SVector)` calls the `norm` implemented in this PR?

thchr:
20% could be a reasonable price to pay though, at least in my view: my original comment was based mainly off some of the earlier benchmarking from this thread.

mcabbott:
Note that I don't think the AD issue is underflow, it's just that the slow path designed to handle underflow happens to produce different results with ForwardDiff. That may change with https://github.com/JuliaDiff/ForwardDiff.jl/issues/547#issuecomment-938078594. I guess you could test `l < nextfloat(0.0,1)` instead of `iszero(l)` if you wanted to work around that, maybe even `l <= 0` would work (as it doesn't look like it's selecting a measure zero subspace).

wsshin:
> 20% could be a reasonable price to pay though, at least in my view: my original comment was based mainly off some of the earlier benchmarking from this thread.

If that is the case, maybe we can keep the current PR's `norm` as the default, and implement `@fastmath(norm)` to access the norm that suffers from underflow and overflow.

@mcabbott, could you invite someone who could make a decision as to whether this idea is acceptable?

mcabbott:
I'm not entirely sure who makes decisions around here!

I did cause a slowdown quite a bit larger than 20% here https://github.com/JuliaArrays/StaticArrays.jl/pull/919#issue-914041080 and fixing it does not appear to be an urgent matter.

Sadly I don't think `@fastmath` is extendable, although perhaps you could add something like that to LinearAlgebra?
```
julia> Base.FastMath.fast_op
Dict{Symbol, Symbol} with 49 entries:
  :sinh   => :sinh_fast
  :cmp    => :cmp_fast
```

wsshin:
> Sadly I don't think `@fastmath` is extendable, although perhaps you could add something like that to LinearAlgebra?

How about something like this?
```julia-repl
julia> using StaticArrays, LinearAlgebra  # StaticArrays is present PR

julia> import Base.FastMath.@fastmath

julia> macro fastmath(expr)
           if expr==:norm
               return :(norm_fast)
           end
       end
@fastmath (macro with 1 method)

julia> norm_fast(x) = √sum(abs2, x)  # unscaled norm
norm_fast (generic function with 1 method)

julia> v = @SVector [1e-300,0]
2-element SVector{2, Float64} with indices SOneTo(2):
 1.0e-300
 0.0

julia> norm(v)  # present PR's scaled norm
1.0e-300

julia> @fastmath(norm)(v)  # unscaled norm
0.0
```

This doesn't support `@fastmath norm(v)`, though.  I wonder if there is a way to support that as well.

wsshin:
I think I've figured out how to implement the `@fastmath` idea properly.  It seems that we can add entries to `Base.FastMath.fast_op` and also define `Bast.FastMath.norm_fast` as follows:
```julia-repl
julia> Base.FastMath.fast_op[:norm] = :norm_fast
:norm_fast

julia> @eval Base.FastMath begin
           using StaticArrays
           norm_fast(x::StaticArray) = √sum(abs2, x)
       end
norm_fast (generic function with 1 method)

julia> v = @SVector [1e-300,0]
2-element SVector{2, Float64} with indices SOneTo(2):
 1.0e-300
 0.0

julia> norm(v)
1.0e-300

julia> @fastmath norm(v)
0.0
```
Because this is doable, here is my proposal:
- The current master`s `norm` is renamed `norm_fast` such that it can be accessed by `@fastmath norm` for the users who don`t care about underflow and overflow.
- `norm` performs `norm_fast` first, and if the result is either `0` or `Inf`, performs scaling to avoid underflow and overflow.

Let me know what you guys think.

mcabbott:
I don't think this is safe, it's a kind of piracy. It could change what happens to other uses of `@fastmath`, depending on which packages are loaded first. And the function it calls is in its own module:
```
julia> @macroexpand @fastmath norm([1,2,3])
:(Base.FastMath.norm_fast([1, 2, 3]))
```
You might persuade Base to make such a change upstream, which this package could extend. `@fastmath maximum` would also make sense. But at present it only applies to scalar functions, so I don't know whether people will like the idea.

thchr:
I think it would be better to just have a single `norm` than to introduce an additional layer of hard-to-discover functionality (such as the `@fastmath` approach).

I think the crucial question is how much a good implementation of an under/overflow checked `norm` slows down relative to the unchecked implementation. If it's ~20%, it's probably fine to just make that the new version of `norm`.

wsshin:
I am not able to find an easy way to avoid type instability mentioned in https://github.com/JuliaArrays/StaticArrays.jl/pull/975#issuecomment-1004359576, so I am introducing `maxabs_nested()` again.  The new code passes tests on Julia nightly.

Here is the benchmark result:
```julia-repl
julia> x = @SVector rand(3)
 0.2000072904640665
 0.1841830247591485
 0.029486864461234608

julia> @btime norm($x)  # with present PR
  2.851 ns (0 allocations: 0 bytes)
0.27348816797799824

julia> @btime norm($x)  # with master
  2.328 ns (0 allocations: 0 bytes)
0.27348816797799824

julia> x = @SVector rand(10);

julia> @btime norm($x)  # with present PR
  4.874 ns (0 allocations: 0 bytes)
2.337172993625254

julia> @btime norm($x)  # with master
  4.347 ns (0 allocations: 0 bytes)
2.337172993625254
```
The present PR's `norm()` is up to 20% slower, which seems acceptable.

wsshin:
Can this be PR approved?  The only failed tests are for Julia nightly on ubuntu-latest - x86, but they don't seem caused by the changes made in the present PR.

wsshin:
Bump.

hyrodium:
I will merge this PR in a few days if there are no objections.

ufechner7:
Well, this is really bad for performance. I guess I have to copy the old code of norm into my project to avoid this performance penalty.

thchr:
Maybe you could include some benchmark comparisons here if you think it ought to be rolled back?

ufechner7:
See https://github.com/JuliaArrays/StaticArrays.jl/issues/1132

You can just check out https://github.com/ufechner7/KiteModels.jl and run Pkg.test(), then add StaticArrays#master and run the tests again...

I don't say that it should be rolled back, I would only say if you introduce such a breaking change there should be a documented way to achieve the old behavior again without having to install an old version of the package.

thchr:
This is not a breaking change in any semver sense though. Additionally, I don't think 12-15% drops for methods that should only very rarely be compute-bottlenecks - like `norm` - should be very controversial.

The PR above has sat unmerged for ~1 year without anyone raising serious objections to that level of performance regression. As you can see from the comments, consideration was given to allowing both the old and new version to live side by side, e.g., via `@fastmath`, but it was deemed technically problematic and noone seriously rooted for having both either.

ufechner7:
Well, in the past I told people, if you want high performance, use StaticArrays.jl . Now I have to tell them, if you want high performance, write your own linear algebra library... Not nice.

wsshin:
@ufechner7, could you point out which unit tests of your KiteModels regress a lot due to this merge?  That will help us to have a more productive discussion.

ufechner7:
Well, you can use a very simple benchmark:
```
using StaticArrays, LinearAlgebra, BenchmarkTools

const KVec3    = MVector{3, Float64}

norm2(vec3) = sqrt(vec3[1]^2 + vec3[2]^2 + vec3[3]^2)

@benchmark norm2(vec3) setup=(vec3 = KVec3(10*rand(3)))
# 2.668 ns ±  0.573 ns StaticArrays 1.5.2
# 3.020 ns ±  0.579 ns StaticArrays#master
```
Or you run https://github.com/ufechner7/KiteModels.jl/blob/main/test/bench3.jl or https://github.com/ufechner7/KiteModels.jl/blob/main/test/bench4.jl , but they are pretty large...

mateuszbaran:
> Well, in the past I told people, if you want high performance, use StaticArrays.jl . Now I have to tell them, if you want high performance, write your own linear algebra library... Not nice.

As far as I know, StaticArrays.jl has never been "best performance possible". More like a middle ground between execution speed, accuracy and ease of use, skewed towards execution speed. https://github.com/JuliaSIMD/LoopVectorization.jl can often produce faster code even for small arrays. Different people prefer different trade-offs in this space.

By the way, I think we could have a `fast_norm` function that follows the old implementation.

wsshin:
@ufechner7, `norm()` needs to be fast, but it also needs to be accurate.  There had been cases like #913 and #1116 where the previous `norm()` produced inaccurate results.  The present merge is an accurate implementation avoiding the problem with minimal performance penalty.  The new behavior also matches the behavior of the [upcoming change in JuliaLang](https://github.com/JuliaLang/julia/issues/1861#ref-pullrequest-165754044).  From what you said in https://github.com/JuliaArrays/StaticArrays.jl/pull/975#issuecomment-1435782404, I believe you also agree that the new behavior has its merits.

If you would like to recover the performance of the old behavior while sacrificing accuracy, I think one way is to use `√sum(abs2, v)`.  This is a relatively compact expression, and yet it clearly shows the user's intention that s/he would like to perform the operation without worrying about underflow or overflow.  Also, its performance is on-par with the previous `norm()` implementation.  With StaticArrays v1.5.12,
```julia-repl
julia> v = @SVector rand(3);

julia> @btime norm($v);
  4.299 ns (0 allocations: 0 bytes)

julia> @btime √sum(abs2, $v);
  4.111 ns (0 allocations: 0 bytes)
```
Hope this is a reasonable solution for you.  But as others pointed out, I don't think the new `norm()` will show a very noticeable performance regression unless the computational complexity of your code is dominated by `norm()`.  So please consider using the more accurate `norm()`; its accuracy could save your code one day!

PR #977, opened 2021-12-10
Remove deprecation due to lu pivot choice change in v1.7

ChrisRackauckas:
`lu` on larger static arrays now hits a deprecation in Base.


thchr:
Shouldn't this be placed behind a `VERSION` check? I.e., seems this would break on earlier Julia versions that don't have `NoPivot` or `RowMaximum`.

A similar change should probably also be made for `qr` around here: https://github.com/JuliaArrays/StaticArrays.jl/blob/6e12349d029c8a26fb0e5c191971816c49df3a5c/src/qr.jl#L85

PR #980, opened 2021-12-16
Named access with .{x,y,z} to elements of SVector and MVector

c42f:
Having access with v.x etc makes the *concrete* SVector interface much
more usable for geometry; indeed, it will "just work like you expect"
without the need to resort to a custom FieldVector type.

This makes the *concrete* SVector/MVector interface fatter, but OTOH I
don't think it can do any harm to people who don't want to use it.

If we do this, it should be made to work at least for SVector{1}
SVector{2} and SVector{3}, but I saw no reason to disallow the use of
these in higher dimensions and overcomplicate the implementation.

Inspired by discussion at

https://discourse.julialang.org/t/ray-tracing-in-a-week-end-julia-vs-simd-optimized-c/72958

I'll add tests and some docs if we decide this is a good idea!

### DIFF
--- a/docs/src/pages/quickstart.md
+++ b/docs/src/pages/quickstart.md
@@ -49,6 +49,23 @@ v1[SVector(3,2,1)] === @SVector [3, 2, 1]
 v1[:] === v1
 typeof(v1[[1,2,3]]) <: Vector # Can't determine size from the type of [1,2,3]
 
+# For geometric and computer graphics applications in dimensions 1 to 4, the
+# conventional dimension names x,y,z,w can be used to access elements of the
+# vector:
+
+u = SA[1,2,3,4]
+
+u.x === u[1]
+u.y === u[2]
+u.z === u[3]
+u.w === u[4]
+
+# The x,y,z and w properties also work to set values in those dimensions:
+m6 = MVector(1,2)
+m6.x = 10
+# The following is now true
+m6[1] === 10
+
 # Is (partially) hooked into BLAS, LAPACK, etc:
 rand(MMatrix{20,20}) * rand(MMatrix{20,20}) # large matrices can use BLAS
 eigen(m3) # eigen(), etc uses specialized algorithms up to 3×3, or else LAPACK
--- a/src/MVector.jl
+++ b/src/MVector.jl
@@ -99,3 +99,24 @@ macro MVector(ex)
         error("Use @MVector [a,b,c] or @MVector([a,b,c])")
     end
 end
+
+# Named field access for the first four elements, using the conventional field
+# names from low-dimensional geometry (x,y,z) and computer graphics (w).
+let dimension_names = QuoteNode.([:x, :y, :z, :w])
+    body = :(getfield(v, name))
+    for (i,dim_name) in enumerate(dimension_names)
+        body = :(name === $(dimension_names[i]) ? getfield(v, :data)[$i] : $body)
+        @eval @inline function Base.getproperty(v::Union{SVector{$i},MVector{$i}},
+                                                name::Symbol)
+            $body
+        end
+    end
+
+    body = :(setfield!(v, name, e))
+    for (i,dim_name) in enumerate(dimension_names)
+        body = :(name === $dim_name ? @inbounds(v[$i] = e) : $body)
+        @eval @inline function Base.setproperty!(v::MVector{$i}, name::Symbol, e)
+            $body
+        end
+    end
+end
--- a/src/SVector.jl
+++ b/src/SVector.jl
@@ -36,10 +36,6 @@ const SVector{S, T} = SArray{Tuple{S}, T, 1, S}
 ## SVector methods ##
 #####################
 
-@propagate_inbounds function getindex(v::SVector, i::Int)
-    v.data[i]
-end
-
 # Converting a CartesianIndex to an SVector
 convert(::Type{SVector}, I::CartesianIndex) = SVector(I.I)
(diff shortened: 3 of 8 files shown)


src/MVector.jl
```diff
@@ -99,3 +99,12 @@ macro MVector(ex)
         error("Use @MVector [a,b,c] or @MVector([a,b,c])")
     end
 end
+
+# Named field access for the first three elements with the conventional
+# field names
+function Base.getproperty(v::Union{MVector}, name::Symbol)
```
andyferris, line 105: Unnecessary Union

src/SVector.jl
```diff
@@ -117,3 +117,12 @@ macro SVector(ex)
         error("Use @SVector [a,b,c], @SVector Type[a,b,c] or a comprehension like [f(i) for i = i_min:i_max]")
     end
 end
+
+# Named field access for the first three elements with the conventional
+# field names
+function Base.getproperty(v::Union{SVector}, name::Symbol)
```
andyferris, line 123: Unnecessary Union

andyferris (approved):
Yes, I can see how this could be convenient for using these as points, and as you say there isn’t any real downside.

src/SVector.jl
c42f: Heh oops, thanks!

The reason this was there was that I had this defined as `Union{SVector,MVector}` because the implementations are the same.

But then I didn't know which file to put these in so I duplicated them instead as they're a concrete interface. Kind of a stupid reason to have duplicate code though I guess...

c42f:
Perhaps I should also add `setproperty!` for `MVector` to go with this.

src/MVector.jl
```diff
@@ -99,3 +99,19 @@ macro MVector(ex)
         error("Use @MVector [a,b,c] or @MVector([a,b,c])")
     end
 end
+
+# Named field access for the first three elements, using the conventional field
+# names from 3D geometry
+function Base.getproperty(v::MVector, name::Symbol)
```
mateuszbaran, line 105: Shouldn't this be `@inline` for safety? It would be quite bad for performance if `name` had to be checked at runtime.

c42f: Sure. It'd really be too bad if performance randomly fell off a cliff because of this!

Another defensive alternative is to explicitly use `getfield` in the `getindex` implementation.  That will be kinder on the compiler anyway and it's quite easy, so I've done it as well.

src/SVector.jl
```diff
@@ -36,10 +36,6 @@ const SVector{S, T} = SArray{Tuple{S}, T, 1, S}
 ## SVector methods ##
 #####################
 
-@propagate_inbounds function getindex(v::SVector, i::Int)
-    v.data[i]
-end
```
c42f, line 41: Haha, this redundant method has been in here for years. I removed it in favour of the linear getindex from `SArray`

mateuszbaran (approved):

c42f:
Absent objections, I'll probably merge this some time tomorrow or so.

It's a neat little feature and should be harmless to anyone who hasn't been involved in type piracy :-)

cdsousa:
> Perhaps this should also include `p.w = p[4]` to complete the set that's used in glsl?

It is also used by the Eigen C++ library.

c42f:
I agree `w` is an obvious option for the forth conventional dimension name. Perhaps it's the best of the available options. If you're a physicist, the most obvious fourth dimension name might be `t` for time. (No it's ok, we're not doing that :laughing:)

I initially decided against `w` because it's not quite so standard — it's more a graphics (homogeneous coordinates) convention than a universal geometry thing...

fredrikekre:
Would it make sense to limit this to length-3 vectors (`.x` for 1D, `.x`, `.y` for 2D and `.x`, `.y`, `.z` for 3D)?

claforte:
Maybe throw @inbounds in there? I think that made a significant speed-up in my toy raytracing implementation, and otherwise people might not think to add it up, i.e. might not realize that .x/.y translates into array look-up...

c42f:
> Would it make sense to limit this to length-3 vectors (.x for 1D, .x, .y for 2D and .x, .y, .z for 3D)?

I don't know! The question is, could it lead to unexpected bugs in other cases?

You know what I think I agree: let's be strict here and relax the rules later if we think it makes sense.  It's much easier to add API than to remove it!

> Maybe throw `@inbounds` in there?

Adding `@inbounds` here shouldn't help because any bounds check should be elided given constant propagation of the *constant* index into the data tuple. Though on that note, it would be better (easier on the compiler) to just implement this in terms of getfield(v,:data) rather than getindex — that avoids `getproperty` being recursive.  I'll do that!

c42f:
Side note... gosh our docs are pretty awful :sweat_smile:

Not something for this PR though.

docs/src/pages/quickstart.md
```diff
@@ -49,6 +49,23 @@ v1[SVector(3,2,1)] === @SVector [3, 2, 1]
...
+# For geometric and computer graphics applications in dimensions 1 to 4, the
+# conventional dimension names x,y,z,w can be used to access elements of the
+# vector:
+
+u = SA[1,2,3,4]
+
+u.x === u[1]
+u.y === u[2]
+u.z === u[3]
+u.w === u[3]
```
Samuel3008, line 61: Should be `u[4]` :)

c42f: Fixed

Issue #981, opened 2021-12-19
Version 1.2.6 breaks callbacks


ufechner7:
The following code works with StaticArrays 1.2.5, but fails with 1.2.6 and newer:
```
using Sundials, StaticArrays
# Tutorial example showing how to use an implicit solver 
# It simulates a falling mass.

const G_EARTH  = [0.0, 0.0, -9.81]

# Type definitions
const SimFloat = Float64
const Vec3     = MVector{3, SimFloat}

# Falling mass.
# State vector y   = mass.pos, mass.vel
# Derivative   yd  = mass.vel, mass.acc
# Residual     res = (y.vel - yd.vel), (yd.acc - G_EARTH)     
function res1(res, yd, y::MVector{S, SimFloat}, p, t) where S
    res[1:3] .= y[4:6] - yd[1:3]
    res[4:6] .= yd[4:6] - G_EARTH
end

vel_0 = Vec3(0.0, 0.0, 50.0)    # Initial velocity
pos_0 = Vec3(0.0, 0.0,  0.0)    # Initial position
acc_0 = Vec3(0.0, 0.0, -9.81)   # Initial acceleration
y0  = vcat(pos_0, vel_0)    # Initial pos, vel
yd0 = vcat(vel_0, acc_0)    # Initial vel, acc

tspan = (0.0, 10.2)         # time span

differential_vars = ones(Bool, 6)
prob = DAEProblem(res1, yd0, y0, tspan, differential_vars=differential_vars)

sol = solve(prob, IDA(), saveat=0.1)
```

Error message:
```
julia> include("src/DAE_Example.jl")
ERROR: LoadError: MethodError: no method matching res1(::Vector{Float64}, ::SizedVector{6, Float64, Vector{Float64}}, ::SizedVector{6, Float64, Vector{Float64}}, ::SciMLBase.NullParameters, ::Float64)
Closest candidates are:
  res1(::Any, ::Any, ::MVector{S, Float64}, ::Any, ::Any) where S at /home/ufechner/repos/KiteViewer/src/DAE_Example.jl:15
Stacktrace:
  [1] (::DAEFunction{true, typeof(res1), Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing})(::Vector{Float64}, ::Vararg{Any, N} where N)
[...]

PR #982, opened 2021-12-19
Make `vec` just return the argument for `MVector` and `SizedVector`.

mateuszbaran:
This should fix part of the problem reported in #981 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.3.0"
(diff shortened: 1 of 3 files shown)


ufechner7:
I don't think that the failure on macOS-latest is related to this commit... So how to proceed?

mateuszbaran:
That looks like a random failure. Now I'm going to wait for someone to review this PR.

mateuszbaran:
@c42f could you check this PR? It's a small bugfix that would be nice to have.

thchr:
This seems like a correct fix to me.

thchr (approved):

PR #983, opened 2022-01-03
note named access for SVector in FieldVector docs

stevengj:
Since #980 allows named access to `SVector` fields, this should be noted in the `FieldVector` docs so that users realize that they don't need a `FieldVector` for these particular names.


c42f:
Thanks, good idea.

Issue #986, opened 2022-01-11
reduce(f, ::SVector) allocates


tpapp:
In JuliaStats/LogExpFunctions.jl#33 the following allocates:

```julia
using StaticArrays, LogExpFunctions, BenchmarkTools
x = randn(10);
s = SVector(x...);
@btime logsumexp($x)
@btime logsumexp($s)
```
while for the vector `x` we get zero allocations.

We have narrowed it down to `reduce` for `SVector` and `foldl` is a workaround, but I have no clear idea why `reduce` does this.

mateuszbaran:
Seems like Julia fails at inlining a call with keyword arguments. There is quite a lot that can be done to force inlining of non-keyword functions but keywords still cause issues.

tpapp:
But it works for `foldl`. Why are `(map)reduce` and `(map)foldl` implemented differently?

mateuszbaran:
Hm, thanks for the question -- I thought there were no significant differences but there indeed was one. I think this should be easily fixable now that I see this.

mateuszbaran:
Could you check if this branch fixes the problem for you: https://github.com/JuliaArrays/StaticArrays.jl/tree/mbaran/reduce-inlining ?

tpapp:
Yes, this fixes the issue fully. Thanks!

PR #987, opened 2022-01-11
Force more inlining of `reduce`

mateuszbaran:
Fixes issue #986 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.3.1"
+version = "1.3.2"
 
 [deps]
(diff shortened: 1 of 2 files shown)


tpapp (approved):
Thanks, this fixes JuliaStats/LogExpFunctions.jl#33.

thchr:
This seems reasonable to me.
(Although, isn't this trick technically more oriented towards [forcing specialization](https://github.com/JuliaLang/julia/pull/33917#discussion_r349730291) than inlining? - I guess the end result will be the indistinguishable when combined with `@inline`).

thchr (approved):

mateuszbaran:
Yes, that's true, and in this particular case because these methods are marked `@inline` the end effect is that they will be inlined wherever possible.

PR #989, opened 2022-02-02
Avoid using deprecated macro `@_inline_meta`

tkf:
`@_inline_meta` is deprecated in 1.8-DEV.

```console
$ julia1.8 --depwarn=error --project -e 'using StaticArrays; zeros(SVector{3,Int})'
ERROR: LoadError: `@_inline_meta` is deprecated, use `@inline` instead.
```

This patch defines `const var"@_inline_meta" = Base.var"@inline"` for 1.8-DEV.

It also removes the import of unused macro `@_pure_meta`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.3.3"
+version = "1.3.4"
 
 [deps]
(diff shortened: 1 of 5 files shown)


mateuszbaran (approved):
Looks fine. I've reported that failure on Julia nightly: https://github.com/JuliaLang/julia/pull/43779#discussion_r797424412 .

PR #995, opened 2022-02-11
Fix eigvals for complex 3x3 matrices

mateuszbaran:
As in the title, currently `eigvals` doesn't work for complex 3x3 matrices (see test for an example). This fixes the problem.


thchr (approved):
This looks right to me; nice! (and test failures on nightly look unrelated to this PR)

PR #996, opened 2022-02-14
map: build output type based on first static argument.

fredrikekre:
A colleague noticed the following inconsistency in the output types:
```julia
julia> struct Vec3{T} <: FieldVector{3, T}
           x::T
           y::T
           z::T
       end

julia> StaticArrays.similar_type(::Type{<:Vec3}, ::Type{T}, s::Size{(3,)}) where {T} = Vec3{T}

julia> v3 = Vec3(1, 2, 3); v = [1, 2, 3];

julia> v3 + v
3-element Vec3{Int64} with indices SOneTo(3):
 2
 4
 6

julia> v + v3
3-element SVector{3, Int64} with indices SOneTo(3):
 2
 4
 6
```

With this patch `similar_type` is called on the first argument that is `<: StaticArray`, which is actually consistent with the 0-dimensional case: https://github.com/JuliaArrays/StaticArrays.jl/blob/1858207b405ea168ecd6810ffa7bbb07e984a2b7/src/mapreduce.jl#L62

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -73,7 +73,7 @@ end
         @_inline_meta
         S = same_size(a...)
         @inbounds elements = tuple($(exprs...))
-        @inbounds return similar_type(typeof(_first(a...)), eltype(elements), S)(elements)
+        @inbounds return similar_type(typeof(a[$first_staticarray]), eltype(elements), S)(elements)
     end
 end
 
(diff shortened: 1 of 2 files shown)


mateuszbaran (approved):

mateuszbaran:
StaticArrays.jl tests on x86 platforms too often run out of memory.

fredrikekre:
Okay, maybe it is time to split up the tests on multiple processes?

mateuszbaran:
Yes, I think that would be the best solution.

PR #985, opened 2022-01-06
Drop support for versions of Julia lower than v1.6

thchr:
We discussed this briefly on Slack: now that the new Julia LTS is v1.6, it would make sense to drop support for earlier versions. There's at least two good reasons for doing so: 

1. CI takes forever to run when it needs to go through so many Julia versions.
2. Development becomes tedious and somewhat fragile because one has to guard what may or may not work on different Julia versions. As a result, there's a lot of `VERSION` checks in StaticArrays currently (in source files and in tests).

I'm putting this up mainly for discussion - but the relevant bits to actually do this should also be here and are split across independent commits. 

I also bumped the minor version of StaticArrays to v1.4: quoting @c42f from Slack:
> [...] we can release patches versions 1.3.x if really necessary for older Julia versions.

One obvious counter-point is that this could cause a lot of downstream "noise" for packages that try to support pre-LTS versions of Julia. A similar fear could be that this could cause some fracturing of the compat versions of StaticArrays in the ecosystem, which could lead to compat conflicts.
Conversely, since StaticArrays is so widely used, this could also be seen as good thing, in that it would help speed up a broader transition to move on to v1.6+.

PR #998, opened 2022-02-21
Improve latency of matrix-`exp`

thchr:
As discussed in https://github.com/JuliaArrays/StaticArrays.jl/pull/985#issuecomment-1044720641, `exp` is very slow to compile for larger `StaticArray`s. This tries to reduce the latency a bit by just pulling out code into smaller functions, aiming to reduce the amount of inlined code.

There's little difference in the latency at small matrix sizes - and a slight reduction in performance for 3×3 matrices. For all other sizes, this PR actually produces a clear a performance improvement however. The latency improvements at larger sizes are also pronounced. Still, 3×3 matrices are definitely more important than bigger matrices, so it's not a no-brainer I guess - but I wanted to put the PR up for discussion anyway.

The summary of timings before/after this PR are (the performance timings depend on which branch is taken; also, this is on v1.7.2):

| Size | `@time` (1st) | `@btime` (branch 1) | `@btime` (branch 2) | Latency win | Performance win |
|---:|:---:|:---:|:---:|:---:|:---:|
| 3×3 | 1.2 s/1.3 s | 62 ns/77 ns | 68 ns/73 ns | ÷ | ÷ |
| 4×4 | 2.9 s/3.2 s | 208 ns/201 ns | 194 ns/161 ns | ÷ | ✓ |
| 10×10 | 46 s/21 s | 1.77 μs/1.62 μs | 8.43 μs/5.80 μs | ✓ | ✓ |
| 15×15 | 47 min/4.5 min | ?/11.9 μs |  ?/37-50 μs | ✓ | ? |

### DIFF
--- a/src/expm.jl
+++ b/src/expm.jl
@@ -77,53 +77,54 @@ function _exp(::Size, _A::StaticMatrix{<:Any,<:Any,T}) where T
     A = S.(_A)
     # omitted: matrix balancing, i.e., LAPACK.gebal!
     nA = maximum(sum(abs.(A); dims=Val(1)))    # marginally more performant than norm(A, 1)
-    ## For sufficiently small nA, use lower order Padé-Approximations
-    if (nA <= 2.1)
-        A2 = A*A
-        if nA > 0.95
-            U = @evalpoly(A2, S(8821612800)*I, S(302702400)*I, S(2162160)*I, S(3960)*I, S(1)*I)
-            U = A*U
-            V = @evalpoly(A2, S(17643225600)*I, S(2075673600)*I, S(30270240)*I, S(110880)*I, S(90)*I)
-        elseif nA > 0.25
-            U = @evalpoly(A2, S(8648640)*I, S(277200)*I, S(1512)*I, S(1)*I)
-            U = A*U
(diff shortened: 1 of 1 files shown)


mateuszbaran:
I wonder how much compilation time reduction is just due to not duplicating `U = A*U` in each `if` branch? Maybe we don't have to decrease performance for the 3x3 case. Also, we could use the old variant for size 3x3 and the new one for larger matrices.

thchr:
I tested various permutations of this quite extensively yesterday (pretty sure I also tested the `A*U` shuffle).
Puzzlingly, I fount that almost all the compilation time comes from the presence of the two loops that rescale `A` in the second branch: without those, compilation time is drastically reduced. I don't understand why the loops make such a difference.

Alternatively, we could also use the new in-function `@inline` in 1.8 to force inlining in the 3x3 case (to avoid code duplication). Just having the code duplicated would indeed also be a solution.

c42f:
> almost all the compilation time comes from the presence of the two loops that rescale `A` in the second branch: without those, compilation time is drastically reduced

Interesting, though confusing! Did you try moving just that part out into a separate function?

PR #1000, opened 2022-02-21
Benchmark on latest Julia and scale back large-matrix benchmarking

thchr:
Following up on https://github.com/JuliaArrays/StaticArrays.jl/pull/985#issuecomment-1046936674 (cc. @KristofferC), this changes the benchmarks to run on the latest released version of Julia.

Additionally, it reduces the amount of benchmarking done for large matrices: those benchmarks currently take up the majority of benchmarking time due to their large compilation time, which doesn't seem worthwhile given that using StaticArrays for such large matrices is effectively counter to the package's focus on small sizes.


c42f (approved):
LGTM

PR #1001, opened 2022-02-22
Make `broadcast` extendable to user array wrapper.

N5N3:
This PR tries to make the current broadcast kernal reusable to more wrapped StaticArray. 
For this goal, we need to get `first_staticarray` outside the `@generated function` to make sure the type match extendable. (Not sure the reason)
Our optimizer should be clever enough to make it zero cost at runtime.

This was discussed in https://github.com/JuliaArrays/StructArrays.jl/pull/215, where we want to make 
```julia
s = StructArray{ComplexF64}((SVector(1., 2., 3.), SVector(0., 0., 0.)))
log.(s)
```
return a `SVector`.

A simple test has been added.

### DIFF
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -302,13 +302,14 @@ end
 @testset "Broadcast with unknown wrapper" begin
     data = (1, 2)
     for T in (SVector{2}, MVector{2})
+        destT = T <: SArray ? SArray : MArray
         a = T(data)
         for b in (WrapArray(a), WrapArray(a'))
-            @test @inferred(b .+ a) isa T
-            @test @inferred(b .+ b) isa T
-            @test @inferred(b .+ (1, 2)) isa T
-            @test @inferred(b .+ a') isa StaticMatrix
-            @test @inferred(a' .+ b) isa StaticMatrix
+            @test @inferred(b .+ a) isa destT
+            @test @inferred(b .+ b) isa destT
+            @test @inferred(b .+ (1, 2)) isa destT
+            @test @inferred(b .+ a') isa destT
+            @test @inferred(a' .+ b) isa destT
             # @test @inferred(b' .+ a') isa StaticMatrix # Adjoint doesn't propagate style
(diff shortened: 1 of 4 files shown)


mateuszbaran (approved):
This looks fine, there should be as little code in generated functions as possible. Constant propagation will have a bit more work to do but it should manage I think.

test/broadcast.jl
```diff
@@ -282,3 +282,30 @@ end
...
+Base.@propagate_inbounds Base.setindex!(A::WrapArray, v::Any, i::Integer...) = setindex!(A.data, v, i...)
+Base.size(A::WrapArray) = size(A.data)
+Broadcast.BroadcastStyle(::Type{WrapArray{T,N,P}}) where {T,N,P} = Broadcast.BroadcastStyle(P)
+StaticArrays.isstatic(A::WrapArray) = StaticArrays.isstatic(A.data)
+StaticArrays.Size(::Type{WrapArray{T,N,P}}) where {T,N,P} = StaticArrays.Size(P)
+function StaticArrays.similar_type(::Type{WrapArray{T,N,P}}, ::Type{t}, s::Size{S}) where {T,N,P,t,S}
+    return StaticArrays.similar_type(P, t, s)
+end
+
+@testset "Broadcast with unknown wrapper" begin
```
mateuszbaran, line 301: Could you add a test where the first array is a transposed static array? I think something like `@test @inferred(transpose(b) .+ transpose(a)) isa T` would be fine.

You could also bump patch version, then I'll merge this and tag a new release.

N5N3: `transpose(b) .+ transpose(a)` wont be a `StaticArray`, as `b` is not a `StaticArray`.
If we want to propagate `BroadcastStyle` more eagerly, we'd better change this behaviour in `LinearAlgebra`.

mateuszbaran: Ah, right, we can't transpose `b` here. Let's leave this as is then.

PR #1002, opened 2022-02-22
Fix `lu`/`qr` pivot deprecations and allow `PivotingStrategy` input

thchr:
This tackles the same problem as #977 (and builds on it), but applies it to `qr` as well, and should also work on Julia v1.6, where the new `LinearAlgebra.PivotingStrategy` is not defined.

At the same time, I added accessors so that `qr(::StaticArray, pivot)` and `lu(...)` can be called with `pivot <: PivotingStrategy` instances.

Closes #977.


mateuszbaran (approved):
This looks reasonable :+1:

PR #1008, opened 2022-03-01
Drop `combine_size`

N5N3:
This PR tries to perform `combine_size` during `Broadcast.instantiate`:
1. For `broadcast`, `combine_axes` will return a `Tuple{Vararg{SOneTo}}`, thus the output size has been static.
2. For `broadcast!`, the `dest`'s `axes` is marked as static (if possible). If there's no size-1 dim in `src`. The current generated kernal will be called.

With this change, we can reduce the compile time for first broadcast by about 1/4:
```julia
julia> a = @SArray zeros(3,3)
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> @time a .+ a;
  0.234903 seconds (452.90 k allocations: 24.654 MiB, 99.75% compilation time) #  on master: 0.308612 seconds (513.89 k allocations: 27.912 MiB, 99.87% compilation time)
```

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -60,8 +60,11 @@ static_check_broadcast_shape(::Tuple{}, ::Tuple{}) = ()
 @inline function Base.copy(B::Broadcasted{StaticArrayStyle{M}}) where M
     flat = Broadcast.flatten(B); as = flat.args; f = flat.f
     argsizes = broadcast_sizes(as...)
-    destsize = combine_sizes(argsizes)
-    _broadcast(f, destsize, argsizes, as...)
+    ax = axes(B)
+    if ax isa Tuple{Vararg{SOneTo}}
+        return _broadcast(f, Size(map(length, ax)), argsizes, as...)
+    end
+    return copy(convert(Broadcasted{DefaultArrayStyle{M}}, B))
 end
 # copyto! overloads
 @inline Base.copyto!(dest, B::Broadcasted{<:StaticArrayStyle}) = _copyto!(dest, B)
@@ -69,12 +72,13 @@ end
 @inline function _copyto!(dest, B::Broadcasted{StaticArrayStyle{M}}) where M
     flat = Broadcast.flatten(B); as = flat.args; f = flat.f
     argsizes = broadcast_sizes(as...)
-    destsize = combine_sizes((Size(dest), argsizes...))
-    if Length(destsize) === Length{Dynamic()}()
-        # destination dimension cannot be determined statically; fall back to generic broadcast!
-        return copyto!(dest, convert(Broadcasted{DefaultArrayStyle{M}}, B))
+    ax = axes(B)
+    if ax isa Tuple{Vararg{SOneTo}}
+        @boundscheck axes(dest) == ax || Broadcast.throwdm(axes(dest), ax)
+        return _broadcast!(f, Size(map(length, ax)), dest, argsizes, as...)
     end
-    _broadcast!(f, destsize, dest, argsizes, as...)
+    # destination dimension cannot be determined statically; fall back to generic broadcast!
+    return copyto!(dest, convert(Broadcasted{DefaultArrayStyle{M}}, B))
 end
 
 # Resolving priority between dynamic and static axes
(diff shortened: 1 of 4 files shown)


thchr (commented):
Someone knowledgeable about the broadcasting machinery should review this probably - but I was reading this and figured I might as well add the few comments I had.

src/broadcast.jl
```diff
@@ -4,99 +4,112 @@
...
 using Base.Broadcast: _bcsm
 # Add a new BroadcastStyle for StaticArrays, derived from AbstractArrayStyle
 # A constructor that changes the style parameter N (array dimension) is also required
 struct StaticArrayStyle{N} <: AbstractArrayStyle{N} end
 StaticArrayStyle{M}(::Val{N}) where {M,N} = StaticArrayStyle{N}()
 BroadcastStyle(::Type{<:StaticArray{<:Tuple, <:Any, N}}) where {N} = StaticArrayStyle{N}()
-BroadcastStyle(::Type{<:Transpose{<:Any, <:StaticArray{<:Tuple, <:Any, N}}}) where {N} = StaticArrayStyle{N}()
-BroadcastStyle(::Type{<:Adjoint{<:Any, <:StaticArray{<:Tuple, <:Any, N}}}) where {N} = StaticArrayStyle{N}()
+BroadcastStyle(::Type{<:Transpose{<:Any, <:StaticArray}}) = StaticArrayStyle{2}()
+BroadcastStyle(::Type{<:Adjoint{<:Any, <:StaticArray}}) = StaticArrayStyle{2}()
```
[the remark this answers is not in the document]
N5N3: Yes.

src/broadcast.jl
```diff
@@ -4,46 +4,93 @@
 
 import Base.Broadcast:
 BroadcastStyle, AbstractArrayStyle, Broadcasted, DefaultArrayStyle, materialize!
+import Base.Broadcast: combine_axes, instantiate, _broadcast_getindex, broadcast_shape, Style
```
mateuszbaran, line 7: I'd suggest using `using` instead of `import` if no new methods are added to a function.

src/broadcast.jl
```diff
@@ -4,46 +4,93 @@
...
 @inline function Base.copy(B::Broadcasted{StaticArrayStyle{M}}) where M
     flat = Broadcast.flatten(B); as = flat.args; f = flat.f
     argsizes = broadcast_sizes(as...)
-    destsize = combine_sizes(argsizes)
-    _broadcast(f, destsize, argsizes, as...)
+    ax = axes(B)
+    if ax isa Tuple{Vararg{SOneTo}}
+        return _broadcast(f, Size(map(length, ax)), argsizes, as...)
+    end
+    return copy(convert(Broadcasted{DefaultArrayStyle{M}}, B))
```
mateuszbaran, line 67: Do we have a test for this case?

mateuszbaran (commented):
I've done some work with broadcasting before but I'm not quite familiar with certain parts of broadcasting touched in this PR.

src/broadcast.jl
```diff
@@ -4,46 +4,93 @@
...
     DefaultArrayStyle(Val(max(M, N)))
 BroadcastStyle(::StaticArrayStyle{M}, ::DefaultArrayStyle{0}) where {M} =
     StaticArrayStyle{M}()
+
+# combine_axes overload (for Tuple)
+@inline static_combine_axes(A, B...) = broadcast_shape(static_axes(A), static_combine_axes(B...))
+static_combine_axes(A) = static_axes(A)
+static_axes(A) = axes(A)
+static_axes(x::Tuple) = (SOneTo{length(x)}(),)
+static_axes(bc::Broadcasted{Style{Tuple}}) = static_combine_axes(bc.args...)
```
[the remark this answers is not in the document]
N5N3: Mainly for `Tuple`s, as their `axes` is a `OneTo`, while we want a `SOneTo` to support, e.g. `(1, 2) .+ SMatrix{1}(1, 2)` (the first `axes` is determined by `(1, 2)`)

src/broadcast.jl
N5N3: Unfortunately, no.
It is a little complex to test this branch as it should never be hitted during `broadcast`. (`DefaultArrayStyle` beats `StaticArrayStyle` for most cases)
We can define some buggy `StaticArray` with none-static `axes` to test this branch, but i', not sure whether we need it.

Maybe this branch should be `error("please file a bug")`?

src/broadcast.jl
N5N3: Cleaned.

src/broadcast.jl
mateuszbaran: OK, an error here is fine.

mateuszbaran (approved):
I think this looks fine :+1: .

mateuszbaran:
Could you bump patch version? I'd tag a new release.

N5N3:
> Could you bump patch version? I'd tag a new release.

Patch version was bumped in #1001. (master is 1.4.2 at present). I guess there's no need to bump again?

mateuszbaran:
Yes, right, I didn't notice.

judober:
I think this affects HybridArrays. See:
https://github.com/JuliaArrays/HybridArrays.jl/issues/50

PR #1011, opened 2022-03-06
Update docstring of `Size`

### LINKED ISSUE #923: zeros(Size) does not work
The docstring of `Size` suggests to use `zeros(Size(3,3))` to construct a StaticArray rather than a normal array.

Running this with StaticArrays v1.2.2 gives
```julia
julia> zeros(Size(3,3))
ERROR: MethodError: no method matching zeros(::Size{(3, 3)})
Closest candidates are:
  zeros(::Union{Integer, AbstractUnitRange}...) at array.jl:498
  zeros(::Tuple{Vararg{Union{Integer, AbstractUnitRange}, N} where N}) at array.jl:500
  zeros(::Type{MVector{N, T} where T}) where N at /home/schrist/.julia/packages/StaticArrays/uH2MB/src/MVector.jl:25
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[16]:1
```

hyrodium:
This PR fixes #923.

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -25,8 +25,9 @@ function dimmatch end
 `Size` is used extensively throughout the `StaticArrays` API to describe _compile-time_
 knowledge of the size of an array. The dimensions are stored as a type parameter and are
 statically propagated by the compiler, resulting in efficient, type-inferrable code. For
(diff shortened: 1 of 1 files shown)


mateuszbaran (approved):

hyrodium:
The CI failure with nightly is not related to this PR.

PR #1012, opened 2022-03-10
remove stale `broadcasted_indices`

thchr:
This is a minor nit found while playing around with JET.jl's `report_package` (the other highlights from that appeared to be false positives as far as I could see, so that's nice).

Justification is in the commit message.


thchr:
CI is having a hard time with some internal out-of-memory errors. I shouldn't be related to this though (but somewhat annoying; same sporadic failures for CI restarted 3 times).

mateuszbaran (approved):

PR #1013, opened 2022-03-12
Fix qr test failure on nightly

thchr:
Failure example is [here](https://github.com/JuliaArrays/StaticArrays.jl/runs/5522395851?check_suite_focus=true) and occurs because of a change to the element type of heterogeneous vectors of arrays (cf. https://github.com/JuliaLang/julia/issues/44586).

The change here just makes the tests loop over a tuple rather than an array (to make sure the element type is unchanged).

### DIFF
--- a/test/qr.jl
+++ b/test/qr.jl
@@ -48,7 +48,7 @@ Random.seed!(42)
         test_qr(arr)
     end
     # some special cases
-    for arr in [
+    for arr in (
                    (@MMatrix randn(3,2)),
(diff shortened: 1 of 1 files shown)


mateuszbaran:
Is that a bug or an intentional change?

mateuszbaran:
OK, that's intentional: https://github.com/JuliaLang/julia/pull/44096/files#r817938189 .

mateuszbaran (approved):

PR #1016, opened 2022-03-22
Code clean and reuse for `constructor`.

### LINKED ISSUE #809: `Scalar(1, 2) === Scalar((1, 2))`?
Currently,

```julia
julia> Scalar(1, 2) === Scalar((1, 2))
true
```

But I guess disallowing `Scalar(1, 2)` is sensible?  It's a subset of #518 but I think just throwing a method error or augment error here is an easy fix.

ref #774

### LINKED ISSUE #911: MMatrix generation with @MMatrix requires equal type for elements
I ran into this:
```julia
@MMatrix [1 1; 1 1.0]
```
errors with `Size mismatch in Static Array parameters. Got size Tuple{2, 2}, dimension 2 and length 1.`
The same works for SMatrix:
```julia
@SMatrix [1 1; 1 1.0]
```
turning all elements into float64. The workaround is this:
```julia
@MMatrix [1.0 1.0; 1.0 1.0]
```
I would expect MMatrix to behave like SMatrix.
Also, for MVector it works:
```julia 
@MVector [1, 1.0]
```

N5N3:
Seperated from #1009.

This PR aims to unify the implement of `StaticArray`'s constructor with a new method `construct_type`.
It returns a `concrete` constructor based on `src <: Union{Tupe, StaticArray, AbstractArray}` and the target type `SA` (if possible).
e.g.
```julia
julia> StaticArrays.construct_type(SMatrix{2},(1,2,3,4))
SMatrix{2, 2, Int64} (alias for SArray{Tuple{2, 2}, Int64, 2})

julia> StaticArrays.construct_type(SMatrix{2},(1.,2,3,4))
SMatrix{2, 2, Float64} (alias for SArray{Tuple{2, 2}, Float64, 2})

julia> StaticArrays.construct_type(SMatrix,(1.,2,3,4)) # the output's size is not static.
ERROR: DimensionMismatch: No precise constructor for SMatrix found. Length of input was 4.
```
Thus, we only need to mannually define the constructor with precise `size`, `eltype` and `ndims`.
Unneeded dispatch has been removed.
Edit1: Since 3rd party `StaticArray` might prefer previous design, the default `construct_type` returns the input `SA` directly to keep compatibility.
Edit2:  `construct_type` for `FieldArray` has been implemented. Now `FieldArray`'s default constructor support eltype promotion.
```julia
julia> struct Point2D{T} <: FieldVector{2,T}
           x::T
           y::T
       end

julia> Point2D(1,1.0)
2-element Point2D{Float64} with indices SOneTo(2):
 1.0
 1.0
```

This PR also restrict the input wrapping within constructor.
If multiple arguments are provided, all of them should be treated as the result's elements, and re-wrapping is not allowed anymore.
For example
```julia
julia> Scalar(1, 2)
ERROR: DimensionMismatch: No precise constructor for Scalar found. Length of input was 2.
julia> Scalar((1, 2))
Scalar{Tuple{Int64, Int64}}(((1, 2),))
julia> SVector{1}(1, 2)
ERROR: DimensionMismatch: No precise constructor for SVector{1} found. Length of input was 2.
julia> SVector{1}((1, 2))
1-element SVector{1, Tuple{Int64, Int64}} with indices SOneTo(1):
 (1, 2)
```
Close #809. 
Close #911. (fix constructor missing)

### DIFF
--- a/src/SizedArray.jl
+++ b/src/SizedArray.jl
@@ -16,7 +16,7 @@ wrap a 2x3 array `a` in a `SizedArray`, use `SizedMatrix{2,3}(a)`.
 struct SizedArray{S<:Tuple,T,N,M,TData<:AbstractArray{T,M}} <: StaticArray{S,T,N}
     data::TData
 
-    function SizedArray{S,T,N,M,TData}(a::TData) where {S,T,N,M,TData<:AbstractArray{T,M}}
+    function SizedArray{S,T,N,M,TData}(a::TData) where {S<:Tuple,T,N,M,TData<:AbstractArray{T,M}}
         require_one_based_indexing(a)
         if size(a) != size_to_tuple(S) && size(a) != (tuple_prod(S),)
             throw(DimensionMismatch("Dimensions $(size(a)) don't match static size $S"))
@@ -24,40 +24,49 @@ struct SizedArray{S<:Tuple,T,N,M,TData<:AbstractArray{T,M}} <: StaticArray{S,T,N
         return new{S,T,N,M,TData}(a)
     end
 
-    function SizedArray{S,T,N,1,TData}(::UndefInitializer) where {S,T,N,TData<:AbstractArray{T,1}}
+    function SizedArray{S,T,N,1,TData}(::UndefInitializer) where {S<:Tuple,T,N,TData<:AbstractArray{T,1}}
         return new{S,T,N,1,TData}(TData(undef, tuple_prod(S)))
     end
-    function SizedArray{S,T,N,N,TData}(::UndefInitializer) where {S,T,N,TData<:AbstractArray{T,N}}
+    function SizedArray{S,T,N,N,TData}(::UndefInitializer) where {S<:Tuple,T,N,TData<:AbstractArray{T,N}}
         return new{S,T,N,N,TData}(TData(undef, size_to_tuple(S)...))
     end
 end
 
-# Julia v1.0 has some weird bug that prevents this from working
-@inline SizedArray(a::StaticArray{S,T,N}) where {S<:Tuple,T,N} = SizedArray{S,T,N}(a)
-@inline function SizedArray{S,T,N}(
-    a::TData,
-) where {S,T,N,M,TData<:AbstractArray{T,M}}
-    return SizedArray{S,T,N,M,TData}(a)
+@inline function SizedArray{S,T,N,M}(a::AbstractArray) where {S<:Tuple,T,N,M}
+    if eltype(a) == T && (M == 1 || M == ndims(a))
+        a′ = M == 1 ? vec(a) : a
+        return SizedArray{S,T,N,M,typeof(a′)}(a′)
+    end
+    return convert(SizedArray{S,T,N,M}, a)
 end
-@inline function SizedArray{S,T}(a::TData) where {S,T,M,TData<:AbstractArray{T,M}}
-    return SizedArray{S,T,tuple_length(S),M,TData}(a)
+
+@inline function SizedArray{S,T,N}(a::AbstractArray) where {S<:Tuple,T,N}
+    M = ndims(a) == N ? N : 1
+    return SizedArray{S,T,N,M}(a)
 end
-@inline function SizedArray{S}(a::TData) where {S,T,M,TData<:AbstractArray{T,M}}
-    return SizedArray{S,T,tuple_length(S),M,TData}(a)
-end
-function SizedArray{S,T,N,N}(::UndefInitializer) where {S,T,N}
+
+@inline (::Type{SZA})(a::AbstractArray) where {SZA<:SizedArray} = construct_type(SZA, a)(a)
+
+# disambiguation
+@inline SizedArray{S,T,N,M}(a::StaticArray) where {S<:Tuple,T,N,M} = construct_type(SizedArray{S,T,N,M}, a)(a.data)
+@inline SizedArray{S,T,N}(a::StaticArray) where {S<:Tuple,T,N} = construct_type(SizedArray{S,T,N}, a)(a.data)
+@inline (::Type{SZA})(a::StaticArray) where {SZA<:SizedArray} = construct_type(SZA, a)(a.data)
+# TODO: Should we respect `TData`?
+SizedArray{S,T,N,M,TData}(a::TData) where {S<:Tuple,T,N,M,TData<:StaticArray{<:Tuple,T,M}} = SizedArray{S,T,N,M}(a.data)
+
+function SizedArray{S,T,N,N}(::UndefInitializer) where {S<:Tuple,T,N}
     return SizedArray{S,T,N,N,Array{T,N}}(undef)
 end
-function SizedArray{S,T,N,1}(::UndefInitializer) where {S,T,N}
+function SizedArray{S,T,N,1}(::UndefInitializer) where {S<:Tuple,T,N}
     return SizedArray{S,T,N,1,Vector{T}}(undef)
 end
-@inline function SizedArray{S,T,N}(::UndefInitializer) where {S,T,N}
+@inline function SizedArray{S,T,N}(::UndefInitializer) where {S<:Tuple,T,N}
     return SizedArray{S,T,N,N}(undef)
 end
-@inline function SizedArray{S,T}(::UndefInitializer) where {S,T}
+@inline function SizedArray{S,T}(::UndefInitializer) where {S<:Tuple,T}
     return SizedArray{S,T,tuple_length(S)}(undef)
 end
-@generated function (::Type{SizedArray{S,T,N,M,TData}})(x::NTuple{L,Any}) where {S,T,N,M,TData<:AbstractArray{T,M},L}
+@generated function (::Type{SizedArray{S,T,N,M,TData}})(x::NTuple{L,Any}) where {S<:Tuple,T,N,M,TData<:AbstractArray{T,M},L}
     if L != tuple_prod(S)
         error("Dimension mismatch")
     end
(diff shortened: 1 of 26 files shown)


test/ambiguities.jl
```diff
@@ -1,7 +1,7 @@
 # Allow no new ambiguities (see #18), unless you fix some old ones first!
 
-const allowable_ambiguities = VERSION ≥ v"1.7" ? 60 :
-                              VERSION ≥ v"1.6" ? 61 : error("version must be ≥1.6")
+const allowable_ambiguities = VERSION ≥ v"1.7" ? 10 :
```
sjkelly, line 3: :fireworks:

N5N3:
At present, this PR breaks `Rotations.jl`'s test.
Then I realize that we can't assume that all `StaticArray`'s constructer don't change input's eltype by default
```julia
julia> R = RotMatrix3 |> one
3×3 RotMatrix3{Bool} with indices SOneTo(3)×SOneTo(3):
 1  0  0
 0  1  0
 0  0  1

julia> AngleAxis(R) # The best guessed constructor should be `AngleAxis` rather than `AngleAxis{Bool}`
3×3 AngleAxis{Float64} with indices SOneTo(3)×SOneTo(3)(0.0, 1.0, 0.0, 0.0):
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0
```
We can limit this constructor inference to 1st class `StaticArray` to keep compatibility.
But I believe the API could be improved for future coda reuse.

src/convert.jl
```diff
@@ -29,21 +165,20 @@ end
...
 length_val(a::Type{T}) where {T<:StaticArrayLike} = length_val(Size(T))
 
+unroll_tuple(a::AbstractArray, ::Length{0}) = ()
+unroll_tuple(a::AbstractArray, ::Length{1}) = @inbounds (a[],)
 @generated function unroll_tuple(a::AbstractArray, ::Length{L}) where {L}
-    exprs = [:(a[$j]) for j = 1:L]
+    exprs = (:(a[$j+Δj]) for j = 0:L-1)
     quote
         @_inline_meta
+        Δj = firstindex(a)
```
hyrodium, line 181: I think this change is a fix for `OffsetArrays.jl`. Could you add tests with the package with the `[extras]` table in `Project.toml`?

src/convert.jl
```diff
@@ -1,14 +1,150 @@
...
+@pure has_size1(::Type{<:StaticMatrix}) = false
+_size1(::Type{<:StaticMatrix{M}}) where {M} = M
+StaticSquareMatrix{N,T} = StaticMatrix{N,N,T}
+@generated function _sqrt(::Length{L}) where {L}
+    N = round(Int, sqrt(L))
+    N^2 == L || throw(DimensionMismatch("Input's length must be perfect square"))
+    return :($N)
+end
+
+const FirstClass = Union{SArray, MArray, SHermitianCompact, SizedArray}
```
hyrodium, line 28: What's the meaning of `FirstClass`? Should this include `FieldArray`? I thought a few more comments would make it easier to review:+1:

src/convert.jl
```diff
@@ -1,14 +1,150 @@
...
+#    If `SA <: StaticMatrix{M}`, the output `Size` is derived based on `length(x)÷M`.
+#    If `SA <: StaticMatrix{M,M} where M`, the output `Size` is derived based on `sqrt(length(x))`.
+#    If `length(SA) == 1 && length（x） > 1`, then we tries to fill `SA` with `x` itself. (rewrapping)
+# 2. Args (`SA(x...)`)
+#    Similar to `Tuple`, but rewrapping is not allowed.
+# 3. StaticArray
+#    Treat `x` as `Tuple` whenever possible. If failed, then try to inherit `x`'s `Size`.
+# 4. AbstractArray
+#    `x` is used to provide eltype. Thus `SA` must be static sized.
+function construct_type(::Type{SA}, x) where {SA<:FirstClass}
```
hyrodium, line 55: I think this can be a docstring, instead of a comment. That change will be helpful in REPL. If we would like to keep this `construct_type` function internal, we can add warning in the docstring.
cf. https://github.com/SciML/ColPrac#changes-that-are-not-considered-breaking

test/SizedArray.jl
```diff
@@ -30,7 +31,7 @@
         @test @inferred(SizedArray{Tuple{2,2}}([1 2;3 4]))::SizedArray{Tuple{2,2},Int,2,2} == [1 2; 3 4]
         # From Array, reshaped
         @test @inferred(SizedArray{Tuple{2,2}}([1,2,3,4]))::SizedArray{Tuple{2,2},Int,2,1} == [1 3; 2 4]
-        @test_throws DimensionMismatch SizedArray{Tuple{4}}([1 2; 3 4])
+        @test_throws DimensionMismatch SizedArray{Tuple{1,4}}([1 2; 3 4])
```
hyrodium, line 34: We can test the original `SizedArray{Tuple{4}}([1 2; 3 4])` now doesn't throw errors.
```suggestion
        @test SizedArray{Tuple{4}}([1 2; 3 4]) == vec([1 2; 3 4])
        @test_throws DimensionMismatch SizedArray{Tuple{1,4}}([1 2; 3 4])
```

test/SizedArray.jl
```diff
@@ -252,3 +253,21 @@ struct OVector <: AbstractVector{Int} end
...
+    @test_throws Exception SizedMatrix{1,1}(1, 2)
+    @test (@inferred(SizedMatrix{1,1}((1, 2)))::SizedMatrix{1,1,NTuple{2,Int}}) == fill((1, 2),1,1)
+    @test (@inferred(SizedMatrix{1,2}((1, 2)))::SizedMatrix{1,2,Int}) == reshape(1:2, 1, 2)
+    @test (@inferred(SizedMatrix{1}((1, 2)))::SizedMatrix{1,2,Int}) == reshape(1:2, 1, 2)
+    @test (@inferred(SizedMatrix{1}(1, 2))::SizedMatrix{1,2,Int}) == reshape(1:2, 1, 2)
+    @test (@inferred(SizedMatrix{2}((1, 2)))::SizedMatrix{2,1,Int}) == reshape(1:2, 2, 1)
+    @test (@inferred(SizedMatrix{2}(1, 2))::SizedMatrix{2,1,Int}) == reshape(1:2, 2, 1)
+
+
+end
```
hyrodium, line 273:
```suggestion
    @test (@inferred(SizedMatrix{2}(1, 2))::SizedMatrix{2,1,Int}) == reshape(1:2, 2, 1)
end
```

hyrodium (commented):
Most of the changes seem great! I have added some comments.

src/convert.jl
```diff
@@ -1,14 +1,150 @@
...
+    Base.ndims(::Type{<:StaticArray{<:Tuple,<:Any,N}}) where {N} = N
+end
+@pure has_eltype(::Type{<:StaticArray{<:Tuple,T}}) where {T} = @isdefined T
+@pure has_eltype(::Type{<:StaticArray}) = false
+@pure has_size(::Type{<:StaticArray{S}}) where {S<:Tuple} = @isdefined S
+@pure has_size(::Type{<:StaticArray}) = false
+@pure has_size1(::Type{<:StaticMatrix{M}}) where {M} = @isdefined M
+@pure has_size1(::Type{<:StaticMatrix}) = false
+_size1(::Type{<:StaticMatrix{M}}) where {M} = M
+StaticSquareMatrix{N,T} = StaticMatrix{N,N,T}
```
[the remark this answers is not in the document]
N5N3: `StaticMatrix{N,N}` is shorter than `StaticSquareMatrix{N}`, and it's clear enough.
I'll just remove it as it's used only once.

src/SizedArray.jl
```diff
@@ -16,48 +16,57 @@ wrap a 2x3 array `a` in a `SizedArray`, use `SizedMatrix{2,3}(a)`.
...
+    M = ndims(a) == N ? N : 1
+    return SizedArray{S,T,N,M}(a)
 end
-function SizedArray{S,T,N,N}(::UndefInitializer) where {S,T,N}
+
+@inline (::Type{SZA})(a::AbstractArray) where {SZA<:SizedArray} = construct_type(SZA, a)(a)
+
+# disambiguation
+@inline SizedArray{S,T,N,M}(a::StaticArray) where {S<:Tuple,T,N,M} = construct_type(SizedArray{S,T,N,M}, a)(a.data)
+@inline SizedArray{S,T,N}(a::StaticArray) where {S<:Tuple,T,N} = construct_type(SizedArray{S,T,N}, a)(a.data)
```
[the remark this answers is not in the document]
N5N3: Added

src/SizedArray.jl
```diff
@@ -16,48 +16,57 @@ wrap a 2x3 array `a` in a `SizedArray`, use `SizedMatrix{2,3}(a)`.
...
+@inline (::Type{SZA})(a::StaticArray) where {SZA<:SizedArray} = construct_type(SZA, a)(a.data)
+# TODO: Should we respect `TData`?
+SizedArray{S,T,N,M,TData}(a::TData) where {S<:Tuple,T,N,M,TData<:StaticArray{<:Tuple,T,M}} = SizedArray{S,T,N,M}(a.data)
```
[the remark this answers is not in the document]
N5N3: Added

src/convert.jl
N5N3: I have renamed it as `SizeEltypeAdaptable` (`FieldArray` only do eltype adaption) Not sure is it a better name though.

oscardssmith:
Is this ready to merge?

N5N3:
I've added a warning to `construct_type`'s docstring that invalid overload of this function might cause infinite recursion.
The main concern might be the influence on downside packages.
The default behavior should be exactly the same, but I only test this PR with `Rotations.jl`, `Geodesy.jl`.

src/util.jl
```diff
@@ -19,6 +19,8 @@ TupleN{T,N} = NTuple{N,T}
 end
 
 # Base gives up on tuples for promote_eltype... (TODO can we improve Base?)
+_NTuple{T} = Tuple{T,Vararg{T}}
```
hyrodium, line 22: The difference between `StaticArrays._NTuple` and `Base.NTuple` is not clear from their names. How about `_TupleOf` instead?
The name `_TupleOf` seems appropriate for the following behavior.
```julia
julia> () isa NTuple{N, Int} where N
true

julia> (1,) isa NTuple{N, Int} where N
true

julia> () isa StaticArrays._NTuple{Int}
false

julia> (1,) isa StaticArrays._NTuple{Int}
true
```

hyrodium:
Sorry for the late review. The rest of the code looks great to me.

src/util.jl
N5N3: Make sense. Thanks for the suggestion.

hyrodium (approved):

PR #1020, opened 2022-04-20
Update version of Documenter.jl

hyrodium:


thchr:
Maybe also bump version of StaticArrays (so we can merge & tag to get your `pinv` work accessible).

thchr (approved):

PR #1021, opened 2022-04-24
Define Base.show for SArray and MArray.

### LINKED ISSUE #692: `repr` roundtripping. "print it like you build it"
I'm not sure if there are any functions in the display/show/repr family that are supposed to obey it, but it would be nice if `a_rt === a` (to the extent that everything is `bitstype`) in the following

```julia
using StaticArrays

a = @SArray [(SA_F32[1,2,3]), SA_F32[2,3,4]]
@show typeof(a)
sa = repr(a)
@show sa
a_rt = eval(Meta.parse(sa))
@show typeof(a_rt)
```
output:
```
typeof(a) = SArray{Tuple{2},SArray{Tuple{3},Float32,1,3},1,2}
sa = "SArray{Tuple{3},Float32,1,3}[[1.0, 2.0, 3.0], [2.0, 3.0, 4.0]]"
typeof(a_rt) = Array{SArray{Tuple{3},Float32,1,3},1}
```

(related a bit to the conversation in https://github.com/JuliaArrays/StaticArrays.jl/issues/691.

tpapp:
This allows `SArray` and `MArray` to be printed (`show`, `repr`, etc) and read back into an object of the same type.

Fixes #692.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -36,6 +36,12 @@ mutable struct MArray{S <: Tuple, T, N, L} <: StaticArray{S, T, N}
     end
 end
 
+function Base.show(io::IO, a::MArray{S,T}) where {S,T}
+    print(io, 'M')
+    _show_shape_size_type(io, size(a), T)
+    print(io, Tuple(a))
+end
+
 @generated function (::Type{MArray{S,T,N}})(x::Tuple) where {S,T,N}
     return quote
         $(Expr(:meta, :inline))
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -29,6 +29,12 @@ struct SArray{S <: Tuple, T, N, L} <: StaticArray{S, T, N}
     end
 end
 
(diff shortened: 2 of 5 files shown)


thchr:
I'm not sure that it's preferable to print the content as a tuple though; e.g.,
```jl
A = @SMatrix [1 2; 3 4]
```
would now `repr` as `SMatrix{2,2,Int}(1, 2, 3, 4)`. That seems OK for `repr` maybe, but it's not great for printing vectors of `A`. E.g.,  the default showing of `[A, A, A]` now becomes:
```
julia> [A, A, A]
3-element Vector{SMatrix{2, 2, Int64, 4}}:
 SMatrix{2,2,Int64}(1, 3, 2, 4)
 SMatrix{2,2,Int64}(1, 3, 2, 4)
 SMatrix{2,2,Int64}(1, 3, 2, 4)
```
How about instead printing this as `SA{T}[...]`?

mcabbott:
Xref #906 for compact printing

PR #1029, opened 2022-05-25
Split tests into two groups for x86 CI

mateuszbaran:
I'm tired of tests failing randomly on CI for x86 architectures, presumably due to too high memory usage. This should hopefully fix the issue by splitting tests into two groups when CI is used on an x86 architecture. The division point was determined by looking at which point memory typically runs out.


mateuszbaran:
It seems to be working. The project coverage change seems random, codecov doesn't show any new uncovered lines introduced by this PR.

test/runtests.jl
```diff
@@ -21,51 +21,60 @@ function addtests(fname)
...
-addtests("Scalar.jl")
-addtests("SUnitRange.jl")
-addtests("SizedArray.jl")
-addtests("SDiagonal.jl")
-addtests("SHermitianCompact.jl")
-if VERSION >= v"1.8.0-beta1"
-    addtests("empty_array_syntax.jl")
+TEST_GROUP = get(ENV, "STATICARRAYS_TEST_GROUP", "all")
+
+println("Using test group: ", TEST_GROUP)
```
thchr, line 26: Should this get printed even if `TEST_GROUP` is `"all"` or `""`?

thchr:
This LGTM and seems like a good fix to improve CI stability.

thchr (approved):

PR #1037, opened 2022-06-06
Fix copy of a complicated `SizedArray`

mateuszbaran:
I've noticed that the following code didn't work:
```julia
M = @MMatrix [1 2; 3 4]
copy(vec(M))
```
and this change fixes it.

### DIFF
--- a/test/SizedArray.jl
+++ b/test/SizedArray.jl
@@ -125,6 +125,9 @@
         sa2 = SizedArray{Tuple{2, 2}, Int}([1, 2, 3, 4])
         @test (@inferred vec(sa2)) isa SizedVector{4, Int}
         @test vec(sa2).data === vec(sa2.data)
+
+        M = @MMatrix [1 2; 3 4]
+        @test copy(vec(M)) isa SizedVector{4, Int}
     end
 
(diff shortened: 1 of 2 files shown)


test/SizedArray.jl
```diff
@@ -125,6 +125,9 @@
         sa2 = SizedArray{Tuple{2, 2}, Int}([1, 2, 3, 4])
         @test (@inferred vec(sa2)) isa SizedVector{4, Int}
         @test vec(sa2).data === vec(sa2.data)
+
+        M = @MMatrix [1 2; 3 4]
+        @test copy(vec(M)) isa SizedVector{4, Int}
```
hyrodium, line 130: A little unrelated to the original issue, but why does `vec(M::MMatrix)` return `SizedVector`? I thought it should be `MVector`. Is this documented behavior, or is there any performance reason for this?

mateuszbaran: `vec` is supposed to return a view of the given matrix (see Julia's docs for `vec`) which `MVector` can't represent, that's why `SizedVector` is used.

hyrodium (approved):

PR #1043, opened 2022-06-16
Resolve the identifier conflicts in `@SArray`

### LINKED ISSUE #1042: Macro hygiene issue in @SVector and @SMatrix
The internal symbol `f`, used in `@Vector`  is not scrambled, leading to a conflict with existing variables at the callsite:

```
f = [1,2,3]
s = @SVector[f[i] for i in 1:3]
```
gives

```
ERROR: MethodError: no method matching getindex(::var"#f#27", ::Int64)
Closest candidates are:
  getindex(::Any, ::Vector{BEAST.Variational.HilbertVector}) at ~/dev/BEAST/src/utils/variational.jl:264
  getindex(::Any, ::Vector{BEAST.Variational.HilbertVector}, ::Vector{BEAST.Variational.HilbertVector}) at ~/dev/BEAST/src/utils/variational.jl:295
  getindex(::Any, ::BEAST.Variational.HilbertVector) at ~/dev/BEAST/src/utils/variational.jl:262
  ...
Stacktrace:
 [1] (::var"#f#27")(i::Int64)
   @ Main ~/.julia/packages/StaticArrays/0T5rI/src/SVector.jl:68
 [2] top-level scope
   @ ~/.julia/packages/StaticArrays/0T5rI/src/SVector.jl:69
```

because:

```
julia> @macroexpand @SVector[f[i] for i in 1:3]
quote
    #= /home/em/krcools/.julia/packages/StaticArrays/0T5rI/src/SVector.jl:68 =#
    let f(i) = begin
                #= /home/em/krcools/.julia/packages/StaticArrays/0T5rI/src/SVector.jl:68 =#
                f[i]
            end
        #= /home/em/krcools/.julia/packages/StaticArrays/0T5rI/src/SVector.jl:69 =#
        (SVector){3}(tuple(f(1), f(2), f(3)))
    end
end
```

A similar issue occurs in `@SMatrix[ ... ]`

N5N3:
Fix #1042. I'm not sure how to add a test though...

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -60,7 +60,7 @@ A convenience macro to construct `MArray` with arbitrary dimension.
 See [`@SArray`](@ref) for detailed features.
 """
 macro MArray(ex)
-    esc(static_array_gen(MArray, ex, __module__))
+    static_array_gen(MArray, ex, __module__)
 end
 
 function promote_rule(::Type{<:MArray{S,T,N,L}}, ::Type{<:MArray{S,U,N,L}}) where {S,T,U,N,L}
--- a/src/MMatrix.jl
+++ b/src/MMatrix.jl
@@ -24,5 +24,5 @@ A convenience macro to construct `MMatrix`.
 See [`@SArray`](@ref) for detailed features.
 """
 macro MMatrix(ex)
-    esc(static_matrix_gen(MMatrix, ex, __module__))
+    static_matrix_gen(MMatrix, ex, __module__)
 end
--- a/src/MVector.jl
+++ b/src/MVector.jl
@@ -15,7 +15,7 @@ A convenience macro to construct `MVector`.
 See [`@SArray`](@ref) for detailed features.
 """
 macro MVector(ex)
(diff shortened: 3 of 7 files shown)


fredrikekre:
Isn't the problem that the macro puts `esc` on the entire expression, rather than just on user input: https://github.com/JuliaArrays/StaticArrays.jl/blob/52fee86bac81668a7700ab82df5b20c6ec4b8a64/src/SArray.jl#L272 ?

N5N3:
In fact, all most all the identifiers here need this `esc`. So I think this would be a much simpler fix.
I'll cut the unneeded escape when I have time if that's perferred.

hyrodium:
>I'm not sure how to add a test though...

How about just adding tests
```julia
        f = [1,2,3]
        @test (@SVector [f[i] for i in 1:3]) == [1,2,3]
```
in `test/SVector.jl`?

N5N3:
Test added. I didn't do this because the name was chosen at random.
But there seems to be no better way.

hyrodium (approved):
Thanks for this PR!

Issue #1023, opened 2022-05-07
StaticArrays load time, possibility for abstract interface package?


oschulz:
StaticArrays (together with ArraysInterface) seems to be a driving factor in package load times nowadays. For example, the load time of FiniteDiff, FiniteDifferences and ForwardDiff (common direct/indirect dependencies themselves) themselves is mainly due to StaticArrays (and ArrayInterface):

```julia
julia> @time_imports using FiniteDiff
      3.3 ms  ┌ Compat
      0.5 ms  ┌ Requires
    548.3 ms  ┌ StaticArrays
      0.2 ms  ┌ IfElse
     33.7 ms    ┌ Static
    737.6 ms  ┌ ArrayInterface
   1312.8 ms  FiniteDiff
```

```julia
julia> @time_imports using FiniteDifferences
      3.6 ms  ┌ Compat
      0.4 ms  ┌ Richardson
    539.9 ms  ┌ StaticArrays
     62.7 ms  ┌ ChainRulesCore
    617.1 ms  FiniteDifferences
```

```julia
julia> @time_imports using ForwardDiff
    538.5 ms    ┌ StaticArrays
    548.4 ms  ┌ DiffResults
      4.1 ms  ┌ Compat
      0.4 ms  ┌ NaNMath
     60.8 ms    ┌ ChainRulesCore
     62.0 ms  ┌ ChangesOfVariables
      0.7 ms  ┌ OpenLibm_jll
      3.0 ms  ┌ DocStringExtensions
      4.6 ms  ┌ IrrationalConstants
      0.6 ms  ┌ CompilerSupportLibraries_jll
      1.4 ms    ┌ LogExpFunctions
     18.9 ms        ┌ Preferences
     19.8 ms      ┌ JLLWrappers
     23.9 ms    ┌ OpenSpecFun_jll
     37.6 ms  ┌ SpecialFunctions
     10.3 ms    ┌ MacroTools
     11.0 ms  ┌ CommonSubexpressions
      0.8 ms  ┌ DiffRules
    845.2 ms  ForwardDiff
```

Even the load time of a "large" package like Optim is completely dominated by StaticArrays and ArrayInterface (indirectly, via FiniteDiff and ForwardDiff):

```julia
julia> @time using StaticArrays
  0.565813 seconds (2.21 M allocations: 175.407 MiB, 0.51% compilation time)

julia> @time using ArrayInterface
[...]

PR #1045, opened 2022-06-17
StaticArraysCore transition

mateuszbaran:
This PR implements transition to StaticArraysCore.jl, an interface package, as discussed here: https://github.com/JuliaArrays/StaticArrays.jl/issues/1023 .


ranocha:
Will the docstrings for `SArray` etc. still be displayed in the API docs?

mateuszbaran:
Good point, they are currently not. I'll try to fix it.

mateuszbaran:
With the last commit the docstrings show for me locally in the documentation.

ranocha (commented):
Nice :+1:

src/StaticArrays.jl
```diff
@@ -21,6 +21,19 @@ import LinearAlgebra: transpose, adjoint, dot, eigvals, eigen, lyap, tr,
...
+# from StaticArraysCore to make transitioning definitions to StaticArraysCore easier.
+using StaticArraysCore: StaticArraysCore, StaticArray, StaticScalar, StaticVector,
+                        StaticMatrix, StaticVecOrMat, tuple_length, tuple_prod,
+                        tuple_minimum, size_to_tuple, require_one_based_indexing
+import StaticArraysCore: SArray, SVector, SMatrix
+import StaticArraysCore: MArray, MVector, MMatrix
+import StaticArraysCore: SizedArray, SizedVector, SizedMatrix
+import StaticArraysCore: check_array_parameters, convert_ntuple
+
+# end of StaticArraysCore imports
```
fredrikekre, line 35: Move the (re)-exports below to this block too maybe?

fredrikekre:
Could you add some docs about this? I.e. describe why it is split like this and that if you just want to add dispatch for StaticArrays you can depend only on the core package (I guess this is the main reason for this split?)

src/StaticArrays.jl
mateuszbaran: Sure, done.

mateuszbaran:
Sure, I'll add some docs about this split. The motivation is primarily covered in this comment: https://github.com/JuliaArrays/StaticArrays.jl/issues/1023#issuecomment-1133756057 .

README.md
```diff
@@ -23,6 +23,11 @@ Further, the abstract `FieldVector` can be used to make fast `StaticVector`s
 out of any uniform Julia "struct".
 Full documentation can be found [here](https://JuliaArrays.github.io/StaticArrays.jl/stable/).
 
+Some of the array types are defined in the interface package [StaticArraysCore.jl](https://github.com/JuliaArrays/StaticArraysCore.jl).
+That package can be used to implement methods specialized for static arrays
+without depending on StaticArrays.jl, which has a non-negligible load time
+(around 0.6s for StaticArrays 1.4 on Julia 1.7).
```
thchr, line 29:
```suggestion
Most of the primary array types exported by StaticArrays.jl are defined in the small interface
package [StaticArraysCore.jl](https://github.com/JuliaArrays/StaticArraysCore.jl). This includes
e.g., the definitions of the abstract type `StaticArray` and the concrete types `SArray`, `MArray`,
and `SizedArray` (as well as their dimension-specific aliases).
This enables downstream packages to implement new methods for these types without depending
on (and hence loading) the entirety of StaticArrays.jl, and thereby to avoid incurring the full
load-time of StaticArrays.jl (which is on the order of 0.6 s for StaticArrays.jl v1.4 on Julia
v1.7).
```

thchr (approved):
This looks good to me.

I remember it being mentioned that `StaticArrays.LU` and `StaticArrays.QR` are also being extended by downstream packages; but it would probably be better to just completely drop those rather than to include them in StaticArraysCore.jl at all.

oschulz:
Is this going to be StaticArrays v1.5.0 ?

oschulz:
Thanks!

PR #1064, opened 2022-08-01
Preserve more `FieldArray`s with parametric `eltype`.

N5N3:
Before this PR, user needs to define `similar_type` for their own `FieldArray`s with parametric `eltype`.
After this PR, these `FieldArray`s would be preserved automatically if there's not `Size` change.  (Inspired by https://github.com/JuliaLang/julia/pull/46213)

This PR also change the default similar type for mutable `FieldArray` to `MArray`.


vtjnash:
This code is likely wrong, as mentioned in the manual for purpose of this function:
https://docs.julialang.org/en/v1/manual/methods/#Building-a-similar-type-with-a-different-type-parameter
and in the issue mentioning why we don't currently provide this method:
https://github.com/JuliaLang/julia/issues/35543#issuecomment-792793809

src/FieldArray.jl
```diff
@@ -125,14 +125,23 @@ Base.cconvert(::Type{<:Ptr}, a::FieldArray) = Base.RefValue(a)
...
-    default_similar_type(T, NewSize, length_val(NewSize))
+function similar_type(::Type{A}, ::Type{T}, S::Size) where {T,A<:FieldArray}
+    # We can preserve FieldArrays in array operations which do not change their `Size` and `eltype`.
+    has_eltype(A) && eltype(A) === I && has_size(A) && Size(A) === S && return A
+    # FieldArrays with parametric `eltype` would be adapted to the new `eltype` automatically.
+    A′ = Base.typeintersect(base_type(A), StaticArray{Tuple{Tuple(S)...},T,length(S)})
+    # But extra parameters are disallowed here. Also we check `fieldtypes` to make sure the result is valid.
+    isconcretetype(A′) && fieldtypes(A′) === ntuple(Returns(T), Val(prod(S))) && return A′
+    # Otherwise, we fallback to `S/MArray` based on it's mutability.
+    if ismutabletype(A)
```
N5N3, line 136: I still think this branch is useful for `FieldArray`s with only one parameter for `eltype`?
Check in L134 seems sufficient for me, as a `concrete` type with matched `fieldtypes` should be a valid return.

src/FieldArray.jl
```diff
@@ -125,14 +125,23 @@ Base.cconvert(::Type{<:Ptr}, a::FieldArray) = Base.RefValue(a)
...
-_fieldarray_similar_type(A, T, NewSize::S, OldSize::S) where {S} = A
-_fieldarray_similar_type(A, T, NewSize, OldSize) =
-    default_similar_type(T, NewSize, length_val(NewSize))
+function similar_type(::Type{A}, ::Type{T}, S::Size) where {T,A<:FieldArray}
+    # We can preserve FieldArrays in array operations which do not change their `Size` and `eltype`.
+    has_eltype(A) && eltype(A) === I && has_size(A) && Size(A) === S && return A
+    # FieldArrays with parametric `eltype` would be adapted to the new `eltype` automatically.
+    A′ = Base.typeintersect(base_type(A), StaticArray{Tuple{Tuple(S)...},T,length(S)})
+    # But extra parameters are disallowed here. Also we check `fieldtypes` to make sure the result is valid.
+    isconcretetype(A′) && fieldtypes(A′) === ntuple(Returns(T), Val(prod(S))) && return A′
```
N5N3, line 134: I still think this branch is useful for `FieldArrays` with only one parameter for `eltype`?
Check in L134 seems sufficient for me, as a concrete type with matched `fieldtypes` should be a valid return.

PR #1068, opened 2022-08-03
Adjoint and Projections

Abhishek-1Bhatt:
Solves a few AD related issues:

- Adjoint of SArray constructor
- SArray projected to SizedArray in pullback due to [this](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/00e17d236e3c09063c5b27a52c550c841d591fd5/src/projection.jl#L230) `reshape` during pullback
- SMatrix tuple dimensions don't match the matrix dimensions when the above is fixed in [this](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/00e17d236e3c09063c5b27a52c550c841d591fd5/src/projection.jl#L368) projection

Example

```julia
using Zygote
using StaticArrays

u0 = @SVector [1.0f0, 1.0f0]
p = @SVector [1.5f0, 1.0f0, 3.0f0, 1.0f0]

function lotka(u, p)
    du1 = p[1]*u[1] - p[2]*u[1]*u[2]
    du2 = -p[3]*u[2] + p[4]*u[1]*u[2]
    @SVector [du1, du2]
    # @SMatrix [du1 du2 du1; du2 du1 du1]
end

function loss(p)
    u = lotka(u0, p)
    sum(1 .- u)
end

loss(p)

grad = Zygote.gradient(loss, p)
```
```
ERROR: Need an adjoint for constructor SVector{2, Float32}. Gradient is of type SizedVector{2, Float32, Vector{Float32}}
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] (::Zygote.Jnew{SVector{2, Float32}, Nothing, false})(Δ::SizedVector{2, Float32, Vector{Float32}})
    @ Zygote C:\Users\user\.julia\packages\Zygote\D7j8v\src\lib\lib.jl:326
  [3] (::Zygote.var"#1928#back#222"{Zygote.Jnew{SVector{2, Float32}, Nothing, false}})(Δ::SizedVector{2, Float32, Vector{Float32}})
    @ Zygote C:\Users\user\.julia\packages\ZygoteRules\AIbCs\src\adjoint.jl:67
  [4] Pullback
    @ C:\Users\user\.julia\packages\StaticArraysCore\gkLqH\src\StaticArraysCore.jl:106 [inlined]
  [5] (::typeof(∂(SVector{2, Float32})))(Δ::SizedVector{2, Float32, Vector{Float32}})
    @ Zygote C:\Users\user\.julia\packages\Zygote\D7j8v\src\compiler\interface2.jl:0
  [6] Pullback
    @ C:\Users\user\.julia\dev\StaticArrays\src\convert.jl:163 [inlined]
  [7] Pullback
    @ c:\Users\user\.julia\dev\buffer.jl:668 [inlined]
  [8] (::typeof(∂(lotka)))(Δ::SizedVector{2, Float32, Vector{Float32}})
    @ Zygote C:\Users\user\.julia\packages\Zygote\D7j8v\src\compiler\interface2.jl:0
  [9] Pullback
    @ c:\Users\user\.julia\dev\buffer.jl:673 [inlined]
 [10] (::typeof(∂(loss)))(Δ::Float32)
    @ Zygote C:\Users\user\.julia\packages\Zygote\D7j8v\src\compiler\interface2.jl:0
 [11] (::Zygote.var"#60#61"{typeof(∂(loss))})(Δ::Float32)
    @ Zygote C:\Users\user\.julia\packages\Zygote\D7j8v\src\compiler\interface.jl:41
 [12] gradient(f::Function, args::SVector{4, Float32})
    @ Zygote C:\Users\user\.julia\packages\Zygote\D7j8v\src\compiler\interface.jl:76
 [13] top-level scope
    @ c:\Users\user\.julia\dev\buffer.jl:679
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,14 +3,15 @@ uuid = "90137ffa-7385-5640-81b9-e52037218182"
 version = "1.5.2"
 
 [deps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-julia = "1.6"
 StaticArraysCore = "1"
+julia = "1.6"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/Project.toml
+++ b/Project.toml
@@ -18,6 +18,8 @@ BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 
 [targets]
-test = ["InteractiveUtils", "Test", "BenchmarkTools", "OffsetArrays"]
+test = ["InteractiveUtils", "Test", "BenchmarkTools", "OffsetArrays", "Zygote", "ForwardDiff"]
--- a/src/chainrule.jl
+++ b/src/chainrule.jl
@@ -1,39 +1,33 @@
-#####
-##### constructors
-#####
+### Projecting a tuple to SMatrix leads to ChainRulesCore._projection_mismatch by default, so overloaded here
+function (project::ChainRulesCore.ProjectTo{<:Tangent{<:Tuple}})(dx::SArray)
+    # for d in 1:ndims(dx)
+    #     if size(dx, d) != get(length(project.elements), d, 1)
+    #         throw(ChainRulesCore._projection_mismatch(axes(project.elements), size(dx)))
+    #     end
+    # end
+    dy = reshape(dx, axes(project.elements))  # allows for dx::OffsetArray
+    dz = ntuple(i -> project.elements[i](dy[i]), length(project.elements))
+    return ChainRulesCore.project_type(project)(dz...)
+end
+### Project SArray to SArray
+function ChainRulesCore.ProjectTo(x::SArray{S,T}) where {S, T}
+    return ChainRulesCore.ProjectTo{SArray}(; element=ChainRulesCore._eltype_projectto(T), axes=S)
+end
 
-ChainRulesCore.@non_differentiable (::Type{T} where {T<:Union{SArray, SizedArray}})(::UndefInitializer, args...)
+function (project::ChainRulesCore.ProjectTo{SArray})(dx::AbstractArray{S,M}) where {S,M}
+    return SArray{project.axes}(dx)
+end
 
-function ChainRulesCore.frule((_, ẋ), ::Type{T}, x::Tuple) where {T<:Union{SArray, SizedArray}}
+### Adjoint for SArray constructor
+
+ChainRulesCore.@non_differentiable (::Type{T} where {T<:SArray})(::UndefInitializer, args...)
+
+function ChainRulesCore.frule((_, ẋ), ::Type{T}, x::Tuple) where {T<:SArray}
     return T(x), T(ẋ)
 end
 
-function ChainRulesCore.rrule(::Type{T}, x::Tuple) where {T<:Union{SArray, SizedArray}}
+function ChainRulesCore.rrule(::Type{T}, x::Tuple) where {T<:SArray}
     project_x = ProjectTo(x)
     Array_pullback(ȳ) = (NoTangent(), project_x(ȳ))
     return T(x), Array_pullback
-end
-
-function (project::ChainRulesCore.ProjectTo{AbstractArray})(dx::AbstractArray{S,M}) where {S,M}
-    # First deal with shape. The rule is that we reshape to add or remove trivial dimensions
-    # like dx = ones(4,1), where x = ones(4), but throw an error on dx = ones(1,4) etc.
-    dy = if axes(dx) === project.axes
-        dx
-    else
-        for d in 1:max(M, length(project.axes))
-            if size(dx, d) != length(get(project.axes, d, 1))
-                throw(_projection_mismatch(project.axes, size(dx)))
-            end
-        end
-        reshape(dx, project.axes)
-    end
-    # Then deal with the elements. One projector if AbstractArray{<:Number},
-    # or one per element for arrays of anything else, including arrays of arrays:
-    dz = if hasproperty(project, :element)
-        T = project_type(project.element)
-        S <: T ? dy : map(project.element, dy)
-    else
-        map((f, y) -> f(y), project.elements, dy)
-    end
-    return dz
 end
--- a/src/chainrule.jl
+++ b/src/chainrule.jl
@@ -16,12 +16,6 @@ end
 
 ### Adjoint for SArray constructor
 
-ChainRulesCore.@non_differentiable (::Type{T} where {T<:SArray})(::UndefInitializer, args...)
-
-function ChainRulesCore.frule((_, ẋ), ::Type{T}, x::Tuple) where {T<:SArray}
-    return T(x), T(ẋ)
-end
-
 function ChainRulesCore.rrule(::Type{T}, x::Tuple) where {T<:SArray}
     project_x = ProjectTo(x)
     Array_pullback(ȳ) = (NoTangent(), project_x(ȳ))
--- a/src/chainrule.jl
+++ b/src/chainrule.jl
@@ -1,14 +1,10 @@
 ### Projecting a tuple to SMatrix leads to ChainRulesCore._projection_mismatch by default, so overloaded here
 function (project::ChainRulesCore.ProjectTo{<:Tangent{<:Tuple}})(dx::SArray)
-    # for d in 1:ndims(dx)
-    #     if size(dx, d) != get(length(project.elements), d, 1)
-    #         throw(ChainRulesCore._projection_mismatch(axes(project.elements), size(dx)))
-    #     end
-    # end
     dy = reshape(dx, axes(project.elements))  # allows for dx::OffsetArray
     dz = ntuple(i -> project.elements[i](dy[i]), length(project.elements))
     return ChainRulesCore.project_type(project)(dz...)
 end
+
 ### Project SArray to SArray
 function ChainRulesCore.ProjectTo(x::SArray{S,T}) where {S, T}
     return ChainRulesCore.ProjectTo{SArray}(; element=ChainRulesCore._eltype_projectto(T), axes=S)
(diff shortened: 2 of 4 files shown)


Abhishek-1Bhatt:
What does `detect_ambiguities` do?
https://github.com/JuliaArrays/StaticArrays.jl/blob/76ce2c54bb1ac86923ff84d66e6b82875cea6808/test/ambiguities.jl#L11

mateuszbaran:
> What does `detect_ambiguities` do?

It's a standard Julia function: https://docs.julialang.org/en/v1/stdlib/Test/#Test.detect_ambiguities .

Abhishek-1Bhatt:
Is it certain that this won't be merged ? I think the alternative would then be to send a pr to ChainRulesCore for the `ProjectTo` and another pr to ChainRules for the `rrule` and `frule`. Also that would add StaticArrays as a dependency to both. @mcabbott any suggestions?

mateuszbaran:
Well, I can't speak for other people with write access. Anyway there is `StaticArraysCore.jl` which should be a sufficient dependency to implement these methods.

hyrodium:
I think it would be okay to add the dependency on [ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl).

* The package is lightweight. (The package does not depend on anything other than the standard library.)
* Its APIs are stable. (The latest version is `v1.15.3`.)

>The ChainRulesCore package provides a light-weight dependency for defining sensitivities for functions in your packages, without you needing to depend on ChainRules itself.

mcabbott:
ChainRulesCore.jl really shouldn't depend on anything. I wish it were smaller. 

I think the policy for ChainRules.jl is only to define rules for things in Base/std.lib., in the hope that everyone else will depend on CRC, and out of fear that otherwise it'll suffer endless mission creep with sort-of important packages... like this one, or FillArrays:  https://github.com/JuliaArrays/FillArrays.jl/pull/153 still has nowhere to live, sadly.

Abhishek-1Bhatt:
I have removed the ambiguity causing methods, which includes the `frule` (1 ambiguity), but that works for me as I really only need `rrule` . Also adding to @hyrodium 's opinion on ChainRulesCore as a dependency, I think that allowing users to have out of the box AD compatibility has become pretty standard  in Julia, so this will be helping in making the feature-set of StaticArrays more complete, lets please get it merged : )

stevengj:
(Needs a rebase.)

stevengj:
I really think that being differentiable is a reasonable expectation for most packages these days, and ChainRulesCore is a lightweight dependency that is very likely to be already installed.

thchr:
It is not all that lightweight though? It takes ~0.3 s to load ChainRulesCore on my laptop versus 0.75 s for StaticArrays itself, so including it as a dependency would add ~40% to the StaticArrays' load time.
As context to this, I think there is already some sentiment that StaticArrays loads too slowly as-is.

It's true that it is very likely to be loaded by other packages though since it is e.g., in SpecialFunctions.jl (but I don't think the decision to include it there was entirely clear-cut either, cf. https://github.com/JuliaMath/SpecialFunctions.jl/issues/310).

Project.toml
```diff
@@ -3,20 +3,23 @@ uuid = "90137ffa-7385-5640-81b9-e52037218182"
...
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 
 [targets]
-test = ["InteractiveUtils", "Test", "BenchmarkTools", "OffsetArrays"]
+test = ["InteractiveUtils", "Test", "BenchmarkTools", "OffsetArrays", "Zygote", "ForwardDiff"]
```
ToucheSir, line 25: Regardless of where these rules end up living, they should be tested with ChainRulesTestUtils and not ad-hoc via Zygote/ForwardDiff/what have you. That library does far more robust testing than most of us would think to write by hand.

test/abstractarray.jl
```diff
@@ -243,6 +243,43 @@ using StaticArrays, Test, LinearAlgebra
...
+            else
+                @SMatrix [du1 du2 du1; du2 du1 du1]
+            end
+        end
+
+        #SVector constructor adjoint
+        function loss(p)
+            u = lotka(u0, p)
+            sum(1 .- u)
+        end
```
ToucheSir, line 265: Also, this seems a rather not minimal and domain-specific test?

hyrodium:
>It is not all that lightweight though? It takes ~0.3 s to load ChainRulesCore on my laptop versus 0.75 s for StaticArrays itself, so including it as a dependency would add ~40% to StaticArrays' load time.

How did you measure the time? In my environment, the load time doesn't change that much.

**On the current master**
```julia
julia> @time_imports using StaticArrays
      2.7 ms  StaticArraysCore
    659.3 ms  StaticArrays

julia> versioninfo()
Julia Version 1.8.0
Commit 5544a0fab76 (2022-08-17 13:38 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 2700X Eight-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, znver1)
  Threads: 1 on 16 virtual cores
```

**With this PR**
```julia
julia> @time_imports using StaticArrays
      0.5 ms  Compat
     73.7 ms  ChainRulesCore
      2.1 ms  StaticArraysCore
    651.7 ms  StaticArrays

julia> versioninfo()
Julia Version 1.8.0
Commit 5544a0fab76 (2022-08-17 13:38 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 2700X Eight-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, znver1)
  Threads: 1 on 16 virtual cores
```

thchr:
> How did you measure the time? In my environment, the load time doesn't change that much.

I just did plain `@time using` on the individual packages (i.e., I didn't check out this PR) - but trying this again, I see that it depends on the load-order of StaticArrays and ChainRules, and also seems to depend on whether or not Revise is loaded. With the right ordering or without Revise, I see times closer to yours:

<details>
<summary>With Revise loaded</summary>

```jl
tchr@mit:~$ julia

julia> @time using ChainRulesCore
  0.283618 seconds (109.83 k allocations: 6.930 MiB, 78.45% compilation time)

  0.643347 seconds (2.27 M allocations: 162.173 MiB, 31.34% compilation time: 100% of which was recompilation)
```

```jl
tchr@mit:~$ julia

  0.679030 seconds (2.06 M allocations: 151.050 MiB, 30.93% compilation time)

julia> @time using ChainRulesCore
  0.065993 seconds (89.80 k allocations: 5.816 MiB)
```
</details>
<details>
<summary>Without Revise loaded</summary>

```jl
tchr@mit:~$ julia --startup-file=no

julia> @time using ChainRulesCore
  0.061022 seconds (96.67 k allocations: 8.075 MiB)

  0.710017 seconds (2.27 M allocations: 162.195 MiB, 2.32% gc time, 31.62% compilation time: 100% of which was recompilation)
```

```jl
tchr@mit:~$ julia --startup-file=no

  0.476433 seconds (2.05 M allocations: 152.639 MiB)

julia> @time using ChainRulesCore
  0.063983 seconds (87.06 k allocations: 5.672 MiB)
```

</details>

The loading time I originally mentioned might then be mainly due to invalidations between ChainRulesCore and Revise, I suppose?

mcabbott:
BTW, some experiments with loading time just before CRC 1.0 last year: https://github.com/JuliaDiff/ChainRulesCore.jl/issues/413 . These are just `@time using`, could perhaps be done more carefully to count the cost of invalidations, Revise, etc. 

Maybe a much lighter CRC is worth some thought. I think the downside of moving everything except `function rrule end` out to CR is essentially testing --- it is never useful without CR being loaded, but it is currently tested without loading CR, which makes a nice tree with no cycles.

It has [148 direct deps](https://juliahub.com/ui/Packages/ChainRulesCore/G6ax7/1.15.5?page=2) so we shouldn't do 2.0 lightly. I wonder how many would actually notice if it were stripped?

src/chainrule.jl
```diff
@@ -0,0 +1,23 @@
...
+    return ChainRulesCore.project_type(project)(dz...)
+end
+
+### Project SArray to SArray
+function ChainRulesCore.ProjectTo(x::SArray{S,T}) where {S, T}
+    return ChainRulesCore.ProjectTo{SArray}(; element=ChainRulesCore._eltype_projectto(T), axes=S)
+end
+
+function (project::ChainRulesCore.ProjectTo{SArray})(dx::AbstractArray{S,M}) where {S,M}
+    return SArray{project.axes}(dx)
```
mcabbott, line 14: To match what projection does for `Array`, I think this ought to use the `element` projector to correct eltype. 

The CRC one also checks that the size differs only in trivial ways (i.e. size-1 trailing dimensions), otherwise errors. Sometimes this is helpful for finding bugs in AD rules. I think this will accept any shape with the right length, which most bugs will still hit...

src/chainrule.jl
```diff
@@ -0,0 +1,23 @@
+### Projecting a tuple to SMatrix leads to ChainRulesCore._projection_mismatch by default, so overloaded here
+function (project::ChainRulesCore.ProjectTo{<:Tangent{<:Tuple}})(dx::SArray)
+    dy = reshape(dx, axes(project.elements))  # allows for dx::OffsetArray
+    dz = ntuple(i -> project.elements[i](dy[i]), length(project.elements))
+    return ChainRulesCore.project_type(project)(dz...)
+end
+
+### Project SArray to SArray
+function ChainRulesCore.ProjectTo(x::SArray{S,T}) where {S, T}
+    return ChainRulesCore.ProjectTo{SArray}(; element=ChainRulesCore._eltype_projectto(T), axes=S)
```
mcabbott, line 10: If anyone is relying on `ProjectTo(x).axes` actually containing `axes(x)`, then this will be very surprising. Maybe this `Tuple{size...}` thing ought to have a different name:

```suggestion
    return ChainRulesCore.ProjectTo{SArray}(; element=ChainRulesCore._eltype_projectto(T), axes=axes(x), static_size=S)
```

src/chainrule.jl
```diff
@@ -0,0 +1,23 @@
+### Projecting a tuple to SMatrix leads to ChainRulesCore._projection_mismatch by default, so overloaded here
+function (project::ChainRulesCore.ProjectTo{<:Tangent{<:Tuple}})(dx::SArray)
```
mcabbott, line 2: What's the MWE that hits this?

And might it be simpler just to do something like `(p::ChainRulesCore.ProjectTo{<:Tangent{<:Tuple}})(dx::SArray) = p(Tuple(dx))`, correct the type & then take the same path as other tuples?

mateuszbaran:
I think `ChainRulesCore` as a Julia 1.9-style package extension in `StaticArrays` would make sense.

gdalle:
I can turn this PR into a package extension if someone can answer these two questions:

- is the code otherwise operational?
- what behavior do we want for Julia < 1.9? either nothing or Requires?

Abhishek-1Bhatt:
The tests are passing in SciMLSensitivity https://github.com/SciML/SciMLSensitivity.jl/pull/680/commits/d354ba79c801babe34fad29d555eec0da02e74e7

ChrisRackauckas:
Fixed in https://github.com/JuliaArrays/StaticArrays.jl/pull/1224

PR #1073, opened 2022-08-10
improve error message of `setindex!`

### LINKED ISSUE #1065: Throw better error when trying to `setindex!` on a `SArray`
Right now, if you try to mutate a static array, it shows the following error: 

```julia
ERROR: setindex!(::SArray{N, T}, value, ::Int) is not defined.
```

which I think could be a bit confusing. I propose that we define `setindex!` to trow an error, something like this:

```julia
setindex!(::SArray, args...) = error("Static arrays can't be mutated") # Or "you are not allowed to mutate static arrays" or similar
```

It seems like an easy change. I would do a pull request myself but I am very unfamiliar with the codebase and I'm not sure where it should go.

Odilf:
This pull request closes #1065 by throwing a better error when attempting to mutate a `StaticArray`

```julia
julia> SVector(1, 2, 3)[1] = 2
ERROR: setindex!(::SVector{3, Int64}, value, ::Int) is not defined.
 Hint: Use `MArray` to create a mutable static array
```

Note: Julia has an [experimental error hints](https://docs.julialang.org/en/v1/base/base/#Base.Experimental.register_error_hint) but I didn't use them because they're still experimental (and it seemed like too much stuff for a simple error message).

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -1,6 +1,6 @@
 # Default error messages to help users with new types and to avoid subsequent stack overflows
 getindex(a::StaticArray, i::Int) = error("getindex(::$(typeof(a)), ::Int) is not defined.")
-setindex!(a::StaticArray, value, i::Int) = error("setindex!(::$(typeof(a)), value, ::Int) is not defined.")
+setindex!(a::StaticArray, value, i::Int) = error("setindex!(::$(typeof(a)), value, ::Int) is not defined.\n Hint: Use `MArray` to create a mutable static array")
 
 #######################################
 ## Multidimensional scalar indexing  ##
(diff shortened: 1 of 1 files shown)


src/indexing.jl
```diff
@@ -1,6 +1,6 @@
 # Default error messages to help users with new types and to avoid subsequent stack overflows
 getindex(a::StaticArray, i::Int) = error("getindex(::$(typeof(a)), ::Int) is not defined.")
-setindex!(a::StaticArray, value, i::Int) = error("setindex!(::$(typeof(a)), value, ::Int) is not defined.")
+setindex!(a::StaticArray, value, i::Int) = error("setindex!(::$(typeof(a)), value, ::Int) is not defined.\n Hint: Use `MArray` to create a mutable static array")
```
hyrodium, line 3:
```suggestion
setindex!(a::StaticArray, value, i::Int) = error("setindex!(::$(typeof(a)), value, ::Int) is not defined.\n Hint: Use `MArray` or `SizedArray` to create a mutable static array")
```

hyrodium (commented):
LGTM! just a minor suggestion.

hyrodium (approved):

PR #1077, opened 2022-08-15
Move FieldArray to Core

mateuszbaran:
Follow-up on https://github.com/JuliaArrays/StaticArraysCore.jl/pull/7 .


Project.toml
```diff
@@ -10,7 +10,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 julia = "1.6"
-StaticArraysCore = "1"
+StaticArraysCore = "1.1"
```
hyrodium, line 13:
```suggestion
StaticArraysCore = "~1.1"
```

https://pkgdocs.julialang.org/v1/compatibility/#Tilde-specifiers

hyrodium (commented):
Is this behavior expected?
```julia
(tmp) pkg> st StaticArrays StaticArraysCore
      Status `~/tmp/Project.toml`
  [90137ffa] StaticArrays v1.5.3
  [1e83bf80] StaticArraysCore v1.1.0

julia> using StaticArrays, StaticArraysCore

julia> StaticArrays.FieldArray == StaticArraysCore.FieldArray
false
```
To prevent this behavior, I thought the `[compat]` table should be updated with a tilde specifier.

mateuszbaran:
That's expected and should be more or less harmless. But indeed a tilde specifier would prevent such combinations of versions in the future.

PR #1086, opened 2022-08-26
implement ConstructionBase interface

aplavin:
Brief motivation: ConstructionBase is the foundational package for (re)constructing and modifying immutable and mutable objects, see https://github.com/JuliaObjects/ConstructionBase.jl.
`constructorof(T)(args)` is the main interface of that package, it works with lots of types by default. However, it needs to be overloaded when some type parameters cannot be determined purely from arguments passed to the constructor. Clearly, this is the case for `SArray`.
This PR implements that interface, making StaticArrays immediately compatible with all those packages and user codes that use ConstructionBase.

A similar PR was proposed 1.5 years ago: https://github.com/JuliaArrays/StaticArrays.jl/pull/901. See for more discussion and more details on the motivation.
Several things have changed since that PR that make me believe there are more reasons for this to be accepted now.

First, the implementation here is definitely cleaner and shorter than in #901. It just forwards `constructorof` to already existing S/MArray constructors, keeping only those type parameters that are necessary for reconstruction. So, it automatically supports all S/MArray constructor variations.
In partucular, this addresses a concern made for #901:
> In generic code I might want to change the size or element type - so I'm a little curious how you handle this in practice with constructorof?

Second, more and more packages are using ConstructionBase: already more than 1k in General (transitively) depend on it, and this number is growing. ConstructionBase has stayed stable and there are no plans for breaking changes or "feature creep". It's an extremely lightweight package with no dependencies and negligible load time.

@rafaqz @jw3126 @andyferris

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "90137ffa-7385-5640-81b9-e52037218182"
 version = "1.5.6"
 
 [deps]
+ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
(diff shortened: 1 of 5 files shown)


mateuszbaran:
I think it would make sense to have this as a package extension: https://github.com/JuliaLang/julia/pull/47695 when a stable version of Julia is released with it.

Issue #1088, opened 2022-09-02
StackOverflowError on FieldVector introduced in v1.4.5


tamasgal:
I have the following code which has worked fine until 1.4.4 and started to produce a StackOverflowError in 1.4.5. I have not done a bisect yet but I guess it's coming from https://github.com/JuliaArrays/StaticArrays.jl/issues/1026

This is the code, which ought to prevent a construction of the `FieldVector` subtype with different types:

```julia
struct Position{T} <: FieldVector{3, T}
    x::T
    y::T
    z::T

    Position(x::T, y::T, z::T) where {T} = new{T}(x, y, z)
end
```

The offending line is the inner constructor. It works with individual errors but gives a `StackOverflowError ` when e.g. a tuple is passed:

```
julia> Position(1,2,3)
3-element Position{Int64} with indices SOneTo(3):
 1
 2
 3

julia> Position((1,2,3))
ERROR: StackOverflowError:
Stacktrace:
     [1] Position{Int64}(::Int64, ::Vararg{Int64})
       @ StaticArrays ~/.julia/packages/StaticArrays/G7IlJ/src/convert.jl:152
     [2] Position{Int64}(x::Tuple{Int64, Int64, Int64})
       @ StaticArrays ~/.julia/packages/StaticArrays/G7IlJ/src/FieldArray.jl:113
--- the last 2 lines are repeated 18647 more times ---
 [37297] StaticArray
       @ ~/.julia/packages/StaticArrays/G7IlJ/src/convert.jl:152 [inlined]
```

If I remove that line, it works but then segfaults on v1.4.4 and below 😉 see here: 

```julia
julia> Position(1,2,3.0)
ERROR: StackOverflowError:
Stacktrace:
     [1] Position(::Int64, ::Vararg{Any})
       @ StaticArrays ~/.julia/packages/StaticArrays/58yy1/src/convert.jl:4
     [2] Position(x::Tuple{Int64, Int64, Float64})
       @ StaticArrays ~/.julia/packages/StaticArrays/58yy1/src/FieldArray.jl:117--- the last 2 lines are repeated 18646 more times ---
```

I guess the solution is to tag a version and limit the StaticArrays version to max 1.4.4 and then create a new one with the removed inner constructor and require at least 1.4.5. Or is there a better way?

mateuszbaran:
This is a small intended change introduced by #1016 to fix some issues.

tamasgal:
[...]

PR #1090, opened 2022-09-09
Avoid possible stackoverflow in FieldArray's constructor

N5N3:
Stackoverflow is always bad as a error message. This PR tries to make it more informative.
Test added.

### DIFF
--- a/src/FieldArray.jl
+++ b/src/FieldArray.jl
@@ -5,8 +5,12 @@ function construct_type(::Type{FA}, x) where {FA <: FieldArray}
     has_size(FA) || error("$FA has no static size!")
     length_match_size(FA, x)
     FA′ = adapt_eltype(FA, x)
-    FA′ === FA && x isa Args && error("the constructor for $FA is missing!")
+    FA′ === FA && x isa Args && _missing_fa_constructor(FA, typeof(x.args))
     return FA′
+end
+@noinline function _missing_fa_constructor(@nospecialize(FA), @nospecialize(AT))
+    Ts = join(("::$T" for T in fieldtypes(AT)), ", ")
+    error("The constructor for $FA($Ts) is missing!")
 end
 
 @propagate_inbounds getindex(a::FieldArray, i::Int) = getfield(a, i)
--- a/src/FieldArray.jl
+++ b/src/FieldArray.jl
@@ -4,7 +4,9 @@
 function construct_type(::Type{FA}, x) where {FA <: FieldArray}
     has_size(FA) || error("$FA has no static size!")
     length_match_size(FA, x)
-    return adapt_eltype(FA, x)
+    FA′ = adapt_eltype(FA, x)
+    FA′ === FA && x isa Args && error("the constructor for $FA is missing!")
+    return FA′
 end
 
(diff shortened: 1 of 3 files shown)


hyrodium:
x-ref https://github.com/JuliaArrays/StaticArrays.jl/issues/1088

src/FieldArray.jl
```diff
@@ -4,7 +4,9 @@
 function construct_type(::Type{FA}, x) where {FA <: FieldArray}
     has_size(FA) || error("$FA has no static size!")
     length_match_size(FA, x)
-    return adapt_eltype(FA, x)
+    FA′ = adapt_eltype(FA, x)
+    FA′ === FA && x isa Args && error("the constructor for $FA is missing!")
```
mateuszbaran, line 8:
```suggestion
    FA′ === FA && x isa Args && error("the constructor for $FA from $x is missing!")
```
In the test `Position1088` does have _a_ constructor, just not the one that is needed, so I think it would be good to have a more clear error message.

N5N3: I tried to expand it as `Args` is not public, now this gives
```julia
julia> Position((1.,2.,3.))
ERROR: The constructor for Position{Float64}(::Float64, ::Float64, ::Float64) is missing!
```
should be more intuitive.

mateuszbaran: Shouldn't it be `Position{Float64}(::typeof(x))`?

N5N3: `x` isa `Args` here, so the `construct_type` was called from `FA(x.args...)`.
Args` was introduced in #1016 to distinguish `SA(x::Tuple)` and `SA(x...)`). It's only used in `construct_type` and would not be passed to the derived constructor.
User should define `Position{T}(x::T, y::T, z::T) where {T}` to fix the constructor missing here.

mateuszbaran: Ah, right, sorry -- I'm not particularly familiar with `FieldArray` stuff.

mateuszbaran (approved):

N5N3:
Well, x isa Args here, which meanings we call construct_type from FA(x.args…). Args is only for internal usage to distinguish SA(x::Tuple) and SA(x…)

------------------ Original ------------------
From: Mateusz Baran ***@***.***&gt;
Date: Sun,Sep 11,2022 10:44 PM
To: JuliaArrays/StaticArrays.jl ***@***.***&gt;
Cc: N5N3 ***@***.***&gt;, Author ***@***.***&gt;
Subject: Re: [JuliaArrays/StaticArrays.jl] Avoid possible stackoverflow in FieldArray's constructor (PR #1090)





 
@mateuszbaran commented on this pull request.
 
 
In src/FieldArray.jl:
 &gt; @@ -4,7 +4,9 @@  function construct_type(::Type{FA}, x) where {FA <: FieldArray}      has_size(FA) || error("$FA has no static size!")      length_match_size(FA, x) -    return adapt_eltype(FA, x) +    FA′ = adapt_eltype(FA, x) +    FA′ === FA &amp;&amp; x isa Args &amp;&amp; error("the constructor for $FA is missing!")  
Shouldn't it be Position{Float64}(::typeof(x))?

PR #1091, opened 2022-09-12
Update advice on working with mutable arrays

npbarnes:
It looks like Julia 0.5 needed some workarounds for broadcasting to be efficient, but I'm pretty sure it's okay to use .+= and ./= in recent versions of Julia. I removed references to Julia 0.5 and 0.6.


thchr (approved):
Lgtm

PR #1094, opened 2022-09-19
Move `Size` to StaticArraysCore.jl

mateuszbaran:
Counterpart to https://github.com/JuliaArrays/StaticArraysCore.jl/pull/13

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -1,11 +1,3 @@
-"""
-    Dynamic()
-
-Used to signify that a dimension of an array is not known statically.
-"""
-struct Dynamic end
-
-const StaticDimension = Union{Int, Dynamic}
 
 """
     dimmatch(x::StaticDimension, y::StaticDimension)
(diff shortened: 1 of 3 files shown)


thchr:
This LGTM.

src/traits.jl
```diff
@@ -19,85 +11,13 @@ function dimmatch end
...
-Size(a::T) where {T<:AbstractArray} = Size(T)
-Size(::Type{SA}) where {SA <: StaticArray} = missing_size_error(SA)
-Size(::Type{SA}) where {SA <: StaticArray{S}} where {S<:Tuple} = @isdefined(S) ? Size(S) : missing_size_error(SA)
-
 Size(::Type{Adjoint{T, A}}) where {T, A <: AbstractVecOrMat{T}} = Size(Size(A)[2], Size(A)[1])
 Size(::Type{Transpose{T, A}}) where {T, A <: AbstractVecOrMat{T}} = Size(Size(A)[2], Size(A)[1])
 Size(::Type{Symmetric{T, A}}) where {T, A <: AbstractMatrix{T}} = Size(A)
 Size(::Type{Hermitian{T, A}}) where {T, A <: AbstractMatrix{T}} = Size(A)
 Size(::Type{Diagonal{T, A}}) where {T, A <: AbstractVector{T}} = Size(Size(A)[1], Size(A)[1])
 Size(::Type{<:LinearAlgebra.AbstractTriangular{T, A}}) where {T,A} = Size(A)
```
devmotion, line 19: Should this be moved to StaticArraysCore as well? Technically it is type piracy.

devmotion: There are some additional definitions below with types from base.

mateuszbaran: `StaticArraysCore` currently doesn't import `LinearAlgebra` so that's why I didn't more them. Is there a use for these methods in `StaticArraysCore` without `StaticArrays`?

I know this is type piracy but it should be harmless since `StaticArrays` in practice "owns" `Size`.

devmotion: > I know this is type piracy but it should be harmless since StaticArrays in practice "owns" Size.

That's why I assumed it is mostly technical. It would only become an issue if some other package wants to use these `Size` functionalities without depending on StaticArrays itself.

Some of the definitions below don't even use LinearAlgebra (or some other standard library), e.g., `Base.prod` or definitions for `Tuple`s.

Generally, I would assume that it's fine for StaticArraysCore to depend on LinearAlgebra as it is a standard library and I think usually any package with StaticArraysCore in the dependency tree will have LinearAlgebra in the dependency tree anyway.

mateuszbaran: Since the PR to core is already merged and tagged, let's leave that to a new PR.

fredrikekre: Basically all of StaticArrays is type piracy, e.g. `*(::SMatrix, ::SVector)` is type piracy since StaticArrays owns neither the function nor the types.

devmotion: I think this might actually not only be a theoretical problem but cause issues with generated functions: https://github.com/JuliaDiff/ForwardDiff.jl/pull/599#issuecomment-1257071309

mcabbott:
Maybe some functions like `length(::Type{StaticArraysCore.SMatrix...` need to move here too? 

Defining them later in StaticArrays seems to cause problems for `@generated` functions, here: https://github.com/JuliaDiff/ForwardDiff.jl/pull/599#issuecomment-1257070104 .

mateuszbaran:
I think moving `length` would be fine. Note though that we can't put too much functionality into SACore.jl, I'd like to see a clear use for each function/method/type that is put there so porting packages as complicated as ForwardDiff.jl may take some time.

PR #1095, opened 2022-09-20
nits from JET

thchr:
Some minor nits found with JET.

The point of changes in `src/SVector.jl` and `src/SArray.jl` is that the if-statement otherwise cannot know that the call to `.head` is meaningful (it still thinks that `ex.args[1] isa Any`). This change just splits it up and type-asserts so the knowledge that `ex isa Expr` can propagate. The code duplication is a little annoying, but I couldn't think of a good way to avoid it.

Since it's a macro it doesn't really matter - but I figured it might be nice anyway in order to reduce noise in reports by JET in the future.

### DIFF
--- a/src/SArray.jl
+++ b/src/SArray.jl
@@ -159,10 +159,13 @@ function static_array_gen(::Type{SA}, @nospecialize(ex), mod::Module) where {SA}
         args = parse_cat_ast(ex)
         return :($SA{$Tuple{$(size(args)...)}}($tuple($(escall(args)...))))
     elseif head === :comprehension
-        if length(ex.args) != 1 || !isa(ex.args[1], Expr) || ex.args[1].head != :generator
+        if length(ex.args) != 1
             error("Expected generator in comprehension, e.g. [f(i,j) for i = 1:3, j = 1:3]")
         end
         ex = ex.args[1]
+        if !isa(ex, Expr) || (ex::Expr).head != :generator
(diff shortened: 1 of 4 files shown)


mateuszbaran (approved):
It would be nice to have coverage for these error messages but other than that it looks fine.

thchr:
Would be happy to add some tests. Only... I don't know how to (if it is possible?) to create a `:comprehension` whose first argument is _not_ a `:generator`? I.e., I'm not sure what would actually fail this test. Any ideas?

mateuszbaran:
`[1 for i in 1:10 for j in 1:10]` has `:flatten` inside `:comprehension` so it would work as an example:
```julia
julia> dump(:([1 for i in 1:10 for j in 1:10]))
Expr
  head: Symbol comprehension
  args: Array{Any}((1,))
    1: Expr
      head: Symbol flatten
      args: Array{Any}((1,))
        1: Expr
          head: Symbol generator
          args: Array{Any}((2,))
            1: Expr
              head: Symbol generator
              args: Array{Any}((2,))
                1: Int64 1
                2: Expr
            2: Expr
              head: Symbol =
              args: Array{Any}((2,))
                1: Symbol i
                2: Expr
```

thchr:
Fantastic, thanks - that'll do it! Added tests in https://github.com/JuliaArrays/StaticArrays.jl/pull/1095/commits/ead793ebb742f97f8054207ea74973634b6d6f2e.

PR #1096, opened 2022-09-24
Remove faulty `isassigned` specialization

ChrisRackauckas:
```
 inserting isassigned(a::StaticArray, i::Int64...) in StaticArrays at C:\Users\mkitti\.julia\packages\StaticArrays\NOLon\src\abstractarray.jl:31 invalidated:
   backedges: 1: superseding isassigned(a::AbstractArray, i::Integer...) in Base at abstractarray.jl:563 with MethodInstance for isassigned(::AbstractMatrix, ::Int64, ::Int64) (4 children)
              2: superseding isassigned(a::AbstractArray, i::Integer...) in Base at abstractarray.jl:563 with MethodInstance for isassigned(::AbstractVecOrMat, ::Int64, ::Int64) (4 children)
   1 mt_cache
```

Yes it fixes an invalidating function, but it's probably just a good idea to follow Base's convention on this one regardless.

### DIFF
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -27,12 +27,6 @@ function Base.summary(io::IO, a, inds::Tuple{SOneTo, Vararg{SOneTo}})
     Base.showarg(io, a, true)
 end
 
-# This seems to confuse Julia a bit in certain circumstances (specifically for trailing 1's)
-@inline function Base.isassigned(a::StaticArray, i::Integer...)
-    ii = LinearIndices(size(a))[i...]
-    1 <= ii <= length(a) ? true : false
-end
-
 Base.IndexStyle(::Type{T}) where {T<:StaticArray} = IndexLinear()
 
 # Default type search for similar_type
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -28,7 +28,7 @@ function Base.summary(io::IO, a, inds::Tuple{SOneTo, Vararg{SOneTo}})
 end
 
 # This seems to confuse Julia a bit in certain circumstances (specifically for trailing 1's)
-@inline function Base.isassigned(a::StaticArray, i::Int...)
+@inline function Base.isassigned(a::StaticArray, i::Integer...)
     ii = LinearIndices(size(a))[i...]
     1 <= ii <= length(a) ? true : false
 end


thchr:
But isn't the current version consistent with Base's interface?

```jl
julia> methods(isassigned)
# 6 methods for generic function "isassigned":
[...]
[5] isassigned(a::AbstractArray, i::Integer...) in Base at abstractarray.jl:563
[...]
```

(Of course, that's precisely why there are invalidations here; I suppose I'm not clear on what you mean by "follow Base's convention" here?)

thchr:
Ah, I read it the change the wrong way around, so ignore the above.

This LGTM.

thchr (approved):

thchr:
It doesn't fix the invalidation though? On this branch, I still see:
```jl
 inserting isassigned(a::StaticArray, i::Integer...) in StaticArrays at C:\Users\tchr\.julia\packages\StaticArrays\psE8q\src\abstractarray.jl:31 invalidated:
   backedges: 1: superseding isassigned(a::AbstractArray, i::Integer...) in Base at abstractarray.jl:563 with MethodInstance for isassigned(::AbstractMatrix, ::Int64, ::Int64) (4 children)
              2: superseding isassigned(a::AbstractArray, i::Integer...) in Base at abstractarray.jl:563 with MethodInstance for isassigned(::AbstractVecOrMat, ::Int64, ::Int64) (12 children)
```
In fact, there are more children invalidated now?

ChrisRackauckas:
Those methods were invalidated before, so it looks like it doesn't end up fixing the invalidations (in all cases, i.e. where `a` is not inferred but `i` is inferred to at least `Integer`), but at least it matches the Base interface now.

thchr:
The culprits are from Base's `_print_matrix(io, @nospecialize(X::AbstractVecOrMat)` and `_show_nonempty(io, inferencebarrier(X), ...)`. I recall them being the cause of many invalidations in StaticArrays unfortunately :/

chriselrod:
What is the point of this method?

thchr:
It looks like it's just a tiny performance thing, since it would otherwise inherit from [`isassigned(a::AbstractArray, i::Integer...)`](https://github.com/JuliaLang/julia/blob/e6d99792e6dd0cffe41e98b8d642cf07e64364c5/base/abstractarray.jl#L584-L595) which is just doing a try/catch block to check.
I don't know where you'd need that performance, if at all, but, yeah...

ChrisRackauckas:
That seems like an over-optimization to me, and the comment above it seems to say the function isn't quite correct too. Just remove it?

chriselrod:
Yeah, let's delete this method.

ChrisRackauckas:
done

chriselrod (approved):

Issue #1106, opened 2022-11-05
Is it possible to map enumeration to `SVector`?


drelatgithub:
When I do the following:
```julia
v = SA[10, 20, 30]
map(x -> x[1] + x[2], enumerate(v))
```
The result is a 3-element `Vector{Int64}` of values 11, 22 and 33. Is it possible to get the result in `SVector` since the length is known at compile time (assuming that the codes above is inside a function, and that `v` has fixed but unknown length)?

Other methods I tried but not did not work:
1. I tried to use comprehension and macro `@SVector`, but it seems that in this case I need `length(v)` but its value is unknown during macro expansion.
2. I also tried `eachindex(v)` but it seems that the result is converted to `MVector` instead of `SVector`. This part seems to have been covered in https://github.com/JuliaArrays/StaticArrays.jl/issues/1032

mateuszbaran:
Yes, mapping such an enumeration should give an `SVector`. I will take a look, maybe it won't be hard to implement.

mateuszbaran:
Since StaticArrays 1.5.10 it is possible to do static mapping of enumeration using `enumerate_static` (instead of `enumerate`):
```julia
julia> using StaticArrays

julia> v = SA[10, 20, 30]
3-element SVector{3, Int64} with indices SOneTo(3):
 10
 20
 30

julia> map(x -> x[1] + x[2], enumerate_static(v))
3-element SVector{3, Int64} with indices SOneTo(3):
 11
 22
 33
```

PR #1107, opened 2022-11-05
Mapping over enumeration of a static array

mateuszbaran:
Implements #1106, at least the basic version when mapping over only that one enumeration.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.9"
+version = "1.5.10"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -58,6 +58,7 @@ export @MVector, @MMatrix, @MArray
 
 export similar_type
 export push, pop, pushfirst, popfirst, insert, deleteat, setindex
+export enumerate_static
 
 export StaticArraysCore
 
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -77,7 +77,13 @@ end
     end
 end
 
-@generated function map(f, a::Base.Enumerate{<:StaticArray})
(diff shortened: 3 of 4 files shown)


mateuszbaran:
Inlining check heuristic is failing on Julia master because there is a "call" substring in the LLVM code for a simple multiplication function. Could some LLVM expert propose a better heuristic that works with Julia master? This is the problematic code: https://github.com/JuliaArrays/StaticArrays.jl/blob/master/test/testutil.jl#L54-L75

```julia
julia> should_be_inlined(x) = x*x
should_be_inlined (generic function with 1 method)

julia> @code_llvm should_be_inlined(1)
;  @ REPL[2]:1 within `should_be_inlined`
define i64 @julia_should_be_inlined_305(i64 signext %0) #0 {
top:
  %thread_ptr = call i8* asm "movq %fs:0, $0", "=r"() #1
  %ppgcstack_i8 = getelementptr i8, i8* %thread_ptr, i64 -8
  %ppgcstack = bitcast i8* %ppgcstack_i8 to {}****
  %pgcstack = load {}***, {}**** %ppgcstack, align 8
  %ptls_field3 = getelementptr inbounds {}**, {}*** %pgcstack, i64 2
  %1 = bitcast {}*** %ptls_field3 to i64***
  %ptls_load45 = load i64**, i64*** %1, align 8
  fence syncscope("singlethread") seq_cst
  %2 = getelementptr inbounds i64*, i64** %ptls_load45, i64 2
  %safepoint = load i64*, i64** %2, align 8
  %3 = load volatile i64, i64* %safepoint, align 8
  fence syncscope("singlethread") seq_cst
; ┌ @ int.jl:88 within `*`
   %4 = mul i64 %0, %0
; └
  ret i64 %4
}
```

mateuszbaran:
If no one has anything against it I'm going to merge it soon.

fredrikekre:
I think this is somewhat strange, and also commits "semi-type-piracy" since it changes the result type of `map(_, ::Enumerate)`. In general, I don't think one should expect result to be static arrays just because you have a static array that is wrapped.

mateuszbaran:
I've reworked it to use `enumerate_static` which is the safest option.

PR #1122, opened 2023-02-05
Fix `rand` with zero-element array

### LINKED ISSUE #1070: `@SMatrix` with zero-element matrix returns a wrong type
The following `@SMatrix rand(4,0)` should return `SMatrix{4, 0, Float64, 0}`.
```julia
julia> using StaticArrays

julia> @SMatrix rand(4,0)
4×0 SMatrix{4, 0, Union{}, 0} with indices SOneTo(4)×SOneTo(0)

julia> @SMatrix rand(4,1)
4×1 SMatrix{4, 1, Float64, 4} with indices SOneTo(4)×SOneTo(1):
 0.11427209965882268
 0.9929547693983929
 0.4317791052041611
 0.7460285383058526
```

hyrodium:
This PR fixes #1070.

**Before this PR**
```julia
julia> using StaticArrays

julia> @SArray rand(2,2,0)
2×2×0 SArray{Tuple{2, 2, 0}, Union{}, 3, 0} with indices SOneTo(2)×SOneTo(2)×SOneTo(0)
```

**After this PR**
```julia
julia> using StaticArrays

julia> @SArray rand(2,2,0)
2×2×0 SArray{Tuple{2, 2, 0}, Float64, 3, 0} with indices SOneTo(2)×SOneTo(2)×SOneTo(0)
```


hyrodium:
The CI failure with Julia nightly is not relevant to this PR.

mateuszbaran:
I think this looks fine, it would just be nice to have a test for `SizedArray` for completeness.

mateuszbaran (approved):

Jollywatt:
Note that this change breaks the guarantee that `zeros(T)::T`, `rand(T)::T` and friends where `T <: StaticArray`, when the eltype is `Any`.

For example, before this change (in v1.5.12):
```julia
julia> zero(MVector{3,Any})::MVector{3,Any}
3-element MVector{3, Any} with indices SOneTo(3):
 0.0
 0.0
 0.0
```
After this change (in v1.5.13):
```julia
julia> zero(MVector{3,Any})::MVector{3,Float64} # Float64 !== Any
3-element MVector{3, Float64} with indices SOneTo(3):
 0.0
 0.0
 0.0
```

Some might argue the new behaviour is preferred, some might argue against —  either way, it should probably be mentioned.

FWIW, it broke some code of mine which initialised an `MVector` with eltype `Any` (used to contain a combination of `Int`s and `BasicSymbolic`).

Jollywatt:
Afterthought: it appears that the checks for `T == Any` in the methods affected in this PR (such as `_zeros` and `_ones`) are there to allow things like
```julia
julia> zeros(MVector{2})
2-element MVector{2, Float64} with indices SOneTo(2):
 0.0
 0.0
```
to default to `Float64`, for convenience. Why is that needed when we also have the following methods defined?
```julia
# Some more advanced constructor-like functions
@inline zeros(::Type{MVector{N}}) where {N} = zeros(MVector{N,Float64})
@inline ones(::Type{MVector{N}}) where {N} = ones(MVector{N,Float64})
```
Would it be better not to hard code
```juia
    T = eltype(SA)
    if T == Any
        T = Float64
    end
```
into those constructors?

mateuszbaran:
Good point, dispatching as you suggest makes sense. @hyrodium could you fix this issue?

PR #1129, opened 2023-02-15
Fix zero(SVector{3,Any}) etc. with multiple dispatch

hyrodium:
This PR fixes the output values commented on https://github.com/JuliaArrays/StaticArrays.jl/pull/1122#issuecomment-1429146870.

**StaticArrays.jl v1.5.12**

```julia
julia> zero(MVector{3,Any})::MVector{3,Any}
3-element MVector{3, Any} with indices SOneTo(3):
 0.0
 0.0
 0.0
```

**StaticArrays.jl v1.5.13**

```julia
julia> zero(MVector{3,Any})::MVector{3,Float64} # Float64 !== Any
3-element MVector{3, Float64} with indices SOneTo(3):
 0.0
 0.0
 0.0
```

**After this PR**

```julia
julia> using StaticArrays

julia> zero(MVector{3,Any})::MVector{3,Float64}
ERROR: MethodError: no method matching zero(::Type{Any})
```

I think the return value on v1.5.13 is problematic, but I'm not sure the return value on v1.5.12 is correct because `zero(Any)` throws an error.

### DIFF
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -1,4 +1,4 @@
-@inline zeros(::Type{SA}) where {SA <: StaticArray{<:Tuple}} = zeros(SA{Float64})
+@inline zeros(::Type{SA}) where {SA <: StaticArray{<:Tuple}} = zeros(Base.typeintersect(SA, AbstractArray{Float64}))
 @inline zeros(::Type{SA}) where {SA <: StaticArray{<:Tuple, T}} where T = _zeros(Size(SA), SA)
 @generated function _zeros(::Size{s}, ::Type{SA}) where {s, SA <: StaticArray}
     T = eltype(SA)
@@ -16,7 +16,7 @@
     end
 end
 
-@inline ones(::Type{SA}) where {SA <: StaticArray{<:Tuple}} = ones(SA{Float64})
+@inline ones(::Type{SA}) where {SA <: StaticArray{<:Tuple}} = ones(Base.typeintersect(SA, AbstractArray{Float64}))
 @inline ones(::Type{SA}) where {SA <: StaticArray{<:Tuple, T}} where T = _ones(Size(SA), SA)
 @generated function _ones(::Size{s}, ::Type{SA}) where {s, SA <: StaticArray}
     T = eltype(SA)
--- a/src/arraymath.jl
+++ b/src/arraymath.jl
@@ -35,7 +35,7 @@ end
 end
 
 @inline fill(val, ::SA) where {SA <: StaticArray{<:Tuple}} = _fill(val, Size(SA), SA)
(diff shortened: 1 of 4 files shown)


hyrodium:
@Jollywatt Do you have any thoughts on this?

src/arraymath.jl
```diff
@@ -38,13 +34,11 @@ end
     end
 end
 
-@inline fill(val, ::SA) where {SA <: StaticArray} = _fill(val, Size(SA), SA)
-@inline fill(val, ::Type{SA}) where {SA <: StaticArray} = _fill(val, Size(SA), SA)
-@generated function _fill(val::U, ::Size{s}, ::Type{SA}) where {U, s, SA <: StaticArray}
+@inline fill(val, ::SA) where {SA <: StaticArray{<:Tuple}} = _fill(val, Size(SA), SA)
+@inline fill(val::U, ::Type{SA}) where {SA <: StaticArray} where U = fill(val, SA{U})
```
Jollywatt, line 38: I don't pretend to know much about this code, but is it safe to add the type parameter `U` to `SA{U}` like that? What if `SA = SVector{N,T} where {N,T}` so that `SA{U} = SVector{U,T} where T`? Then the eltype `U` would be in the `N` slot instead of the `T` slot.

Would something like
```julia
@inline fill(val::U, ::Type{SA}) where {SA <: StaticArray{S}} where {U,S} = fill(val, SA{U})
```
at least ensure that the first type parameter of `SA` is filled, so that `SA{U}` fills the second slot?

Jollywatt:
I think having `zeros(MVector{3,Any})` call `zeros(Any)` and then error is the most correct. You can always do `zeros{MVector{3})` which defaults to `Float64`, mirroring `zeros(dims...)`.

Less correct would be to return a `MVector{3,Any}` whose elements are `Float64` zeros (or whatever the preferred fallback type is). It's correct in the sense that it returns an object of the type you asked for, but incorrect in the sense that it's not particularly good practice to have `Any` eltype, not to mention there already existing an obvious way of accomplish it with `MVector{3,Any}(zeros(MVector{3,Float64}))` if you really wanted to.

Definitely incorrect would be for `zeros(MVector{3,Any})` to return a `MVector{3,Float64}`, which is not a subtype (because Julia types are "invariant, not covariant"). The proper way to have a "default eltype" is to implement `zeros(::Type{MVector{S}}) where S = zeros(MVector{S,Float64})`.

**TL;DR** I agree with this PR.

src/arraymath.jl
mateuszbaran: I agree, applying `U` to an arbitrary subtype of `StaticArray` is not safe.

hyrodium: Thanks for the comments! I have replaced `SA{U}` with `Base.typeintersect(SA, AbstractArray{U})`.

mateuszbaran (approved):

PR #1131, opened 2023-02-17
Avoid allocation when computing norm of MVector

### LINKED ISSUE #1126: StaticArrays 1.5.13 breaks unit tests in other packages.
The latest version of StaticArrays is causing a lot of failures in the unit tests of the package KiteModels, see for example:
https://github.com/ufechner7/KiteModels.jl/actions/runs/4157073359/jobs/7191296947

The issues disappeared when downgrading to 1.5.12 as I did in the current master version of KiteModels.

One type of error are memory allocations that appear with 1.5.14, but not with 1.5.12 or older versions.

mateuszbaran:
This is one way to fix #1126, another would be to convert any static array to `SArray` before the norm is calculated. I don't have a strong preference one way or another.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.15"
+version = "1.5.16"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -226,7 +226,7 @@ _inner_eltype(x::Number) = typeof(x)
 end
 
 @inline maxabs_nested(a::Number) = abs(a)
-function maxabs_nested(a::AbstractArray)
+@inline function maxabs_nested(a::AbstractArray)
     prod(size(a)) == 0 && (return _init_zero(a))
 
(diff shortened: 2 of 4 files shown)


fredrikekre (approved):

ufechner7:
Any idea why the tests with Julia nightly fail?

mateuszbaran:
The failure on Julia nightly is actually quite amusing. A test was marked as `@test_broken`, intended to indicate that an expression result type is not inferred. At some point the issue was fixed, but the test wasn't failing (as would be expected when `@test_broken` succeeds) because the expression returned an `SVector` instead of a `Bool`, which is incorrect. So the test was always accepting it as broken for the wrong reason. Julia decided to change `@test_broken` to interpret not returning a `Bool` as a failure, so it broke the incorrect `@test_broken` we have in StaticArrays.jl.

ufechner7:
Thanks for the detailed explanation!
Now it only needs to get merged...

ufechner7:
Thanks for merging!
I tested #master, and it works.
On the other hand, on my Laptop I see a 12% performance penalty compared with the older version of StaticArrays. I will re-test on my Desktop which should give more reliable results.

mateuszbaran:
You're welcome :slightly_smiling_face: .

The new norm is slightly slower but on the other hand it is more accurate (see discussion in #975 ).

PR #1133, opened 2023-02-19
Relax recursion check in `construct_type`

### LINKED ISSUE #1118: Constructor missing?
Hello,
I have the following example:

```julia
a = SVector{1, Int64}(1)
# this works
SVector{2, Tuple{SVector{1, Int64}, SVector{1, Int64}}}((a,a),(a,a))
# this doesn't work in Julia 1.8.3 but used to work in earlier versions
SVector{1, Tuple{SVector{1, Int64}, SVector{1, Int64}}}((a,a)) 
```

Do I miss something here, or is the constructor missing?

Greetz, max

N5N3:
If `need_rewrap` returns `true`, then `x` would be re-wrapped before the next call.
Thus there's actually no stackoverflow risk in this case.
fix #1118.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.18"
+version = "1.5.19"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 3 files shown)


mateuszbaran (approved):
I think this should be fine :+1: . Could you merge master (or rebase) and bump version?

N5N3:
bumped

PR #1143, opened 2023-03-23
fix vector indexing

pepijndevos:
This fixes `mul!(@MArray([0.0]), Diagonal([1]), @MArray([-14691.819508873506]))` in line with the same change in getindex.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.19"
+version = "1.5.20"
 
 [deps]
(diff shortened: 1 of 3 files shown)


mateuszbaran:
Sorry for overlooking this, it looks fine, could you just add a test and bump version?

pepijndevos:
welp it looks like this introduced new ambiguities? What does that mean?

pepijndevos:
It seems this is a 1.10 problem so JFYI but I'll bump the version number and call it good, I don't think I introduced these ambiguities

mateuszbaran:
Don't worry about these ambiguities, it's an unrelated issue. Tests on Julia nightly don't have to pass.

mateuszbaran (approved):

PR #1146, opened 2023-03-25
Replace AbstractTriangular by union

dkarrasch:
The diff looks exactly like it should. This PR backports some minor changes to the pre-1.0 series, at which many packages are still stuck because of old compat restrictions. I'm convinced, however, that most if not all could bump their compat restriction to v1.x. Without that bump, however, all these packages will stop working on Julia v1.10, after https://github.com/JuliaLang/julia/pull/26307 is merged.

Before we merge this, let's wait for a confirmation that we're on the right track.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,8 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.0'
-          - '1.1'
-          - '1.2'
-          - '1.3'
-          - '1.4'
+          - '1.6'
+          - '1'
           - 'nightly'
         os:
           - ubuntu-latest
@@ -31,13 +28,6 @@ jobs:
           # See https://github.com/marketplace/actions/setup-julia-environment
           # MacOS not available on x86
           - {os: 'macOS-latest', arch: 'x86'}
-          # Don't test on all versions
-          - {os: 'macOS-latest', version: '1.1'}
-          - {os: 'macOS-latest', version: '1.2'}
-          - {os: 'macOS-latest', version: '1.3'}
-          - {os: 'windows-latest', version: '1.1'}
-          - {os: 'windows-latest', version: '1.2'}
-          - {os: 'windows-latest', version: '1.3'}
     steps:
       - uses: actions/checkout@v1
       - uses: julia-actions/setup-julia@latest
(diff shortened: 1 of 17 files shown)


mateuszbaran (approved):
This looks fine to me but I will wait with merging a couple of days to let people comment on this. I think people should just move to StaticArrays.jl at this point, I haven't seen a good reason not to. I guess the packages stuck on StaticArrays.jl 0.12 are just no longer maintained?

dkarrasch:
Perhaps we should include Julia v1.6 in CI on this branch.

mateuszbaran:
Yes, sure, adding Julia 1.6 to CI here is a good idea.

mateuszbaran:
Hm, it looks like a bit more needs to get backported to make the tests running on newer Julia versions. BTW, I'd still like to keep testing on Julia 1.0 on this branch.

dkarrasch:
Honestly, I think we should just merge and release. The last pre-v1 version of this package has 53 ambiguities, this PR has only 52 (at least locally). Packages that are bound to <v1 of StaticArrays.jl already need to live with those 53 ambiguities on Julia v1.6 (and perhaps beyond). The way how to resolve the ambiguities is to bump the compat restriction.

dkarrasch:
Hm, another option could be to release a v0.12.6 with a Julia compat bounded from above by `<v1.6`... Clearly, the v0.12 series has not been maintained (not even tested) on Julia v1.6+, and several tests fail as they currently stand. One could argue that the pre-v1 series is "broken" on Julia v1.6+, and should therefore not be used. The question then is how to nudge package maintainers to lift their compat for StaticArrays.jl. The test adoptions are a pain!

mateuszbaran:
Hm, yes, this doesn't look good.

> Hm, another option could be to release a v0.12.6 with a Julia compat bounded from above by `<v1.6`

As far as I can see, this doesn't solve any problem because it won't un-block changing `AbstractTriangular` in Julia 1.10+ because those packages couldn't use StaticArrays.jl 0.12.6 with newer versions of Julia.

> The question then is how to nudge package maintainers to lift their compat for StaticArrays.jl.

Probably just opening issues on github or writing e-mails to maintainers? This would really be the best solution but I doubt all of these packages are still actively maintained.

dkarrasch:
Boy, this is so broken (in terms of functionality, it can still be loaded though)... and apparently, the packages that are still stuck on pre-v1 don't need any of the broken functionality, so they don't see it. I think this can be merged and released now.

mateuszbaran:
That's indeed quite a few broken things. Let's release it though and see what happens.

PR #1150, opened 2023-04-03
Allow mutation of non-isbits eltypes

MasonProtter:
For large `MArray`s, this'll be slower than if we had a proper `MutableTuple`, but it does work and it won't screw up the garbage collector or do anything unsafe like the previous version did.

### DIFF
--- a/src/MArray.jl
+++ b/src/MArray.jl
@@ -36,9 +36,9 @@ end
     else
         # This one is unsafe (#27)
         # unsafe_store!(Base.unsafe_convert(Ptr{Ptr{Nothing}}, pointer_from_objref(v.data)), pointer_from_objref(val), i)
-        error("setindex!() with non-isbitstype eltype is not supported by StaticArrays. Consider using SizedArray.")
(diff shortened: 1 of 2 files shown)


MasonProtter:
That's so bizarre, this test works totally fine for me locally, and the operations it's doing should be legal.

serenity4:
This works for me too, perhaps we can give it another try? What were the issues? (logs are no longer available)

I made an attempt as well, with some special-casing for certain mutable types (for which I believe the concerns of #27 do not apply): https://github.com/JuliaArrays/StaticArrays.jl/compare/master...serenity4:setindex-nonisbits. That seems to greatly improve performance for large arrays:

```julia
julia> mutable struct Mut; x::Float64; end

julia> @btime setindex!(v, $("ho"), 2) setup = v = @MArray ["ha" for i in 1:100];
  408.465 ns (3 allocations: 1.62 KiB)

julia> @btime setindex!(v, $(Mut(3.0)), 2) setup = v = @MArray [Mut(rand()) for i in 1:100];
  7.032 ns (0 allocations: 0 bytes)
```

PR #1154, opened 2023-04-20
small nits from JET report

thchr:
A few nits from JET.jl: mostly just to reduce noise in future reports.

In `/src/mapreduce.jl` the issue is a missing default for a kwarg, in `/src/traits.jl` the issue is that `Val{L}` is not instantiated (i.e., ought to have been `Val{L}()`) and beyond that that there isn't even a `unroll_tuple(::Any, ::Val)` method to call.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -316,8 +316,8 @@ _mean_denom(a, ::Type{Val{D}}) where {D} = size(a, D)
 @inline maximum(f::Function, a::StaticArray; dims=:) = _mapreduce(f, max, dims, _InitialValue(), Size(a), a)
 
 # Diff is slightly different
-@inline diff(a::StaticArray; dims) = _diff(Size(a), a, dims)
-@inline diff(a::StaticVector) = diff(a;dims=Val(1))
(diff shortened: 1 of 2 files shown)


src/mapreduce.jl
```diff
@@ -316,8 +316,8 @@ _mean_denom(a, ::Type{Val{D}}) where {D} = size(a, D)
 @inline maximum(f::Function, a::StaticArray; dims=:) = _mapreduce(f, max, dims, _InitialValue(), Size(a), a)
 
 # Diff is slightly different
-@inline diff(a::StaticArray; dims) = _diff(Size(a), a, dims)
-@inline diff(a::StaticVector) = diff(a;dims=Val(1))
+@inline diff(a::StaticArray; dims=Val(1)) = _diff(Size(a), a, dims)
```
mateuszbaran, line 319: Julia Base doesn't have a default value either: https://github.com/JuliaLang/julia/blob/c068048f06619c0cfa35d76739632a5cff14836d/base/multidimensional.jl#L1013 . Maybe it was intentional?

thchr: Ah, then it is indeed probably intentional; will revert.

mateuszbaran (approved):
LGTM :+1:

Issue #1155, opened 2023-04-24
`SVector` type unstable for Julia v1.9 and `--check-bounds=no`


sloede:
As described in https://github.com/JuliaLang/julia/issues/49472, using StaticArrays.jl with Julia v1.9 and `--check-bounds=no` causes the `SVector` constructor to be type unstable even for the simplest invocations.

To reproduce, start Julia v1.9  (RC1 or later were tested and reproduce this problem) with `--check-bounds=no` and then run the following:
```julia
julia> using StaticArrays

julia> @code_warntype SVector(1)
MethodInstance for (SVector)(::Int64)
  from (::Type{SA})(x...) where SA<:StaticArray @ StaticArrays ~/hackathon/perf/mwe-staticarrays/rc1/tmp-depot/packages/StaticArrays/4uslg/src/convert.jl:160
Static Parameters
  SA = SVector
Arguments
  #self#::Type{SVector}
  x::Tuple{Int64}
Body::Any
1 ─      nothing
│   %2 = $(Expr(:static_parameter, 1))::Core.Const(SVector)
│   %3 = StaticArrays.Args(x)::StaticArrays.Args{Tuple{Int64}}
│   %4 = StaticArrays.construct_type(%2, %3)::Any
│   %5 = (%4)(x)::Any
└──      return %5
```

_Without_ `--check-bounds=no`, we get the expected, type stable output:
```julia
julia> using StaticArrays

julia> @code_warntype SVector(1)
MethodInstance for (SVector)(::Int64)
  from (::Type{SA})(x...) where SA<:StaticArray @ StaticArrays ~/hackathon/perf/mwe-staticarrays/rc1/tmp-depot/packages/StaticArrays/4uslg/src/convert.jl:160
Static Parameters
  SA = SVector
Arguments
  #self#::Type{SVector}
  x::Tuple{Int64}
Body::SVector{1, Int64}
1 ─      nothing
│   %2 = $(Expr(:static_parameter, 1))::Core.Const(SVector)
│   %3 = StaticArrays.Args(x)::StaticArrays.Args{Tuple{Int64}}
│   %4 = StaticArrays.construct_type(%2, %3)::Core.Const(SVector{1, Int64})
│   %5 = (%4)(x)::SVector{1, Int64}
└──      return %5
```

[...]

PR #1156, opened 2023-04-24
Mark typeintersect pure for Julia 1.9

vchuravy:
Fixes the concrete issue in #1155.

Julia 1.9 favors concrete-eval over constant propagation.
Furthermore the effect system was powered up and [many `@pure` annotation were removed](https://github.com/JuliaLang/julia/pull/44776).

With `--check-bounds=no` [concrete-eval is disabled](https://github.com/JuliaLang/julia/blob/8419b5042690fda143f366d09e4b0ee3f275efc7/base/compiler/abstractinterpretation.jl#L804)
and we rely entirely on const-prop, which seems to struggle whithout the `@pure` annotation.

@aviatesk would it be feasible to consider the effects for directing const-prop? We have two situations
where the reliance on concrete-eval has shown itself to be problematic. GPU code with custom method-tables,
and `--check-bounds=no`. 

In Julia 1.10 [`@pure` is deprecated](https://github.com/JuliaLang/julia/pull/48682) and the testcase in #1155
fails again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         version:
           - '1.6'
+          - '~1.9.0-0'
           - '1'
           - 'nightly'
         os:
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.21"
+version = "1.5.22"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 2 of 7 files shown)


aviatesk:
> @aviatesk would it be feasible to consider the effects for directing const-prop? 

Yes, we can do something like https://github.com/JuliaLang/julia/pull/46471 (or even set up a separate system to track const-prop' profitability). 

Having said that, `typeintersect` is not inferrable by const-prop', since the main logic is implemented by C.

vchuravy:
> Having said that, typeintersect is not inferrable by const-prop', since the main logic is implemented by C.

Right it used to have an `@pure` annotation for that reason.

ranocha (commented):
Thanks a lot for the fix, @vchuravy! Would it make sense to add some tests for this? This would basically need to call Julia again but with the appropriate command line flags since testing uses bounds checking enabled by default.

ranocha:
@vchuravy I made a PR adding some tests to your branch. Once it is merged, this PR looks good to me. Thanks a lot for investigating this issue and creating a workaround so quickly!

ranocha (approved):
This looks good to me and is a great fix for some performance problems we encounter with Julia v1.9.

mateuszbaran (approved):
LGTM. Could you just add Julia 1.9-dev to CI testing and bump patch version?

sloede:
Great, thanks a lot to everyone involved for creating a fix so fast. Once a new version of StaticArrays.jl has been released, I will run some performance benchmarks on Julia v1.9 with `--check-bounds=no` 🐎

PR #1158, opened 2023-04-27
Improve accuracy of 2x2 symmetric/Hermitian eigendecomposition

mateuszbaran:
This PR significantly improves worst-case accuracy of eigenvalues in 2x2 symmetric and Hermitian eigendecomposition. For matrix `m1 = SA[1.6590891025248637 -2.7087777909606814e-7; -2.7087777909606814e-7 1.659089317183428]`, we had
```julia
julia> norm(m1 - Array(eigen(m1)))
2.706275120796337e-10

julia> @btime eigen($m1)
  10.599 ns (0 allocations: 0 bytes)
```
and after this change:
```julia
julia> norm(m1 - Array(eigen(m1)))
1.1102224952296907e-16

julia> @btime eigen($m1)
  10.842 ns (0 allocations: 0 bytes)
```
so it's slightly slower but much more accurate.

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -159,14 +159,11 @@ end
 @inline function _eig(::Size{(2,2)}, A::LinearAlgebra.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
     a = A.data
     TA = eltype(A)
-
+    
     @inbounds if A.uplo == 'U'
         if !iszero(a[3]) # A is not diagonal
             t_half = real(a[1] + a[4]) / 2
-            d = real(a[1] * a[4] - a[3]' * a[3]) # Should be real
-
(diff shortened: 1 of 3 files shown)


mateuszbaran:
I will merge this soon if no one is against.

src/eigen.jl
```diff
@@ -159,14 +159,11 @@ end
 @inline function _eig(::Size{(2,2)}, A::LinearAlgebra.RealHermSymComplexHerm{T}, permute, scale) where {T <: Real}
     a = A.data
     TA = eltype(A)
-
+
```
hyrodium, line 162:
```suggestion

```
We don't need these spaces here.

hyrodium (approved):
LGTM! I added a minor suggestion.

PR #1161, opened 2023-05-12
Fix norm of StaticArrays with non-finite elements

### LINKED ISSUE #1135: `isapprox` returns wrong results with `nans=true`
```jl
using StaticArrays # v1.5.16

v = [1.0, NaN]
s = @SVector [1.0, NaN]

# These should be false
isapprox(v, v) # OK
isapprox(v, s) # OK
isapprox(s, v) # OK
isapprox(s, s) # OK

# But these should return true
isapprox(v, v, nans=true) # OK (no StaticArrays)
isapprox(v, s, nans=true) # wrong - returns false
isapprox(s, v, nans=true) # wrong - returns false
isapprox(s, s, nans=true) # wrong - returns false
```

gerlero:
Fixes #1135 by making the functionality introduced in #975 play nicer with arrays that contain `NaN` or `(±)Inf` elements. This removes an unexpected dependence of the return of `norm()` on the ordering of the elements in these cases, and also makes the results consistent with regular Julia arrays.

(Thanks to @rmkn85 for the help in tracing the cause of this one)

EDIT: fix typos

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.24"
+version = "1.5.25"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -231,7 +231,7 @@ end
 
     m = maxabs_nested(a[1])
     for j = 2:prod(size(a))
-        m = @fastmath max(m, maxabs_nested(a[j]))
+        m = max(m, maxabs_nested(a[j]))
     end
 
     return m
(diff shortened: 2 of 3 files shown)


gerlero:
CI failures with Julia nightly are because of #1162

gerlero:
Re the Nightly CI fails, #1162 is now fixed (I can update the PR branch if desired) but #1155 still remains.

Both appear unrelated, so I'd like to see this reviewed anyway if at all possible.

mateuszbaran (approved):
LGTM. Could you just bump patch version? I will wait a couple of days before merging in case someone has some remarks.

gerlero:
> LGTM. Could you just bump patch version? I will wait a couple of days before merging in case someone has some remarks.

Thanks! Sure; done.

hyrodium (approved):
LGTM!

Issue #1164, opened 2023-05-18
Poor type inference with `@SVector` macro and `zeros`


sjkelly:
See for example:

```
julia> using StaticArrays

julia> function test_scale(n::Integer)
           n .* @SVector zeros(6)
       end
test_scale (generic function with 1 method)

julia> using Cthulhu

julia> Cthulhu.@descend test_scale(5)
test_scale(n::Integer) @ Main REPL[4]:1
1 function test_scale(n::Int64::Integer)::SVector{6, Float64}
2     (n::Int64 .* @SVector zeros(6)::Type{SVector{6}})::SVector{6, Float64}
```

where:

```
@SVector zeros(6)::Type{SVector{6}}
```

but in `zeros` assumes `Float64`

PR #1165, opened 2023-05-18
make `@SVector` and `@SMatrix` calls default to `Float64` to match Base

sjkelly:
This makes calls to `zeros`, `ones`, `rand`, and friend default to `Float64`
similar to Base. This can be a source of type instability when used
in the single argument version, and it is not transparent as to why.
Making `Float64` the default eliminates this source of error for the user.

Closes: https://github.com/JuliaArrays/StaticArrays.jl/issues/1164


mateuszbaran (approved):
LGTM. Could you just bump patch version?

PR #1167, opened 2023-06-02
only precompile on Julia version less than v1.9

sjkelly:
In the post 1.9 era, we do not have to be prescriptive
about precompiles within the package. We can save some
overall loading time/meory use and allow down stream
packages to determine the precompilation sigantures.

### DIFF
--- a/src/StaticArrays.jl
+++ b/src/StaticArrays.jl
@@ -133,7 +133,9 @@ include("flatten.jl")
 include("io.jl")
 include("pinv.jl")
 
-include("precompile.jl")
-_precompile_()
+@static if VERSION < v"1.9"
+    include("precompile.jl")
+    _precompile_()
+end
 
 end # module


timholy:
For a package like this, I think it makes a lot of sense. Any numbers to share?

sjkelly:
Sure, here you go:

Master:
```
  3.980512 seconds (903.36 k allocations: 61.053 MiB, 1.14% gc time, 0.83% compilation time)
```
```
  0.263459 seconds (875.64 k allocations: 58.681 MiB, 16.92% gc time, 0.63% compilation time)
```

PR:
```
  2.111673 seconds (869.26 k allocations: 58.837 MiB, 2.12% gc time, 1.53% compilation time)
```
```
  0.241016 seconds (841.54 k allocations: 56.465 MiB, 18.34% gc time, 0.64% compilation time)
```

oscardssmith (approved):

timholy:
Discounting the precompilation time, it looks like it's a savings of ~20ms on load time. Since this is fairly small, there's one issue I should make sure you know: if package `Owner` (the one that defines all the required types and methods) precompiles MethodInstance `mi`, then any other packages that call that `mi` just link against it; no actual code gets stored in those "downstream" packages. In contrast, if `Owner` doesn't precompile `mi`, but it's needed by 5 downstream packages, *each* of those will precopmile their own copy of `mi`. Then when Julia loads them, each of those `mi`s gets loaded, but only the first-to-load gets used.

I explain this simply to point out that you can speed up loading of one package but the same intervention can possibly slow down later packages.

sjkelly:
@timholy That is a clear explanation, thank you. I did not know that this would be duplicated in down streams! I will rework this to instead use PrecompileTools.jl so this can be set via preferences.

PR #1170, opened 2023-06-29
remove redundant `transpose`/`adjoint` methods for `Lower/UpperTriangular`

thchr:
@dkarrasch pointed out that these methods [have existed in Base](https://github.com/JuliaLang/julia/blob/7eb358e79967efb9a07b53f592c329139da7706a/stdlib/LinearAlgebra/src/triangular.jl#L399-L406C75) since at least v1.6, so there's no reason for them to be duplicated here (and StaticArrays assumes v1.6 already).


thchr:
The nightly CI errors seem to be the same as in other recent PRs (e.g., https://github.com/JuliaArrays/StaticArrays.jl/pull/1165/checks) and so probably unrelated.

thchr:
Somehow, I managed to merge this twice? Not sure how that is possible, but GitHub was seemingly having some trouble loading the page at the time I was merging.

I assume (and hope) this doesn't affect anything though, aside from looking silly in the log.

Issue #1171, opened 2023-06-29
Change (or drop) Statistics to package extension


thchr:
StaticArrays depends on Statistics which depends on SparseArrays. But SparseArrays is no longer in the sysimage, and this means it now contributes significantly (~70%) to the loading time of StaticArrays:
```jl
julia> @time_imports using StaticArrays
               ┌ 9.1 ms SparseArrays.CHOLMOD.__init__() 73.00% compilation time
    425.4 ms  SparseArrays 16.16% compilation time
      0.6 ms  Statistics
      0.7 ms  StaticArraysCore
    185.3 ms  StaticArrays
```

But Statistics is only used in order to extend (performance optimize?) `mean`, cf.:
https://github.com/JuliaArrays/StaticArrays.jl/blob/376efdbf3bbab34d624dfbee88a9ae83282f0718/src/StaticArrays.jl#L11
https://github.com/JuliaArrays/StaticArrays.jl/blob/376efdbf3bbab34d624dfbee88a9ae83282f0718/src/mapreduce.jl#L304-L310

So we could either use the new package extension functionality to load Statistics only conditionally - or we could entirely drop this (it's not entirely clear to me why it wouldn't be better to leave this to the [default `mean` implementation](https://github.com/JuliaStats/Statistics.jl/blob/bb7063d77a47e3fdc782d33b586d02c12908791e/src/Statistics.jl#L179) which e.g. dispatches to `mapreduce`)? 

Per discussion w/ @dkarrasch.

thchr:
In fact, it seems that there is no noticeable performance advantage to using the extensions of `mean` from StaticArrays:

```jl
using StaticArrays, Statistics, BenchmarkTools
static_mean(A; kws...)    = mean(A; kws...)
static_mean(f, A; kws...) = mean(f, A; kws...)
base_mean(A; kws...)      = @invoke mean(A::AbstractArray; kws...)
base_mean(f, A; kws...)   = @invoke mean(f, A::AbstractArray; kws...)

S2 = @SMatrix rand(2,2)
S3 = @SMatrix rand(3,3)
M2 = @MMatrix rand(2,2)
M3 = @MMatrix rand(3,3)

@btime static_mean($S2) # 1.741 ns (0 allocations: 0 bytes)
@btime base_mean($S2)   # 1.985 ns (0 allocations: 0 bytes)

@btime static_mean($S3) # 2.004 ns (0 allocations: 0 bytes)
[...]

PR #1172, opened 2023-06-30
Handle Statistics dep as Pkg extension

dkarrasch:
I think this is a no-brainer. It doesn't cost anything in terms of functionality, but reduces package load time significantly on Julia v1.9+. One could still decide later on to remove this or solve this via fallbacks etc., see #1171.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,12 +1,18 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.26"
+version = "1.5.27"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[weakdeps]
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[extensions]
+StaticArraysStatisticsExt = "Statistics"
 
 [compat]
 julia = "1.6"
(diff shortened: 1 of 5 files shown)


Project.toml
```diff
@@ -1,13 +1,19 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.5.26"
+version = "1.5.27"
```
fredrikekre, line 3:
```suggestion
version = "1.6.0"
```
perhaps?

ext/StaticArraysStatisticsExt.jl
```diff
@@ -0,0 +1,16 @@
...
+
+import Statistics: mean
+
+using StaticArrays
+using StaticArrays: _InitialValue, _reduce, _mapreduce
+
+_mean_denom(a, ::Colon) = length(a)
+_mean_denom(a, dims::Int) = size(a, dims)
+_mean_denom(a, ::Val{D}) where {D} = size(a, D)
+_mean_denom(a, ::Type{Val{D}}) where {D} = size(a, D)
```
thchr, line 11: I realized this isn't actually used anywhere, afaict.
```suggestion
```

thchr:
I think we should probably still let `mean` work on earlier Julia versions too. We could use Requires for that (like [here](https://github.com/thchr/StaticArrays.jl/commit/1cf071f4c467d697879801424f6c78a445c61fb9)): it wouldn't load Requires on later versions.

That aside, I agree this is the best approach and worthwhile.

dkarrasch:
We don't need Requires to do that. `mean` just works on older Julia versions. On older versions, the weak dependency is interpreted as a "normal" dependency, as it is right now. That's why it is still listed as a dependency. On older versions, the weak dependencies are ignored, on newer versions normal dependencies are ignored if they are also listed as a weak dependency. IIUC @KristofferC

KristofferC:
LGTM

thchr:
> We don't need Requires to do that. `mean` just works on older Julia versions. On older versions, the weak dependency is interpreted as a "normal" dependency, as it is right now. That's why it is still listed as a dependency. On older versions, the weak dependencies are ignored, on newer versions normal dependencies are ignored if they are also listed as a weak dependency. IIUC @KristofferC

Ah, I hadn't noticed the `include` on earlier version: thanks for explaining.

thchr (approved):

PR #1174, opened 2023-07-11
help Julia with type inference by providing `ntuple` with `Val`

nsajko:
As per https://github.com/JuliaArrays/StaticArraysCore.jl/pull/21


mateuszbaran (approved):
LGTM. Could you bump patch version?

PR #1175, opened 2023-07-12
make conversion of `Length` to `Int` safer

nsajko:
Now `Int(Length{StaticArraysCore.Dynamic()}())` throws a `MethodError` instead of returning a non-`Int` value.


nsajko:
I came upon this while removing `@pure`s. The `@pure` on that line is not necessary, I can remove it now or in a future PR where more `@pure`s will be deleted additionally.

mateuszbaran (approved):
LGTM. Could you bump the patch version?

PR #1185, opened 2023-07-17
fix typos

spaette:
src/lu.jl
test/inv.jl
test/linalg.jl
test/pinv.jl


spaette:
The change in `test/pinv.jl` was that of a test name.

Issue #1019, opened 2022-04-03
Error when @SMatrix randn(rng,ComplexF64,(2,2))


jisutich:
I try to construct static random matrix by using `@SMatrix`. And `@SMatrix randn(ComplexF64,2,2)` works. But when try to further seed the random number generator like 
```
using StatsBase
using Statistics
using Random
using StaticArrays

rng = MersenneTwister(123)
@SMatrix randn(rng,ComplexF64,(2,2))
```
I got the error
```
ERROR: TypeError: in Type, in parameter, expected Type, got a value of type MersenneTwister
```

hyrodium:
You can just call `randn` function without the `@SMatrix` macro.
```julia
julia> rng = MersenneTwister(123)
MersenneTwister(123)

julia> randn(rng, SMatrix{2,2,ComplexF64})
2×2 SMatrix{2, 2, ComplexF64, 4} with indices SOneTo(2)×SOneTo(2):
 0.841646+1.44828im   -0.280494-0.470023im
 0.807976+0.324856im   0.693649-0.0533746im
```

x-ref: https://github.com/JuliaArrays/StaticArrays.jl/issues/391

jisutich:
Yes this works. Thanks a lot!

PR #1210, opened 2023-10-26
Extend macros with rand to support custom samplers

### LINKED ISSUE #804: Generating random StaticArrays works with SamplerTypes but not other Samplers
Here's an example taken from the Julia docs on random number generation [here](https://docs.julialang.org/en/v1/stdlib/Random/#Generating-values-from-a-type-1)
```julia
struct Die
    nsides::Int
end
Random.rand(rng::AbstractRNG, ::Random.SamplerType{Die}) = Die(rand(rng, 4:20))
Random.rand(rng::AbstractRNG, d::Random.SamplerTrivial{Die}) = rand(rng, 1:d[].nsides)
Base.eltype(::type{Die}) = Int
```
This lets us conveniently generate arrays of random dice,
```julia
julia> rand(Die,3)
3-element Array{Die,1}:
 Die(11)
 Die(17)
 Die(4)

```
or arrays of random dice rolls
```julia
julia> rand(Die(6), 3)
3-element Array{Int64,1}:
 3
 1
 5

```
With StaticArrays we can generate random dice
```julia
julia> @SVector rand(Die, 3)
3-element SArray{Tuple{3},Die,1,3} with indices SOneTo{3}:
 Die(20)
 Die(10)
 Die(13)

```
but not random dice rolls
```julia
julia> @SVector rand(Die(6), 3)
ERROR: ArgumentError: Static Array parameter T must be a type, got Die(6)
Stacktrace:
...
```
The same error occurs when using the samplers from Distributions.jl, for example.

mateuszbaran:
Fixes #804 .


hyrodium:
I have some questions about the implementation.

* How can we create `SizedArray` version of `rand(1:5, 3)`?
* How can we create something like `@SVector rand(rng, 1:5, 3)`? (x-ref: https://github.com/JuliaArrays/StaticArrays.jl/issues/1019#issuecomment-1086898833)

These are why I proposed `strand` function in the comment. (https://github.com/JuliaArrays/StaticArrays.jl/issues/804#issuecomment-1780880103)

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.6.5"
+version = "1.6.6"
```
hyrodium, line 3:
```suggestion
version = "1.7.0"
```
This PR is a new feature "supports custom samplers" rather than a bug fix, so I think it would be better to bump the minor version.

mateuszbaran:
> * How can we create `SizedArray` version of `rand(1:5, 3)`?

I'd just do `SizedVector{3}(rand(1:5, 3))`.

> * How can we create something like `@SVector rand(rng, 1:5, 3)`? (x-ref: [Error when @SMatrix randn(rng,ComplexF64,(2,2)) #1019 (comment)](https://github.com/JuliaArrays/StaticArrays.jl/issues/1019#issuecomment-1086898833))
> 
> These are why I proposed `strand` function in the comment. ([#804 (comment)](https://github.com/JuliaArrays/StaticArrays.jl/issues/804#issuecomment-1780880103))

I don't see this PR as a full replacement of `strand`, just a quick fix for the most commonly requested use case. If you want to work on `strand`, I can review your PR.

test/arraymath.jl
```diff
@@ -1,6 +1,13 @@
 using StaticArrays, Test
 import StaticArrays.arithmetic_closure
 
+struct TestDie
+    nsides::Int
+end
+Random.rand(rng::AbstractRNG, ::Random.SamplerType{TestDie}) = TestDie(rand(rng, 4:20))
```
hyrodium, line 7:
```suggestion
```
This method can be removed because we don't use `rand(TestDie)` in the test.

src/SVector.jl
```diff
@@ -78,7 +78,12 @@ function static_vector_gen(::Type{SV}, @nospecialize(ex), mod::Module) where {SV
             if length(ex.args) == 2
                 return :($f($SV{$(esc(ex.args[2])), Float64})) # default to Float64 like Base
             elseif length(ex.args) == 3
-                return :($f($SV{$(escall(ex.args[3:-1:2])...)}))
+                if f === :rand && ex.args[3] isa Int && ex.args[3] > 0
+                    # supports calls like rand(Type, n) and rand(sampler, n)), but only if n > 0
+                    return :(_rand(Random.GLOBAL_RNG, $(esc(ex.args[2])), Size($(esc(ex.args[3]))), $SV{$(esc(ex.args[3]))}))
```
hyrodium, line 83: How about adding the type of the element of the vector?
```suggestion
                if f === :rand && ex.args[3] isa Int && ex.args[3] ≥ 0
                    # supports calls like rand(Type, n) and rand(sampler, n))
                    return :(_rand(Random.GLOBAL_RNG, $(esc(ex.args[2])), Size($(esc(ex.args[3]))), $SV{$(esc(ex.args[3])), Random.gentype($(esc(ex.args[2])))}))
```
This change supports a 0-element vector like this.
```julia
julia> @SVector rand(1:3, 0)
0-element SVector{0, Int64} with indices SOneTo(0)
```

I'm not sure which `Random.gentype` or `eltype` would be preferred. [The documentation](https://docs.julialang.org/en/v1/stdlib/Random/#A-simple-sampler-without-pre-computed-data) recommends adding a method to `eltype`, but the docstring of `eltype` does not say anything about samplers, and `Random.gentype` is not well-documented and may not be stable API. However, perhaps some other packages may define their `gentype` method and the result is different from `eltype`.

I think the purposes of these functions are different; `eltype` for iterators, `gentype` for samplers.
It would be better to make the `gentype` function stable, and encourage people to define their `gentype` method (not `eltype`) to create their sampler.

mateuszbaran: OK, it should work I think.

src/SMatrix.jl
```diff
@@ -73,7 +73,12 @@ function static_matrix_gen(::Type{SM}, @nospecialize(ex), mod::Module) where {SM
             if length(ex.args) == 3
                 return :($f($SM{$(escall(ex.args[2:3])...), Float64})) # default to Float64 like Base
             elseif length(ex.args) == 4
-                return :($f($SM{$(escall(ex.args[[3,4,2]])...)}))
+                if f === :rand && ex.args[3] isa Int && ex.args[3] > 0 && ex.args[4] isa Int && ex.args[4] > 0
+                    # supports calls like rand(Type, n, m) and rand(sampler, n, m)), but only if n, m > 0
+                    return :(_rand(Random.GLOBAL_RNG, $(esc(ex.args[2])), Size($(esc(ex.args[3])), $(esc(ex.args[4]))), $SM{$(esc(ex.args[3])), $(esc(ex.args[4]))}))
```
hyrodium, line 78:
```suggestion
                if f === :rand && ex.args[3] isa Int && ex.args[3] ≥ 0 && ex.args[4] isa Int && ex.args[4] ≥ 0
                    # supports calls like rand(Type, n, m) and rand(sampler, n, m))
                    return :(_rand(Random.GLOBAL_RNG, $(esc(ex.args[2])), Size($(esc(ex.args[3])), $(esc(ex.args[4]))), $SM{$(esc(ex.args[3])), $(esc(ex.args[4])), Random.gentype($(esc(ex.args[2])))}))
```

hyrodium (commented):
Could you please also update `@SArray` and `@MArray` macros? Other parts LGTM!

mateuszbaran:
> Could you please also update `@SArray` and `@MArray` macros? Other parts LGTM!

Sure. When working on it, I had an idea for supporting passing `rng` as well. What do you think?

hyrodium:
Great!!:rocket: But I found some degression.

**Before this PR**
```julia

julia> using StaticArrays

julia> @SArray rand(3)
 0.3872905952155763
 0.9519315858388775
 0.42729797815514925
```

**After this PR**
```julia
julia> using StaticArrays

julia> @SArray rand(3)
ERROR: LoadError: BoundsError: attempt to access 2-element Vector{Any} at index [3]
Stacktrace:
```

If we can handle `rng` in `@SArray`, it would be nice to have these features in `@SVector` and `@SMatrix`.

mateuszbaran:
OK, I think I've fixed that issue and also added `rng` support to `SVector` and `SMatrix`.

src/SArray.jl
```diff
@@ -201,12 +201,43 @@ function static_array_gen(::Type{SA}, @nospecialize(ex), mod::Module) where {SA}
...
                 return :($f($SA{$Tuple{},$Float64}))
-            end
-            return quote
-                if isa($(esc(ex.args[2])), DataType)
-                    $f($SA{$Tuple{$(escall(ex.args[3:end])...)},$(esc(ex.args[2]))})
-                else
-                    $f($SA{$Tuple{$(escall(ex.args[2:end])...)}})
+            elseif f !== :rand || length(ex.args) == 2
+                return quote
+                    if isa($(esc(ex.args[2])), DataType)
```
stevengj, line 206: Rather than dispatching with an `if` statement, it seems a lot easier to simply rewrite to a custom function call and then using ordinary dispatch.   See my comment https://github.com/JuliaArrays/StaticArrays.jl/issues/804#issuecomment-1834474456

I used a similar strategy in ChangePrecision.jl — https://github.com/JuliaMath/ChangePrecision.jl/blob/1efe76bfa9e4bf47f9bfe247bc6fe5568ddad64b/src/ChangePrecision.jl#L20-L24 — and it's very pleasant and flexible in my experience.

mateuszbaran: Thanks, that's a very good suggestion.

mateuszbaran: @hyrodium implemented it in #1213

hyrodium (approved):

PR #1224, opened 2023-12-09
Add ChainRules Overloads to StaticArrays

avik-pal:
These rules were a part of SciMLSensitivity https://github.com/SciML/SciMLSensitivity.jl/blob/f1aa09244e45598874e1a14a008356a2b2b241a1/src/staticarrays.jl#L23

I will add some tests. But any objections to moving them here?

cc @ChrisRackauckas

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,6 @@ uuid = "90137ffa-7385-5640-81b9-e52037218182"
 version = "1.7.1"
 
 [deps]
-ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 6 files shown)


avik-pal:
It seems there is an older PR https://github.com/JuliaArrays/StaticArrays.jl/pull/1068

I have moved the dependency into a weakdep and added more minimal tests based on the discussion there. Let me know what else needs to be done.

ChrisRackauckas (approved):

oxinabox:
If someone wants to add this to our Downstream tests in ChainRulesCore CI we can make sure not to accidentally break it

PR #1231, opened 2024-01-12
Bring over tests from ChainRulesCore

avik-pal:
See https://github.com/JuliaArrays/StaticArrays.jl/pull/1228.

I am rearranging the fields a bit to ensure Transpose works.


oxinabox (approved):
TY

Issue #1244, opened 2024-02-14
Force specialization of `dimension_mismatch_fail()` on type parameter? 


trahflow:
Currently the function [`dimension_mismatch_fail(SA::Type, a::AbstractArray)`](https://github.com/JuliaArrays/StaticArrays.jl/blob/ff8ed9b64f41833fe22af81bd2859d5a0b9c3d18/src/convert.jl#L195C1-L197C4) does not specialize on the `SA` parameter, which makes code that uses StaticArrays inside GPU kernels incompatible (unless @inbounds is used in front of the constructor/convert statements).
See the discussion here: https://discourse.julialang.org/t/problems-with-linearalgebra-functions-within-kernelabstractions-and-cuda/97566/5

Would it be ok to force specialization on the `SA` parameter? I could provide a quick PR.

mateuszbaran:
Resolved by #1245

PR #1245, opened 2024-02-20
Force specialization of dimension_mismatch_fail() on type argument.

trahflow:
Following #1244 I just went ahead with this small change.

This should improve GPU-compatibility.
Original discussion: https://discourse.julialang.org/t/problems-with-linearalgebra-functions-within-kernelabstractions-and-cuda/97566/5


mateuszbaran:
If you bump the patch version, I could tag it just after merging and CI.

PR #1246, opened 2024-02-22
NamedTuple constructor support for FieldArray subtypes

cadojo:
This PR was originally posted to [`StaticArraysCore.jl`](https://github.com/JuliaArrays/StaticArraysCore.jl/pull/25), but was moved to here because `StaticArraysCore` lacks the method definitions necessary to make this `NamedTuple` functionality work. 

I have found `NamedTuple` to be a performant way to initialize `FieldArray` subtypes. See, for example, the following code in [`AstrodynamicalModels.jl`](https://github.com/cadojo/AstrodynamicalModels.jl/blob/8c811923c2d56b7927bf4c62acc2fbfa1bcd91d4/src/AstrodynamicalModels.jl#L107). I think it may also be useful sometimes to go in reverse: convert a `FieldArray` subtype back into a `NamedTuple`. **Do you agree this may be useful, and without undesirable side effects?** I got this idea from `SciML`'s [`Base.NamedTuple` method](https://github.com/SciML/LabelledArrays.jl/blob/d652e95b08b28260d36ca1ce34af1d02fd5506d0/src/LabelledArrays.jl#L59) for `LabelledArrays.jl`. Thanks for reading!

```julia
"""
A mutable vector, with labels, for 6DOF Cartesian states.
"""
Base.@kwdef mutable struct CartesianState{F} <: FieldVector{6,F}
    x::F = 0.0
    y::F = 0.0
    z::F = 0.0
    ẋ::F = 0.0
    ẏ::F = 0.0
    ż::F = 0.0

    CartesianState{F}(::UndefInitializer) where {F} = new{F}()
    CartesianState(::UndefInitializer) = CartesianState{Float64}(undef)

    CartesianState{F}(x, y, z, ẋ, ẏ, ż) where {F} = new{F}(x, y, z, ẋ, ẏ, ż)
    CartesianState(x, y, z, ẋ, ẏ, ż) = new{promote_type(typeof(x), typeof(y), typeof(z), typeof(ẋ), typeof(ẏ), typeof(ż))}(x, y, z, ẋ, ẏ, ż)
    CartesianState{F}(state::NamedTuple) where {F} =
        let
            (; x, y, z, ẋ, ẏ, ż) = merge((; x=zero(F), y=zero(F), z=zero(F), ẋ=zero(F), ẏ=zero(F), ż=zero(F)), state)
            CartesianState{F}(x, y, z, ẋ, ẏ, ż)
        end
    CartesianState(state::NamedTuple) = CartesianState{Float64}(state)
end
```

### DIFF
--- a/src/FieldArray.jl
+++ b/src/FieldArray.jl
@@ -31,3 +31,7 @@ similar_type(::Type{A}, ::Type{T}, S::Size) where {N, T, A<:FieldArray{N, T}} =
 _fieldarray_similar_type(A, T, NewSize::S, OldSize::S) where {S} = A
 _fieldarray_similar_type(A, T, NewSize, OldSize) =
     default_similar_type(T, NewSize, length_val(NewSize))
+
+# Convenience constructors for NamedTuple types 
+Base.convert(::Type{NamedTuple}, array::FieldArray) = Base.NamedTuple(array)
+Base.NamedTuple(array::FieldArray) = Base.NamedTuple{propertynames(array)}(array)
--- a/src/FieldArray.jl
+++ b/src/FieldArray.jl
@@ -33,5 +33,4 @@ _fieldarray_similar_type(A, T, NewSize, OldSize) =
     default_similar_type(T, NewSize, length_val(NewSize))
 
 # Convenience constructors for NamedTuple types 
-Base.convert(::Type{NamedTuple}, array::FieldArray) = Base.NamedTuple(array)
 Base.NamedTuple(array::FieldArray) = Base.NamedTuple{propertynames(array)}(array)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.9.2"
+version = "1.9.3"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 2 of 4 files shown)


mateuszbaran:
I think this is useful.

mateuszbaran:
Could you just add a test for `convert` and bump patch version?

src/FieldArray.jl
```diff
@@ -31,3 +31,7 @@ similar_type(::Type{A}, ::Type{T}, S::Size) where {N, T, A<:FieldArray{N, T}} =
 _fieldarray_similar_type(A, T, NewSize::S, OldSize::S) where {S} = A
 _fieldarray_similar_type(A, T, NewSize, OldSize) =
     default_similar_type(T, NewSize, length_val(NewSize))
+
+# Convenience constructors for NamedTuple types 
+Base.convert(::Type{NamedTuple}, array::FieldArray) = Base.NamedTuple(array)
```
fredrikekre, line 36: I don't think this is a good idea since `convert` is called implicitly in some cases. Note also that `convert(::Tuple, ::Array)` isn't defined either, which would be similar to this.

mateuszbaran: OK, then let's only define the constructor here.

cadojo: Done!

cadojo:
Is this okay as-is, despite the `julia-nightly` failure? I am assuming that is a Julia development branch quirk and not a problem with this branch.

mateuszbaran (approved):

mateuszbaran:
There is no julia-nighty failure on the latest commit but even if there was, it would be fine as long as the failure is not related to changes in the PR.

Issue #1263, opened 2024-06-23
Correctness issues with SHermitianCompact


lxvm:
I've been trying to use `SHermitianCompact` to speed up some interpolation and linear algebra, but I find it quite a difficult type to work with because it doesn't actually guarantee its values are Hermitian. The simplest example is multiplying a `SHermitianCompact` matrix by a complex number, which will generically result in a non-Hermitian matrix.
```julia
julia> using StaticArrays, LinearAlgebra

julia> y = rand(SHermitianCompact{3,Float64,6});

julia> x = rand(ComplexF64);

julia> norm(y*x - SMatrix(y)*x)
2.1996254891748803

julia> using Pkg

julia> Pkg.status("StaticArrays")
Status `/local/home/lxvm/projects/autobz/Project.toml`
  [90137ffa] StaticArrays v1.9.5
```
A summary of the problems I've encountered include:
* (As shown above) multiplication of a `SHermitianCompact` by a complex number should promote to a `SMatrix`
* The diagonal of a `SHermitianCompact` should always be real-valued, so that the `ishermitian` method, which is hard-coded to return true, is correct
* The `issymmetric` method should be true when the element type is real, and otherwise there should be a numerical check

I think I can correct all of the above without changing the storage layout of `SHermitianCompact` in a small PR. Do the proposed changes seem reasonable and am I missing any others?

mateuszbaran:
Multiplication by complex numbers indeed looks bugged. A small PR with a fix would be appreciated :slightly_smiling_face: .

PR #1264, opened 2024-06-24
Complex multiplication with SHermitianCompact

lxvm:
This pr adds some methods to dispatch complex scalar multiplication/division with `SHermitianCompact` to a correct method using `SMatrix`. Tests are also added. For other details see #1263

### DIFF
--- a/src/SHermitianCompact.jl
+++ b/src/SHermitianCompact.jl
@@ -151,21 +151,25 @@ LinearAlgebra.issymmetric(a::SHermitianCompact) = true
     end
 end
 
-@inline Base.:*(a::Number, b::SHermitianCompact) = SHermitianCompact(a * b.lowertriangle)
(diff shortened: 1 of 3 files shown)


src/SHermitianCompact.jl
```diff
@@ -153,9 +153,19 @@ end
 
 @inline Base.:*(a::Number, b::SHermitianCompact) = SHermitianCompact(a * b.lowertriangle)
 @inline Base.:*(a::SHermitianCompact, b::Number) = SHermitianCompact(a.lowertriangle * b)
+@inline Base.:*(a::Complex, b::SHermitianCompact) = a * SMatrix(b)
+@inline Base.:*(a::SHermitianCompact, b::Complex) = SMatrix(a) * b
```
mateuszbaran, line 157: Maybe it would be better to make your code the default for `Number` and return `SHermitianCompact` only when the number argument is `Real`.

mateuszbaran (approved):
LGTM. Could you just bump patch version in Project.toml?

PR #1266, opened 2024-06-25
improved definitions of ishermitian/issymmetric

lxvm:
As mentioned in #1263, the `SHermitianCompact` type can represent matrices which are neither Hermitian nor symmetric, so this pr fixes the definitions of `ishermitian` and `issymmetric` to capture this and adds tests to verify the logic on some of these unusual cases. I think this would be preferable to making `SHermitianCompact` guarantee that it is in fact Hermitian, i.e. by always returning real values on the diagonal, because the latter would be breaking.

### DIFF
--- a/src/SHermitianCompact.jl
+++ b/src/SHermitianCompact.jl
@@ -1,9 +1,11 @@
 """
     SHermitianCompact{N, T, L} <: StaticMatrix{N, N, T}
 
-A [`StaticArray`](@ref) subtype that represents a Hermitian matrix. Unlike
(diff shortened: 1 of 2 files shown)


mateuszbaran:
LGTM. I wouldn't necessarily add a guarantee that `SHermitianCompact` is hermitian but I think we should fix rand for this type so that it only returns hermitian matrices.

mateuszbaran:
OK, let's keep it as-is then.

mateuszbaran (approved):

PR #1267, opened 2024-06-26
refactor inv to also work for lu

lxvm:
This pr refactors the `inv` code so that `inv(::StaticArrays.LU)` behaves like `inv(::LinearAlgebra.LU)`, allowing the construction of the inverse from the LU factorization. I'm not sure if I handled inlining correctly and I would be happy to implement any feedback.


mateuszbaran (approved):
LGTM. Could you just bump patch version?

Issue #1119, opened 2023-01-16
Support `init` keyword of `sum`


ranocha:
StaticArrays.jl does not seem to support the keyword `init` of `sum`. It would be nice to be able to use this Base interface.

With Julia v1.8 and StaticArrays v1.5.11, I get

```julia
julia> using StaticArrays

julia> x = SVector(1, 2)
2-element SVector{2, Int64} with indices SOneTo(2):
 1
 2

julia> sum(identity, x; init = 0)
ERROR: MethodError: no method matching sum(::typeof(identity), ::SVector{2, Int64}; init=0)
Closest candidates are:
  sum(::Union{Function, Type}, ::StaticArray{<:Tuple, T}; dims) where T at ~/.julia/packages/StaticArrays/B0HhH/src/mapreduce.jl:283 got unsupported keyword argument "init"
  sum(::Any, ::StaticArray{<:Tuple, T}; dims) where T at ~/.julia/packages/StaticArrays/B0HhH/src/mapreduce.jl:282 got unsupported keyword argument "init"
  sum(::Any, ::AbstractArray; dims, kw...) at reducedim.jl:995
  ...
Stacktrace:
 [1] kwerr(::NamedTuple{(:init,), Tuple{Int64}}, ::Function, ::Function, ::SVector{2, Int64})
   @ Base ./error.jl:165
 [2] top-level scope
   @ REPL[3]:1

help?> sum
search: sum sum! summary cumsum cumsum! isnumeric VersionNumber issubnormal get_zero_subnormals set_zero_subnormals

  sum(f, itr; [init])

  Sum the results of calling function f on each element of itr.

  The return type is Int for signed integers of less than system word size, and UInt for unsigned integers of less
  than system word size. For all other arguments, a common return type is found to which all arguments are
  promoted.

  The value returned for empty itr can be specified by init. It must be the additive identity (i.e. zero) as it is
  unspecified whether init is used for non-empty collections.

  │ Julia 1.6
  │
  │  Keyword argument init requires Julia 1.6 or later.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> sum(abs2, [2; 3; 4])
  29

  Note the important difference between sum(A) and reduce(+, A) for arrays with small integer eltype:

  julia> sum(Int8[100, 28])
  128
  
  julia> reduce(+, Int8[100, 28])
  -128

  In the former case, the integers are widened to system word size and therefore the result is 128. In the latter
[...]

PR #1281, opened 2024-10-18
Let `count`, `sum`, and `prod` take an `init` kwarg

thchr:
Allows StaticArrays' `count` implementation to take an `init` kwarg, just as `Base.count` (since Julia v1.6).

Also bumps StaticArrays to v1.9.8, to enable releasing the fix.

EDIT: Updated to also add `init` to `prod` and `sum`, fixing #1119 in the process.

### DIFF
--- a/.github/workflows/Invalidations.yml
+++ b/.github/workflows/Invalidations.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: julia-actions/setup-julia@v2
       with:
-        version: '1'
+        version: 'lts'
     - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


mateuszbaran (approved):
LGTM. The invalidation action is failing but that appears to be due to some changes in Julia 1.11. Maybe we should run that action on Julia 1.10 for now or report the issue somewhere?

thchr:
Do you want me to change CI to do that here?

mateuszbaran:
Yes, I think it would be a good idea to change CI script to `lts` here.

PR #1284, opened 2024-10-29
fix `mul_generic` implementation

aviatesk:
`a` isn't defined globally, thus it should not be interpolated.


mateuszbaran (approved):
LGTM

PR #1289, opened 2024-12-08
define propertynames() to be consistent with getproperty()

aplavin:


src/MVector.jl
```diff
@@ -18,6 +18,8 @@ end
 let dimension_names = QuoteNode.([:x, :y, :z, :w])
     body = :(getfield(v, name))
     for (i,dim_name) in enumerate(dimension_names)
+        @eval @inline Base.propertynames(v::Union{SVector{$i},MVector{$i}}) = ($(first(dimension_names, i)...),)
```
nsajko, line 21: `propertynames` has two methods, I think both have to be added:

```suggestion
        @eval @inline Base.propertynames(v::Union{SVector{$i},MVector{$i}}, ::Bool = false) = ($(first(dimension_names, i)...),)
```

nsajko: https://docs.julialang.org/en/v1/manual/interfaces/#man-instance-properties

nsajko:
An issue with this PR is that it only overloads `propertynames` when `getproperty` is overloaded, leading to inconsistency. `:data ∉ propertynames(SVector(1, 2, 3, 4))`, but `:data ∈ propertynames(SVector(1, 2, 3, 4, 5))`. Either have both calls include `:data` or neither IMO.

src/MVector.jl
```diff
@@ -18,6 +18,9 @@ end
 let dimension_names = QuoteNode.([:x, :y, :z, :w])
     body = :(getfield(v, name))
     for (i,dim_name) in enumerate(dimension_names)
+        @eval @inline Base.propertynames(v::Union{SVector{$i},MVector{$i}}, private::Bool = false) =
+            private ? ($(first(dimension_names, i)...), :data) : ($(first(dimension_names, i)...),)
```
nsajko, line 22: FWIW IMO it's nicer to deduplicate like this:

```suggestion
        @eval @inline function Base.propertynames(v::Union{SVector{$i},MVector{$i}}, private::Bool = false)
            named_dims = $(first(dimension_names, i)...)
            private ? (named_dims..., :data) : named_dims
        end
```

nsajko (approved):

mateuszbaran (approved):
LGTM. Could you just bump patch version? I will merge and tag.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.9.8"
+version = "1.9.9"
```
mateuszbaran, line 3:
```suggestion
version = "1.9.11"
```
We're already at 1.9.10.

PR #1300, opened 2025-03-28
inv, size, and \ for QR objects

### LINKED ISSUE #1192: Cannot use qr factorization with documented functionalities
The docs when using `?StaticArrays.qr` in the Repl writes that: `The following functions are available for the QR objects: inv, size, and \.`
However, this seems to not be true for the array types in StaticArrays.

```julia
using StaticArrays

A = randn(4, 3)
b = randn(4)
StaticArrays.qr(A)\b # works fine

A_static = SMatrix{4, 3}(A)
b_static = SVector{4}(b)
StaticArrays.qr(A_static)\b_static # not working
```

It would be great to have this functionality implemented.

gustafsson:
Fixes #1192


src/triangular.jl
```diff
@@ -61,3 +61,39 @@ end
...
+
+inv(R::UpperTriangular{T, <:StaticMatrix}) where T =
+    (@inline; UpperTriangular(_inv_upper_triangular(R.data)))
+
+_inv_upper_triangular(R::StaticMatrix{n, m}) where {n, m} =
+    checksquare(R)
+
+@generated function _inv_upper_triangular(R::StaticMatrix{n, n, T}) where {n, T}
+    ex = quote
+        R_inv = MMatrix{n,n,T}(undef)
```
mateuszbaran, line 74: You should either use `similar` (to properly handle non-isbits `T`) or the same trick as `_A_ldiv_B`. Also, make sure the method works to `T` being an integer.

src/util.jl
```diff
@@ -74,10 +74,12 @@ TrivialView(a::AbstractArray{T,N}) where {T,N} = TrivialView{typeof(a),T,N}(a)
...
-     ip[p] = 1:length(p)
-     similar_type(p)(ip)
+@inline function invperm(p::StaticVector{N,T}) where {N,T<:Integer}
+    ip = zeros(MVector{N,T})
+    @inbounds for i in SOneTo(N)
+        j = p[i]
+        1 <= j <= N && iszero(ip[j]) || throw(ArgumentError("argument is not a permutation"))
+        ip[j] = i
+    end
+    SVector(ip)
```
mateuszbaran, line 84: Quite likely someone relied on this being faster thanks to not performing the checks so I'm not sure about this change.

src/qr.jl
```diff
@@ -62,6 +64,9 @@ function identity_perm(R::StaticMatrix{N,M,T}) where {N,M,T}
     return similar_type(R, Int, Size((M,)))(ntuple(x -> x, Val{M}()))
 end
 
+is_identity_perm(p::StaticVector{M}) where {M} = all(i->i==p[i], 1:M)
```
mateuszbaran, line 67:
```suggestion
is_identity_perm(p::StaticVector{M}) where {M} = all(i->i==p[i], SOneTo(M))
```

mateuszbaran (commented):
Thank you for the contribution. I have a few comments.

PR #1306, opened 2025-05-11
stop pirating `size(::Type{Union{}})`

simeonschaub:
JET complained about this method when analyzing my package and it's a
fairly easy fix

### DIFF
--- a/src/SDiagonal.jl
+++ b/src/SDiagonal.jl
@@ -14,8 +14,10 @@ SDiagonal(x...) = Diagonal(SVector(x...))
 SDiagonal(a::SVector) = Diagonal(a)
 SDiagonal(a::StaticMatrix{N,N,T}) where {N,T} = Diagonal(diag(a))
 
-size(::Type{<:SDiagonal{N}}) where {N} = (N,N)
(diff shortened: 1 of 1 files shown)


nsajko:
How about a test, for example this one:

```julia
function test_we_do_not_own_the_call(func, arg_types)
    function f(method)
        StaticArrays != parentmodule(method)
    end
    @test all(f, methods(func, arg_types))
end
 
@testset "type piracy" begin
    @testset "PR #1306: type piracy of `size`" begin
        test_we_do_not_own_the_call(size, Tuple{Type{Union{}}})
    end
end
```

simeonschaub:
That's a good idea! Unfortunately there is still https://github.com/JuliaArrays/StaticArrays.jl/blob/e23a2f5d0ceb27bab266908e11c3d2b1a395aab9/src/abstractarray.jl#L4 which isn't that easy to fix. Arguably, `length` and `size` should not really be defined on types at all, but that is another discussion.

nsajko (approved):

mateuszbaran (approved):
LGTM, thanks!

mateuszbaran:
`StaticArrayLike` is just a union, so fixing it would also be doable.

PR #1310, opened 2025-07-15
fix: fix n-dimensional `view` into n-dimensional `SArray` with `Any` eltype

AayushSabharwal:
As of the latest release, trying to `view` into an `SArray` with an eltype of `Any` such that the view also returns an `SArray` would wrap the returned array in a scalar. `_maybewrapscalar` assumes that if the `eltype` of the array matches the type of the view, it must be a scalar view. This assumption breaks down when the `eltype` is `Any`. MWE:

```julia
julia> A = SVector{4, Any}(1,2,3,4)

julia> I = SVector{2}(3, 1)

julia> view(A, I)
Scalar{Any}((Any[3, 1],))

julia> view(A, [3, 1])
2-element view(::SVector{4, Any}, [3, 1]) with eltype Any:
 3
 1
```

I've verified that all the added tests fail on the latest release and pass with this PR.

### DIFF
--- a/test/indexing.jl
+++ b/test/indexing.jl
@@ -253,4 +253,16 @@ using StaticArrays, Test
         @test lastindex(ind[3]) === 2
         @test size(ind[3]) === (2,)
     end
+
+    @testset "Array view into `Any` eltype `SArray`" begin
+        A = SVector{4, Any}(1,2,3,4)
+        v = @inferred view(A, SA[3, 1])
+        @test v == SVector{2, Any}(3, 1)
+        A = SMatrix{2, 2, Any}(1, 2, 3, 4)
+        v = @inferred view(A, @SArray(fill(1, 1, 1)))
+        @test v == SMatrix{1, 1, Any}(1)
+        A = SArray{Tuple{2, 2, 2}, Any}(1, 2, 3, 4, 5, 6, 7, 8)
+        v = @inferred view(A, @SArray(fill(1, 1, 1, 1)))
+        @test v == SArray{Tuple{1, 1, 1}, Any}(1)
+    end
 end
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
I'm not sure what to make of the CI failure or how it relates to the changes in this PR.

mateuszbaran:
This looks reasonable.

Regarding the failure, IIRC Julia 1.6 doesn't support the new `[1;;;]` syntax but you can use something like `@SArray fill(1, 1, 1)` instead in the test.

test/indexing.jl
```diff
@@ -253,4 +253,16 @@ using StaticArrays, Test
         @test lastindex(ind[3]) === 2
         @test size(ind[3]) === (2,)
     end
+
+    @testset "Array view into `Any` eltype `SArray`" begin
+        A = SVector{4, Any}(1,2,3,4)
+        v = @inferred view(A, SA[3, 1])
+        @test v == SVector{2, Any}(3, 1)
+        A = SMatrix{2, 2, Any}(1, 2, 3, 4)
+        v = @inferred view(A, @SArray(fill(1, 1)))
```
mateuszbaran, line 262:
```suggestion
        v = @inferred view(A, @SArray(fill(1, 1, 1)))
```
I think this needs one more `1`: two for dimensions and one for for actual element? (similarly in the next test)

AayushSabharwal: Looks like I confused myself - thanks for catching that.

mateuszbaran (approved):

AayushSabharwal:
Thanks for merging! Would it be possible to cut a release soon 😅 this is blocking some work in ModelingToolkit.jl.

mateuszbaran:
Sure, I will make a release soon.

PR #1325, opened 2025-12-17
Delete some functions already defined in Base

JamesWrigley:
This fixes some invalidations. On Julia 1.12 the number of invalidations go down from 1628 to 1498. As a dependency of CUDA.jl and in combination with https://github.com/JuliaData/SentinelArrays.jl/pull/119 the number of invalidations from loading CUDA go from 11175 to 10432.

One potentially dodgy change is removing the `getindex()` method, but I benchmarked it against the default one in Base and got the same results (~3 ns). When running the tests I also found that one of the `@test_broken` tests was passing so I fixed that too, but it's an allocation test so it may just be a 1.12 thing.


JamesWrigley:
Closing because the inference failures are a bit too tricky for me to debug right now.

PR #1326, opened 2025-12-21
Compute inverse of `Cholesky` with a single solve

devmotion:
Alternative to #1258 that instead of averaging out the errors of the two solves computes the inverse with a single solve and a symmetric matrix multiplication.

Checks of symmetry in the tests pass locally, and the example in the tests is even slightly faster:

## This PR

```julia
julia> using StaticArrays, LinearAlgebra, Chairmarks

julia> @be cholesky((x -> x * x' + I)(@SMatrix(randn(3,3)))) inv
Benchmark: 3702 samples with 4741 evaluations
 min    4.561 ns
 median 5.018 ns
 mean   5.419 ns
 max    11.680 ns
```

## master

```julia
julia> using StaticArrays, LinearAlgebra, Chairmarks

julia> @be cholesky((x -> x * x' + I)(@SMatrix(randn(3,3)))) inv
Benchmark: 3756 samples with 3276 evaluations
 min    6.983 ns
 median 7.326 ns
 mean   7.712 ns
 max    53.406 ns
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.9.15"
+version = "1.9.16"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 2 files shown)


mateuszbaran (approved):
I think this is a good idea.

PR #1327, opened 2025-12-22
Make CI pass on Julia 1.12

mateuszbaran:
It would be better to have these tests working but it's better to disable them than having non-functional CI.


devmotion:
Maybe there'd be a bit less noise if tests would run on `pre` instead of `nightly`?

mateuszbaran:
`pre` would be less noisy, though I personally don't care too much either way. `pre` allows for spotting issues sooner but then someone has to actually investigate these failures.

PR #1332, opened 2026-02-16
Implement wrapped-SMatrix + UniformScaling

THargreaves:
## Context

`UniformScaling` operations on structured wrappers backed by `StaticMatrix` were falling back to generic `LinearAlgebra` methods. That path uses mutable copies, which caused type leakage from `SMatrix`-backed wrappers to `MMatrix`-backed wrappers. In some cases (for example `I - Symmetric(SMatrix(...))`), it could also fail due to mutation of immutable storage.

I don't think there is an easy way to solve this problem in general due to how these operations are defined in base, so instead I've just written some appropriate wrappers.

## What this changes

This PR adds `UniformScaling` specialisations for structured wrappers over `StaticMatrix` in `src/linalg.jl`:
  - `Symmetric`
  - `Hermitian`
  - `UpperTriangular`
  - `LowerTriangular`
  - `UnitUpperTriangular` (result type `UpperTriangular`)
  - `UnitLowerTriangular` (result type `LowerTriangular`)

I only implement methods for `A + J` and `J - A` for brevity, with `uniformscaling.jl:L156` handling the reverse argument orders.

I include a special case for the some of Hermitian + complex UniformScaling which removes the wrapper to be type stable, consistent with base LinearAlgebra.jl.

## Tests

I've extended `test/linalg.jl` (`Interaction with UniformScaling`) to cover all of the cases mentioned.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -40,6 +40,50 @@ import Base: +, -, *, /, \
     end
 end
 
+@inline _plus_uniform_parent(A, λ) = _plus_uniform(Size(A), parent(A), λ)
+@inline _minus_uniform_parent(A, λ) = _plus_uniform(Size(A), -parent(A), λ)
+@inline _sym_uplo(A) = A.uplo == 'U' ? :U : :L
+
+# Self-adjoint wrappers over StaticMatrix with UniformScaling
+@inline +(A::Symmetric{<:Any,<:StaticMatrix}, J::UniformScaling) =
+    Symmetric(_plus_uniform_parent(A, J.λ), _sym_uplo(A))
+@inline -(J::UniformScaling, A::Symmetric{<:Any,<:StaticMatrix}) =
+    Symmetric(_minus_uniform_parent(A, J.λ), _sym_uplo(A))
+
+@inline +(A::Hermitian{<:Any,<:StaticMatrix}, J::UniformScaling) =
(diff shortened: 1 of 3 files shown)


src/linalg.jl
```diff
@@ -40,6 +40,50 @@ import Base: +, -, *, /, \
     end
 end
 
+@inline _plus_uniform_parent(A, λ) = _plus_uniform(Size(A), parent(A), λ)
+@inline _minus_uniform_parent(A, λ) = _plus_uniform(Size(A), -parent(A), λ)
+@inline _sym_uplo(A) = A.uplo == 'U' ? :U : :L
+
+# Self-adjoint wrappers over StaticMatrix with UniformScaling
+@inline +(A::Symmetric{<:Any,<:StaticMatrix}, J::UniformScaling) =
```
mateuszbaran:
```suggestion
@inline +(A::Symmetric{T,<:StaticMatrix{N,M,T}}, J::UniformScaling) where {N,M,T} =
```
I generally prefer to repeat type constraints in method signatures, that way there is less chance for ambiguity. Could you edit other methods similarly?

mateuszbaran:
I think this looks reasonable up to the argument type as explained above. Could you also bump patch version?

THargreaves:
No problem, both of those changes made.

mateuszbaran (approved):

PR #1335, opened 2026-03-09
Fix norm_sqr init for Unitful element types

aplavin:
\`norm_sqr\` used \`_init_zero(v)\` as the mapreduce init, which computes \`float(norm(zero(eltype)))\`. For Unitful quantities this has units of e.g. \`m\`, but \`norm_sqr\` produces values with units of \`m²\` — causing \`DimensionError\` when adding.

Fix: use \`norm_sqr(zero(...))\` instead of \`norm(zero(...))\` for the init, so it has the correct squared units.

This fixes \`norm\` for arrays of Unitful SVectors and \`isapprox\` for Unitful SVectors with explicit \`atol\`.


mateuszbaran (approved):
Thanks, LGTM.

PR #1336, opened 2026-03-09
Fix pinv for Symmetric/Hermitian wrapped SMatrix

aplavin:
\`pinv(Symmetric(smatrix))\` failed because the \`pinv\` codepath tried to construct a \`Symmetric\` from the SVD result, which doesn't work.

Fix: unwrap \`HermOrSym\` to the underlying \`StaticMatrix\` before calling \`_pinv\`.

### DIFF
--- a/src/pinv.jl
+++ b/src/pinv.jl
@@ -7,6 +7,8 @@
     return _pinv(A_S, atol, rtol)
 end
 
+@inline pinv(A::LinearAlgebra.HermOrSym{T,<:StaticMatrix}; kwargs...) where T = pinv(similar_type(A)(A); kwargs...)
+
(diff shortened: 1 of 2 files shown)


src/pinv.jl
```diff
@@ -7,6 +7,8 @@
     return _pinv(A_S, atol, rtol)
 end
 
+@inline pinv(A::LinearAlgebra.HermOrSym{T,<:StaticMatrix}; kwargs...) where T = pinv(similar_type(A)(A); kwargs...)
```
mateuszbaran:
```suggestion
@inline pinv(A::LinearAlgebra.HermOrSym{T,<:StaticMatrix}; kwargs...) where T = pinv(convert(similar_type(A), A); kwargs...)
```
This should avoid making a copy when `A.data` is already mutable.

mateuszbaran (approved):

PR #1337, opened 2026-03-09
Fix zero() for LinearAlgebra wrappers over StaticArrays

aplavin:
\`zero(Symmetric(smatrix))\`, \`zero(Diagonal(svector))\`, etc. failed because \`zero\` was not defined for LinearAlgebra wrapper types over StaticArrays.

Fix: add \`zero\` methods for \`Symmetric\`, \`Hermitian\`, \`UpperTriangular\`, \`LowerTriangular\`, \`UnitUpperTriangular\`, \`UnitLowerTriangular\`, \`Transpose\`, \`Adjoint\`, and \`Diagonal\` wrapping StaticArrays.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -155,7 +155,6 @@ end
 for WR in (:UpperTriangular, :LowerTriangular, :UnitUpperTriangular, :UnitLowerTriangular, :Transpose, :Adjoint)
     @eval @inline Base.zero(a::$WR{<:Any,<:StaticArray}) = $WR(zero(parent(a)))
 end
-@inline Base.zero(a::Diagonal{<:Any,<:StaticVector}) = Diagonal(zero(parent(a)))
 
 @inline _construct_sametype(a::Type, elements) = a(elements)
 @inline _construct_sametype(a, elements) = typeof(a)(elements)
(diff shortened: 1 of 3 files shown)


src/linalg.jl
```diff
@@ -148,6 +148,15 @@ end
...
 @inline Base.zero(a::Type{SA}) where {SA <: StaticArray} = zeros(SA)
 
+# Wrappers over StaticArray (see StaticMatrixLike definition)
+for WR in (:Symmetric, :Hermitian)
+    @eval @inline Base.zero(a::$WR{<:Any,<:StaticMatrix}) = $WR(zero(parent(a)), _sym_uplo(a))
+end
+for WR in (:UpperTriangular, :LowerTriangular, :UnitUpperTriangular, :UnitLowerTriangular, :Transpose, :Adjoint)
+    @eval @inline Base.zero(a::$WR{<:Any,<:StaticArray}) = $WR(zero(parent(a)))
+end
+@inline Base.zero(a::Diagonal{<:Any,<:StaticVector}) = Diagonal(zero(parent(a)))
```
mateuszbaran: `zero` is already special-cased for `Diagonal` with `SVector`, and works fine with mutable static arrays, so this method doesn't seem to be needed.

aplavin: just didn't notice it

mateuszbaran (approved):

mateuszbaran:
Could you bump the patch version here? I would make a new release with your patches.

aplavin:
sure! thanks a lot for being so responsive

PR #1343, opened 2026-04-28
Add Mooncake.jl extension

yebai:
- Add `StaticArraysMooncakeExt` package extension porting Mooncake.jl's existing `MooncakeStaticArraysExt`. The extension teaches Mooncake that `SArray{S,T,N,L}` with IEEE float / complex IEEE float element types is a leaf-like differentiable container (its tangent type is itself), and adds `frule!!` / `rrule!!` rules for `Mooncake._new_` (which IR normalisation lowers `SArray(...)` constructor calls to) and `getindex(::SArray, ::Int)`.
- Tests live under `test/ext/mooncake/` with their own `Project.toml` and a self-activating driver script, mirroring how Mooncake's own repo organises ext tests. A new `extra` CI job (matrix-templated for future ext labels) runs them on Julia 1.10 and the current stable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ext/StaticArraysMooncakeExt.jl
+++ b/ext/StaticArraysMooncakeExt.jl
@@ -1,0 +1,158 @@
+module StaticArraysMooncakeExt
+
+using Mooncake
+using Random: AbstractRNG
+using Base: IEEEFloat
+using StaticArrays: StaticArrays, SArray
+
+using Mooncake: @foldable
+
+import Mooncake:
+    MaybeCache,
+    IncCache,
+    SetToZeroCache,
+    NoFData,
+    NoRData,
+    CoDual,
+    Dual,
+    MinimalCtx,
+    primal,
+    tangent,
+    zero_fcodual,
+    tangent_type,
+    fdata_type,
+    rdata_type,
+    zero_tangent_internal,
+    randn_tangent_internal,
+    set_to_zero_internal!!,
+    increment_internal!!,
+    _add_to_primal_internal,
+    tangent_to_primal_internal!!,
+    primal_to_tangent_internal!!,
+    _dot_internal,
+    _scale_internal,
+    zero_rdata,
+    zero_rdata_from_type,
+    can_produce_zero_rdata_from_type,
+    _verify_rdata_value,
+    __verify_fdata_value,
+    _new_
+
+# Element types treated as differentiable: real and complex IEEE floats.
+const _SElt = Union{IEEEFloat,Complex{<:IEEEFloat}}
+
+# An SArray with a supported element type uses *itself* as its tangent type.
+# It is immutable and stores only by-value scalar data, so its fdata is empty
+# and its rdata carries the full tangent. This mirrors how Mooncake handles
+# `Complex{<:IEEEFloat}` in `src/rules/complex.jl`.
+
+@foldable function tangent_type(::Type{SArray{S,T,N,L}}) where {S,T<:_SElt,N,L}
+    return SArray{S,T,N,L}
+end
+
+@foldable function tangent_type(
+    ::Type{NoFData}, ::Type{SArray{S,T,N,L}}
+) where {S,T<:_SElt,N,L}
+    return SArray{S,T,N,L}
+end
+
+# Non-parametric alias used as a constraint, analogous to `CF` in complex.jl.
+const _SAFloat = SArray{S,T,N,L} where {S,T<:_SElt,N,L}
+
+@foldable fdata_type(::Type{T}) where {T<:_SAFloat} = NoFData
+@foldable rdata_type(::Type{T}) where {T<:_SAFloat} = T
+
+tangent(::NoFData, t::_SAFloat) = t
+
+# Core tangent operations.
+
+zero_tangent_internal(p::_SAFloat, ::MaybeCache) = zero(p)
+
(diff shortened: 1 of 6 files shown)


yebai:
@ChrisRackauckas, this should fix https://github.com/SciML/SciMLSensitivity.jl/issues/1425

vchuravy:
Speaking from my experience maintaining Enzyme.jl, I would encourage this to be an extension in Mooncake for StaticArrays.jl

StaticArrays has been very stable for a while, while the AD landscape is rapidly evolving. Looking back at the Enzyme extension in KernelAbstraction, it has caused the majority of patch version releases in the 0.9 series. 

Furthermore, the extension uses Mooncake internals quite a lot and barely requires special knowledge about StaticArrays, so it would make sense if this code lives where Mooncake experts can have an eye on it and there is less risk of it becoming stale.

vchuravy:
> Unfortunately, it has to be here or closed.

Why can it not be an extension in Mooncake? Extensions are perfectly symmetrical.

yebai:
Mooncake's policy means it covers Julia Base only. Existing pkg extensions maybe moved out in the future. 

Unfortunately I won't have time to read further comments. I'll fix review comments next week if any. Please feel free to close this if it is out of scope.

vchuravy:
> Mooncake's policy means it covers Julia Base only. Existing pkg extensions maybe moved out in the future.

That is your prerogative, but I would ask you to reconsider. A PR and an extension to another package is an implicit ask: "Please take on the maintenance of this piece of code." This is not a small ask, and the internals of an AD package are complicated. I have co-maintained Enzyme.jl, ForwardDiff.jl and even helped with Mooncake and it's interaction with the Julia compiler, yet when I look at this code, I am not confident that I could maintain it. 

When deciding where a package extension lives I try to look at the amount of internals of each package it is using. The current extension is mostly Mooncake and very little StaticArrays, so it would very naturally for me be a Mooncake extension.

wsmoses:
I believe this was already discussed and considered by @gdalle in https://github.com/JuliaArrays/StaticArrays.jl/pull/1342

"""
Closing this PR for now since it is unclear to me that augmenting StaticArrays this way is the right design choice. This code should live in Mooncake
"""

ext/StaticArraysMooncakeExt.jl
```diff
@@ -0,0 +1,158 @@
...
+    return SArray{S,T,N,L}
+end
+
+# Non-parametric alias used as a constraint, analogous to `CF` in complex.jl.
+const _SAFloat = SArray{S,T,N,L} where {S,T<:_SElt,N,L}
+
+@foldable fdata_type(::Type{T}) where {T<:_SAFloat} = NoFData
+@foldable rdata_type(::Type{T}) where {T<:_SAFloat} = T
+
+tangent(::NoFData, t::_SAFloat) = t
```
mateuszbaran, line 65: Aside from whether this should live in Mooncake or StaticArrays, was this even reviewed by a human before submission? These definitions look like they should be exactly identical to defaults.

Is this really the kind of pull requests you would welcome in your projects?

yebai: Yes, this is actually needed -- the default tangent for Mooncake is not StaticArrays. Mooncake's automatically derived tangents are usually a NamedTuple of all field members of structs. 

The tests for this PR run successfully locally, and I reviewed the code myself before creating it.

yebai:
I'm closing this PR because I won't have time to continue discussions. The code here can serve as a template for anyone who wants to use StaticArrays with Mooncake.

ext/StaticArraysMooncakeExt.jl
mateuszbaran: If it's all necessary, then it's not much better. Nearly duplicating so much code for each array-like type is going to cause huge maintenance issues. Maybe it could all be generated by a macro that lives in Mooncake.jl?

Anyway, if you don't have the time to address issues raised in a review, making PRs is just a waste of time.

yebai: > If it's all necessary, then it's not much better. Nearly duplicating so much code for each array-like type is going to cause huge maintenance issues. Maybe it could all be generated by a macro that lives in Mooncake.jl?

Will Tebbutt primarily designed Moocanke’s tangent-type system while working with me. Automating this via metaprogramming is difficult, as subtypes of AbstractArray can vary considerably. This reflects lessons learned from Zygote, ChainRules, and other autograd systems. Will has nearly a decade of experience with automatic differentiation in Julia, so I trust his judgment on this.

> if you don't have the time to address issues raised in a review, making PRs is just a waste of time.

The level of disagreement has made it difficult to sustain focused technical discussion.

Related: https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/custom_tangent_type/#Writing-Custom-Tangent-Types

mateuszbaran: > Automating this via metaprogramming is difficult, as subtypes of AbstractArray can vary considerably. 

I have about 8 years of experience writing and maintaining Julia code, and in my opinion automation via metaprogramming would be the easier approach in the long run. This extension is its current form breaks major good engineering practices, for example not accessing internal functions of other packages and only using public API. Unless "internal" in function name is only decorative and they are not internal at all?

PR #1348, opened 2026-07-25
Report element count in generator-too-short error

### LINKED ISSUE #1207: Bad error message: "Expected exactly 5 elements, but generator stopped at 5"
I think this error message could be improved
```julia
julia> SVector{5}(i for i in 1:5 if i != 2)
ERROR: Generator produced too few elements: Expected exactly 5 elements, but generator stopped at 5
```

JohnCobbler:
The too-short generator error printed the failing `CartesianIndex`
coordinate via `shape_string(i)`, not the number of values actually
produced. For `SVector{5}(i for i in 1:5 if i != 2)` that showed
"stopped at 5" even though only 4 elements were yielded.

Report the produced count as `LinearIndices(inds)[i] - 1` instead, and
drop the unused `shape_string(::CartesianIndex)` method.

Fixes #1207

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.9.18"
+version = "1.9.19"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


mateuszbaran:
Thanks for the fix, LGTM. I've disabled one test that was caused by an LLVM bug most likely: https://github.com/JuliaLang/julia/issues/62368 .

mateuszbaran (approved):

PR #1350, opened 2026-08-23
Fix eigen for 3x3 Symmetric{BigFloat} static matrices

### LINKED ISSUE #1349: `eigen` returns wrong results for 3×3 `Symmetric{BigFloat}` static matrices
Yes, I know `BigFloat` allocates, so using `SMatrix` with it kind of
defeats the purpose, but we're all for interop here, right?  And this
is a pretty ***huge*** error.  (Hypothetically, I guess it would also be
an error for some other non-allocating but huge-exponent-range
type, too.)

For a 3×3 `Symmetric` static matrix of `BigFloat`, `eigen` and
`eigvals` disagree; `eigen` is wrong:

```julia

julia> A = Symmetric(SMatrix{3,3}(BigFloat[4 1 2; 1 5 3; 2 3 6]))
3×3 Symmetric{BigFloat, SMatrix{3, 3, BigFloat, 9}} with indices SOneTo(3)×SOneTo(3):
 4.0  1.0  2.0
 1.0  5.0  3.0
 2.0  3.0  6.0

julia> eigvals(A)
3-element SVector{3, BigFloat} with indices SOneTo(3):
 2.194397167422408554607930640304764781172696375270516435906510101161316445735186
 3.386770156607549226057975534383083133866174193346971868254782083048971298796766
 9.418832675970042219334093825312152084961129431382511695838707815789712255468117

julia> eigen(A).values
3-element SVector{3, BigFloat} with indices SOneTo(3):
 3.384615384615384615384615384615384615384615384615384615384615384615384615384562
 5.6153846153846153846153846153846153846153846153846153846153846153846153846153
 6.0
 ```

`eigvals` is correct (it matches the `Float64` answer to ~15 digits).
`eigen` is not, and the eigenvectors are wrong to match:
`norm(A*E.vectors - E.vectors*Diagonal(E.values))` is about `5.1`,
rather than something near zero. There's nothing special about this
matrix — I first hit it on one with a degenerate spectrum, but a
generic one like the above does it just as well.

Scope, as far as I can tell:

* It's only `BigFloat`; `Float16`, `Float32` and `Float64` are all
  fine (residuals of a few eps).
* It's only 3×3; 2×2 is fine, and 4×4 and up are handed off to
  `LinearAlgebra`, so they're fine too (given a generic backend such
  as GenericLinearAlgebra).
* Lowering the precision doesn't help — the residual is ~5.1 at
  `setprecision(BigFloat, 24)`, `53` and `256` alike.

I think the problem is in this line in the 3×3 symmetric solver
(`src/eigen.jl:289`):

```julia
max_iteration = 2 * (1 + precision(T) - exponent(floatmin(T)))
```

For `BigFloat`, `exponent(floatmin(BigFloat))` is
-4611686018427387904, so that expression comes to 9223372036854776322
— just 515 past `typemax(Int64)` (9223372036854775807) — and wraps
around to a large negative number. Both loops that use it
(`src/eigen.jl:293` and `:339`) are then `for iteration in
1:max_iteration` over an empty range, so the QR iteration never runs
and the routine returns its un-iterated intermediate as though it had
converged. Since `precision(T)` contributes only ~10²–10³ and the
`floatmin` exponent contributes ~4.6e18, changing the precision can't
get you back under the limit — which is why `setprecision` makes no
difference.

For the other types the same expression is small and harmless: 52, 302
and 2152 for `Float16`, `Float32` and `Float64`.

### Claude's suggested fix

The following is just a copy-paste from Claude's response to my request for a fix.  None of the text is mine, but it looks right to me, and I think it would compile away at least for the important cases of `Float*`:

```diff
-    max_iteration = 2 * (1 + precision(T) - exponent(floatmin(T)))
+    emin = max(exponent(floatmin(T)), -100_000)
+    max_iteration = 2 * (1 + precision(T) - emin)
```

  The reason it can be this simple is something I checked before recommending it: the bound is only a safety valve. The loop breaks as soon
  as it genuinely converges — ~6 iterations, and flat across 64 to 4096 bits of precision. I'd worried that un-overflowing the bound would
  just trade a wrong answer for a hang, but a cap of 10⁷ returns in 367 µs, so exact convergence really is reached. Nothing needs to change
  about the algorithm or the convergence test.

  Verified on a deved copy:

  - Float16/Float32/Float64 bounds come out unchanged — still 52, 302, 2152. No behavior change for existing users.
  - BigFloat goes from a negative (overflowed) bound to a finite 200514.
  - Residual drops from ~5.1 to ~17 eps, matching the other element types; eigen and eigvals now agree. Confirmed on the generic, degenerate,
    and identity matrices.

  -100_000 sits below the floatmin exponent of every hardware type and Float128 (-16382), which is why nothing existing shifts. That constant
  is the arbitrary part — widening the arithmetic to Int128 and capping would work equally well, so I framed it in the draft as a suggestion
  rather than the answer.

  One incidental confirmation: setting the cap to 0 reproduces the unpatched behavior exactly, residual 5.1 on the nose. That's about as
  direct a confirmation of "zero iterations run" as you could ask for.

### Version info

```
StaticArrays v1.9.19  (current release)
```

```julia
julia> versioninfo()
Julia Version 1.12.7
Commit 6d172b025e4 (2026-08-15 08:05 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: macOS (arm64-apple-darwin25.5.0)
  CPU: 12 × Apple M2 Max
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, apple-m2)
  GC: Built with stock GC
Threads: 8 default, 1 interactive, 8 GC (on 8 virtual cores)
Environment:
  JULIA_EDITOR = code-insiders
```

moble:
Claude did this, including writing the following (I'm on vacation).  But I've reviewed the code and the following text and it looks right to me.  Note that the new lines will compile away for the important cases of `Float*` and probably other fixed-precision types (not `BigFloat`, but I don't think this is a bottleneck for it).

---

Fixes #1349.

The iteration bound in the 3×3 symmetric solver overflows `Int64` for `BigFloat`:

```julia
max_iteration = 2 * (1 + precision(T) - exponent(floatmin(T)))
```

MPFR's minimum exponent is exactly `-2^62`, so `2 * 2^62` is exactly `typemax(Int64) + 1` and the
expression wraps to a large negative number. Both loops that use it (`src/eigen.jl:293` and `:339`)
then run `for iteration in 1:max_iteration` over an empty range, so the QR sweep never executes and
the routine returns its un-iterated intermediate as though it had converged.

Because it clears the limit by only `2 * (1 + precision(T)) + 1`, this happens at *every* `BigFloat`
precision — 515 over at the default 256 bits, and still 7 over at 2 bits — which is why reducing the
precision doesn't avoid it.

The failure is silent, and `eigen` ends up disagreeing with `eigvals` (which uses a different,
closed-form path and is correct):

```julia
julia> setprecision(BigFloat, 64);

julia> A = Symmetric(SMatrix{3,3}(BigFloat[4 1 2; 1 5 3; 2 3 6]));

julia> eigvals(A)               # correct
3-element SVector{3, BigFloat} with indices SOneTo(3):
 2.1943971674224085548
 3.38677015660754922484
 9.41883267597004221992

julia> eigen(A).values          # wrong
3-element SVector{3, BigFloat} with indices SOneTo(3):
 3.38461538461538461527
 5.61538461538461538408
 6.0
```

With this change the two agree, and `norm(A*V - V*Diagonal(λ))` drops from about `5.1` to about
`17 * eps(BigFloat)`, in line with the other element types.

### The fix

```julia
emin = max(exponent(floatmin(T)), -100_000)
max_iteration = 2 * (1 + precision(T) - emin)
```

A few notes on why this is safe:

* **Hardware types are unaffected.** `-100_000` is below the `floatmin` exponent of every hardware
  type and of `Float128` (`-16382`), so their bounds are unchanged: still 52, 302 and 2152 for
  `Float16`, `Float32` and `Float64`.
* **It costs nothing at run time.** For concrete hardware types the expression constant-folds
  exactly as before — `code_typed` returns the literal `2152` for `Float64` both with and without
  the change, and a 3×3 `Float64` `eigen` benchmarks the same (~387 ns, 0 allocations). It doesn't
  fold for `BigFloat`, since `precision(BigFloat)` is a runtime global, but that was already true;
  the change adds one integer `max` to a path that is about to do MPFR arithmetic.
* **The bound is only a safety valve.** The loop exits as soon as it actually converges — about 6
  iterations here, independent of precision — so nothing relies on the larger bound. Raising the cap
  to `10^7` still returns in ~370 µs.

The specific constant is arbitrary.

### Tests

Adds a `3×3 BigFloat` testset covering the reconstruction, agreement with `eigvals`, agreement with
the `Float64` answer, and a reduced precision. All four assertions fail on `master` and pass here.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StaticArrays"
 uuid = "90137ffa-7385-5640-81b9-e52037218182"
-version = "1.9.19"
+version = "1.9.20"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -286,7 +286,8 @@ end
     b22 = a22
 
     # Givens reflections, B' = G^T * B * G, preserve tridiagonal matrices
-    max_iteration = 2 * (1 + precision(T) - exponent(floatmin(T)))
+    emin = max(exponent(floatmin(T)), -100_000)
+    max_iteration = 2 * (1 + precision(T) - emin)
 
     if abs(b12) <= abs(b01)
         saveB00, saveB01, saveB11 = b00, b01, b11
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -286,6 +286,11 @@ end
     b22 = a22
 
     # Givens reflections, B' = G^T * B * G, preserve tridiagonal matrices
+    #
+    # `emin` is clamped because it can overflow for extended-range numbers.  This limit
+    # compiles away for standard fixed-precision types, and does not change `max_iteration`
+    # for them.  It corrects the behavior for `BigFloat`; see
+    # https://github.com/JuliaArrays/StaticArrays.jl/issues/1349
     emin = max(exponent(floatmin(T)), -100_000)
     max_iteration = 2 * (1 + precision(T) - emin)
 
--- a/test/eigen.jl
+++ b/test/eigen.jl
@@ -329,4 +329,21 @@ using StaticArrays, Test, LinearAlgebra
         λ, V = eigen(A)
         @test A * V ≈ V * Diagonal(λ)
     end
+
+    @testset "3×3 BigFloat" begin
+        # The iteration bound used to overflow `Int64` for `BigFloat`, so the QR sweep
+        # never ran and `eigen` returned its un-iterated intermediate (#1349).
+        m = BigFloat[4 1 2; 1 5 3; 2 3 6]
+        A = Symmetric(SMatrix{3,3}(m))
+        λ, V = eigen(A)
+        @test A * V ≈ V * Diagonal(λ)
+        @test λ ≈ eigvals(A)
+        @test λ ≈ eigvals(Symmetric(Float64.(m)))
+        # the overflow happened at every precision, so check a small one too
+        setprecision(BigFloat, 24) do
+            B = Symmetric(SMatrix{3,3}(BigFloat[4 1 2; 1 5 3; 2 3 6]))
+            μ, W = eigen(B)
+            @test B * W ≈ W * Diagonal(μ)
+        end
+    end
 end


mateuszbaran:
Thanks, the fix looks reasonable. Could you write a short comment in the code explaining the logic behind `emin` (or at least a link the the issue about it) and bump the patch version?

moble:
Thanks.  Good suggestions.  I've implemented them now.

mateuszbaran (approved):

PR #1352, opened 2026-09-13
Extend to_indices to accept any integer in StaticArray index

barche:
When indexing an array using a `StaticArray` as indices, the return type depends on the element type of the `StaticArray`, resulting in a slowdown when the type differs from `Int`, as seen in the normals calculation code in GeometryBasics.jl (see PR https://github.com/JuliaGeometry/GeometryBasics.jl/pull/285). This change makes sure any kind of integer behaves the same way and always returns an `SVector` as the indexing result.


mateuszbaran (approved):

ffreyer:
This seems to have broken
```julia
v = SVector(1,2,3,4)
v[v .== 1]
```
With 1.9.20 this returns `[1]` (base array), with 1.9.21 it errors:
```
ERROR: MethodError: no method matching index_sizes(::Base.LogicalIndex{Int64, SVector{4, Bool}})
The function `index_sizes` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  index_sizes()
   @ StaticArrays ~/.julia/packages/StaticArrays/y4kA2/src/indexing.jl:83
  index_sizes(::Int64, Any...)
   @ StaticArrays ~/.julia/packages/StaticArrays/y4kA2/src/indexing.jl:84
  index_sizes(::S, Any...) where S<:Size
   @ StaticArrays ~/.julia/packages/StaticArrays/y4kA2/src/indexing.jl:81
  ...

Stacktrace:
 [1] _getindex(::IndexLinear, ::SVector{4, Int64}, ::StaticArrays.StaticIndexing{Base.LogicalIndex{Int64, SVector{4, Bool}}})
   @ StaticArrays ~/.julia/packages/StaticArrays/y4kA2/src/indexing.jl:265
 [2] getindex(A::SVector{4, Int64}, I::SVector{4, Bool})
   @ Base ./abstractarray.jl:1342
 [3] top-level scope
   @ REPL[4]:1
```

mateuszbaran:
Thanks for reporting, it's odd we don't test for that. I'll fix it in a moment.

